PR #8, opened 2025-09-09
feat/port media chrome components

cjpillsbury:


packages/core/media-store/rollup.config.js
```diff
@@ -0,0 +1,62 @@
+const typescript = require('@rollup/plugin-typescript');
```
cjpillsbury, line 1: I think we should probably migrate everything to rollup or similar. I was able to use tsup for the core libraries before, but I've ended up fighting things along the way and that also means we're using two build tools instead of one (learning curve + maintainability of monorepo).

packages/react/react/src/components/TimeRange.tsx
```diff
@@ -0,0 +1,98 @@
...
+    ...props,
+  };
+
+  return baseProps;
+};
+
+export type useTimeRangeProps = typeof useTimeRangeProps;
+type TimeRangeProps = ReturnType<useTimeRangeProps>;
+
+export const renderTimeRange = (
```
cjpillsbury, line 76: NOTE: These (and their equivalents in web components) are just simple implementations for testing full component to state flow. They do not attempt to solve our "compound component" problem.

packages/react/react/src/components/TimeRange.tsx
```diff
@@ -0,0 +1,98 @@
...
+        const seekTime = ratio * state.duration;
+        state.requestSeek(seekTime);
+      }}
+    />
+  );
+};
+
+export type renderTimeRange = typeof renderTimeRange;
+
+export const TimeRange = toConnectedComponent(
```
cjpillsbury, line 99: NOTE: Our compound component architecture may require using something other than this higher order function, or updates to this higher order function to (optionally) support complex component state, or something else altogether.

packages/react/react/src/skins/styles.module.css
```diff
@@ -67,6 +67,71 @@
   flex-grow: 1;
 }
 
+/* Time Range UI/Styles */
+.TimeRange {
```
cjpillsbury: NOTE: These styles are definitely "wrong", even for the simple use case.

PR #10, opened 2025-09-11
feat: make time range compound component

### LINKED ISSUE #13: Add TimeRange React compound component
Add TimeRange React compound component

luwes:
fix #13

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,4 +102,4 @@ bun.lockb
 # Monorepo specific
 .rush/
 .turbo/
-.lerna/
+.lerna/.playwright-mcp/
--- a/examples/react-demo/src/main.tsx
+++ b/examples/react-demo/src/main.tsx
@@ -5,18 +5,19 @@ import { MediaProvider, Video, MediaSkinDefault } from '@vjs-10/react';
 const DemoPlayer = () => {
   return (
     <MediaProvider>
-      <MediaSkinDefault>
-        {/* @ts-ignore */}
-        {/* <Video
-          muted
-          src="https://stream.mux.com/A3VXy02VoUinw01pwyomEO3bHnG4P32xzV7u1j1FSzjNg/high.mp4"
-        /> */}
-        <Video
-          style={{ display: 'block', width: '100%', maxWidth: '600px', aspectRatio: '16/9' }}
-          muted
-          src="https://stream.mux.com/a4nOgmxGWg6gULfcBbAa00gXyfcwPnAFldF8RdsNyk8M.m3u8"
-        />
-      </MediaSkinDefault>
+      <div style={{ width: '100%', maxWidth: '800px', margin: '0 auto' }}>
+        <MediaSkinDefault>
+          {/* @ts-ignore */}
+          {/* <Video
+            muted
+            src="https://stream.mux.com/A3VXy02VoUinw01pwyomEO3bHnG4P32xzV7u1j1FSzjNg/high.mp4"
+          /> */}
+          <Video
+            muted
+            src="https://stream.mux.com/a4nOgmxGWg6gULfcBbAa00gXyfcwPnAFldF8RdsNyk8M.m3u8"
+          />
+        </MediaSkinDefault>
+      </div>
     </MediaProvider>
   );
 };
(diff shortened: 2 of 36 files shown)


packages/react/react/package.json
```diff
@@ -17,6 +17,7 @@
   ],
   "scripts": {
     "build": "rollup -c",
+    "dev": "rollup -c -w",
```
cjpillsbury, line 20: ❤️

packages/react/react/src/components/TimeRange.tsx
```diff
@@ -1,105 +1,371 @@
...
+    'aria-valuemax': 100,
+    'aria-valuenow': sliderFill,
     'aria-valuetext': `${currentTimeText} of ${durationText}`,
-    /** data attributes */
     'data-current-time': state.currentTime,
     'data-duration': state.duration,
-    /** external props spread last to allow for overriding */
+    style: {
+      ...props.style,
+      '--slider-fill': `${Math.round(sliderFill)}%`,
```
cjpillsbury, line 172: thought: If we go the route of custom CSS properties for these things, a nitpick improvement here - put the spread second to make it easy to override default behaviors.

packages/react/react/src/components/TimeRange.tsx
```diff
@@ -1,105 +1,371 @@
...
 };
 
-export type useTimeRangeProps = typeof useTimeRangeProps;
-type TimeRangeProps = ReturnType<useTimeRangeProps>;
+type useTimeRangeRootState = typeof useTimeRangeRootState;
+type useTimeRangeRootProps = typeof useTimeRangeRootProps;
+type TimeRangeRootState = ReturnType<useTimeRangeRootState>;
+type TimeRangeRootProps = ReturnType<useTimeRangeRootProps>;
+
+export const renderTimeRangeRoot = (props: TimeRangeRootProps, state: TimeRangeRootState) => {
```
cjpillsbury, line 192: per prior out of band convo, we'll likely want to do some architecture expansion so that less of this lives in the render function (would require updates to `toConnectedComponent` or similar). Ideally (imo), we want to land on something more like the simplicity of the non-compound components, roughly like:

```tsx
export const renderTimeRangeRoot = (props: TimeRangeRootProps, state: TimeRangeRootState) => {
  return <div {...props}>{props.children}</div>
};
```

packages/react/react/src/components/TimeRange.tsx
```diff
@@ -1,105 +1,371 @@
...
+
+// ============================================================================
+// THUMB COMPONENT
+// ============================================================================
+
+export const useTimeRangeThumbProps = (props: React.HTMLAttributes<HTMLDivElement>) => {
+  return {
+    ...props,
+    style: {
+      ...props.style,
```
cjpillsbury, line 270: same spread order callout/comment as above, re: overridable styles if this is where the dust settles

packages/react/react/src/components/TimeRange.tsx
```diff
@@ -1,105 +1,371 @@
...
+  useTimeRangeProgressProps,
+  renderTimeRangeProgress,
+  'TimeRange.Progress'
+);
+
+// ============================================================================
+// EXPORTS
+// ============================================================================
+
+export const TimeRange = Object.assign(
```
cjpillsbury, line 354: nitpick: Is `Object.assign` needed at all? Can we just:
```ts
export const TimeRange = {
  Root: TimeRangeRoot,
  Track: TimeRangeTrack,
  Thumb: TimeRangeThumb,
  Pointer: TimeRangePointer,
  Progress: TimeRangeProgress,
} as const;
```
?

packages/react/react/src/components/TimeRange.tsx
```diff
@@ -1,105 +1,371 @@
...
 import { timeRangeStateDefinition } from '@vjs-10/media-store';
 
-export const useTimeRangeState = (_props: any) => {
+import { toConnectedComponent, toContextComponent } from '../utils/component-factory';
+
+const formatTime = (time: number): string => {
+  const minutes = Math.floor(time / 60);
+  const seconds = Math.floor(time % 60);
+  return `${minutes}:${seconds.toString().padStart(2, '0')}`;
+};
```
sampotts, line 11: a definite nit: but I'd imagine this could be shared with other time display components and even non-React stuff eventually. Perhaps the same for below.

PR #14, opened 2025-09-15
feat: add compound html timerange component

luwes:


packages/react/react/src/utils/component-factory.tsx
```diff
@@ -7,18 +7,20 @@ export type StateHookFn<TProps = any, TState = any> = (props: TProps) => TState;
...
+  state: TState
 ) => TResultProps;
 
 export type RenderFn<TProps = any, TState = any> = (
   props: TProps,
-  state: TState,
+  state: TState
 ) => React.ReactElement;
 
+const Context = React.createContext<any | null>(null);
```
cjpillsbury, line 18: non-blocking: If there's an easy way to make this optional, that would be 👌. Feel free to e.g. change the args into an "options-style object" if that makes things easier.

packages/react/react/src/components/TimeRange.tsx
```diff
@@ -237,8 +182,8 @@ const TimeRangeRoot = toConnectedComponent(
 // TRACK COMPONENT
 // ============================================================================
 
-export const useTimeRangeTrackProps = (props: React.PropsWithChildren<{ [k: string]: any }>) => {
-  const { setTrackRef } = useTimeRangeRootContext();
+export const useTimeRangeTrackProps = (props: React.PropsWithChildren<{ [k: string]: any }>, context: any) => {
```
cjpillsbury, line 185: question(non-blocking) - is this "context" or the context's "value"?

packages/react/react/src/utils/component-factory.tsx
luwes: cool yes, that a nice optimization. I'll leave it for a next iteration.

packages/react/react/src/components/TimeRange.tsx
luwes: the context value 
`const context = React.useContext(Context);`

bad naming on my part

PR #23, opened 2025-09-24
feat: add core range, time and volume range

luwes:
This is a proof of concept for extracting a core class for both the React and HTML time range.
It's very rough without many things abstracted away.

Used a prefixed underscore for some state props so they can be minified by the bundler.

### DIFF
--- a/packages/core/core/src/components/range.ts
+++ b/packages/core/core/src/components/range.ts
@@ -55,7 +55,7 @@ export class Range {
     const state = this.#state.get();
 
     let _pointerWidth = 0;
-    if (state._hovering && state._pointerRatio !== null) {
+    if (state._hovering) {
       _pointerWidth = state._pointerRatio;
     }
 


packages/react/react/src/components/TimeRange.tsx
```diff
@@ -40,141 +20,69 @@ export const useTimeRangeRootState = (
...
 
   const methods = useMemo(() => timeRangeStateDefinition.createRequestMethods(mediaStore.dispatch), [mediaStore]);
 
   const { requestSeek } = methods;
-  const [pointerPosition, setPointerPosition] = useState<number | null>(null);
-  const [hovering, setHovering] = useState(false);
-  const [dragging, setDragging] = useState(false);
-  const [trackRef, setTrackRef] = useState<HTMLDivElement | null>(null);
+
+  const [_coreState, onCoreStateChange] = useState<CoreTimeRangeState | null>(null);
```
luwes: I used this to trigger re-renders, maybe there is a better way?

PR #72, opened 2025-10-21
docs: readmes v0

cjpillsbury:


eslint.config.mjs
```diff
@@ -47,4 +47,32 @@ export default antfu(
       'style/max-len': 'off',
     },
   },
+  {
+    files: ['**/*.md/**'],
```
cjpillsbury, line 51: quick fix for readmes only linting errors. It's likely we're too aggressive in our current setup, but this may also be too aggressive in the other direction.

PR #163, opened 2025-11-05
chore(site): add trademark notice to footer

heff:


site/src/components/Footer.astro
```diff
@@ -66,15 +66,14 @@ const { class: className } = Astro.props;
   </div>
 
   <p class="text-sm text-center sm:text-left text-dark-40 dark:text-light-40 px-3">
-    © 2010&ndash;{new Date().getFullYear()}, Video.js contributors
+    Video.js is a free and open source HTML5 <a href="/" class="intent:decoration-wavy">video player</a>.
```
decepulis, line 69: This doesn't actually render a decoration. You're missing the `underline`
```suggestion
    Video.js is a free and open source HTML5 <a href="/" class="underline intent:decoration-wavy">video player</a>.
```

decepulis: And also, did you mean to link back to the home page, here? I don't really understand why. If you didn't, here's this
```suggestion
    Video.js is a free and open source HTML5 video player.
```

site/src/components/Footer.astro
```diff
@@ -66,15 +66,14 @@ const { class: className } = Astro.props;
...
-    Video.js is a free and open source HTML5 video player framework
-    <br />
-    Sponsored by <a
+    &copy; 2010&ndash;{new Date().getFullYear()} Video.js contributors &nbsp;|&nbsp; Sponsored by <a
       href="https://www.mux.com?utm_source=videojs&utm_campaign=vjs10"
       class="underline intent:decoration-wavy"
     >
       Mux
     </a>
+    <br />"VIDEO.JS" is a registered trademark of Brightcove Inc.
```
decepulis: To my taste, this flows better as the second line, but that's such a not-big-deal I almost didn't mention it

decepulis: Also to my taste, VIDEO.JS is a bit heavy, but, I assume that's legally necessary

site/src/components/Footer.astro
heff: It's for SEO, assuming as the site evolves the footer will go to other pages that then link back to the homepage. And the link not being a critical one I left out the underline.

site/src/components/Footer.astro
heff: image

I could be ok with that.

site/src/components/Footer.astro
decepulis: There's already a link to the home page in the footer. I think we're covered!

site/src/components/Footer.astro
```diff
@@ -66,11 +66,10 @@ const { class: className } = Astro.props;
   </div>
 
   <p class="text-sm text-center sm:text-left text-dark-40 dark:text-light-40 px-3">
-    © 2010&ndash;{new Date().getFullYear()}, Video.js contributors
+    Video.js is a free and open source HTML5 <a href="/" class="intent:decoration-wavy">video player</a>.
+    <br />The term <strong>VIDEO.JS</strong> is a registered trademark of Brightcove Inc.
```
decepulis, line 70: `<strong />` looks a bit weird. Looks like it's applying `font-weight: 700` but we don't actually load a 700 weight for that font. So the browser's faking it.

You can fix that by 
1. changing this line to `<span class="font-semibold">VIDEO.JS</span>` 
2. or by adding `b, strong { font-weight: var(--font-weight-semibold); } ` to `site/src/styles/globals.css`
3. or by changing `weights: ['400 600'],` to `weights: ['400 700'],` in `site/astro.config.mjs`

I like option 2 the best, but, they're all valid!

decepulis: If you merge in main, you'll find that I already did `2`!

site/src/components/Footer.astro
heff: SEO works by using the keywords you want to rank for as link text. A significant piece of how search engines understand what your page is about is by what words other pages (even your own) use when linking to it. The current footer link uses "Home" as the link text, which isn't exactly a keyword we care about. And funnily enough is one of the most competitive keywords in the world because almost everyone uses it to point back to their home page.

There's a lot more/better we can do when we really care to optimize, but I try to find opportunities whenever I can.

site/src/components/Footer.astro
heff: Great, thanks! Rebased.

site/src/components/Footer.astro
decepulis: Huh yeah makes sense to me!

PR #175, opened 2025-11-07
fix(skins): slightly more idiomatic Tailwind, added custom properties

sampotts:
- Add `--vjs-border-radius` to all skins so that border radius can be baked in but controlled a little easier by consumers. 
- Add `--vjs-font-family` to HTML skins so we set a default font family. The assumption being that the Tailwind users would have a font stack set to inherit from. I'm open to discussion on this.
- Make the Tailwind _slightly_ more idiomatic. I think it's hard because we have the nested elements for tooltip content and icons. Perhaps this logic to hide/show should be elsewhere (e.g. via a hook) and not in the CSS. That's probably what I'd expect in React at least.


packages/react/src/skins/frosted/styles.ts
```diff
@@ -12,24 +12,25 @@ function cn(...classes: (string | undefined)[]): string {
 const styles: FrostedSkinStyles = {
   MediaContainer: cn(
     'vjs', // Scope preflight
-    'vjs:relative vjs:isolate vjs:@container/root vjs:group/root vjs:overflow-clip vjs:bg-black',
+    'vjs:relative vjs:isolate vjs:@container/root vjs:group/root vjs:overflow-clip vjs:bg-black vjs:rounded-(--vjs-border-radius,2rem)',
```
mihar-22, line 15: I think we  should [extend the theme](https://tailwindcss.com/docs/border-radius#customizing-your-theme) to support this value? Something like `--radius-chrome` which would be `vjs:rounded-chrome`? Alternative could be brand, theme, or vjs. Similar comment with font family and other config vars.

packages/react/src/skins/frosted/styles.ts
```diff
@@ -12,24 +12,25 @@ function cn(...classes: (string | undefined)[]): string {
...
-    // Ensure the nested video inherits the radius.
+    // Ensure the nested video inherits the radius
     'vjs:[&_video]:w-full vjs:[&_video]:h-full',
   ),
   Overlay: cn(
     'vjs:opacity-0 vjs:delay-500 vjs:duration-300 vjs:rounded-[inherit] vjs:absolute vjs:inset-0 vjs:pointer-events-none vjs:bg-gradient-to-t vjs:from-black/50 vjs:via-black/20 vjs:to-transparent vjs:transition-opacity vjs:backdrop-saturate-150 vjs:backdrop-brightness-90',
-    // Hide when playing (for now).
+    // ------------------------------------
     //  FIXME: This is crude temporary logic, we'll improve it later I guess with a [data-show-controls] attribute or something?
     'vjs:has-[+.controls_[data-paused]]:opacity-100 vjs:has-[+.controls_[data-paused]]:delay-0 vjs:has-[+.controls_[data-paused]]:duration-100',
```
mihar-22, line 30: Usage of `has` here is a code smell to me that we need a `Overlay` (maybe different name, Scrim?) component with it's own data attributes.

packages/react/src/skins/frosted/styles.ts
```diff
@@ -44,16 +45,20 @@ const styles: FrostedSkinStyles = {
...
+    // ------------------------------------
     'vjs:@container/controls vjs:absolute vjs:inset-x-3 vjs:bottom-3 vjs:rounded-full vjs:flex vjs:items-center vjs:p-1 vjs:gap-0.5 vjs:text-white',
     // Animation
     'vjs:transition vjs:will-change-[transform,scale,filter,opacity] vjs:origin-bottom vjs:ease-out',
-    //  FIXME: Temporary className hook for above logic in the overlay. Can be removed once have a proper way to handle controls visibility.
-    'controls',
+    // ------------------------------------
     //  FIXME: Temporary hide/show logic
     'vjs:scale-90 vjs:opacity-0 vjs:blur-sm vjs:delay-500 vjs:duration-300',
     'vjs:has-[[data-paused]]:scale-100 vjs:has-[[data-paused]]:opacity-100 vjs:has-[[data-paused]]:blur-none vjs:has-[[data-paused]]:delay-0 vjs:has-[[data-paused]]:duration-100',
```
mihar-22, line 58: Similar to overlay, code smell to me that we need a `Controls` component, which we actually do anyway.

packages/react/src/skins/frosted/styles.ts
```diff
@@ -76,8 +81,8 @@ const styles: FrostedSkinStyles = {
     'vjs:[&_.icon]:shrink-0 vjs:[&_.icon]:transition-opacity vjs:[&_.icon]:duration-150 vjs:[&_.icon]:ease-linear vjs:[&_.icon]:drop-shadow-[0_1px_0_var(--tw-shadow-color)] vjs:[&_.icon]:shadow-black/25',
   ),
   PlayButton: cn(
-    'vjs:[&_.pause-icon]:opacity-100 vjs:[&[data-paused]_.pause-icon]:opacity-0',
-    'vjs:[&_.play-icon]:opacity-0 vjs:[&[data-paused]_.play-icon]:opacity-100',
+    'vjs:[&_.pause-icon]:opacity-100 vjs:data-paused:[&_.pause-icon]:opacity-0',
```
mihar-22: I'd expect these to be set on the icon directly using `group/btn` or something. Similar comment for other buttons/icons.

packages/react/src/skins/frosted/styles.ts
```diff
@@ -89,17 +94,17 @@ const styles: FrostedSkinStyles = {
   PauseTooltip: cn('pause-tooltip'),
```
mihar-22, line 90: We should be able to remove these semantic classes now as the compiler won't need them since we discovered the design system loader. Similar comment with others like it in this file.

mihar-22 (commented):
Awesome, thank you! Only a few notes in addition to review:

1. Worth a discussion with the team on whether we'll use `media-` or `vjs-` prefix. We should be consistent with our custom element prefix. 
2. Can we drop `styles.ts` and move all styles directly into markup? Good step towards being more idiomatic, don't think `styles.ts` is a recognized pattern in TW-land. I understand we did it for early compiler goals and testing.
3. Can we drop manual addition of `vjs:` utility prefix? Create a new temporary utility in skin files which calls `cn` and adds prefix for us. Temporary because we'll have an official vanilla CSS import not generted with TW compiler as it is today. I'm assuming the TW compiler doesn't require this prefix to be statically analyzable when set?

packages/react/src/skins/frosted/styles.ts
cjpillsbury: Agree with the general sentiment. If MediaContainer had all properties by default, would that work?

sampotts: Yeah, I've anticipated all of this controls hide/show logic this is temporary. Having a dedicated component would be ideal. Backdrop might be another suitable name perhaps (aligns with the native `<dialog>` backdrop etc).

packages/react/src/skins/frosted/styles.ts
sampotts: Agreed. This is temporary logic again.

packages/react/src/skins/frosted/styles.ts
sampotts: I was aiming to avoid having folks extend or change their configuration ideally. It just creates a complexity that's a bit too much for some and then we get support requests asking why it doesn't look right. That said, it works for Shadcn but perhaps that's because it's a wider component library rather than a single component. Probably a discussion point.

packages/react/src/skins/frosted/styles.ts
sampotts: Yeah, we can possibly remove them later once we have the compiler but they're currently needed to hide/show the correct content. 

For me, the most idiomatic approach (in React, at least) for the tooltip content would be conditional rendering: 

```ts
{isPlaying ? t('PAUSE') : t('PLAY')}
```

(assuming we have a `t` function for i18n at some point).

packages/react/src/skins/frosted/styles.ts
sampotts: It also raises the question of how the compiler would handle it but perhaps we just have known aspects that we have to handle separately.

packages/react/src/skins/frosted/styles.ts
sampotts: Sure, that should be possible. I'll look into it.

packages/react/src/skins/frosted/styles.ts
mihar-22: Ye I think similar to shadcn we would just have these included either manually or codemod when the skin is added via CLI. Worth a discussion for sure. I think if we're building a design system as part of Tailwind then this is something we should do idiomatically and avoid littering vars inline. It's also something the compiler could/should extract from the config into vanilla CSS. Again, love to chat more about it :)

packages/react/src/skins/frosted/styles.ts
mihar-22: Ye we would need a dedicated component to avoid relying on parent state to style this elmenet which is breaking encapsulation.

packages/react/src/skins/frosted/styles.ts
mihar-22: Cool, worth a callout here in terms of not being idiomatic or breaking encapsulation. I wasn't expecting to fix in this PR. I'll have tickets up for these eventually.

packages/react/src/skins/frosted/styles.ts
mihar-22: Cool! I imagine few ways we can handle this:

1. Use the `render` function on the `PlayButton` to access state and render tooltip content accordingly inside. I don't mind this but it does feel funny with tooltip inside the button.
2. Use a hook like you've mentioned. We can just pluck state from media store or package up a more specific hook so devs don't need to guess which state they need. Imperative approach and more barebones.
3. `PlayTooltip` component and use data attrs for showing/hiding. Declarative approach and exposes styling hooks.

I like 3 because it's consitent with other tooltips structurally and we can expose additional data attributes for styling. Doesn't preclude us from having the hook as well. It's similar to how we have different versions of "toggle" buttons like play, mute, captions, etc. I originally wasn't a fan but the more I think about it, makes sense.

packages/react/src/skins/frosted/styles.ts
sampotts: Done.

Issue #209, opened 2025-11-29
Bug: Fullscreen not working on iOS Safari (latest version)


lifeofguenter:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which package(s) or projects are affected?

Core (@videojs/core)

### Browser/OS/Node environment

Latest iOS + Safari

### Description

On the demo site itself, when opening on iOS + Safari I am unable to resize the video to fullscreen.

### Steps to reproduce

1. Go to v10.videojs.org
2. Play video
3. Tap fullscreen icon

### Current Behavior

Nothing happens

### Expected Behavior

Video resized into fullscreen

### What version of the package are you using?

demo site

PR #211, opened 2025-12-01
fix(core): fixed fullscreen on ios safari

LachlanRumery:
This PR is aiming to fix the issue where fullscreen on safari does not work (issue #209). The issue related to requesting full screen on the container, but in IOS Safari you need to request full screen on the video element. 

Therefore, I added a check to determine if the user is on safari & IOS and if so request full screen on the video element.

```js
const isIOS = typeof navigator !== 'undefined' && /iPad|iPhone|iPod/.test(navigator.userAgent);
const isSafari = typeof navigator !== 'undefined' && navigator.userAgent.includes('Safari');

...


if (isIOS && isSafari) {
    // Safari (IOS Support)
    const videoElement = media?._playbackEngine?.element || media;
    (videoElement as any).webkitEnterFullscreen();
}
```

Let me know if you need me to make any changes or want me to explain something a bit better.

### DIFF
--- a/packages/core/src/store/mediators/fullscreenable.ts
+++ b/packages/core/src/store/mediators/fullscreenable.ts
@@ -66,7 +66,7 @@ export const fullscreenable = {
           // Enter fullscreen
           if (container.requestFullscreen) {
             container.requestFullscreen();
-          } else if (media._playbackEngine.element.webkitEnterFullscreen) {
+          } else if (media._playbackEngine?.element?.webkitEnterFullscreen) {
             // Safari support (IOS)
             media._playbackEngine.element.webkitEnterFullscreen();
           } else if (container.webkitRequestFullscreen) {


luwes:
Thanks for the contribution! This will need a few changes, I think we want to avoid user agent checks for this case and use feature detection checks. Similar to https://github.com/muxinc/media-chrome/blob/main/src/js/utils/fullscreen-api.ts

packages/core/src/store/mediators/fullscreenable.ts
```diff
@@ -58,16 +58,19 @@ export const fullscreenable = {
...
+      if (!container || !media || !globalThis?.document) return;
 
       try {
         if (value) {
           // Enter fullscreen
           if (container.requestFullscreen) {
             container.requestFullscreen();
+          } else if (media._playbackEngine.element.webkitEnterFullscreen) {
+            // Safari support (IOS)
+            media._playbackEngine.element.webkitEnterFullscreen();
```
luwes: Gonna approve and merge since it works but I'm not confident we want to access the _playbackEngine prop like that in this place. Just making a note.

luwes (approved):
Thanks for the contribution!

packages/core/src/store/mediators/fullscreenable.ts
```diff
@@ -58,16 +58,19 @@ export const fullscreenable = {
...
-      if (!container || !globalThis?.document) return;
+      const { container, media } = stateOwners;
+      if (!container || !media || !globalThis?.document) return;
 
       try {
         if (value) {
           // Enter fullscreen
           if (container.requestFullscreen) {
             container.requestFullscreen();
+          } else if (media._playbackEngine.element.webkitEnterFullscreen) {
```
luwes:
```suggestion
          } else if (media._playbackEngine?.element?.webkitEnterFullscreen) {
```

packages/core/src/store/mediators/fullscreenable.ts
luwes: Actually this doesn't seem right, I'm gonna do some more tests first.
Ideally it should be `media.webkitEnterFullscreen();`

luwes (changes requested):
I need to test this first.

PR #213, opened 2025-12-09
feat(compiler): in package source react skin -> html skin (module)

cjpillsbury:
Ref: #244 

# Overview
This PR is an incomplete implementation of our v0 goals of compiler, which is to:
Take a source file that defines an in-source/package VJS React Skin definition and produce an "appropriately equivalent" version of an in-source/package VJS HTML Skin definition (including styles).

## Example:
Given the in-source definition of react [`FrostedSkin`](https://github.com/videojs/v10/blob/main/packages/react/src/skins/frosted/FrostedSkin.tsx) (and, by association, its tailwind-based [styles module](https://github.com/videojs/v10/blob/main/packages/react/src/skins/frosted/styles.ts)), produce an "appropriately equivalent" HTML skin, roughly resembling the current HTML [`<media-skin-frosted>`](https://github.com/videojs/v10/blob/main/packages/html/src/skins/frosted/index.ts) (and, by association, its "vanilla css"[styles](https://github.com/videojs/v10/blob/main/packages/html/src/skins/frosted/styles.css)).

## Tasks
- [x] React -> HTML Skin Equivalent Module Compilation
- [ ] TS Module Tailwind -> Vanilla CSS Equivalent Module Compilation
- [ ] A basic CLI for use in the monorepo
  - [ ] reading from + writing to the file system

## High Level notes on complexity
- Needs to be resilient to changes in component architecture and implementation details (e.g. changes in popover/tooltips for HTML; changes in where/how styles are defined for HTML; etc)
- Needs to anticipate different planned use cases ("eject"/"template" use cases, different style definition preferences, "composed" components compilation; eventual additional target frameworks/libraries)


eslint.config.mjs
```diff
@@ -23,6 +23,7 @@ export default antfu(
     // https://github.com/antfu/eslint-config/blob/main/src/globs.ts#L56
     ignores: [
       '**/CLAUDE.md',
+      '**/ARCHITECTURE.md',
```
cjpillsbury, line 26: Low priority thing I'd like to discuss: antfu's linting rules for markdown-owned code blocks is too aggressive for common use cases - it requires syntactically correct code when often what you want is pseudo-code (code with `...` and the like), snippets that are technically syntactically invalid (`<ReactComponentExample>` that isn't e.g. in a return function or assigned to anything), or the like. Whether we convert this to a warning, remove the rule, or try to do something fancier, I think we should fix this so we don't "throw the baby out with the bathwater" on linting markdown.

mihar-22:
superseded by #1696

Issue #218, opened 2025-12-17
Media Store Foundation


mihar-22:
⚠️ Design is continously evolving, see #279 for latest store design that the Media Store will build on top of. The functions listed below are more generic now.

- [x] Store definition and creation (`createMediaStore`)
- [x] Type definitions (features, requests, state)
- [x] Feature slice support (`defineMediaFeature`, `createMediaFeature`)
- [x] Lifecycle Hooks (`onSetup`, `onMediaAttach`, `onError`)
- [x] Request Action API (`dispatch`, `dispatchAsync`, `isPending`, status)
- [x] Request Queue (media attached queue + media can play queue)
- [x] Media attach/detach lifecycle
- [x] Media interface pattern (`{ media: HTMLMediaElement }`, supporting multiple like `{ media: ..., hls: ... }`)
- [x] TypeScript cleanup (remove `any` types, proper inference)

## References

- https://github.com/videojs/v10/discussions/185
- [Notion: Media Store Design Doc](https://www.notion.so/mux/Video-js-Media-Store-2be97a7f89d080ec87c8e1f47485763f?source=copy_link)

PR #279, opened 2025-12-22
`@videojs/store`

mihar-22:
Ref: #218 

See [README](https://github.com/videojs/v10/blob/ra/videojs-store/packages/store/README.md) for overview. This is _not_ the Media Store, this is the generic store that it will build on.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -16,7 +16,7 @@
       "cache": false
     },
     "test": {
-      "outputs": ["coverage/**"]
+      "outputs": []
     },
     "lint": {
       "outputs": []


packages/store/README.md
```diff
@@ -4,9 +4,836 @@
...
+A reactive store for managing state owned by external systems. Built for media players, streaming libraries, and real-time systems where you don't own the state.
 
-Coming soon.
+```bash
+npm install @videojs/store
+```
+
+## Why?
+
+[Traditional state management](#how-its-different) assumes you own the state. But when working with a `<video>` element, Web Sockets, streaming libraries, and real-time systems, the external system is the authority. You observe it, send requests to it, and react to its changes.
```
heff, line 15: I like this this angle. 👍 Some providers that use this might still have their own state baked in? ex. the hypothetical playlist example

packages/store/README.md
```diff
@@ -4,9 +4,836 @@
...
+
+  // Async shorthand
+  async seek(time, { target, signal }) {
+    target.currentTime = time;
+    await onEvent(target, 'seeked', signal);
+  },
+
+  // Full config when needed
+  play: {
+    key: 'playback',
```
heff, line 126: What's the point of a "key"? I see you can flush it together. How would that be used?

packages/store/README.md
```diff
@@ -4,9 +4,836 @@
...
+    guard: canPlay,
```
heff: Mentioned in the other doc this is wrong here. Could go under seek.

packages/store/README.md
```diff
@@ -4,9 +4,836 @@
...
+
+## Multiple Targets
+
+Targets can contain multiple references. Useful when adapting to different systems.
+
+```ts
+interface Media {
+  media: HTMLMediaElement;
+  hls?: Hls;
+  dash?: DashJS;
```
heff: We don't need these, the media els will provide standard APIs for the features the UI needs. And even if there's a world where we add them for edge case needs, I don't love the story it's telling in this readme.

Is there a different example to use here for other "state owners", or does that change how this might be shaped?

packages/store/README.md
```diff
@@ -4,9 +4,836 @@
...
+
+```ts
+const qualitySlice = createSlice<Media, QualityState, QualityRequests>({
+  initialState: {
+    levels: [],
+    currentLevel: -1
+  },
+
+  getSnapshot: ({ target, initialState }) => {
+    if (target.hls) {
```
heff: Example of what we get to avoid with the standard media APIs.

packages/store/README.md
mihar-22: Good point. They don't have to coordinate at the store/slice level. Playlist data (`items`, `currentIndex`) and actions (`next()`, `prev()`) can live in a context provider, orchestrating by calling store requests (`loadSource`, `play`).

This keeps the store focused on media concerns. Playlist logic (UI concerns) and coordination lives in the provider.

That said, Playlist and Ads could end up as interesting hybrids. Hard to say until we get further along. We might find a natural point where there's enough media coordination to move some logic into a slice.

packages/store/README.md
mihar-22: They solve race conditions and coordinate requests. Ultimately, keys serve two purposes:

**1. Supersession** - Same key requests cancel previous ones:

```typescript
// User drags seek slider rapidly
store.request.seek(10);  // queued
store.request.seek(15);  // 10 cancelled
store.request.seek(20);  // 15 cancelled
// Only seek(20) executes
```

**2. Coordination** - Different requests can share a key:

```typescript
requests: {
  play:  { key: 'playback', handler: ... },
  pause: { key: 'playback', handler: ... },
}

store.request.play();   // queued
store.request.pause();  // play cancelled, pause queued
```

Play and pause are mutually exclusive. Same key ensures only the latest intent wins.

**Flushing by key** - Execute a specific queued request immediately:

```typescript
queue.flush('playback');  // Execute now, don't wait for scheduler
```

Useful if you scheduled with a delay but need to force it through (e.g., user clicks away, flush pending state).

packages/store/README.md
mihar-22: Updated examples!

However, I think it depends on the target. For `HTMLMediaElement`, a guard is optional since `play()` can trigger loading and handles its own rejections. For embeds (YouTube, Vimeo), you likely need a guard because calling methods before their SDK is ready can just fail.

I also think guards like this are nice because they push timing complexity into the store. Without one, UI components need to handle "is it ready?" and deal with rejections. With one, UI just calls `play()` and it works or gets cancelled cleanly.

The neat part is I assume we will have a consistent Media contract so each implementation can provide it's own `canPlay` check.

---

In our Notion doc, you had mentioned not being able to seek before metadata has loaded, here's how we solve it with our store :)

```ts
const hasMetadata: Guard = ({ target, signal }) => {
  if (target.readyState >= HAVE_METADATA) return true;
  return onEvent(target, 'loadedmetadata', signal);
};

requests: {
  seek: {
    guard: hasMetadata,
    handler(time, { target }) {
      target.currentTime = time;
    },
  },
}
```

User scrubs to 10, 20, 30 before metadata loads:

```markdown
seek(10) → queued, guard waiting
seek(20) → 10 cancelled, queued, guard waiting
seek(30) → 20 cancelled, queued, guard waiting
metadata loads → seek(30) executes
```

packages/store/README.md
mihar-22: Agreed! Maybe Ads? I'm not sure what it would look like right now so just removing. I imagine there's good reasons for multiple external state owners but ye...

packages/store/README.md
mihar-22: Yep, updated! Thank you.

heff:
I'm not arguing against guards, I like the idea of them in general, like for seeking. 👍

The term "canplay" already has meaning in web video playback, and doesn't mean "don't call play() until then". So unless you want to confuse the heck out of me, let's find a different term.

But that said, I hate "ready" events for underlying medias and have always pushed back on relying on them. We've already solved for this in the custom youtube/vimeo embed components and I wouldn't expect us to undo that at this point, so neither the UI or the store have to worry about this already.

heff:
I'm just thinking about the ecosystem we're building here, not just for ourselves but for plugin/extension authors. What pre-existing utilities will they have to work with so they don't have to require additional external dependencies for their extension. Are we saying:

* We don't have any state utils you can use for your own extension

* We have state utils if you're doing some specific thing

* We have state utils for whatever you're trying to do

I'm putting that out there to think about, not as a new requirement on this thing.

PR #300, opened 2026-01-08
[RFC] Player API Design

### LINKED ISSUE #301: Discovery: Player API Design
Design a unified API for Media and Container concerns with coordinated state management.

## Problem

Video players have two concerns that must coordinate but target different elements:

- **Media** (HTMLMediaElement): playback, volume, time
- **Container** (HTMLElement): fullscreen, idle, keyboard shortcuts

**Coordination examples:**
- Keyboard shortcuts on container → media actions
- Fullscreen prefers container, falls back to media on iOS
- Idle timer resets when media plays

## Need

Separate stores with explicit cross-store access pattern.

## Design Goals

- **Composability** — Media slices, container slices, and adapters compose without tight coupling
- **Tree-shaking** — Pay only for what you use
- **Extensibility** — Extend without forking
- **Progressive complexity** — From skin to fully custom

## Related

- Media API Design: #304 (split from this issue)

mihar-22:
closes #301

Adds planning document for the Player API Design — coordinating Media and Container concerns.

## Summary

- Dual-store coordination pattern for Media + Container concerns
- Factory that composes both stores
- Player slices with cross-store access to media

## Goals

- **Composability** — Build players by combining independent pieces
- **Tree-shaking** — Bundle only what you use
- **Extensibility** — Extend without forking
- **Progressive complexity** — From skin to fully custom

## Document

- `.claude/plans/player-api-design.md` ([view](https://github.com/videojs/v10/blob/media-arch-plans/.claude/plans/player-architecture.md))

---

This is an RFC for team review. See linked issue #301 for problem statement.

### DIFF
--- a/.claude/plans/player-api/examples.md
+++ b/.claude/plans/player-api/examples.md
@@ -239,9 +239,9 @@ customElements.define('my-container', MediaContainerElement);
 ```ts
 import { createMedia, features, VjsElement } from '@videojs/html';
 
-const { MediaMixin } = createMedia([features.playback, features.time]);
+const { ProviderMixin, MediaController } = createMedia([features.playback, features.time]);
 
-class VjsAudioController extends MediaMixin(VjsElement) {
+class VjsAudioController extends ProviderMixin(VjsElement) {
   // Programmatic control, no UI
 }
 
--- a/.claude/plans/player-api/index.md
+++ b/.claude/plans/player-api/index.md
@@ -120,15 +120,29 @@ const {
 
 ```ts
 const {
-  ProviderElement, // Ready-to-use element
-  ProviderMixin, // Store provider (Media + Player stores)
-  ContainerMixin, // Attaches container to player store
-  PlayerController, // Player state + requests (like usePlayer)
+  // Ready-to-use elements
+  ProviderElement, // Both stores, auto-attaches media
+  ContainerElement, // Attaches container to player store
 
-  // Escape hatches (advanced)
-  MediaProviderMixin, // Media store only
-  MediaController, // Media state + requests
+  // Mixins (customization)
+  ProviderMixin, // Both stores
+  ContainerMixin, // Container attachment
+  MediaProviderMixin, // Media store only (escape hatch)
+
+  // Controllers
+  PlayerController, // Player state + requests (like usePlayer)
+  MediaController, // Media state + requests (escape hatch)
 } = createPlayer(presets.website);
+```
+
+### Returns (HTML) — createMedia
+
+```ts
+const {
+  ProviderElement, // Ready-to-use element
+  ProviderMixin, // Customization
+  MediaController, // Media state + requests
+} = createMedia([features.playback, features.time]);
 ```
 
 ### usePlayer (React)


.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,620 @@
...
+**These concerns need to coordinate:**
+
+1. Keyboard shortcuts on container must trigger media actions (`Space` → `togglePlay`)
+2. Fullscreen prefers container but falls back to media on iOS
+3. Idle detection resets when media plays
+4. Gestures on container seek the media
+
+**We need:**
+
+- Separate stores for media and container (clean separation)
```
heff: With the container store still depending so heavily on the media store and coordinating, it feels like the "clean separation" isn't super real. While on the other side, if there's a single store for both it gives one clear place for devs and plugin authors to extend state, instead of navigating multiple stores. That'll sometimes be on top of a learning curve when trying to extend the UI and Media at the same time as the State.

You sure we don't just want one store for the player?

.claude/plans/media-provider-architecture.md
```diff
@@ -0,0 +1,913 @@
+# Media Provider Architecture
+
+Abstraction layer for consistent interaction with different media (native video, HLS, DASH, YouTube, Vimeo).
+
+## The Problem
+
+Different media have different APIs:
```
heff: Ok, interesting approach, but I hate this one as much as I thought I would, for all the reasons we already talked about. :) So let's chat IRL if this is still the direction you want to push for.

What you should focus on at this point is convincing me and others why the media chrome approach of standardizing Media APIs (at the media component) doesn't work for us long term. Or why your approach here is so much better. I'm not seeing an attempt to do that yet. In this line the term "media" is conflated to mean both the video element and engines, which then dodges the key argument that we need here, because in MC land all "medias" intentionally do have the same APIs.

When you're free I can walk you through the [media elements](https://github.com/muxinc/media-elements) and [media ui extensions](https://github.com/video-dev/media-ui-extensions) to give you any context you're missing.

mihar-22: Haha, yep. Based on our IRL chats, I knew you weren't initially going to like this! However, you're totally right. The "design decisions" section in the plans mentions bundle size and trade-offs with monolithic but doesn't actually make the argument 🤦

Let me unpack it briefly. Happy to chat more IRL and then bring our notes back here! Also, perfect place for us to get feedback from Christian as he's working on SPF and this is tightly related.

**TLDR:** This is related to SPF (composable streaming engine) and runtime performance. The goal is to support a truly modular and composable design whilst not sacrificing the easy path.

---

**Note:** This is advanced architecture. Most users will never see it — use-case presets that you've been designing handle common scenarios out of the box:

```ts
// Almost _all_ users will at most see something like this:
createPlayer(presets.backgroundVideo);
```

## SPF

SPF is composable at the **engine level**:

```ts
// Completely made up crude example
const engine = createEngine([
  audioTracks(),
  renditions(),
  thumbnails({ sprite: true }),
  // textTracks(), // excluded
  abr({ strategy: 'bandwidth' }),
]);
```

The adapter model lets player composition mirror engine composition:

```ts
// Engine: compose what you need
const engine = createEngine([audioTracks(), renditions()]);

// Player: mirrors the engine — no textTracks composed, no text slice
createPlayer(engine.slices);
```

You don't bundle what you didn't compose. TypeScript enforces the symmetry.

If the player layer instead exposed a monolithic contract that includes text track APIs, something has to implement it. SPF would then include text track code anyway (defeating composition).

## Runtime Performance

Runtime cost isn't just bundle size. Even with monolithic engines like hls.js where all capabilities ship anyway, there's runtime overhead. A monolithic player layer has to:

- Subscribe to all capability events (`LEVEL_SWITCHED`, `AUDIO_TRACK_SWITCHED`, `SUBTITLE_TRACK_SWITCH`, etc.)
- Sync state on every event
- Maintain request handlers and subscriptions for each capability
- Boilerplate for each capability (error handling, cleanup, etc.)

That happens on every `store.attach()`, whether or not the UI uses it. If all you have is a play button and progress bar, you're still paying for quality menu infra.

With adapters, you pay for what you use.

```ts
// Minimal — just playback
createPlayer([slices.playback]);

// Add quality when the UI needs it
createPlayer([slices.playback, slices.quality(engine.quality)]);
```

Composition is explicit, you wire up exactly what you need.

## The Simple Path Stays Simple

For users who want everything:

```ts
createPlayer(presets.from(engine));
```

`presets.from()` inspects the engine for capabilities and creates matching slices. Full capability in one line. Composition is the escape hatch.

```ts
presets.from(engine); // inspects the engine, returns store slices

// Sample engine contract which can be dynamically created and exposed
interface MediaEngineCapabilities {
  quality?: QualityAdapter;
  audio?: AudioAdapter;
  text?: TextAdapter;
  // ...
}
```

mihar-22:
**Note:** The code here at times is illustrative — APIs and specific names/designs will evolve as we implement, discuss, and iterate.

.claude/plans/player-architecture.md
mihar-22: ## Key Notes from Conversation

**Naming**
- Should we rename "container store" → **"player store"** — makes it the obvious place to extend?
- Player store has access to: media store + container DOM element
- Media store remains focused: just media target, clean boundary

**Use Case-Based Architecture**
```tsx
// Use case defines state/features, skin is just UI
<BackgroundVideoProvider>       {/* ← state/features */}
  <FrostedVideoSkin>            {/* ← UI (contains container internally) */}
    <Video src="..." />         {/* ← media */}
  </FrostedVideoSkin>
</BackgroundVideoProvider>
```

**Skins Get Simpler**
- Skin only exports UI + styling (no store config)
- Use case owns all player configuration
- "Ejecting" = just expanding the UI, no store dance

**Open Questions**
- Where does "in-between" functionality go? (not clearly media, not clearly UI). Do we have any such examples?
- Plugin author experience: how many concepts to learn?

## Two-Store Concerns (from Heff)

1. **Plugin author simplicity** — Extending the player shouldn't require navigating multiple stores or understanding which store owns what. One clear entry point for extension.

2. **Container ≠ Provider** — The container must remain purely UI. If container becomes a provider, we blur the clean separation:
   ```
   State (Provider) → UI (Container/Skin) → Media (Video)
   ```
   Container providing state breaks this mental model and makes skin ejection confusing.

**Implication for two-store design:**

If we keep two stores, the architecture must still present as:
```tsx
<PlayerProvider>        {/* ← state lives here (both stores) */}
  <Skin>                {/* ← UI only, no store creation */}
    <Video />           {/* ← media */}
  </Skin>
</PlayerProvider>
```

Container inside the skin just *attaches* to an existing store — it doesn't provide one. The "player store" naming helps: plugin authors extend the player store, container is just the DOM attachment point.

mihar-22: @heff I heard your feedback and went back to the drawing board to simplify.

**TLDR:** Rename "container" → "player" everywhere. `createPlayer` returns simple hooks. Two stores underneath, one API for users.

---

**Note:** Most users never see any of this. Use-case presets handle common scenarios:

```ts
// <Container> is out of convenience here, can be imported, or inside a skin
const { Provider, Container } = createPlayer(presets.backgroundVideo);
```

"Preset" is sounding really nice compared to "use-case config" or just "config" in general. Pre-built, ready to use.  

## Surface API

```ts
import { createPlayer, slices, presets } from '@videojs/react';

// Shorthand config (slices array)
const { Provider, Container, usePlayer, useMedia } = createPlayer(presets.all);

// Config object — extensible with future configuration options.
createPlayer({
  slices: [
    slices.playback,
    slices.fullscreen,
    // ...
  ],
});
```

```tsx
<Provider>
  <Container>
    <Video />
    <Controls />
  </Container>
</Provider>
```

```ts
// usePlayer overloads
const player = usePlayer();              // full store
const paused = usePlayer(s => s.paused); // selector
const play = usePlayer().request;        // request

player.state.paused;
player.request.play();

// Media escape hatch (rare)
const isMediaFullscreen = useMedia(s => s.isFullscreen);
```

## Exports

```ts
import { slices, presets, media, createPlayer, createMedia } from '@videojs/react';

// Slices — flat, type-discriminated
slices.playback
slices.fullscreen
slices.keyboard

// Presets — slice arrays for common use cases
presets.backgroundVideo   // playback, time, loop
presets.newsArticle       // autoplay, muted
presets.all               // everything

// Shorthand
createPlayer(presets.all);

// Config object — extensible
createPlayer({
  slices: presets.all,
  // future: devTools, middleware, etc.
});

// Extend preset
createPlayer({
  slices: [...presets.backgroundVideo, slices.keyboard],
});

// Advanced — direct media slices for standalone use
media.playback
media.fullscreen

createMedia([media.playback]);  // headless, no UI
```

## Naming

**`createPlayer` vs `createPlayerStore`**

"Store" is an internal concept. Users think "I'm creating a player" — stores are how we build it:

```ts
createPlayer([...]);      // what they're making
createPlayerSlice({...}); // store concept, shows up here
```

Progressive reveal:

| Level | Sees "store"? |
|-------|---------------|
| Use preset | No — `createBackgroundVideoPlayer()` |
| Custom player | No — `createPlayer([...])` |
| Use hooks | No — `usePlayer(s => s.paused)` |
| Write slice | Yes — `createPlayerSlice({...})` |

"Store" shows up when you're authoring slices. That's when targets, state, requests, subscriptions matter.

**Simplified from original:**

| Before | After |
|--------|-------|
| `useMediaSelector` | `useMedia` |
| `useContainerSelector` | `usePlayer` |
| `useMediaRequest` | `useMedia().request` |
| `useContainerRequest` | `usePlayer().request` |

## Extending

One place to extend: player slices.

```ts
// This can be typed to specific player target + media store
const myFeature = createPlayerSlice({
  initialState: { ... },
  request: {
    doThing: (_, { target }) => {
      target.container;              // container DOM element
      target.media.state.paused;     // read media
      target.media.request.play();   // control media (traced, queued)
    }
  }
});

interface PlayerTarget {
  media: Store<MediaTarget>;
  container: HTMLElement;
}
```

Media slices are rare — only if adding core media capability.

## Under the Hood

Still two stores, still separate mixins (advanced):

```
createPlayer()
├── createStore({ slices: [...media] })       → MediaMixin
└── createStore({ slices: [...container] })   → ContainerMixin
```

Slices have type discriminator:

```ts
interface MediaSlice { type: 'media'; ... }
interface PlayerSlice { type: 'player'; ... }
```

`createPlayer` filters, builds both stores, returns unified hooks.

## Why Two Stores (internally)

1. **Different targets:** Media slices target `MediaTarget`. Player slices target `PlayerTarget`. One store means every slice handles both, with null checks everywhere.

2. **Attachment timing:** `<Video>` and `<Container>` mount at different times, potentially different places in the tree. Two stores = two clean attachment points.

3. **Config dependency:** Player slices configure against typed media store. Media store exists first. Player slices depend on it.

4. **Observability:** Player slices interact with media *through the store*. Debugging, tracing, request queuing. `video.play()` directly bypasses all of that.

5. **Media without UI:** Headless player, audio-only, programmatic control. Media store+slices works standalone.

6. **Type-safe slice contracts:** Player slices can declare required media capabilities. TypeScript catches mismatched configs at compile time, not runtime.

heff: I'm liking the direction 👍 The concepts are feeling more easily accessible at a lower level.

> Users think "I'm creating a player" — stores are how we build it:

Yes but...we're also immediately teaching the user "_your_ player has three independent parts, State, UI and Media". So it's worth pointing out we're not being super consistent in our communication to the user when calling just the State piece the "player".
The term Player is always going to be complicated, and I still feel torn, but I can get behind the use of it here as the primary entry point.

"Media" I think we need to be more careful about. In isolation here it feels ok, but there's a whole third of the vjs ecosystem we're also calling "media", that is separate from all of this. So what `createMedia()` does won't be immediately clear. Feels to me like we should bring the term Store in here, like `createMediaStore` and `mediaStore`. It is a more internal and advanced concept in this case.

> `createPlayerSlice({...}); // store concept, shows up here`

Feels like we should call it `createPlayerStoreSlice`. More verbose, but I'm assuming "PlayerSlice" has no real meaning to anyone.

mihar-22: Heard! Documented concerns & decisions, updated plan with everything so far. Resolving conversation but feel free to open new comments/reviews. Thank you, heff :)

.claude/plans/media-provider-architecture.md
mihar-22: Moving to: https://github.com/videojs/v10/pull/305

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+Unified API for Media and Container concerns. Two stores internally, one API for users.
+
+## Quick Start
+
+### React
+
+```tsx
+import { createPlayer, presets } from '@videojs/react';
+
+const { Provider, Container, usePlayer } = createPlayer(presets.website);
```
heff: I'd feel a lot more comfortable prefixing these with "Player": `PlayerProvider` and `PlayerContainer`. Reads clearer (to me), less potential conflict, more symmetry with the HTML side.

luwes: I'm torn on this one. Could also do 

```tsx
const { usePlayer, ...Player } = createPlayer();` 

return <Player.Provider />;
```

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+### Returns (React)
+
+```ts
+const {
+  Provider, // Creates both stores
+  Container, // Attaches player store to element
+
+  usePlayer, // Player state + requests
+  useMedia, // Media state + requests (escape hatch)
+  useTasks, // Task state (loading, error, etc.)
```
luwes: What is this again?

If we have an error, isn't there an `error` prop in the store. 
`readyState` would indicate video states https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState

.claude/plans/player-architecture.md
heff: I can get behind that.

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+```
+
+### HTML
+
+```ts
+import { createPlayer, presets } from '@videojs/html';
+
+const { PlayerElement } = createPlayer(presets.website);
+
+PlayerElement.define('my-player');
```
heff: Does this do anything more than the standard web component registration?

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+```
+
+```html
+<my-player>
+  <video src="video.mp4"></video>
```
heff: The react version of this has a UI example. Can we write out what the same details would look like in HTML?

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+  return <button onClick={paused ? play : pause} />;
+}
+```
+
+### HTML
+
+```ts
+import { createPlayer, presets } from '@videojs/html';
+
+const { PlayerElement } = createPlayer(presets.website);
```
heff: I'm not feeling good calling this the PlayerElement. Especially with the UI-less example below, people will assume the UI is baked in. PlayerProviderElement would be the most accurate at this point.

heff: I'm also not sure this would be the primary example, when they could just import a `<website-player-provider>`.

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+
+When media element and fullscreen target need different DOM locations.
+
+```ts
+import { createPlayer, presets } from '@videojs/html';
+
+const { MediaMixin, ContainerMixin } = createPlayer(presets.website);
+
+class MediaProvider extends MediaMixin(LitElement) {}
+class MediaContainer extends ContainerMixin(LitElement) {}
```
heff: Related to another comment somwhere

ContainerMixin => Container ✅ 
MediaMixin  => ~~Media~~ Provider ❌

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+  PlayerElement, // Ready-to-use element with .define()
+  PlayerMixin, // Combined mixin for custom elements
+  MediaMixin, // Media store only (advanced)
+  ContainerMixin, // Player store only (advanced)
+
+  SelectorController, // Bound to player store
+  RequestController,
+  TasksController,
+
+  MediaController, // Bound to media store (escape hatch)
```
heff: Is this intended to be the imperative API for the media store?

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+### Returns (HTML)
+
+```ts
+const {
+  PlayerElement, // Ready-to-use element with .define()
+  PlayerMixin, // Combined mixin for custom elements
+  MediaMixin, // Media store only (advanced)
+  ContainerMixin, // Player store only (advanced)
+
+  SelectorController, // Bound to player store
```
heff: The term "SelectorController" sounds off to me. What are you aiming for this to be for users?

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+
+`createPlayer` filters slices by type, builds both stores, returns unified API.
+
+---
+
+## Skins and Presets
+
+**Skins are tied to presets.** Stores don't live in or extend from skins.
+
+Skins are organized under their preset:
```
heff: I had this question myself. Makes for long URLs but makes sense. 👍

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+
+```
+packages/react/src/
+└── presets/
```
heff: I think we could just drop "presets" as part of the URL, but that's not a strong opinion.

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+```tsx
+import { createPlayer, presets } from '@videojs/react';
+import { FrostedSkin } from '@videojs/react/presets/website';
+
+const { Provider, Container } = createPlayer(presets.streaming);
+
+<Provider>
+  <Container>
+    <Video />
+    <FrostedSkin /> {/* UI components only, no store */}
```
heff: The Skin needs to wrap the Media for layout reasons.

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+To customize, create your own player with a different preset:
+
+```tsx
+import { createPlayer, presets } from '@videojs/react';
+import { FrostedSkin } from '@videojs/react/presets/website';
+
+const { Provider, Container } = createPlayer(presets.streaming);
+
+<Provider>
+  <Container>
```
heff: We landed on the Container being inside the skin.

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,731 @@
...
+### HTML
+
+#### Declarative (Skin)
+
+```html
+<script type="module">
+  import '@videojs/html/presets/website/skins/frosted.js';
+</script>
+
+<vjs-frosted-skin>
```
heff: We still need a provider in this example.

.claude/plans/player-architecture.md
mihar-22: Convenience, alternative is `customElements.define('my-player', PlayerElement)`. Saves being verbose.

.claude/plans/player-architecture.md
mihar-22: I think with stuff like this just let people decide, it's simpler. Someone might do what Wes is doing which is neat, someone else might just do `{ Provider: PlayerProvider` if for some reason they need that clarity or need to avoid collision. No real need to be overly descriptive since it's being returned from `createPlayer` which signals what it  belongs to. You can also jump to definition in IDE and it'll take you there which I do alot.

.claude/plans/player-architecture.md
mihar-22: Skin is the provider here. This is where we landed in terms of keeping HTML feeling idiomatic (without logical provider elements). Any reversal in your mind there?

.claude/plans/player-architecture.md
mihar-22: You're suggesting dropping `Mixin`?

.claude/plans/player-architecture.md
mihar-22: Removed.

.claude/plans/player-architecture.md
mihar-22: Added

.claude/plans/player-architecture.md
mihar-22: Removed to move queue conversation.

.claude/plans/player-architecture.md
mihar-22: Changed to `PlayerController`. Examples included (check them out to see usage). 

These are a [`ReactiveController`](https://lit.dev/docs/composition/controllers/). They are a way for users to consume the player store in their own custom elements, same as React hooks enable on that side.

mihar-22: These are a [`ReactiveController`](https://lit.dev/docs/composition/controllers/) for context.

.claude/plans/player-architecture.md
mihar-22: Yep, I've added an example.

mihar-22: Mirrors `useMedia` on React side.

.claude/plans/player-architecture.md
mihar-22: 1. Updated, added simplified example just importing registration. 
2. Fair callout, renamed to `ProviderElement`, same argument above for returns.

```ts
const Player = createPlayerStore(presets.website);

Player.ProviderElement // if you prefer namespaced.
```

mihar-22:
@heff updated plan to better align, renamed `createPlayer` and `createMedia` to `createPlayerStore` and `createMediaStore`. All decisions around naming have been added to plan.

.claude/plans/player-architecture.md
mihar-22: `createPlayer()` -> `createPlayerStore()`

.claude/plans/player-architecture.md
mihar-22: [Explained here.](https://github.com/videojs/v10/pull/300#discussion_r2692805282)

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,778 @@
...
+        <Video src="video.mp4" />
+        <Controls />
+      </Container>
+    </Provider>
+  );
+}
+
+function Controls() {
+  const paused = usePlayer((s) => s.paused);
+  const { play, pause } = usePlayer().request;
```
decepulis: Without understanding underlying architecture, it's not initially obvious to me why we have two different patterns for state and requests. (I mean, I know you can also usePlayer().state, so, we have three patterns!)

I'd love to get an intuitive understanding of why we did it this way so that we can build an intuition in the docs.
Or, if it's not too late, maybe we could talk some more about APIs like `usePlayerRequest` and `usePlayerState`. I see they've been considered before, but, since I'm tuning in at the end, I'm not sure why they were discarded 😬

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,778 @@
...
+The Lit equivalent of `usePlayer`. Use in custom elements to access player state and requests.
+
+```ts
+import { createPlayerStore, presets, VjsElement } from '@videojs/html';
+
+const { PlayerController } = createPlayerStore(presets.website);
+
+class VjsPlayButton extends VjsElement {
+  // Selector subscribes — triggers requestUpdate() when paused changes
+  #paused = new PlayerController(this, (s) => s.paused);
```
luwes: `#paused = new PlayerController(this, (s) => s.paused);`

Is there an alternative? Something like:
Creating a new instance for each subscription doesn't feel right.

```ts
#player = new PlayerController(this);
#paused = this.#controller.observe((s) => s.paused)
```

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,778 @@
...
+```
+
+---
+
+## Surface API
+
+### createPlayerStore
+
+```ts
+// Shorthand — preset or slice array
```
littlespex: Is the name "slice" set in stone? If not, it would be helpful to have some explanation of what a "slice" is, and what it's a slice of. After reading the entire doc I get general idea that these are feature sets, but I was only able to draw that conclusion after carefully reading all of the code examples. Thinking of them as features made things click for me, with `presets` being pre-bundled collections of features, etc. Are "slices" a design pattern specific to this new architecture? Doing a search for `slice`, specifically in a JS/TS context, only brings up the references to `Array|String|TypedArray.prototype.slice`. If this is an established pattern, can we get some external links?

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,778 @@
...
+
+The Lit equivalent of `usePlayer`. Use in custom elements to access player state and requests.
+
+```ts
+import { createPlayerStore, presets, VjsElement } from '@videojs/html';
+
+const { PlayerController } = createPlayerStore(presets.website);
+
+class VjsPlayButton extends VjsElement {
+  // Selector subscribes — triggers requestUpdate() when paused changes
```
luwes: 👍

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,778 @@
...
+  }
+}
+```
+
+Both versions expose `.state` and `.request`. Only the selector version subscribes:
+
+```ts
+// No selector — read state, make requests, no subscription
+#player = new PlayerController(this);
+this.#player.state.paused;      // current value (may be stale if not subscribed)
```
luwes: I hope when accessing this with the whole prop chain `this.#player.state.paused` it is fresh.
Why would it not be?

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,778 @@
...
+#paused = new PlayerController(this, (s) => s.paused);
+this.#paused.value;             // selected value, always fresh
+this.#paused.state.paused;      // full state also available
+this.#paused.request.play();    // requests also available
+```
+
+| React                        | Lit                                         |
+| ---------------------------- | ------------------------------------------- |
+| `usePlayer()`                | `new PlayerController(this)`                |
+| `usePlayer(s => s.paused)`   | `new PlayerController(this, s => s.paused)` |
```
luwes: same concern, hope we can use a cached controller, call a method to subscribe.

.claude/plans/player-architecture.md
heff: Re: provider naming, I'm all for giving people options. Let's look at your example which has the vjs Provider as a child of `function App() {`. It obviously won't be the only provider our users use. I see then in their future "oh shoot, naming conflict". Why not just avoid that?

mihar-22: I hear you, but I strongly expect the function will be called in a dedicated module like `player.tsx` or `player-store.ts`. I can't see collisions being a real issue based on my own experience, and the workaround is simple, just rename on destructure (e.g., if you are throwing it in `App.tsx` for some reason).

Let's start concise and revisit if it comes up in feedback or docs, easy to refactor. Keeping it minimal makes the library feel simpler and more approachable.

.claude/plans/player-architecture.md
mihar-22: Welcome, @decepulis! 

## Why

I originally had hooks separated by concern, but after discussing with Heff, we decided to simplify.

One hook = less to think about.

With separate hooks, users have to decide: "Do I need state or request? Both? Which import?" With `usePlayer`, there's one answer to "how do I interact with the player?"

Zustand works this way: `useBearStore` gives you state and actions together, not `useBearState` and `useBearActions`. 

We aren't "flat" like Zustand but the separation in the return (`state`, `request`) is intentional atm — state = snapshot (read), request = method (write). Different mental models, but accessed from one place. You learn one hook, then discover the shape. 

Important to callout difference between `.state` and selector:

```ts
usePlayer().state // this doesn't trigger re-renders (static)
usePlayer(s => s.paused) // this does (reactive)
```

❓ That said, answering your question here does raise a question: could it just be flat? I do like `player.play()` over `request.play()`, leans more into users not having to know state vs. request. One answer is the split avoids naming collisions. State keys and request methods are separate objects on slices, so they can share names without conflict. Whether that's worth the extra nesting is worth discussing. 

## Conflicted

> I'm still torn between `usePlayer` and `usePlayerStore`. Leaning toward `usePlayerStore` for ecosystem familiarity (Zustand uses `useBearStore`, etc.), but there's a valid argument for concise — and the factory is already `createPlayerStore`, so "Store" is assumed.

## Questions

- Does one hook with multiple return shapes feel like too much to hold in your head at once? Would "flat" be more appealing `player = usePlayer()` and then `player.*` for state/requests?
- Is there a library or pattern you've used where the separate hooks felt clearer? What makes `usePlayerState` and `usePlayerStore` appealing to you?

.claude/plans/player-architecture.md
mihar-22: Ah, yes. I like it. 

I was thinking the following could help with that, but still sucks if you need both subscription and then request methods. 

```ts
new PlayerController(this, s => ({ paused: s.paused, ended: s.ended }));
```

`observe` sounds good, we could also do `.select` or `.watch` (align with [lit signals](https://lit.dev/docs/data/signals/)) naming:

```ts
#player = new PlayerController(this); // state, request, watch

// Watch
#paused = this.#player.watch(s => s.paused);

// Read
this.#paused.value;

// Remove selector from here
new PlayerController(this, s => s);
```

Could be `subscribe` as well to align with the store, but it looks funny and you usually expect a unsubscribe function returned, not value.

.claude/plans/player-architecture.md
mihar-22: Hey, Casey!

I've been having so many background conversations, and exploring all over that I forgot the obvious. First-time reader experience. Sorry about that.

I've gone ahead and added TOC, intro to those concepts you called out, and link to our store package. Let me know if I can help unpack anything.

👉  [Updated](https://github.com/videojs/v10/blob/media-arch-plans/.claude/plans/player-architecture.md)

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,821 @@
...
+### HTML
+
+```html
+<script type="module" src="@videojs/html/presets/website/skins/frosted.js"></script>
+
+<vjs-website-provider>
+  <vjs-frosted-skin>
+    <video src="video.mp4"></video>
+  </vjs-frosted-skin>
+</vjs-website-provider>
```
mihar-22: @heff, questions:

1. Since the import path makes it clear we're importing from the preset, should we consider just having `vjs-frosted-skin` without the provider?

2. **Issue:** related to my comment in Notion-ish... my concern is having minimal/frosted skins per preset will lead to a lot of confusion. In this case, how do we differentiate skin element names per preset? `vjs-website-frosted-skin` starting to sound silly to me.

I can see a world where there's just a single default skin per preset: `vjs-website-skin`.

What's your thoughts here?

mihar-22: Another argument in favour of simplicity is that Video.js and Plyr did really well because the default path was obvious. No decision required. No browsing skins. No feature comparisons. 

```html
<!-- Preset has both your player configuration and skin setup here -->
<script type="module" src="@videojs/html/presets/website.js"></script>

<vjs-website-skin>
  <video src="video.mp4"></video>
</vjs-website-skin>
```

mihar-22: One more thing I realized... if we allow multiple skins per preset, the store has to support the skin with the most features. So what does "minimal" actually mean? Minimal UI components? Feels confusing to me.

Preset, features, and skin should be the same thing. If you want streaming features with fewer controls, that's eject plus build your own. One preset, one skin, one bundle.

.claude/plans/player-architecture.md
mihar-22: I'm torn on this one Wes, because there's some real trade-offs.

💪  **Benefits of live state:**
- Single source of truth - Target IS the state, we're just reading it
- Always correct - Never lying about what the target has right now
- Forgives bad slice wiring - If subscribe misses an event, state still works (note: subscribers are now out of sync which is bad)
- Predictable - No wondering "is this fresh?" - it always is

❌  **Making state truly live has trade-offs:**
- Side-effectful reads - Reads should be pure. Live state that syncs makes reads have side effects. Footgun for debugging, testing, and reasoning about code.
- Calls `getSnapshot()` on every access (perf cost if used in render paths, harder to avoid in React).
- Need separate snapshot property for React (`useSyncExternalStore` needs reference stability). Two properties with different behaviors = confusing API
-  Inconsistency or race condition, subscriber callback receives cached state, but `store.state` returns live. Could see different values.
- Testing harder - Can't easily mock/control what state returns without controlling target

😐  **Current design (cached, event-driven) feels like right trade-off:**
- `HTMLMediaElement` fires events for property changes (`volumechange`, `pause`, etc.)
- Staleness window is non-existent between property change and event dispatch (event dispatch and store patch is sync)
- Simpler mental model, better perf

**Options:**
1. Keep current - trust event wiring
2. Add explicit getter or method for live reads when needed
3. Make state live + add snapshot for cached (confusing)

---

> `getSnapshot`: A function that returns a snapshot of the data in the store that’s needed by the component. While the store has not changed, repeated calls to `getSnapshot` must return the same value. If the store changes and the returned value is different (as compared by [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is)), React re-renders the component.

.claude/plans/player-architecture.md
mihar-22: Maybe let's discuss this one IRL first? Prob best I move this conversation to a separate RFC for Presets.

.claude/plans/player-architecture.md
decepulis: > I'm still torn between usePlayer and usePlayerStore

Honestly? You're not gonna get a strong opinion from me. I have a mild opinion towards `usePlayer` since I didn't encounter the concept of a store until I started working with Svelte, so, I don't want to assume that word is going to mean anything to a React user. But, like, we're already using the term in the factory. So. Idk. I'm not gonna be much help here.

> Does one hook with multiple return shapes feel like too much to hold in your head at once?
> Is there a library or pattern you've used where the separate hooks felt clearer? What makes usePlayerState and usePlayerStore appealing to you?

> [!NOTE]
> If folks can learn zustand, they can learn this pattern, too. This conversation is still helpful -- No matter what we decide, I'll want to explain this well in the docs -- but ultimately, these concerns are non-blocking for me.

My ideal world would be `const { paused, play } = usePlayer()`. But of course, that world doesn't exist. First of all, because it doesn't offer the option between reactive and non-reactive state. (And I'm not sure there's a clean way to offer that†.) And second of all, because of namespace collisions. 

Given those constraints, what would I rather see? I feel friction remembering 1. that there are three types of *things* you get from the store (unavoidable) and 2. the three syntaxes of the overloaded function (avoidable). 

One solution? Three functions. `usePlayerState`, `usePlayerCurrent`††, and `usePlayerRequest`. Sure, you still have to make a decision, but at least you don't have to remember syntax.

But you would still have to remember three function names. So... what if...
```
const { state, current, request } = usePlayer() // ††
const reactivePaused = state(s => s.paused)
```
Then, the three modes are unavoidable and self-documenting. 

Or we might encourage the shorthands
```
const play = usePlayer().request.play
const reactivePaused = usePlayer().state(s => s.paused)
const notReactivePaused = usePlayer().current.paused
```

It's still not _ideal_. It's still weird that `state()` is a function and `request` isn't. But. Whatever.

> <sup>† claude says this is possible with some proxy magic, like in [valtio](https://github.com/pmndrs/valtio?tab=readme-ov-file). I'm not familiar with js proxies, so, I haven't wrapped my mind around this yet. Nor do I have to, because the namespace collision concern still remains.</sup>

> <sup>†† is the non-reactive state more like a ref, in that it changes over time without triggering re-renders, or more like a snapshot, where it's read once at render-time? This matters in event handlers. If the value changes over time, we could call it `ref` or `current`. If it doesn't, maybe `snapshot`</sup>

.claude/plans/player-architecture.md
decepulis: Yeah the term "slice" caught me, too. Is there a term we could use that might have more intrinsic meaning? How do you think about the term "feature"?

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,821 @@
...
+```
+
+## Surface API
+
+### createPlayerStore
+
+```ts
+// Shorthand — preset or slice array
+createPlayerStore(presets.website);
+createPlayerStore([slices.playback, slices.fullscreen]);
```
decepulis: For what it's worth, I'd prefer
```suggestion
createPlayerStore({ slices: presets.website });
createPlayerStore({ slices: [slices.playback, slices.fullscreen] });
```
Just feels less-magic. When the time comes to expand, folks know what's what and where to go.
Weak preference, though. I see the value in shorthands, too.

sampotts: Looks like it's an option below.

.claude/plans/player-architecture.md
```diff
@@ -0,0 +1,821 @@
...
+#### Declarative (Skin)
+
+```tsx
+import { createPlayerStore, presets, Video } from '@videojs/react';
+import { FrostedSkin } from '@videojs/react/presets/website';
+
+const { Provider } = createPlayerStore(presets.website);
+
+<Provider>
+  <FrostedSkin>
```
sampotts: I'm guessing we'll want to have some logic in the skin to check for the presence of the required slices and warn if something's missing and fail silently. I'm sure you had something in another doc for this but I thought I'd raise it.

sampotts: I wonder if it's worth bundling the preset slices and related skins together to avoid that or would there ever be a scenario where we don't want to do that?

.claude/plans/player-architecture.md
decepulis: Yeah. To clarify, I'm proposing _not_ having a shorthand! It's more verbose, but also more instructional. Might be a worthwhile trade-off.

.claude/plans/player-architecture.md
decepulis: I asked the same question in an internal conversation! 

lol and then about an hour later, I answered my own question like this:

> What if the user wants to add more slices? Or what if the user wants custom controls outside the chrome of the player?

> Conceivably we could pass a config object to the skin so that users could add additional slices to address the former concern. We could have the skin detect if it’s in a provider (or pass some withoutProvider prop) for the latter concern. 

> But. Neither of those solutions feels particularly better than what we have here.

@heff and @mihar-22 agreed with my train of thought, there.

sampotts: Good point! My concern was really around a mismatch of slices vs UI but perhaps it's just a case of docs and warning users about it at run time.

.claude/plans/player-architecture.md
luwes: Cool, `watch()` sounds good

.claude/plans/player-architecture.md
luwes: ooh, yes I meant fresh in the sense the state is synced on all events. not always getting it live from the target.

.claude/plans/player-architecture.md
luwes: I'm pretty sure we don't want to structure skins per / under preset. The way we did it in the MC themes seems more natural. The skin might support a feature / state slice but it's not a requirement.

> I can see a world where there's just a single default skin per preset: vjs-website-skin.

Why are we coupling preset to skin so tightly?

mihar-22: Saw your question, Wes. Thank you! We can unpack this one IRL + I'll open a new RFC for presets. It's too much for this PR.

.claude/plans/player-architecture.md
mihar-22: Great callout! That's exactly what [RFC: Slice Accessor Design](https://github.com/videojs/v10/pull/307) was trying to solve. I was initially of the same mind, skins should own the slices since only they know the UI they're targeting.

However, after speaking with Heff about the larger problems we're trying to solve, the idea of designing skins to target a "preset" or specific "use case" started to make a lot more sense. The relationship would be 1 preset → many skins, which keeps store configuration out of the skin layer.

That said, I still want to leave this open for discussion. I'll be opening a new RFC focused on preset discovery — I spoke with Heff again over the weekend, so things are little clearer now. We can unpack it all there.

.claude/plans/player-architecture.md
mihar-22: Spent half a day on the weekend exploring [Valtio](https://github.com/pmndrs/valtio?tab=readme-ov-file) and various implementations, all because of the simple `usePlayer` example and tiny footnote lol. I really appreciate this comment because it made it clearer for me what type of DX we should be aiming for. 

I'll leave a final comment on this RFC before I merge with updated design/changes.

.claude/plans/player-architecture.md
heff: I talked with both Rahim and Wes about this Friday and at the time I was able to bring both around without changes here.

A key detail being I'm not suggesting there be a "frosted" skin for every preset. I wrote up a note at the bottom of the design doc outlining how I saw the actual skin options coming together.

PR #307, opened 2026-01-15
[RFC] Primitives API & Feature Access

### LINKED ISSUE #306: Discovery: Feature Accessor Design
## Problem

1. **Optional feature composition** — Stores are composed of features at runtime. Consumers can't know upfront whether a feature exists.

2. **Primitive UI components** — Primitives need specific features, but features are composed outside their scope.

3. **Cross-feature access** — Features need to read or react to other features' state.

mihar-22:
Closes #306

## Summary

Folded feature accessor design into the player-api RFC as `primitives.md`.

### DIFF
--- a/rfc/player-api/architecture.md
+++ b/rfc/player-api/architecture.md
@@ -127,18 +127,7 @@ const fullscreen = createPlayerFeature({
 
 ## Feature Registry
 
-Each store maintains a registry of its features as a `ReadonlyMap<symbol, AnyFeature>` keyed by `feature.id`.
-
-```ts
-store.features.has(playbackFeature.id); // boolean
-store.features.get(playbackFeature.id); // AnyFeature | undefined
-```
-
-Used by `hasFeature()` type guard for primitives. See [primitives.md](primitives.md).
-
-## Proxy Access
-
-Feature authors access media via the flat proxy on `target.media`. See [primitives.md](primitives.md) for `hasFeature()`, `getFeature()`, and `subscribe()` usage.
+Each store maintains `features: ReadonlyMap<symbol, AnyFeature>` keyed by `feature.id`. Used by `hasFeature()` — see [primitives.md](primitives.md).
 
 ## State Unification
 
--- a/rfc/player-api/decisions.md
+++ b/rfc/player-api/decisions.md
@@ -207,23 +207,13 @@ function DebugPanel() {
 
 ### Two Stores (Not One)
 
-**Decision:** Maintain two internal stores (Media Store + Player Store).
-
-**Rationale:**
-
-- **Different targets:** Media features target `HTMLMediaElement`. Player features target container element.
-- **Different attachment timing:** `<Video>` and `<Container>` mount at different times.
-- **Config dependency:** Player features configure against typed media store. Media store must exist first.
-- **Observability:** Player→media interactions go through store. Enables debugging, tracing, request queuing.
-- **Standalone media:** Headless player, audio-only, programmatic control. Media store works alone.
+**Decision:** Maintain two internal stores (Media Store + Player Store). See [architecture.md](architecture.md) for details.
 
 **Trade-off:** Two stores exist internally, but feature authors access media via `target.media` proxy — same flat API as components.
 
 ### Container ≠ Provider
 
 **Decision:** Container is purely UI attachment. Provider owns state.
-
-Container inside skin attaches to existing store — doesn't provide one.
 
 ## Validation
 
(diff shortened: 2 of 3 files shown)


rfc/feature-availability-design.md
```diff
@@ -78,14 +104,16 @@ Guards receive `{ target, signal }`, not state. Check capability on target direc
...
+  const { volume, volumeAvailability } = usePlayer().state;
 
   if (volumeAvailability === 'unsupported') return null;
   if (volumeAvailability === 'unavailable') return <Slider disabled />;
   return <Slider value={volume} />;
 }
 ```
 
+---
+
```
luwes, line 124: This feature availability doc looks good to me 👍

.claude/plans/slice-accessor-design.md
```diff
@@ -0,0 +1,645 @@
...
+### Required Slices
+
+Slice is necessary for functionality — throw if missing.
+
+**UI component:**
+
+```ts
+function PlayButton() {
+  const playback = store.getSlice(playbackSlice);
+  if (!playback) throw new StoreError('MISSING_SLICE');
```
luwes: I envisioned the slices to be a setup and store internal thing only. 
Is there a way to design the store that we don't have to work with slice specific API's in the UI components?

.claude/plans/slice-accessor-design.md
```diff
@@ -0,0 +1,645 @@
...
+### Optional Slices
+
+Slice enhances functionality — graceful no-op if missing.
+
+**UI component:**
+
+```ts
+function TimeSlider() {
+  const time = store.getSlice(timeSlice);
+  if (!time) throw new StoreError('MISSING_SLICE');
```
luwes:
```suggestion
  if (store.state.currentTime != null) throw new StoreError('MISSING_SLICE');
```

mihar-22:
@decepulis `useSlice` is a both an internal and advanced API for anyone creating their own slices or primitives. The specific problems it's solving are outlined at the top of `slice-accessor-design` doc. Probably 99% of users will never see it. For everyone else, `usePlayer` is all you need.

.claude/plans/slice-accessor-design.md
mihar-22: I'm not sure how we could do that because the slices are optional and configured outside the scope of our own primitives. We don't know if someone has properly configured the store to work with a certain UI component (e.g., volume slice -> volume slider). 

In the tech preview it was the same story, just with a bit more indirection via a [definition](https://github.com/videojs/v10/blob/main/packages/__tech-preview__/core/src/store/definitions/play-button.ts). This approach simplifies the API so you're accessing the slice directly and includes error handling.

mihar-22: It's still an internal API in the sense that we will only interact with it for building our primtive UI components or for cross-slice access. Users will only ever see it if there're building their own slices or distributing UI.

.claude/plans/slice-accessor-design.md
mihar-22: Sorry, I'm not understanding what this diff is suggesting.

.claude/plans/slice-accessor-design.md
luwes: Yes, I guess what bothers me a little bit is that we're adding more API's for this feature / slice abstraction.
I was thinking it would be more like we check for browser API's, check if the needed method / prop exists.

```js
const paused = store.state.paused;
if (paused === undefined) throw new StoreError('Missing paused');
```

.claude/plans/slice-accessor-design.md
luwes: checking if features exist based on prop existence instead of explicit slice / feature check.

mihar-22: Understood! Issue would be types wouldn't be available here to know what `currentTime` is.

mihar-22: I'll resolve this one because top comment is having similar discussion

mihar-22:
Thanks @luwes and @decepulis! Both of you calling it out made me go back to the drawing board and simplify the design. 

I hope both of you are happy with this design! I'm merging to keep us moving but please feel free to leave more comments for us to discuss. Happy to update any time.

### API surface

**Primitives (component authors):**

```ts
function PlayButton() {
  const player = usePlayer(); // UnknownPlayer
  
  // Type guard
  if (!hasFeature(player, features.playback)) {
    throwMissingFeature(features.playback, { displayName: 'PlayButton' });
  }
  
  // Typed here now
  return <button onClick={player.play}>{player.paused ? '▶' : '⏸'}</button>;
}

// Optional features
const volume = getFeature(player, features.volume);
volume.setVolume?.(0.5); // safe if missing
```

**Feature authors — same flat API:**

```ts
const fullscreen = createPlayerFeature({
  request: {
    enterFullscreen: (_, { target }) => {
      const mediaFS = getFeature(target.media, media.fullscreen);
      mediaFS.enterFullscreen?.(); // flat, not mediaFS.request.enterFullscreen()
    },
  },
});
```

### Key changes

- **`target.media` is now a flat proxy (`UnknownMedia`)** — feature authors use same API as components
- **Three utilities:** `hasFeature` (guard), `getFeature` (optional), `throwMissingFeature` (critical)
- **`StoreProxy<T>` interface** — generic base for all proxies
- **Folded into `rfc/player-api/primitives.md`**

PR #308, opened 2026-01-15
feat(store): sync queue

mihar-22:
Remove microtask batching from the Queue. Tasks now execute immediately when enqueued.

**Why:** Gesture-gated browser APIs (`play()`, `requestFullscreen()`, `requestPictureInPicture()`) require user gesture context. Microtask delays could lose that context, causing these operations to fail.

**What changed:**
- Tasks become `pending` synchronously (no queued intermediate state)
- Same-key supersession now aborts in-flight tasks instead of canceling queued ones
- Handlers should be abort-aware for proper cleanup

**Removed:** `#queued` record, `#scheduleFlush()`, `#flushAll()`, microtask batching

**Kept:** Task lifecycle (pending → success/error), abort signals, subscriptions, all public API

### DIFF
--- a/packages/store/src/core/tests/integration/store.test.ts
+++ b/packages/store/src/core/tests/integration/store.test.ts
@@ -357,7 +357,7 @@ describe('immediate execution', () => {
 
     expect(store.state.paused).toBe(true);
 
-    await store.request.play();
+    store.request.play();
 
     // Validates: handler ran → play() called → event fired → state synced
     expect(target.paused).toBe(false);


packages/store/src/core/tests/integration/store.test.ts
```diff
@@ -314,3 +324,68 @@ describe('state syncing', () => {
...
+      },
+    });
+
+    const store = createStore({ slices: [playbackSlice] });
+    const target = new MockMedia();
+    store.attach(target);
+
+    expect(store.state.paused).toBe(true);
+
+    await store.request.play();
```
luwes:
```suggestion
    store.request.play();
```

can we remove this `await`, just to verify the paused prop is updated sync

mihar-22: Good catch! Removed, tested, working :)

PR #323, opened 2026-01-22
feat: implement player API

mihar-22:
Implementing the Player API RFC (`rfc/player-api/`).

## Summary

- Two-store architecture (Media + Player)
- Proxy-based tracking for efficient subscriptions
- `createPlayer` / `createMedia` factories for React and HTML/Lit
- Primitives API (`hasFeature`, `getFeature`, etc.)

See `.claude/plans/player-api.md` for implementation plan.

PR #324, opened 2026-01-22
feat: add play button component

### LINKED ISSUE #265: UI: Play Button Component
Implement the PlayButton component for toggling media playback. Includes core interaction logic, DOM props builder, and React/HTML implementations.

luwes:
fix #265

This pull request introduces a comprehensive implementation plan for the new `PlayButton` UI component, which will serve as a model for future UI components across the project. The plan details a multi-layered approach, covering core logic, DOM integration, React and HTML custom element implementations, and accessibility/testing requirements. Additionally, a minor fix is made to the `commitlint` script in `package.json`.

Component architecture and implementation plan:

* Added a detailed multi-phase plan for the `PlayButton` component, outlining interfaces, implementation details, and file structure across the Core, DOM, React, and HTML packages. This plan establishes conventions for component state, props, event handling, accessibility, and testing. (.claude/plans/components/play-button.md)

Accessibility and testing standards:

* Defined accessibility requirements (ARIA roles, keyboard navigation, disabled state) and a shared testing strategy to ensure behavioral parity across React and HTML implementations. (.claude/plans/components/play-button.md)

Development workflow improvement:

* Fixed the `commitlint` script in `package.json` to remove the unused `$1` argument, ensuring consistent commit message linting. (package.json)

### DIFF
--- a/.claude/plans/components/play-button.md
+++ b/.claude/plans/components/play-button.md
@@ -1,0 +1,281 @@
+# PlayButton Component Plan
+
+**Issue:** [#265 UI: Play Button Component](https://github.com/videojs/v10/issues/265)
+**Status:** Ready for implementation (all blockers closed)
+
+## Summary
+
+Implement the PlayButton component for toggling media playback across three layers:
+1. **Core** (`@videojs/core`) - Runtime-agnostic interaction logic
+2. **DOM** (`@videojs/core/dom`) - DOM props builder and keyboard handling
+3. **React** (`@videojs/react`) - React component and hook
+4. **HTML** (`@videojs/html`) - Web Component custom element
+
+## Current State
+
+**Existing Infrastructure:**
+- `playbackFeature` exists in `packages/core/src/dom/store/features/playback.ts`
+- Store React bindings (`createStore`, `useSnapshot`, `useRequest`) complete
+- Store Lit bindings (controllers, mixins) complete
+- Tech-preview packages contain reference implementations
+
+**Gap:** No UI components exist in production packages. PlayButton establishes patterns for all future components.
+
+---
+
+## Phase 1: Core Layer (`@videojs/core`)
+
+**File:** `packages/core/src/core/components/play-button.ts`
+
+### Interfaces
+
+```ts
+interface PlayButtonProps {
+  disabled?: boolean;
+  label?: string;  // Custom aria-label override
+}
+
+interface PlayButtonCallbacks {
+  onToggle?: () => void;
+  onPlayRequest?: () => void;
+  onPauseRequest?: () => void;
+  onPlay?: () => void;
+  onPause?: () => void;
+  onError?: (error: Error) => void;
+}
+
+interface PlayButtonState {
+  paused: boolean;
+  ended: boolean;
+  label: string;  // Computed: "Play" | "Pause" | "Replay"
+}
+```
+
+### Implementation
+
+1. Implement `PlayButtonCore` class or factory:
+   - `constructor(options: PlayButtonCoreOptions)` - receives store, props, callbacks
+   - `toggle()` method - handles the request flow
+   - `getState()` - returns computed state
+
+2. Label computation:
+   ```ts
+   function getPlayButtonLabel(props: PlayButtonProps, state: { paused: boolean; ended: boolean }): string {
+     if (props.label) return props.label;
+     if (state.ended) return 'Replay';
+     return state.paused ? 'Play' : 'Pause';
+   }
+   ```
+
+3. Request flow:
+   ```
+   User clicks → onToggle → onPlayRequest/onPauseRequest → request.play/pause → onPlay/onPause/onError
+   ```
+
+---
+
+## Phase 2: DOM Layer (`@videojs/core/dom`)
+
+**File:** `packages/core/src/dom/components/play-button.ts`
+
+### Data Attributes
+
+```ts
+interface PlayButtonDataAttributes {
+  'data-paused'?: '';
+  'data-ended'?: '';
+  'data-disabled'?: '';
+}
+```
+
+### Props Builder
+
+```ts
+function getPlayButtonProps(params: {
+  props: PlayButtonProps;
+  state: PlayButtonState;
+  onToggle: () => void;
+}): PlayButtonDOMProps;
+```
+
+Output includes:
+- `role: 'button'`
+- `tabindex: 0` (for non-button elements)
+- `aria-label: state.label`
+- `data-paused`, `data-ended`, `data-disabled` (boolean attributes)
+- `onClick` handler
+- `onKeyDown` handler (Enter/Space)
+
+### Keyboard Handling
+
+- Enter and Space both activate
+- Prevent default for Space (avoids scroll)
+- Respect disabled state
+
+---
+
+## Phase 3: React Package (`@videojs/react`)
+
+**Files:**
+- `packages/react/src/components/play-button.tsx`
+- `packages/react/src/components/index.ts`
+
+### Hook
+
+```tsx
+function usePlayButton(options?: UsePlayButtonOptions): {
+  props: PlayButtonDOMProps;
+  state: PlayButtonState;
+};
+```
+
+### Component
+
+```tsx
+function PlayButton(props: PlayButtonProps): JSX.Element;
+
+namespace PlayButton {
+  export type Props = PlayButtonProps;
+}
+```
+
+Features:
+- `render` prop for custom rendering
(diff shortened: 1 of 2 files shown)


.claude/plans/components/play-button.md
```diff
@@ -0,0 +1,281 @@
...
+
+**File:** `packages/core/src/dom/components/play-button.ts`
+
+### Data Attributes
+
+```ts
+interface PlayButtonDataAttributes {
+  'data-paused'?: '';
+  'data-ended'?: '';
+  'data-disabled'?: '';
```
sampotts: Out of interest, why not use the `disabled` attribute?

.claude/plans/components/play-button.md
```diff
@@ -0,0 +1,281 @@
...
+}): PlayButtonDOMProps;
+```
+
+Output includes:
+- `role: 'button'`
+- `tabindex: 0` (for non-button elements)
+- `aria-label: state.label`
+- `data-paused`, `data-ended`, `data-disabled` (boolean attributes)
+- `onClick` handler
+- `onKeyDown` handler (Enter/Space)
```
sampotts, line 107: nit: Enter and space will trigger a click event when using a `<button>` so would we end up with two events?

.claude/plans/components/play-button.md
```diff
@@ -0,0 +1,281 @@
...
+function getPlayButtonProps(params: {
+  props: PlayButtonProps;
+  state: PlayButtonState;
+  onToggle: () => void;
+}): PlayButtonDOMProps;
+```
+
+Output includes:
+- `role: 'button'`
+- `tabindex: 0` (for non-button elements)
```
sampotts, line 102: I'm assuming we would _recommend_ folks use a `<button>` and our default component would use one?

.claude/plans/components/play-button.md
mihar-22: I see there's actually some more nuance here. This part in [Base UI](https://github.com/mui/base-ui/blob/master/packages/react/src/use-button/useButton.ts#L71-L78) does manage disabled but carefully. They're actively removing native disabled in composite widgets (toolbars, menus, tabs).

**Native disabled:**
- Removes from tab order entirely
- Unfocusable
- Screen reader might skip it completely

Based on what I'm gathering, user should be able to tab through and hear "play button, disabled" rather than it vanishing from keyboard navigation. So we would set `aria-disabled="true"` to provide a hint, and I guess `data-disabled` is for styling/consistency/docs.

.claude/plans/components/play-button.md
mihar-22: Trade off with the above on managing disabled ourselves is that we need to hook into [these events](https://github.com/mui/base-ui/blob/master/packages/react/src/use-button/useButton.ts#L86-L90C9) and manage disabled ourselves. 

If we agree on how we treat disabled, then we'll need to abstract this somehow for all buttons.

.claude/plans/components/play-button.md
```diff
@@ -0,0 +1,281 @@
...
+### Base ButtonElement
+
+- Shadow DOM with slot
+- Keyboard event handling (Enter/Space)
+- `handleEvent` pattern
+
+### PlayButtonElement
+
+- Uses `SnapshotController` for state
+- Uses `RequestController` for play/pause
```
mihar-22, line 165: I simplified some of the controllers/hooks [here](https://github.com/videojs/v10/pull/326), we'll just need Claude to re-run implementation plan. 

I imagine we have to wait for https://github.com/videojs/v10/pull/323 partially so we have feature accessors (e.g., `getFeature(store)`) before implementing this component. I'll create a PR tomorrow prob to get that smaller piece through so UI is unblocked.

.claude/plans/components/play-button.md
```diff
@@ -0,0 +1,281 @@
...
+- Dispatches callbacks as custom events: `toggle`, `playrequest`, `pauserequest`, `play`, `pause`, `error`
```
mihar-22: Events should be kebab-case (e.g., `play-request`).

.claude/plans/components/play-button.md
```diff
@@ -0,0 +1,281 @@
...
+
+### Registration
+
+```ts
+customElements.define('media-play-button', PlayButtonElement);
```
mihar-22, line 171: Worth discussing with the team to confirm what prefix we're going with. I'll need to update existing Claude plans and what not to match. 

**Few arguments come to mind for `vjs-` over `media-`:**
1. You don't end up with funny names like `media-frosted-video-skin` (media+video).
2. We used `media-` in Media Chrome and Vidstack. Prevent collision with either. Look distinct too.
3. `vjs` is iconic. I would recognize it instantly. It also popups when you inspect old player in class list.

Beyond that if the team decides `media-`, I'm down either way.

.claude/plans/components/play-button.md
```diff
@@ -0,0 +1,281 @@
...
+### Implementation
+
+1. Implement `PlayButtonCore` class or factory:
+   - `constructor(options: PlayButtonCoreOptions)` - receives store, props, callbacks
+   - `toggle()` method - handles the request flow
+   - `getState()` - returns computed state
+
+2. Label computation:
+   ```ts
+   function getPlayButtonLabel(props: PlayButtonProps, state: { paused: boolean; ended: boolean }): string {
```
mihar-22: Reuse existing PlayButtonState, not new type definition.

.claude/plans/components/play-button.md
```diff
@@ -0,0 +1,281 @@
...
+```
+
+### Props Builder
+
+```ts
+function getPlayButtonProps(params: {
+  props: PlayButtonProps;
+  state: PlayButtonState;
+  onToggle: () => void;
+}): PlayButtonDOMProps;
```
mihar-22, line 97: Wondering if we should go a step further to simplify and just have `getRootProps` on core component class. 

Similar to Zag.js, we can just define our own `ElementProps` in core (jsx), then map accordingly in html/react packages? Just a thought to co-locate more code and simplify.

.claude/plans/components/play-button.md
sampotts: That all makes sense! Good points. A related nit though; if we're going with `aria-disabled`, why don't we just use that for styling etc too?

luwes: Thanks for the feedback. Updated to remove `data-disabled` here. 
We can use `disabled` or `aria-disabled` to style.

.claude/plans/components/play-button.md
luwes: Good call, I think I still prefer `media-` for it's more native, generic feeling.

I'm wondering now if we could pull off prefix-less if there are 2 or more words in the name.
Why not just `play-button`, `time-slider`?

With scoped registries on the horizon.

https://caniuse.com/wf-scoped-custom-element-registries
https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry
https://github.com/justinfagnani/webcomponents/blob/scoped-registries/proposals/Scoped-Custom-Element-Registries.md

decepulis: I can only offer my opinion as a matter of taste. I don't really have any facts to bring to the table. But. Of the three options, I really like `vjs-`. 

To me, using `media-` on media chrome felt like a branding exercise. "Web components are native HTML. They're timeless. Free from the constraints of component libraries. The media components you wish the platform had built in. Let's use timeless, platform-seeming names, too."

Meanwhile, from a brand standpoint, video.js 10 feels like more its own thing, rather than a thing of the platform. It has a bright brand on its website. It has an eye-catching skin. It demands that you learn the provider, the skin, and the media source. Everything about it feels opinionated. That just feels very `vjs-` to me.

MIND YOU. I haven't done truly native HTML dev in almost a decade. I don't know what naming conventions are idiomatic in web components. What's more important to me than a brand expression is idiomatic web component behavior. Good DX. If `vjs-` stands in the way of that, I take my opinion back. (And, maybe in the near future, scoped registries with generic names will be good DX. So. That's another potential wrench in my spiel.)

.claude/plans/components/play-button.md
```diff
@@ -0,0 +1,281 @@
...
+  label?: string;  // Custom aria-label override
+}
+
+interface PlayButtonCallbacks {
+  onToggle?: () => void;
+  onPlayRequest?: () => void;
+  onPauseRequest?: () => void;
+  onPlay?: () => void;
+  onPause?: () => void;
+  onError?: (error: Error) => void;
```
heff, line 44: I can't think of a reason for error handling at this level. Is there a clear one? If not I suggest we drop things we don't know we need.

Issue #266, opened 2025-12-17
UI: Time Component


mihar-22:
Implement the Time component for displaying media time values (current time, duration, remaining). Includes core logic, DOM props builder, and React/HTML implementations.

PR #454, opened 2026-02-04
docs(design): add time component design

mihar-22:
Refs #266

## Summary

Design doc for the Time component — displays current time, duration, or remaining time.

### DIFF
--- a/internal/design/time-display.md
+++ b/internal/design/time-display.md
@@ -3,7 +3,7 @@ status: draft
 date: 2025-02-04
 ---
 
-# Time Component
+# Time Display
 
 Displays current time, duration, or remaining time.
 


internal/design/time-component.md
```diff
@@ -0,0 +1,443 @@
...
+
+Video players need to display time in various formats:
+
+1. **Current playback position** — updates continuously
+2. **Total duration** — static once loaded
+3. **Remaining time** — duration minus current, often with negative sign
+
+Requirements:
+
+- Formatting options: `1:30:45` (digital), `1h 30m` (narrow), etc.
```
heff: suggest: drop the `1h 30m` option for beta, or completely

I don't feel like I've seen it used.

internal/design/time-component.md
```diff
@@ -0,0 +1,443 @@
...
+# Time Component
+
+Displays current time, duration, or remaining time with formatting aligned to `Intl.DurationFormat`.
+
+## Problem
+
+Video players need to display time in various formats:
+
+1. **Current playback position** — updates continuously
+2. **Total duration** — static once loaded
```
heff: note: Static at least until the source changes, and ignoring live durations (which is fine for beta).

internal/design/time-display.md
```diff
@@ -0,0 +1,443 @@
...
+Video players need to display time in various formats:
+
+1. **Current playback position** — updates continuously
+2. **Total duration** — static once loaded
+3. **Remaining time** — duration minus current, often with negative sign
+
+Requirements:
+
+- Formatting options: `1:30:45` (digital), `1h 30m` (narrow), etc.
+- Conditional hour display (hide when duration < 1 hour)
```
heff, line 21: Side note, don't miss [these utils](https://github.com/muxinc/media-chrome/blob/main/src/js/utils/time.ts) when building. The digits logic has been used in multiple players now so we don't need to reinvent the wheel.

internal/design/time-display.md
```diff
@@ -0,0 +1,443 @@
...
+
+// Composed
+<Time.Group format="digital">
+  <Time.Value type="current" />
+  <Time.Separator />
+  <Time.Value type="duration" />
+</Time.Group>
+```
+
+### HTML
```
heff, line 43: issue: No html imports example

There's different implications than react imports so important to write them out and consider them

internal/design/time-component.md
```diff
@@ -0,0 +1,443 @@
...
+#### Props
+
+| Prop           | Type                               | Default     | Description                   |
+| -------------- | ---------------------------------- | ----------- | ----------------------------- |
+| `format`       | `'digital' \| 'short' \| 'narrow' \| 'long'` | `'digital'` | Display format preset.        |
+| `hours`        | `'numeric' \| '2-digit'`           | `'numeric'` | Hour formatting.              |
+| `minutes`      | `'numeric' \| '2-digit'`           | `'2-digit'` | Minute formatting.            |
+| `seconds`      | `'numeric' \| '2-digit'`           | `'2-digit'` | Second formatting.            |
+| `hoursDisplay` | `'auto' \| 'always'`               | `'auto'`    | When to show hours.           |
+| `fallback`     | `string`                           | `'--:--'`   | Text when time unknown.       |
```
heff: issue(not-blocking): I can't see how this fallback prop is gonna work with a composed component.

Feels like fallback could just be per Time component. Or the fallback default could just be `--`.

internal/design/time-component.md
```diff
@@ -0,0 +1,443 @@
...
+
+### Group
+
+Container for composed time displays. Provides formatting context to children.
+
+#### Props
+
+| Prop           | Type                               | Default     | Description                   |
+| -------------- | ---------------------------------- | ----------- | ----------------------------- |
+| `format`       | `'digital' \| 'short' \| 'narrow' \| 'long'` | `'digital'` | Display format preset.        |
```
heff: suggest: Rename "short" vs. "narrow" to make them more distinguishable

If they're official standards I can get behind it.

internal/design/time-component.md
```diff
@@ -0,0 +1,443 @@
...
+| `hours`        | `'numeric' \| '2-digit'`           | `'numeric'` | Hour formatting.              |
```
heff: suggest: "leading zero" as a boolean, instead of "numeric" vs "2-digit"

"numeric" isn't a very distinguishable name for a number thing

internal/design/time-component.md
```diff
@@ -0,0 +1,443 @@
...
+| `minutes`      | `'numeric' \| '2-digit'`           | `'2-digit'` | Minute formatting.            |
+| `seconds`      | `'numeric' \| '2-digit'`           | `'2-digit'` | Second formatting.            |
+| `hoursDisplay` | `'auto' \| 'always'`               | `'auto'`    | When to show hours.           |
+| `fallback`     | `string`                           | `'--:--'`   | Text when time unknown.       |
+| `as`           | `ElementType`                      | `'span'`    | Container element (React).    |
```
heff: question: I thought we weren't doing `as` in favor of `render`?

internal/design/time-component.md
```diff
@@ -0,0 +1,443 @@
...
+
+| Prop          | Type                               | Default     | Description                    |
+| ------------- | ---------------------------------- | ----------- | ------------------------------ |
+| `type`        | `'current' \| 'duration' \| 'remaining'` | `'current'` | Which time to display.         |
+| `format`      | `'digital' \| 'short' \| 'narrow' \| 'long'` | Inherited | Display format (overrides Group). |
+| `hours`       | `'numeric' \| '2-digit'`           | Inherited   | Hour formatting.               |
+| `minutes`     | `'numeric' \| '2-digit'`           | Inherited   | Minute formatting.             |
+| `seconds`     | `'numeric' \| '2-digit'`           | Inherited   | Second formatting.             |
+| `hoursDisplay`| `'auto' \| 'always'`               | Inherited   | When to show hours.            |
+| `signDisplay` | `'auto' \| 'always' \| 'never'`    | `'auto'`    | Negative sign for remaining.   |
```
heff: suggest: `negativeSign` instead of `signDisplay`

Just feels like the setting should have "negative" in the name somewhere, unless it actually applies to more things.

internal/design/time-component.md
```diff
@@ -0,0 +1,443 @@
...
+
+```html
+<span aria-hidden="true">/</span>
+```
+
+## Formatting
+
+### Format Presets
+
+Aligned with `Intl.DurationFormat` style options:
```
heff: suggest: Don't align with this, reduce the options

At least two of these feel unnecessary, especially for Beta. The way I see it is we need
* One for display in the UI (colon-separate, no leading zero, hours place dependent on duration or preference)
* One for accessibility? Screen readers still can't handle digital time display?

internal/design/time-component.md
```diff
@@ -0,0 +1,443 @@
...
+| `hours`       | `'numeric' \| '2-digit'`           | Inherited   | Hour formatting.               |
+| `minutes`     | `'numeric' \| '2-digit'`           | Inherited   | Minute formatting.             |
+| `seconds`     | `'numeric' \| '2-digit'`           | Inherited   | Second formatting.             |
+| `hoursDisplay`| `'auto' \| 'always'`               | Inherited   | When to show hours.            |
+| `signDisplay` | `'auto' \| 'always' \| 'never'`    | `'auto'`    | Negative sign for remaining.   |
+| `fallback`    | `string`                           | Inherited   | Text when time unknown.        |
+| `label`       | `string`                           | Auto        | Custom aria-label.             |
+| `render`      | `ReactElement \| (props, state) => ReactElement` | — | Custom render element. |
+
+**Note:** When used standalone (outside Group), formatting props default to: `format="digital"`, `hours="numeric"`, `minutes="2-digit"`, `seconds="2-digit"`, `hoursDisplay="auto"`, `fallback="--:--"`.
```
heff: > minutes="2-digit

suggest: don't pad minutes with the leading zero by default if it's less than 10 minutes

> "The duration is 01:30"
> "The duration is 1:30"

Neither is great in a sentence. I think this format will only be used in UI, where I've always defaulted to no leading zero for space reasons.

internal/design/time-display.md
```diff
@@ -0,0 +1,443 @@
...
+| `data-type`  | `current`, `duration`, or `remaining`.   |
+
+#### Renders
+
+```html
+<time datetime="PT1M30S" aria-label="Current time">1:30</time>
+
+<!-- Remaining with negative sign -->
+<time datetime="PT4M30S" aria-label="Remaining">
+  <span aria-hidden="true">-</span>4:30
```
heff, line 182: question: In HTML are we sprouting this in the light dom?

internal/design/time-component.md
```diff
@@ -0,0 +1,443 @@
...
+### React
+
+```tsx
+import { Time } from '@videojs/react';
+
+// Standalone
+<Time.Value type="current" />
+
+// Composed
+<Time.Group format="digital">
```
heff: thought: I'm not sure there's an answer, but I'm missing the clarity in MC that every component can just be dropped in the control bar and it works. Maybe that can still be true with this Group example, it's just not clear yet.

heff (commented):
I like a lot of this. I Iike that it makes the time separator a clearly composed piece. That was always awkward thing in MC.

At the same time I'm seeing a few key features in here that neither Video.js or MC have ever actually needed, and that make some of the settings interface a degree more complicated, so it feels like too much for beta at least and may prevent us from re-using some of the code we've already solved this with in other places.

internal/design/time-component.md
```diff
@@ -0,0 +1,443 @@
...
+
+1. **Current playback position** — updates continuously
+2. **Total duration** — static once loaded
+3. **Remaining time** — duration minus current, often with negative sign
+
+Requirements:
+
+- Formatting options: `1:30:45` (digital), `1h 30m` (narrow), etc.
+- Conditional hour display (hide when duration < 1 hour)
+- Proper accessibility (screen readers need context, not just "1:30")
```
luwes: 👍

internal/design/time-display.md
```diff
@@ -0,0 +1,443 @@
...
+### HTML
+
+```html
+<!-- Standalone -->
+<media-time type="current"></media-time>
+
+<!-- Composed -->
+<media-time-group format="digital">
+  <media-time type="current"></media-time>
+  <media-time-separator></media-time-separator>
```
luwes, line 56: (non-blocking): what's the reason for making this a custom element instead of a simple character?

internal/design/time-component.md
```diff
@@ -0,0 +1,443 @@
...
+
+#### Props
+
+| Prop           | Type                               | Default     | Description                   |
+| -------------- | ---------------------------------- | ----------- | ----------------------------- |
+| `format`       | `'digital' \| 'short' \| 'narrow' \| 'long'` | `'digital'` | Display format preset.        |
+| `hours`        | `'numeric' \| '2-digit'`           | `'numeric'` | Hour formatting.              |
+| `minutes`      | `'numeric' \| '2-digit'`           | `'2-digit'` | Minute formatting.            |
+| `seconds`      | `'numeric' \| '2-digit'`           | `'2-digit'` | Second formatting.            |
+| `hoursDisplay` | `'auto' \| 'always'`               | `'auto'`    | When to show hours.           |
```
luwes: YAGNI (🪵): until this is a user requested feature feels safe to leave off.

internal/design/time-component.md
luwes: Feels like a lot of options to start out with. Maybe start out with 2? One for showing, other for accessibility.

internal/design/time-display.md
mihar-22: The negative symbol `span`? Yes it would.  It's internal to the value (semantically part of time, not compositional). Similar to how Media Chrome handles it?

mihar-22: If you mean `<time>` then no, that would only be on the React side. HTML now is aligned with Media Chrome.

internal/design/time-display.md
mihar-22: Noted in doc! Ty.

internal/design/time-display.md
mihar-22: Added!

internal/design/time-display.md
mihar-22: This is an open design question! Small here but important because it'll likely determine how we approach it elsewhere in our library.

**Pros:**
| Benefit | Description |
|---------|-------------|
| Consistency | Fits the component model — composable with Group and Value |
| Automatic a11y | Always applies `aria-hidden="true"` — prevents mistakes |
| Styling | Can be targeted with CSS (`media-time-separator { ... })`, default spacing |
| Discoverable | Part of the API surface — users know it exists, API reference generated |

**Cons:**
| Drawback | Description |
|----------|-------------|
| Overhead | Custom element for a single "/" character feels heavy |
| Users may ignore it | They might just use <span>/</span> anyway |
| Registration cost | Another element to define/register |
| Overkill? | Could just document "use a span with aria-hidden" |

---

I'm leaning towards keeping it as a custom element for DX. Arguments for feel stronger than against. Implementation cost is very low. What do you think?

internal/design/time-component.md
mihar-22: Done, removed and added smarter defaults.

internal/design/time-component.md
mihar-22: Removed for now! Sticking with digital as default.

internal/design/time-component.md
mihar-22: I think you're pointing at the drop-in experience. If I'm understanding correctly, for us that experience has shifted to docs and skins. 

We just need to be mindful of this and make it easy to control, either by:
- Baking into examples CSS variables for key sizing (padding, height, border-radius)
- Or, simply building good examples so users can extract into vars as needed

mihar-22: That said, `<media-time-group>` does give a clear styling target if you want it to be a consistent "box" alongside other controls.

internal/design/time-component.md
mihar-22: Done

internal/design/time-component.md
mihar-22: Done

internal/design/time-component.md
mihar-22: Done, dropped per-unit formatting props in favor of smart defaults (hours never padded, minutes padded when hours shown, seconds always padded). No numeric vs 2-digit choice needed.

Can revisit if users ask for more control.

internal/design/time-display.md
mihar-22: [Noted here](https://github.com/videojs/v10/pull/454/changes#diff-a695dbf7fd9f8dcd469e50933a240021464277930a937ebb5d5e1717c557bb69R329-R333)

internal/design/time-component.md
mihar-22: Removed all except `hoursDisplay`

internal/design/time-display.md
```diff
@@ -0,0 +1,521 @@
...
+
+## Anatomy
+
+### React
+
+```tsx
+import { Time } from '@videojs/react';
+
+// Standalone
+<Time.Value type="current" />
```
heff, line 33: praise: I do really like this new feature of an easily accessible value that can be used anywhere 👍

internal/design/time-display.md
```diff
@@ -0,0 +1,521 @@
...
+  <Time.Value type="current" />
+  <Time.Separator />
+  <Time.Value type="duration" />
+</Time.Group>
+```
+
+### HTML
+
+```ts
+import '@videojs/html/ui/time';
```
heff, line 46: thought(not-blocking): This is gonna be less minifiable because every custom el is gonna get registered. But probably ok in this case because we expect almost everyone to use all the pieces. Something to keep in mind with web components at least.

internal/design/time-component.md
```diff
@@ -0,0 +1,521 @@
+---
+status: draft
+date: 2025-02-04
+---
+
+# Time Component
```
heff: suggest(blocking): Since this is UI focused and the primary use case will be the full UI component, can we call it Time Display (like Media Chrome)? "Component" is vague, and Time is a concept.

ex. The name "Time Component" wouldn't fit well in the BaseUI list of components, which are all clear purposeful nouns.

Describing as blocking because it's setting a pattern for future UI components names.

internal/design/time-display.md
heff: thought(not-blocking): The background concern here is just for other frameworks that still might not like sprouting HTML. I'm generally ok to try it out either way.

heff (commented):
Looking cleaner, thanks for the updates.

internal/design/time-display.md
mihar-22: Bundlers will sort that part out but we will need optimized CDN bundles!

internal/design/time-component.md
mihar-22: done!

internal/design/time-display.md
mihar-22: Cool, we can test it. Historically it's okay as long as it's a tick or so after frameworks finish hydrating. See how it goes.

internal/design/time-display.md
mihar-22: I guess we're landing on keeping it for reasons listed in doc and pros here, but if you have any thoughts on this or change of heart let me know :)

internal/design/time-component.md
mihar-22: Removed `hoursDisplay` too now, all gone

Issue #237, opened 2025-12-17
Controls Feature


mihar-22:
Blocked by:
- #218
- #229
- #230

PR #456, opened 2026-02-05
docs(design): controls

mihar-22:
ref: https://github.com/videojs/v10/issues/237

## Summary

Design document for the Controls component and `controlsFeature` for Video.js 10.

- Defines user activity tracking and controls visibility management
- Specifies `<media-controls>` and `<media-controls-group>` components
- Documents feature API (`userActive`, `controlsVisible`, `setUserIdleTimeout`)
- Covers accessibility (conditional `role="group"`), default styles, and behavior
- Lists descoped features with rationale (lock API, CSS variables, toolbar navigation)

## References

- Researched Media Chrome, Vidstack, Base UI, and Radix for patterns
- Aligned with WAI-ARIA group role and toolbar pattern guidelines

### DIFF
--- a/internal/design/ui/controls-design.md
+++ b/internal/design/ui/controls-design.md
@@ -24,7 +24,7 @@ Split into **feature** (state management) and **component** (UI):
 |---------|----------|----------------|
 | Activity tracking | `controlsFeature` | Pointer/keyboard events, idle timer (internal) |
 | Visibility computation | `controlsFeature` | `controlsVisible = userActive \|\| paused` |
-| Auto-hide timing | `<media-controls>` | Configures idle timeout via `hide-delay` |
+| Auto-hide timing | `controlsFeature` | Idle timer, configurable delay |
 | Layout | `<media-controls-group>` | Visual grouping of controls |
 
 ## Quick Start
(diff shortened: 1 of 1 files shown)


internal/design/ui/controls-design.md
```diff
@@ -0,0 +1,449 @@
...
+}
+```
+
+Activity tracking and input type detection are handled internally by the feature.
+
+## Component API
+
+### `<media-controls>`
+
+Container for player controls. Manages visibility based on feature state.
```
heff, line 130: issue(blocking): I'm finding the relationship between Container, media-controls, and controls-group confusing.

internal/design/ui/controls-design.md
```diff
@@ -0,0 +1,449 @@
...
+| Attribute | Description |
+|-----------|-------------|
+| `[data-visible]` | Present when controls should be visible |
+
+### `<media-controls-group>`
+
+Visual grouping container. No special behavior — pure layout.
+
+```html
+<media-controls-group>
```
heff, line 143: suggest: If this is the main Control Bar I think we should call it that. It's a well known concept, will make styling easier, and can provide some specific benefits otherwise.

PR #457, opened 2026-02-05
feat(core): add poster component

### LINKED ISSUE #240: UI: Poster Component
Implement the Poster component for displaying a preview image before media playback. Supports CSS-based enter/exit animations. Includes core logic, DOM props builder, and React/HTML implementations.

## References

- [Vidstack: Poster](https://www.vidstack.io/docs/player/components/display/poster)
- [Media Chrome: Poster Image](https://www.media-chrome.org/docs/en/components/media-poster-image)

mihar-22:
Closes #240

## Summary

Add `media-poster` web component and `Poster` React component for displaying video poster images. Shows before playback starts, hides after via `data-visible` attribute.

### Usage

**HTML:**
```html
<media-poster>
  <img src="poster.jpg" alt="Video description" />
</media-poster>
```

**React:**
```tsx
<Poster src="poster.jpg" alt="Video description" />
```

**CSS:**
```css
media-poster {
  position: absolute;
  inset: 0;
  pointer-events: none;
}

media-poster:not([data-visible]) {
  display: none;
}

media-poster img {
  width: 100%;
  height: 100%;
  object-fit: contain;
}
```

### DIFF
--- a/packages/core/src/core/ui/poster/poster-core.ts
+++ b/packages/core/src/core/ui/poster/poster-core.ts
@@ -5,9 +5,9 @@ export interface PosterState {
 }
 
 export class PosterCore {
-  getState(playback: MediaPlaybackState): PosterState {
+  getState(media: MediaPlaybackState): PosterState {
     return {
-      visible: !playback.started,
+      visible: !media.started,
     };
   }
 }
--- a/packages/core/src/core/ui/poster/tests/poster-core.test.ts
+++ b/packages/core/src/core/ui/poster/tests/poster-core.test.ts
@@ -3,13 +3,13 @@ import { describe, expect, it, vi } from 'vitest';
 import type { MediaPlaybackState } from '../../../media/state';
 import { PosterCore } from '../poster-core';
 
-function createMockPlayback(overrides: Partial<MediaPlaybackState> = {}): MediaPlaybackState {
+function createMediaState(overrides: Partial<MediaPlaybackState> = {}): MediaPlaybackState {
   return {
     paused: true,
     ended: false,
     started: false,
     waiting: false,
-    play: vi.fn().mockResolvedValue(undefined),
+    play: vi.fn(async () => {}),
     pause: vi.fn(),
     ...overrides,
   };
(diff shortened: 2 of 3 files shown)


internal/design/ui/poster.md
```diff
@@ -0,0 +1,208 @@
...
+We want a simpler approach: expose minimal state, let the user control the image.
+
+## Solution
+
+Minimal components that:
+
+1. Expose `data-visible` for CSS-based show/hide
+2. Do nothing else
+
+**HTML:** Wrapper element that accepts `<img>` as child.
```
heff, line 32: question: Or `<picture>`?

internal/design/ui/poster.md
```diff
@@ -0,0 +1,208 @@
...
+**React:** Renders `<img>` directly — no wrapper needed.
+
+### Usage
+
+#### HTML
+
+```html
+<media-poster>
```
heff, line 40: praise: I think this is better on the html side. It can help the image load faster compared to waiting on the custom element, and makes the image markup more flexible.

If I think of this as the poster "layer" more than the poster itself it helps me wrap my head around the naming and markup more.

internal/design/ui/poster.md
```diff
@@ -0,0 +1,208 @@
...
+import { Poster } from '@videojs/react';
+
+<Poster src="poster.jpg" alt="Video description" />;
+```
+
+#### CSS
+
+```css
+media-poster:not([data-visible]) {
+  display: none;
```
heff, line 57: question: I assume the answer is no, but should the poster not just hide all its children by default to make it super simple?

internal/design/ui/poster.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Visibility Logic
+
+```ts
+visible = !playback.started;
+```
+
+The poster is visible until playback has started. Once `started` becomes `true` (user plays or seeks), the poster hides and stays hidden.
+
+**Note:** `started` persists — pausing doesn't reset it. The poster only shows on initial load or after a new source is loaded.
```
heff, line 146: note: to be extra specific in case it matters, it's when a new source is loaded and the [video is not paused](https://github.com/muxinc/media-chrome/blob/main/src/js/media-store/state-mediator.ts#L373). 

Just need to avoid potential weirdness with the original poster showing when you don't intend it to, like when switching between playlist items that autoplay.

internal/design/ui/poster.md
mihar-22: Yep, any elements works.

internal/design/ui/poster.md
mihar-22: I get the desire to do it, but like our other components we have to leave it up to the user because they might want to animate (fade) or do something specific. It let's them tie it back to their design system in a way that makes sense for their app.

Issue #273, opened 2025-12-18
UI: Fullscreen Button Component


mihar-22:
Implement the FullscreenButton component for toggling fullscreen mode. Includes core interaction logic, DOM props builder, and React/HTML implementations.

PR #459, opened 2026-02-05
feat(core): add fullscreen button component

mihar-22:
closes https://github.com/videojs/v10/issues/273

## Summary

Adds FullscreenButton component across core, html, and react packages, building on the fullscreen feature from #458.

- **FullscreenButtonCore**: Shared business logic with state management for fullscreen toggle
- **FullscreenButtonElement**: Web Component for html package
- **FullscreenButton**: React component with proper TypeScript types

## Implementation

- Uses `FullscreenState` and `selectFullscreen` from #458
- Button state picks types from `FullscreenState`: `fullscreen` (boolean), `availability` (FeatureAvailability)
- Data attributes derived automatically from state keys: `data-fullscreen`, `data-availability` — no explicit mapping needed
- `toggle()` wrapped in try-catch (fullscreen requests can fail)
- Follows same patterns as MuteButton and PlayButton

## Files

- `packages/core/src/core/ui/fullscreen-button/` — Core logic and tests (15 tests)
- `packages/html/src/ui/fullscreen-button/` — Web Component
- `packages/html/src/define/ui/fullscreen-button.ts` — Custom element registration
- `packages/react/src/ui/fullscreen-button/` — React component

## Testing

- 15 unit tests for FullscreenButtonCore
- All 125 core tests pass

## Depends On

- #458 (fullscreen + pip features)

### DIFF
--- a/packages/core/src/core/ui/fullscreen-button/fullscreen-button-data-attrs.ts
+++ b/packages/core/src/core/ui/fullscreen-button/fullscreen-button-data-attrs.ts
@@ -2,5 +2,5 @@ export const FullscreenButtonDataAttributes = {
   /** Present when fullscreen mode is active. */
   fullscreen: 'data-fullscreen',
   /** Indicates fullscreen availability (`available` or `unsupported`). */
-  fullscreenAvailability: 'data-availability',
+  availability: 'data-availability',
 } as const;
--- a/.claude/skills/component/references/videojs.md
+++ b/.claude/skills/component/references/videojs.md
@@ -46,11 +46,19 @@ interface PlayButtonProps {
 ### State Interface
 
 ```ts
+// When keys match the feature state, use Pick (preserves JSDoc on IDE hover)
 interface PlayButtonState extends Pick<PlaybackState, 'paused' | 'ended' | 'started'> {}
+
+// When renaming keys, use Pick for matching keys and add JSDoc for renamed ones
+interface FullscreenButtonState extends Pick<FullscreenState, 'fullscreen'> {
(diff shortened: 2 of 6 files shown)


packages/core/src/core/ui/fullscreen-button/fullscreen-button-data-attrs.ts
```diff
@@ -0,0 +1,6 @@
+export const FullscreenButtonDataAttributes = {
+  /** Present when fullscreen mode is active. */
+  fullscreenActive: 'data-fullscreen',
```
heff: suggest: In MC we went with a boolean `isfullscreen`, with "is" because it reads a more clearly when you're in the fullscreen button context. Any reason not to do that here? Could also be fullscreen-active to match the prop.

mihar-22: Good callout! I think `data-is-fullscreen` would be inconsistent with what we have. None of our existing data attrs use an `is-` prefix. Play button uses `data-paused`, Mute button uses `data-muted`. Base UI follows the same convention – Toggle uses `data-pressed`, Checkbox uses `data-checked`. Short state words, no prefix.

By that convention, `data-fullscreen` feels right. It names the specific state the same way the others do. The `-active` suffix would also be redundant; presence of the attribute already implies active, same as `data-paused` or `data-pressed`.

I've also updated the state to align: `fullscreenActive` → `fullscreen` on `FullscreenState`, and the button state simplifies to just `fullscreen` and `availability` in the fullscreen button context, data attributes naturally reflect this way.

PR #464, opened 2026-02-05
feat(site): extract api reference from components

### LINKED ISSUE #259: Docs: API Reference
Generated API documentation for all packages. Auto-generated from TypeScript source using documentation tooling.

**Blocked by:**
- Relevant package implementations (per package)

**References:**
- https://github.com/mui/base-ui/tree/master/scripts/api-docs-builder

## Packages

- [ ] `@videojs/core` — Store, features, requests, contracts
- [ ] `@videojs/dom` — Props builders, data attributes, CSS variables
- [ ] `@videojs/react` — Hooks, components, providers
- [ ] `@videojs/html` — Custom elements, controllers

## Extraction

For each component, extract and generate JSON:

- **Props** — From `{Component}Props` interface
- **State** — From `{Component}State` interface
- **Callbacks** — From `{Component}Callbacks` interface
- **Data Attributes** — From `{Component}DataAttributes` enum/interface
- **CSS Variables** — From `{Component}CssVars` enum/interface

## Tasks

- [ ] Set up `typescript-api-extractor` in build pipeline
- [ ] Create component handler for extracting props, state, callbacks
- [ ] Create enum handler for data attributes and CSS variables
- [ ] Generate JSON output per component
- [ ] Create hook handler for extracting parameters and return values
- [ ] Integrate with docs site to render API tables

## Example Output
```json
{
  "name": "PlayButton",
  "description": "Button for toggling media playback.",
  "props": {
    "disabled": {
      "type": "boolean",
      "default": "false",
      "description": "Whether the button should ignore user interaction"
    },
    "label": {
      "type": "string",
      "description": "Custom aria-label"
    }
  },
  "state": {
    "paused": {
      "type": "boolean",
      "description": "Whether media is paused"
    },
    "ended": {
      "type": "boolean",
      "description": "Whether media has ended"
    }
  },
  "callbacks": {
    "onToggle": {
      "type": "() => void",
      "description": "Fires on any toggle interaction"
    },
    "onPlay": {
      "type": "() => void",
      "description": "Fires when play request succeeds"
    }
  },
  "dataAttributes": {
    "data-paused": {
      "description": "Present when media is paused"
    },
    "data-disabled": {
      "description": "Present when button is disabled"
    }
  },
  "cssVariables": {}
}
```

decepulis:
Closes #259 for basic cases

> [!NOTE]
> Does not include support for unimplemented items like component components or css custom properties

## how?
  - Add api-docs-builder script that extracts props, state, and data attributes from TypeScript source files (packages/core and packages/html) and outputs validated JSON to src/content/generated-api-reference/
  - Add Astro components (ApiRefSection, ApiPropsTable, ApiStateTable, ApiDataAttrsTable, PropRow) that render the generated JSON as interactive, framework-aware reference tables
  - Replace hand-written reference docs and examples with auto-generated API reference sections

## unrelated
  - Remove old HTML/React examples and manually-maintained reference pages for fullscreen-button, time-slider, and volume-slider (play-button and mute-button updated to use the new system)

### DIFF
--- a/site/scripts/api-docs-builder/src/core-handler.ts
+++ b/site/scripts/api-docs-builder/src/core-handler.ts
@@ -63,7 +63,11 @@ export function extractCore(filePath: string, program: ts.Program, componentName
  *
  * Looks for: static readonly defaultProps = { label: '', disabled: false }
  */
-function extractDefaultProps(filePath: string, program: ts.Program, componentName: string): Record<string, string> {
+export function extractDefaultProps(
+  filePath: string,
+  program: ts.Program,
+  componentName: string
+): Record<string, string> {
   const sourceFile = program.getSourceFile(filePath);
   if (!sourceFile) return {};
 
@@ -109,7 +113,7 @@ function extractDefaultProps(filePath: string, program: ts.Program, componentNam
 /**
  * Get a string representation of a property value.
  */
-function getPropertyValue(node: ts.Expression, sourceFile: ts.SourceFile): string | undefined {
+export function getPropertyValue(node: ts.Expression, sourceFile: ts.SourceFile): string | undefined {
   if (ts.isStringLiteral(node)) {
     return `'${node.text}'`;
   }
--- a/site/scripts/api-docs-builder/src/data-attrs-handler.ts
+++ b/site/scripts/api-docs-builder/src/data-attrs-handler.ts
@@ -86,7 +86,7 @@ export function extractDataAttrs(
 /**
  * Extract JSDoc comment from a property assignment.
  */
-function getJsDocComment(node: ts.PropertyAssignment, sourceFile: ts.SourceFile): string {
+export function getJsDocComment(node: ts.PropertyAssignment, sourceFile: ts.SourceFile): string {
   // Get leading comment ranges
   const fullText = sourceFile.getFullText();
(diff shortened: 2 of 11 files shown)


site/src/types/api-reference.ts
```diff
@@ -0,0 +1,44 @@
...
+export const StateDefSchema = z.object({
+  type: z.string(),
+  description: z.string().optional(),
+});
+
+export const DataAttrDefSchema = z.object({
+  description: z.string(),
+});
+
+export const ComponentApiReferenceSchema = z.object({
```
mihar-22, line 26: Will this need to change or will it hold up when we add support for Compounds? Either we have a new schema that contains Component API Schems, or we extend this with parts. 

We'll also need to discuss and establish a pattern in core for how Compound Parts will be discovered.

mihar-22 (approved):
This is looking really good! Incredible foundation for us to layer on CSS Vars and  Compound Components. I'm more concerned with how this will evolve to support the latter, might require significant changes. Either way, awesome start :)

site/src/types/api-reference.ts
decepulis: Definitely needs to change for compound.

PR #468, opened 2026-02-06
feat(site): generated multipart component api reference

### LINKED ISSUE #465: Compound component API reference
Extending the work of #259 for compound components like #266

decepulis:
## References
- closes #465
- see also #259

##  Summary

  Extends the API docs builder and site components to support multi-part compound components (like <Time.Group>, <Time.Value>, <Time.Separator>), and adds reference pages for new components.

> [!NOTE]
> We know from Tech Preview that APIs differ across frameworks — we might have compound components in React but not in HTML (e.g., tooltips, popovers). This PR does not address that case.

###  Core package

  - time-core.ts — Renamed TimeCoreProps -> TimeProps, TimeValueState -> TimeState for cleaner API reference output. Aliased the imported TimeState from media/state as MediaTimeState to resolve the name collision. (fyi @mihar-22)

###  API docs builder (site/scripts/api-docs-builder/)

  - Multi-part discovery — New parts-handler.ts parses index.parts.ts from the React package to identify component parts. Each value export becomes a part; type-only exports are skipped.
  - Part matching algorithm — Derives kebab-case segments from source paths (e.g., ./time-value -> value), matches each to its HTML element file ({component}-{part}-element.ts).
  The part with no sub-element file becomes the primary part and inherits the shared core props/state/data-attrs.
  - Type alias expansion — New formatDetailedType() in formatter.ts resolves ExternalTypeNode references against all exports, so aliases like TimeType are expanded to their
  underlying union ('current' | 'duration' | 'remaining') in the output.
  - Part descriptions — Extracts JSDoc descriptions from React component files for each part.
  - Refactored shared logic — buildProps(), buildState(), buildDataAttrs() extracted as reusable functions from the former monolithic builder.
  - shortType for state — StateDef now supports shortType, matching what PropDef already had.
  - extractHtml() accepts optional elementName — For sub-part elements whose class name isn't {ComponentName}Element.
  - Suppressed noisy warnings — typescript-api-extractor logs warnings for unhandled type flags (never, ESSymbol, etc.); these are now silenced.

###  Astro components (site/src/components/docs/api-reference/)

  - New ApiReference.astro — Unified component that handles both single-part and multi-part rendering. Replaces the old per-section ApiRefSection.astro approach. Multi-part components get per-part h2 headings with framework-conditional names (<Time.Group /> for React, <media-time-group> for HTML).
  - New DetailRow.astro — Extracted shared expandable row logic from PropRow.astro. Now used by both PropRow and the new StateRow for consistent expand/collapse behavior.
  - New StateRow.astro — State table rows now support expandable details (type + description), matching the props table UX.
  - Updated ApiStateTable.astro — Uses StateRow component; accepts componentName prop for unique IDs.
  - Slimmed PropRow.astro — Delegates to DetailRow instead of duplicating expand/collapse logic.

###  Reference pages & TOC

  - New pages: fullscreen-button.mdx, time.mdx (multi-part)
  - Updated pages: play-button.mdx, mute-button.mdx — migrated from ApiRefSection to ApiReference, added anatomy sections, fixed HTML tag names (play-button -> media-play-button)
  - remarkConditionalHeadings.js — Now reads generated API reference JSON to inject heading metadata into the TOC for component-rendered headings (supports both single-part and
  multi-part structures with framework-conditional labels)
  - Sidebar — Added fullscreen-button and time to the Components section

###  Other

  - renderInlineMarkdown.ts — Added HTML escaping for </> in code spans (prevents <video> from being parsed as HTML)
  - globals.css — Code font size uses clamp() for better scaling in tables
  - Schema — PartApiReferenceSchema added; ComponentApiReferenceSchema extended with optional parts record
  - New tests for parts-handler, core-handler, html-handler, and formatter


packages/core/src/core/ui/time/time-core.ts
```diff
@@ -3,21 +3,21 @@ import { isFunction } from '@videojs/utils/predicate';
 import { formatTime, formatTimeAsPhrase, secondsToIsoDuration } from '@videojs/utils/time';
 import type { NonNullableObject } from '@videojs/utils/types';
 
-import type { TimeState } from '../../media/state';
+import type { TimeState as MediaTimeState } from '../../media/state';
```
mihar-22, line 6: Nice. Maybe let's just make this the default naming convention for core media state types? Prevent collisions and it's clearer:

- `TimeState` -> `MediaTimeState`
- `FullscreenState` -> `MediaFullscreenState`
- ...

I can follow up!

site/scripts/api-docs-builder/src/index.ts
```diff
@@ -183,10 +225,178 @@ function buildComponentApiReference(source: ComponentSource, program: ts.Program
...
+ * Discover parts and match them to HTML element files.
+ *
+ * Matching algorithm:
+ * 1. Parse `index.parts.ts` for named exports -> part names and source paths
+ * 2. Derive kebab segment from source: `./time-value` -> strip `./time-` prefix -> `value`
+ * 3. For each part, look for `{name}-{kebab}-element.ts` in HTML dir (e.g., `time-group-element.ts`)
+ * 4. The part with NO matching `{name}-{kebab}-element.ts` but where `{name}-element.ts` exists -> primary part
+ * 5. Primary part gets: shared core file, shared data-attrs, main element (`{name}-element.ts`)
+ */
+function discoverParts(source: ComponentSource, program: ts.Program): PartSource[] {
```
mihar-22, line 238: I don't think the discovery algorithm will hold, relies on file naming conventions that I'm just not 100% sure will be reliable or consistent enough. Prefer if we have a more intentional way of documenting or coding this information.

Let's just deal with it when it happens. I don't want to think too hard about it right now. Rather let the code keep naturally evolving and align builder with it.

site/CLAUDE.md
mihar-22, line 1: I really love that you're storing this information in a site specific `CLAUDE.md` and updating it. Awesome!!!

1. I should update our library "component" and "docs" skills to align with the work you've been doing. 

2. I'm thinking we should update our dev workflow in root `CLAUDE.md` to include updating docs API reference tables? It could possibly just be a lint-staged workflow. We should probably ticket up "Automating API reference updates"?

mihar-22 (approved):
This is looking really good!! Awesome work 👏 

---

Thoughts on the "On this page" section for compound components? I like our simple components version. Thinking it could say "API reference" then list out parts (similar to Base UI), something like:

- API reference
  - Time.Value
    - Props
    - State
    - Data attributes

If we don't like three levels of nesting immediately showing (noisy), we could do the following showing only compound parts and progressively disclosing nested children as they become active:

- API reference
  - Time.Value (show props, state, data attrs once active – scroll or click)
  - Time.Group
  - Time.Separator

<details>
  <summary>🖼️  Click to see our TOC</summary>
Image
</details>

<details>
  <summary>🖼️ Click to see Base UI TOC</summary>

Image

</details>

We can ticket this up separately to track and discuss! If you have other ideas then awesome.

---

Merging to keep us rolling!

decepulis:
@mihar-22 

> Thoughts on the "On this page" section for compound components? I like our simple components version.

I get the friction you're bumping against. I don't like what I have committed here. I was just looking for a way to get everything into H2 and H3 so it would show up in the TOC. I think your suggestions probably have something worth following up on

packages/core/src/core/ui/time/time-core.ts
decepulis: I don't have a strong preference on the convention. Just that we have one! Convention > config. I'm happy to defer to your opinions here.

site/scripts/api-docs-builder/src/index.ts
decepulis: Oh yeah. It's definitely painted in a corner. See related api reference prs for places it still needs to improve. I trust AI to save us once we hit these cases though. And that'll be easier once we have components that break the mold.

site/CLAUDE.md
decepulis: On the first bit, yeah for sure. I've got a lot of ideas for the repo skills. Just leaving that for later for now. 

Right now the generate script runs on site build. That might be good enough? If you have a case for running it on staged, though, I'm down to hear it.

Issue #304, opened 2026-01-14
Media API Design


mihar-22:
Split from #301.

Design a unified abstraction layer for consistent interaction with different media engines (native video, HLS, DASH, YouTube, Vimeo).

## Problem

Different streaming engines have incompatible APIs:

```ts
hls.currentLevel = 2;                    // HLS
dash.setQualityFor('video', 2);          // DASH
// Native video has no quality API
```

## Need

Provider + Adapter pattern that's tree-shakeable but easy to use.

## Design Goals

- **Composability** — Independent pieces that combine cleanly
- **Tree-shaking** — Bundle only what you use
- **Extensibility** — Extend without forking
- **Progressive complexity** — From skin to fully custom

## Related

- Parent issue: #301

PR #507, opened 2026-02-10
feat: add media API + HLS video components

luwes:
Fix https://github.com/videojs/v10/issues/304
Implementation https://github.com/videojs/v10/pull/467

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -21,10 +21,10 @@
       "development": "./dist/dev/dom.js",
       "default": "./dist/default/dom.js"
     },
-    "./dom/media/hls": {
-      "types": "./dist/dev/dom/media/hls.d.ts",
-      "development": "./dist/dev/dom/media/hls.js",
-      "default": "./dist/default/dom/media/hls.js"
+    "./dom/media/*": {
+      "types": "./dist/dev/dom/media/*.d.ts",
+      "development": "./dist/dev/dom/media/*.js",
+      "default": "./dist/default/dom/media/*.js"
     }
   },
   "main": "dist/default/index.js",


internal/design/media-api.md
```diff
@@ -0,0 +1,221 @@
...
+  'webkitbeginfullscreen',
+  'webkitendfullscreen',
+  'webkitpresentationmodechanged',
+] as const;
+
+type MediaTarget = HTMLMediaElement | HTMLVideoElement | HTMLAudioElement;
+
+type MediaTargetConstructor = Constructor<MediaTarget>;
+
+declare class Media extends EventTarget {
```
mihar-22: The `Media` contract currently looks close to a full `HTMLMediaElement` API surface. That does not fully line up with the earlier “subset” framing in the doc. I do not yet know what we want the minimal contract to be, but I assume it is something like `play`/`pause`, `currentTime`, `paused`, `ended`, `seekable`, maybe `duration`, `volume`/`muted`, and a small required event set. Feels like something we should spell out concretely so “subset” is mechanically true. Christian would have thoughts here.

In addition, if someone types against this, it reads as a hefty contract and could be misleading if implementations are allowed to omit large parts of it. We may want `Media` plus optional capability extensions, or at least make a lot more of this optional in the type layer.

internal/design/media-api.md
```diff
@@ -0,0 +1,221 @@
...
+| ----------- | --------------------------------------------- | ------------------------------------------------------------------------------- |
+| Stream type | `streamType`, `streamtypechange`              | Know if content is live or on-demand and adapt UI (seek bar, live badge, etc.). |
+| Live edge   | `liveEdgeStart`, constrained `seekable.end()` | “At live” detection and “seek to live” for live/HAS.                            |
+| Renditions  | `videoRenditions`, `audioRenditions`          | List and select quality levels for HAS.                                         |
+
+- **HTMLMediaElement subset** — A conforming implementation may implement only a subset of HTMLMediaElement (e.g. core playback: `play`, `pause`, `currentTime`, `duration`, `seekable`, `paused`, `ended`). The type system and documentation will define which subset is required for which integration.
+- **Custom extensions** — Implementations may expose the custom extensions only when applicable (e.g. `liveEdgeStart` for live; renditions for HAS).
+
+```ts
+declare const MediaEvents = [
```
mihar-22, line 54: The `MediaEvents` list is very broad and includes vendor prefixed events. If the goal is minimal conformance, I think we should separate “required events” from “forwarded events”, or otherwise make it clear what a minimal implementation must emit versus what it may forward.

internal/design/media-api.md
```diff
@@ -0,0 +1,221 @@
...
+type MediaTarget = HTMLMediaElement | HTMLVideoElement | HTMLAudioElement;
+
+type MediaTargetConstructor = Constructor<MediaTarget>;
+
+declare class Media extends EventTarget {
+  // Custom extensions (unified media API)
+  streamType: 'live' | 'on-demand';
+  liveEdgeStart: number;
+  videoRenditions: VideoRenditionList;
+  audioRenditions: AudioRenditionList;
```
mihar-22: I think it would be great if we further split the contracts, so it feels more deliberate and extensible (new engines can adopt specific contracts):

```ts
export interface RenditionsCapability {
  readonly videoRenditions: readonly unknown[];
  readonly audioRenditions: readonly unknown[];
}

interface Media extends Partial<RenditionsCapability>, ...
```

Same for other capabilities.

mihar-22: For the platform gap fields, I assume these should be optional so that `HTMLMediaElement` can satisfy the base interface without adapters.

internal/design/media-api.md
```diff
@@ -0,0 +1,221 @@
...
+  'waitingforkey',
+  'resize',
+  'enterpictureinpicture',
+  'leavepictureinpicture',
+  'webkitbeginfullscreen',
+  'webkitendfullscreen',
+  'webkitpresentationmodechanged',
+] as const;
+
+type MediaTarget = HTMLMediaElement | HTMLVideoElement | HTMLAudioElement;
```
mihar-22: `MediaTarget` being a complete DOM element feels like it makes the contract inherently DOM-shaped. I am not sure we want that to be a requirement, especially if the goal is a minimal subset and we want to avoid leaking DOM APIs into the store surface (for Media specifically).

mihar-22: Related, what is the SSR story here. If `MediaTarget` is a concrete DOM type, that can pull DOM into the types and runtime assumptions in ways that are awkward for SSR or Node-only environments.

internal/design/media-api.md
```diff
@@ -0,0 +1,221 @@
...
+
+interface HlsMediaBase {
+  attach?(target: MediaTarget): void;
+  detach?(): void;
+}
+
+// The mixin is used by the web component because it needs to extend HTMLElement!
+export const HlsMediaMixin = <T extends Constructor>(Super: T) => {
+  class HlsMedia extends (Super as Constructor<HlsMediaBase>) {
+    engine = new Hls();
```
mihar-22, line 198: Love the engine property. I can see this as a generic shape and optional property on the base contract like:

```ts
interface Media<Engine = unknown> {
  engine?: Engine;
}

// Typed access to lower-level APIs
function isHlsMedia(media: Media): media is Media<HlsMediaEngine>
```

internal/design/media-api.md
```diff
@@ -0,0 +1,221 @@
...
+
+    attach(target: MediaTarget): void {
+      super.attach?.(target);
+      this.engine.attachMedia(target);
+    }
+
+    detach(): void {
+      super.detach?.();
+      this.engine.detachMedia();
+    }
```
mihar-22: Something about this feels off to me, like engine wiring up details leaking into media. Should this be pushed down into an engine-specific contract?

```ts
interface MediaEngine<Target = unknown> {
  attach(target: Target): void;
  detach(): void;
}

class HlsMediaEngine implements MediaEngine<HTMLVideoElement> {
  // ...
}
```

Makes sense to me since engines inherently need some way to bind to a renderer (a video element, an MSE pipeline, canvas, etc.)

internal/design/media-api.md
```diff
@@ -0,0 +1,221 @@
...
+
+  // Proxy handlers
+  get(prop: keyof MediaTarget): any;
+  set(prop: keyof MediaTarget, val: any): void;
+  call(prop: keyof MediaTarget, ...args: any[]): any;
+}
+
+export class Media extends MediaMixin(EventTarget, HTMLMediaElement) {}
+export class Video extends MediaMixin(EventTarget, HTMLVideoElement) {}
+export class Audio extends MediaMixin(EventTarget, HTMLAudioElement) {}
```
mihar-22: I was under the assumption that the HTML media elements would satisfy the Media API out of the box. Is this "mixin" extending capabilities somehow? Is this intended for React somehow? Maybe I missed something.

mihar-22: Oh I see, maybe the target/attach/detach. I think comment above might help separate those concerns so this isn't needed.

internal/design/media-api.md
mihar-22: After reading through the doc, I understand why this is this way. I confused it with being a "store target". It seems to be more of a `MediaEngineTarget` as I'm understanding it.

internal/design/media-api.md
```diff
@@ -0,0 +1,221 @@
...
+## Summary
+
+This RFC proposes a **unified media API** that extends the [HTMLMediaElement](https://html.spec.whatwg.org/multipage/media.html#htmlmediaelement) to support different media types (video, HLS, DASH, YouTube, etc.) in a consistent way. The API is designed to:
+
+- **Extend** the existing HTMLMediaElement surface rather than replace it. Implementations may expose the **full** HTMLMediaElement surface or a **subset** of it; the unified contract is "HTMLMediaElement-like" where the subset is sufficient for the integration (e.g. play/pause, currentTime, seekable).
+- **Support multiple integration points:** for HTML, web components that implement this API; for React and other frameworks, an API that extends `EventTarget` and exposes the same surface (full or subset).
+- **Add custom extensions** for behavior not covered by HTMLMediaElement, specifically:
+  - **Stream type** — distinguish live vs. on-demand content.
+  - **Live edge** — model the “live edge” of live streams (window, seekable end, seek-to-live).
+  - **Renditions list** — discover and select video/audio quality levels (renditions) for HTTP Adaptive Streaming (HAS).
```
mihar-22, line 12: This section is great 👍 Love the extensibility callouts and supporting a subset of the API. It mentions "full" or "subset", but I think "superset" is where this doc is going as well (i.e., gaps/extensions)

internal/design/media-api.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Gaps in the platform
+
+Current HTMLMediaElement does not address:
+
+1. **Stream type** — UIs need to know if content is live or on-demand (e.g. show/hide seek bar, live badge, DVR controls). Today this is inferred from `duration` or format-specific logic, which is brittle (e.g. ended live vs. VOD).
+2. **Live edge** — For live/HAS, “live” is a window, not a single time. UIs need a clear notion of “at live” and “seek to live” that respects HOLD-BACK, target duration, and segment boundaries.
+3. **Renditions** — Users and UIs need to see available quality levels and select a rendition; today this is player-specific and not standardized.
+
+By extending the media element (or an EventTarget-based equivalent) with **stream type**, **live edge**, and **renditions list** APIs, we give UIs a single, predictable way to implement live indicators, seek-to-live, quality selectors, and stream-type–aware chrome.
```
mihar-22, line 31: Another great section 👍 Appreciate the hints at extensibility and identifying specific gaps.

internal/design/media-api.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Design principles
+
+- **Extension of HTMLMediaElement** — New behavior is added via partial interfaces and constraints on existing attributes (e.g. `seekable`), not by replacing the element.
```
mihar-22, line 35: This is spot on. I think it might've been missed above in the interface design. Called out in few places below but reading this over twice I see this was the intended design goal. Nice 👍

mihar-22 (commented):
Thank you for writing this up, Wes! I’m aligned with the design goals in the doc and the general direction the shapes are taking. I do see a few opportunities to refine things and tighten alignment with those goals.

Happy to dig deeper on any of my comments, but those are the main things that jumped out at me 🙂

internal/design/media-api.md
luwes: I changed the naming to MediaApiProxy, that's the real use of this mixin. It adds properties and methods of the HTMLMediaElement API and proxies these to a target object. In most cases a video element but it could be an iframe API too.

packages/react/tsdown.config.ts
```diff
@@ -7,7 +7,13 @@ type BuildMode = 'dev' | 'default';
...
 
 const createConfig = (mode: BuildMode): UserConfig => ({
-  entry: 'src/**/index.ts',
+  entry: {
+    index: './src/index.ts',
+    'media/hls-video': './src/media/hls-video.tsx',
+    'presets/audio/index': './src/presets/audio/index.ts',
+    'presets/background/index': './src/presets/background/index.ts',
+    'presets/video/index': './src/presets/video/index.ts',
+  },
```
sampotts: nit: Could we simplify by adding a barrel file in media?

luwes: wouldn't the media elements then be all bundled together? Or how do you mean?

sampotts: True, maybe `/src/media/hls-video/index.ts` in that case? Just wondering if this list will get pretty large later on.

internal/design/media-api.md
luwes: Added this minimal interface that is only required. 

```ts
export type MediaBaseApi = {
  play: () => Promise<void>;
  paused: boolean;
};
````

I think we only want the play() method and a paused prop + play event for the very minimal gif player for example.

PR #523, opened 2026-02-13
feat(react): initial skin scaffolding

sampotts:
Just for React only at this point. If we're happy with the direction, I'll roll it out to the html side too. 

- Add the initial exports for the different use cases for skins.
- Add script to copy CSS to `dist` (watch and build). 
- Move media exports alongside the presets. 
- Add new paths to bundle size checks. 

Imports would look like this:

```tsx
import { VideoSkin, Video } from '@videojs/react/video';
import '@videojs/react/video/skin.css';

import { AudioSkin, Audio } from '@videojs/react/audio';
import '@videojs/react/audio/skin.css';

import { BackgroundVideoSkin, BackgroundVideo } from '@videojs/react/background';
import '@videojs/react/background/skin.css';

// ...and for other skins (only minimal for now)...

import { MinimalVideoSkin, Video } from '@videojs/react/video';
import '@videojs/react/video/minimal-skin.css';

import { MinimalAudioSkin, Audio } from '@videojs/react/audio';
import '@videojs/react/audio/minimal-skin.css';
```

### DIFF
--- a/packages/react/tsdown.config.ts
+++ b/packages/react/tsdown.config.ts
@@ -6,7 +6,7 @@ type BuildMode = 'dev' | 'default';
 
 const buildModes: BuildMode[] = ['dev', 'default'];
 
-const createConfig = (mode: BuildMode, isWatch: boolean): UserConfig => ({
+const createConfig = (mode: BuildMode): UserConfig => ({
   entry: 'src/**/index.ts',
   platform: 'browser',
   format: 'es',
@@ -42,7 +42,4 @@ const createConfig = (mode: BuildMode, isWatch: boolean): UserConfig => ({
   ],
 });
 
-export default defineConfig((cliOptions) => {
-  const isWatch = !!cliOptions.watch;
-  return buildModes.map((mode) => createConfig(mode, isWatch));
-});
+export default defineConfig(buildModes.map((mode) => createConfig(mode)));


packages/react/src/presets/audio/minimal-skin.css
mihar-22, line 1: Are we planning on having vanilla CSS files ready first? I'd assume Tailwind and hand written vanilla only as stop gap if compiler isn't ready. If it is Tailwind first then we can remove these probably and get that side setup.  

Thoughts?

packages/react/package.json
```diff
@@ -16,12 +16,30 @@
...
   },
   "scripts": {
-    "build": "tsdown",
-    "build:watch": "tsdown --watch ./src --no-clean",
-    "dev": "pnpm run build:watch",
+    "build": "tsdown && pnpm css:copy",
+    "build:watch": "concurrently 'tsdown --watch ./src --no-clean' 'pnpm css:watch'",
+    "css:copy": "rsync -a --include='*/' --include='*.css' --exclude='*' src/presets/ dist/default/presets/",
+    "css:watch": "chokidar 'src/presets/**/*.css' --initial -c 'pnpm run css:copy'",
+    "dev": "pnpm build:watch",
```
mihar-22: Love to steer far away from concurrent and separate processes for tasks as they always find a way of breaking down. Doesn't scale well either. We should be able to hook into [`tsdown`](https://tsdown.dev) by either adding the CSS into `entry` or by using [hooks](https://tsdown.dev/advanced/hooks).

However, if we're not having CSS files just yet and first order of business is Tailwind then we can just remove it all.

packages/react/tsdown.config.ts
```diff
@@ -1,13 +1,22 @@
...
 
 const buildModes: BuildMode[] = ['dev', 'default'];
 
+const presetsEntries = Object.fromEntries(
+  globSync('src/presets/**/index.ts').map((file) => {
+    const key = file.replace('src/', '').replace(/\.ts$/, '');
+    return [key, file];
+  })
+);
+
```
mihar-22: `tsdown` supports globs out of the box :) However, I'd recommend simplifying to make it readable in the future. For example:

```ts
entry: {
  audio: './src/presets/audio/index.ts',
  video: './src/presets/video/index.ts',
  background: './src/presets/background/index.ts',
}
```

P.S: we have `unbundle: true` in `tsdown` which will naturally preserve modules and not bundle them together.

package.json
```diff
@@ -17,7 +17,7 @@
   "scripts": {
     "prepare": "simple-git-hooks",
     "postinstall": "pnpm link:opencode && pnpm link:agents",
-    "build:packages": "turbo run build --filter='./packages/*'",
+    "build:packages": "turbo run build --filter='./packages/*' --filter='!@videojs/sandbox'",
```
mihar-22: 👍

.size-limit.json
mihar-22, line 1: 🗒️ Leaving note for others since I discussed with Sam that the bundle size increase in this PR is misleading. 

It's bundling each export subpath separately and adding it all up instead of bundling it together with root then diffing. In other words it should be ((root + subpath) - root) to get the actual size difference, otherwise it's missing out on shared minification and compression.

I'll try to fix this soon.

mihar-22 (changes requested):
Awesome work, thanks for doing this!

packages/react/src/presets/audio/minimal-skin.css
sampotts: I'm assuming we'd use hand-rolled vanilla CSS files first and then the Tailwind compiled versions come later. Probably a good subject to discuss.

packages/react/package.json
sampotts: I tried a few different ways of handling this: 
- `tsdown` has a `copy` option but it only works for builds and doesn't watch the source files and ideally we'd want the copy to execute when the source CSS changes or it'll be a worse DX. I suspect the hooks would be the same unless we can explicitly tell it to watch the CSS files too. 
- I tried a couple of rollup plugins but none seemed to actually work in that they'd only copy when other files changed, not the source changes, so kinda similar to the above.

packages/react/tsdown.config.ts
sampotts: Can do, I guess I just wanted to avoid an extra step every time we add a new preset but perhaps it's not going to happen often enough to be a pain. 

Does this also need updating in that case? 
https://github.com/videojs/v10/blob/1c916efed456c8a6df240776520fea1d892164c1/packages/html/tsdown.config.ts#L9,L14

packages/react/src/presets/audio/minimal-skin.css
sampotts: I assume the docs are going to have to at least use the hand-rolled vanilla CSS unless we want to get back into the `vjs` prefix stuff we had in tech preview.

mihar-22: I guess it depends if we're supporting eject or not in Beta, I don't see it in Sarah's doc so maybe not.

If that's right then starting vanilla makes sense :)

packages/react/package.json
sampotts: This would be ideal really:
https://github.com/rolldown/tsdown/issues/720

mihar-22: Gave it a quick test:

1. Need to watch `*.css` files: `"build:watch": "tsdown --watch ./src --watch ./src/**/*.css --no-clean"`
2. Add the following to `tsdown.config.ts`

```ts
 copy: [
    {
      from: 'src/presets/audio/*.css',
      to: `dist/${mode}/audio`,
    },
  ],
```

packages/react/tsdown.config.ts
mihar-22: That one we should probably keep because it would be a pain to maintain. This one is up to you, just a callout that it can be simplified as there won't be additional presets for a while.

mihar-22: One difference: output directory here would become `dist/${mode}/*/audio` instead of one down in presets which could help clean up Node exports a small touch. 

Alternative is we just go `entry: "src/**/index.ts"` and call it a day?

packages/react/src/presets/audio/minimal-skin.css
mihar-22: Starting vanilla is pretty cool actually because it gives us a real reference to test the compiler against.

packages/react/package.json
mihar-22: ☝️ The copy example `to` depends where you land on the comment below.

packages/react/src/presets/audio/minimal-skin.css
sampotts: Yeah, I think @cpillsbury was keen to have an "expected CSS" to compare against so that would be the aim 👍🏼

packages/react/package.json
sampotts: Nice one. I'll give that another shot!

sampotts: Unfortunately, `--watch ./src --watch ./src/**/*.css` seems to break it watching anything _other_ than the `.css` files. I was pulling (what's left) of my hair out earlier trying to get this working 😆

mihar-22: Ah sorry, should've tested that. I see in types it supports an array. Let me check `--watch ./src,./src/**/*.css` in a second.

sampotts: I tried without luck but maybe you have more luck. The only alternative I could come up with was a custom plugin to add the css files to the watch. It seems like a bit of a bug with rolldown.

.size-limit.json
mihar-22: Fixed on main, we just need to rebase here

PR #600, opened 2026-02-24
fix(site): work around Astro SSR false "Invalid hook call" warnings

### LINKED ISSUE #596: fix(site): Invalid hook call due to dual React instances (18 + 19)
## Problem

Running `pnpm dev:site` throws **"Invalid hook call. Hooks can only be called inside of the body of a function component."**

The root cause is two React instances coexisting in the pnpm store:

| Workspace | `react` resolves to |
|-----------|-------------------|
| `site/` | **18.3.1** |
| `packages/react/` | **19.2.3** |
| `packages/store/` | **19.2.3** |

When site components import from `@videojs/react` or `@videojs/store`, hooks registered with React 19 are called inside a React 18 render tree. React's internal module identity check fails → "Invalid hook call."

The `dedupe: ['react', 'react-dom']` in the Vite config cannot resolve this because pnpm physically installs two separate versions.

## Fix

Upgrade `react` and `react-dom` to `^19.0.0` in `site/package.json` so pnpm resolves a single React 19 instance. Update the React Compiler target from `'18'` to `'19'` in `site/astro.config.mjs`.

All site dependencies support React 19 (`@astrojs/react`, `@base-ui-components/react`, `@mux/mux-uploader-react`, `@nanostores/react`, `lucide-react`, `@testing-library/react`, `@types/react`).

decepulis:
## Summary

- Replace `return null` with `return <></>` in three React components that trigger false SSR warnings
- Workaround for [withastro/astro#12283](https://github.com/withastro/astro/issues/12283)

## Problem

Astro's SSR renderer logs false "Invalid hook call" warnings when a React component with hooks returns `null` instead of JSX. Three components were affected, producing 3 warnings per docs page request:

1. **`PreferenceSync`** — rendered once per page (Base layout)
2. **`PreferenceUpdater`** — rendered twice per docs page (DocsSidebar in desktop + mobile)
3. **`TableOfContents`** — when heading list is empty

## How we confirmed

- Replacing `return null` → `return <></>` in PreferenceSync dropped errors from 3 → 2
- Same for PreferenceUpdater → 2 → 0
- Reverting any single change increases the count back by 1
- Ruled out: dual React instances, React Compiler, @videojs imports, Astro version (tested 5.17.3)

## Test plan

- [x] `pnpm -F site test` — 244 tests pass
- [x] `pnpm dev:site` — 0 "Invalid hook call" warnings on guide and reference pages

Closes #596

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -413,10 +413,10 @@ importers:
     dependencies:
       '@astrojs/mdx':
         specifier: ^4.3.7
-        version: 4.3.13(astro@5.17.3(@netlify/blobs@10.5.0)(@types/node@22.19.3)(@vercel/functions@2.2.13)(jiti@2.6.1)(lightningcss@1.30.2)(rollup@4.53.3)(tsx@4.21.0)(typescript@5.9.3)(yaml@2.8.2))
+        version: 4.3.13(astro@5.16.5(@netlify/blobs@10.5.0)(@types/node@22.19.3)(@vercel/functions@2.2.13)(jiti@2.6.1)(lightningcss@1.30.2)(rollup@4.53.3)(tsx@4.21.0)(typescript@5.9.3)(yaml@2.8.2))
       '@astrojs/netlify':
         specifier: ^6.6.4
-        version: 6.6.4(@netlify/api@14.0.13)(@types/node@22.19.3)(@vercel/functions@2.2.13)(astro@5.17.3(@netlify/blobs@10.5.0)(@types/node@22.19.3)(@vercel/functions@2.2.13)(jiti@2.6.1)(lightningcss@1.30.2)(rollup@4.53.3)(tsx@4.21.0)(typescript@5.9.3)(yaml@2.8.2))(jiti@2.6.1)(lightningcss@1.30.2)(rollup@4.53.3)(tsx@4.21.0)(yaml@2.8.2)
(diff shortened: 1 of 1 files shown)


site/src/components/docs/TableOfContents/index.tsx
```diff
@@ -12,7 +12,9 @@ export function TableOfContents({ headings }: TableOfContentsProps) {
   const activeId = useActiveHeading(filteredHeadings);
 
   if (filteredHeadings.length === 0) {
-    return null;
+    // Workaround: Astro SSR logs false "Invalid hook call" when a React
+    // component with hooks returns null. See withastro/astro#12283.
+    return <></>;
```
sampotts, line 17: This feels like a bug with Astro, `null` is valid. You may need to add a comment to get Biome to ignore this?

decepulis: Oh yeah totally a bug with Astro. I didn't notice Biome was warning about it though. Thanks for the heads up!

sampotts: Ah maybe Biome is cool with it now. I think I originally switched it to `null` as it (or maybe ESLint back then) was complaining.

decepulis: I've been seeing these rule of hooks errors for AGES. I don't think you exclusively started this? (#597 was happening, too, so, who knows)

PR #607, opened 2026-02-25
feat: add background video preset

luwes:
questions: 
- is media-container still useful for background video
- is the store provider (background-video-player) useful? what will it be used for in the future?

### DIFF
--- a/packages/html/src/define/background/skin.css
+++ b/packages/html/src/define/background/skin.css
@@ -1,3 +1,7 @@
+background-video-player {
+  display: contents;
+}
+
 background-video-skin {
   display: block;
   position: relative;
--- a/packages/sandbox/templates/html-background/main.ts
+++ b/packages/sandbox/templates/html-background/main.ts
@@ -9,7 +9,7 @@ import '@videojs/html/background/video';
 const html = String.raw;
 
 document.getElementById('root')!.innerHTML = html`
-  <background-video-player style="display: contents;">
+  <background-video-player>
     <background-video-skin>
       <background-video slot="media" src="https://stream.mux.com/Sc89iWAyNkhJ3P1rQ02nrEdCFTnfT01CZ2KmaEcxXfB008/low.mp4"></background-video>
     </background-video-skin>


packages/html/src/define/background/skin.ts
```diff
@@ -1 +1,40 @@
...
+}
+
+function namedNodeMapToObject(namedNodeMap: NamedNodeMap) {
+  const obj: Record<string, string> = {};
+  for (const attr of namedNodeMap) {
+    obj[attr.name] = attr.value;
+  }
+  return obj;
+}
+
```
mihar-22: Can we replace with utility from `utils/dom`, same name/function/behaviour is exported.

packages/html/src/media/background-video/index.ts
```diff
@@ -12,17 +10,22 @@ function getTemplateHTML(attrs: Record<string, string>) {
         inset: 0;
         width: 100%;
         height: 100%;
-        object-fit: inherit;
+        object-fit: var(--media-object-fit, inherit);
+        object-position: var(--media-object-position, 50% 50%);
```
mihar-22, line 16: 1. If this is the same styles used by hls video as well, can we share them? 
2. Can we share `getTemplateHTML` in general to save on size and avoid duplicating?

packages/html/src/media/background-video/index.ts
```diff
@@ -12,17 +10,22 @@ function getTemplateHTML(attrs: Record<string, string>) {
...
       }
     </style>
     <slot></slot>
     <video ${serializeAttributes(attrs)}></video>
   `;
 }
 
-export class BackgroundVideo extends CustomMediaMixin(HTMLElement, { tag: 'video' }) {
+// Don't extend CustomMediaMixin to save some bytes, background videos don't need the full Media API.
+export class BackgroundVideo extends HTMLElement {
```
mihar-22, line 25: Nice 👍

packages/sandbox/templates/html-background/main.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import '@videojs/html/background/skin.css';
+import '@videojs/html/background/player';
+import '@videojs/html/background/skin';
+import '@videojs/html/background/video';
+
+const html = String.raw;
+
+document.getElementById('root')!.innerHTML = html`
+  <background-video-player style="display: contents;">
```
mihar-22: Should we bake display contents into skin.css?

packages/html/src/define/background/skin.css
```diff
@@ -0,0 +1,23 @@
+background-video-skin {
+  display: block;
+  position: relative;
+  width: 100%;
+  height: 100%;
+  --media-object-fit: cover;
+  object-fit: var(--media-object-fit);
```
mihar-22, line 11: Assuming this is so it cascades down to underlying media? Feels more natural to use var default `var(x, cover)` but understandable if it's sharing.

packages/html/src/define/background/skin.ts
luwes: Will do

packages/html/src/media/background-video/index.ts
luwes: For background video it's different because we always want to expand the media to bounds of its container.
I'd wait with DRYing up this one.

packages/sandbox/templates/html-background/main.ts
luwes: Good call, in the first iteration I didn't use the external CSS file. I thought we wanted the styles in the shadow DOM of the skin but I guess we're doing both? so then this is possible yes.

Issue #622, opened 2026-02-26
UI: Playback Rate Button Component


decepulis:
Implement the PlaybackRateButton component for cycling through available playback rates. Includes core logic, DOM props builder, and React/HTML implementations.

Displays the current playback rate and cycles to the next rate on click (e.g., 1x → 1.5x → 2x → 0.5x → 1x).

## References

- [Media Chrome: Playback Rate Button](https://www.media-chrome.org/docs/en/components/media-playback-rate-button)

PR #624, opened 2026-02-26
docs(design): PlaybackRateButton component spec

decepulis:
## Summary

- Design doc for the PlaybackRateButton component (#622)
- Three decisions documented:
  1. **Store** — Extend `MediaPlaybackState` with `playbackRate` + `changePlaybackRate()` rather than a new feature slice
  2. **ReactiveElement** — Add a `converter` option to `PropertyDeclaration` for custom attribute-to-property conversion (needed for the space-separated `rates` attribute, but general-purpose)
  3. **Component** — Three-layer (Core + HTML + React) cycle button, headless, default rates `[1, 1.2, 1.5, 1.7, 2]`, rate exposed via `data-rate`

Looking for feedback on all fronts, especially Store and ReactiveElement where reasonable alternatives exist

## Doc

[`internal/design/ui/playback-rate-button.md`](https://github.com/videojs/v10/blob/design/playback-rate-button/internal/design/ui/playback-rate-button.md)

### DIFF
--- a/internal/design/ui/playback-rate-button.md
+++ b/internal/design/ui/playback-rate-button.md
@@ -11,110 +11,60 @@ Button for cycling through playback speeds. Displays the current rate and advanc
 
 Users want to speed up or slow down video playback. A button that cycles through preset rates is a common pattern — compact, one-click, no menu required.
 
-Three things are needed:
+Two things are needed:
 
 1. **Store feature** — `playbackRate` isn't tracked in the store yet. No state, no selector, no action.
-2. **ReactiveElement enhancement** — `ReactiveElement` can't handle array attributes. If we want an interface like `rates="1 1.25 1.5 1.75 2"`, then prop needs a way to handle that.
-3. **Button component** — Three-layer implementation (Core + HTML + React) following existing button patterns.
+2. **Button component** — Three-layer implementation (Core + HTML + React) following existing button patterns.
 
-## Store: Extend `MediaPlaybackState`
+## Store: New `MediaPlaybackRateState`
 
-Playback rate should go in the existing playback feature — it's a property of playback, not a separate concern. (Very open to pushback; alternatives discussed below.)
+Playback rate gets its own feature and state slice — separate from `MediaPlaybackState`.
 
-### State Addition
+### State
 
 ```ts
-export interface MediaPlaybackState {
-  paused: boolean;
-  ended: boolean;
-  started: boolean;
-  waiting: boolean;
-  playbackRate: number; // new
-  play(): Promise<void>;
-  pause(): void;
-  changePlaybackRate(rate: number): void; // new
+export interface MediaPlaybackRateState {
+  readonly playbackRates: readonly number[];
+  playbackRate: number;
+  setPlaybackRate(rate: number): void;
 }
 ```
 
-### Feature Addition
+`playbackRates` is a readonly array of available rates, statically set to `[1, 1.2, 1.5, 1.7, 2]` for now. Later, providers can populate this dynamically (e.g., YouTube's `getAvailablePlaybackRates`).
+
+### Feature
 
 ```ts
-// In playbackFeature
-state: ({ target }): MediaPlaybackState => ({
-  // ... existing state
-  playbackRate: 1,
-  changePlaybackRate(rate: number) {
-    target().media.playbackRate = rate;
+const DEFAULT_RATES: readonly number[] = [1, 1.2, 1.5, 1.7, 2];
+
+export const playbackRateFeature = createFeature({
+  state: ({ target }): MediaPlaybackRateState => ({
+    playbackRates: DEFAULT_RATES,
+    playbackRate: 1,
+    setPlaybackRate(rate: number) {
+      target().media.playbackRate = rate;
+    },
+  }),
+
+  attach({ target, signal, set }) {
+    const { media } = target;
+
+    const sync = () => set({ playbackRate: media.playbackRate });
+
+    listen(media, "ratechange", sync, { signal });
   },
-}),
-
-attach({ target, signal, set }) {
-  const { media } = target;
-
-  const sync = () =>
-    set({
-      // ... existing sync
-      playbackRate: media.playbackRate,
-    });
-
-  // ... existing listeners
-  listen(media, 'ratechange', sync, { signal });
-},
+});
 ```
 
-### Why Not a Separate Feature
+### Why a Separate Feature
 
-**Alternatives:**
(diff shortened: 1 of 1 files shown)


internal/design/ui/playback-rate-button.md
```diff
@@ -0,0 +1,269 @@
...
+### `PropertyDeclaration` Change
+
+```ts
+export interface PropertyDeclaration {
+  readonly type?: typeof String | typeof Boolean | typeof Number;
+  readonly attribute?: string;
+  /** Custom attribute-to-property conversion. When set, `type` is ignored. */
+  readonly converter?: (value: string | null) => unknown;
+}
+```
```
mihar-22: This is correct, for context we're keeping our elements compatible with Lit (see [Properties](https://lit.dev/docs/v1/components/properties/#declare)).  We should make sure it aligns with their [converter API](https://lit.dev/docs/v1/components/properties/#conversion-converter). 

However, if we don't expose `rates` now then we don't need this. Keeping it simple.

internal/design/ui/playback-rate-button.md
```diff
@@ -0,0 +1,269 @@
...
+
+One branch before the existing type checks:
+
+```ts
+if (decl.converter) {
+  value = decl.converter(newValue);
+} else if (decl.type === Boolean) {
+  // ... existing logic unchanged
+}
+```
```
mihar-22: 👍

internal/design/ui/playback-rate-button.md
```diff
@@ -0,0 +1,269 @@
...
+- **Custom getter** — `get rates()` parsing the attribute manually. Class getters are non-enumerable, so `setProps(this)` silently skips `rates` and the `satisfies` constraint needs `Exclude`. Two one-off exceptions to standard patterns.
+- **`AttributeTokenList`** — Media Chrome's approach. A `DOMTokenList`-like class that syncs a space-separated attribute, with `add`/`remove`/`toggle` per token. More infrastructure, and still needs conversion to `number[]` for the Core.
+- **`Array` as a built-in type** with `JSON.parse` default — One of Lit's approaches, alongside `converter`. Produces `rates='[1, 1.5, 2]'` in HTML, which is awkward for HTML authors. Honestly, though, my favorite of the rejected ideas.
+
+### Rationale
+
+The `converter` option is ~10 lines of implementation in `ReactiveElement`, follows one of [Lit's established pattern](https://lit.dev/docs/components/properties/), and is general purpose — any future property needing custom attribute parsing (comma-separated lists, enums with validation, etc.) uses the same mechanism.
```
mihar-22: 👍 

Great noted here as per my callout above.

internal/design/ui/playback-rate-button.md
```diff
@@ -0,0 +1,269 @@
...
+  /* With rate display */
+}
+<PlaybackRateButton
+  render={(props, state) => <button {...props}>{state.rate}x</button>}
+/>;
+
+{
+  /* Custom rates */
+}
+<PlaybackRateButton rates={[0.5, 1, 1.5, 2]} />;
```
mihar-22: I don't think we should expose custom rates yet as it requires small conversation. Reason being that some Media like embeds only support specific playback rates (e.g., [YouTube#getAvailablePlaybackRates](https://developers.google.com/youtube/iframe_api_reference#getAvailablePlaybackRates)).

mihar-22: I'd recommend having readonly `playbackRates` array in the store feature which is statically set atm.

internal/design/ui/playback-rate-button.md
```diff
@@ -0,0 +1,269 @@
...
+- **Extend `playbackFeature`** — Add `playbackRate` and `changePlaybackRate` to existing state.
+
+**Decision:** Extend `playbackFeature`.
+
+**Rationale:**
+
+- Conceptually cohesive — rate is a property of playback. The native API puts it on the same `HTMLMediaElement` alongside `play()`/`pause()`.
+- Already listens to related events — adding `ratechange` fits naturally.
+- Reduces selector proliferation — a separate `selectPlaybackRate` for one property is heavy. Consumers using `selectPlayback` get rate for free.
+- Rate changes are rare — minimal re-render concern from including it in the playback slice.
```
mihar-22: Thinking we should separate this state into `MediaPlaybackRateState`.  Feels self-contained, but I can see why "MediaPlaybackState" is enticing.  I'm leaning towards more granular/capability contracts.

Reason: this specific state is usually used separately in the UI, and it'll contain code in the future to check if playback rate is supported by the underlying Media. Example being Vimeo required Pro in the past, might still be the case.

internal/design/ui/playback-rate-button.md
```diff
@@ -0,0 +1,269 @@
...
+
+```ts
+static override properties = {
+  rates: {
+    converter(value) {
+      const defaults = PlaybackRateButtonCore.defaultProps.rates;
+      if (!value) return defaults;
+      const parsed = value.split(/\s+/).map(Number).filter((n) => !Number.isNaN(n));
+      return parsed.length > 0 ? parsed : defaults;
+    },
```
mihar-22: not required if we don't expose

internal/design/ui/playback-rate-button.md
```diff
@@ -0,0 +1,269 @@
...
+```ts
+export interface MediaPlaybackState {
+  paused: boolean;
+  ended: boolean;
+  started: boolean;
+  waiting: boolean;
+  playbackRate: number; // new
+  play(): Promise<void>;
+  pause(): void;
+  changePlaybackRate(rate: number): void; // new
```
luwes:
```suggestion
  setPlaybackRate(rate: number): void; // new
```

nit(non-blocking): I'd go for the shorter, more common naming of using `set` as prefix: `setPlaybackRate()`. Curious though if there are reasons not to. 

I also think we should rename `changeVolume` to `setVolume` and keep this consistent across the store API. Lastly `toggleMute` should be `toggleMuted`. cc @mihar-22

luwes: I know last things are out of scope for this PR, just wanted to put it somewhere.

mihar-22: Yep sounds good to me @luwes, we can ticket up and go for it. I don't have any issues with that change in another PR for places where it's not aligned

internal/design/ui/playback-rate-button.md
decepulis: Good shout. Not exposing `rates` but made a note for later

internal/design/ui/playback-rate-button.md
decepulis: Agreed

internal/design/ui/playback-rate-button.md
decepulis: Agreed

internal/design/ui/playback-rate-button.md
decepulis: Not exposed

internal/design/ui/playback-rate-button.md
decepulis: Changed to `setPlaybackRate`

Issue #501, opened 2026-02-10
UI: Thumbnail Component


mihar-22:
Implement the Thumbnail component for displaying video thumbnail images. Parses VTT sprite sheets and renders cropped regions. Includes core logic, DOM props builder, and React/HTML implementations.

## References

- [Vidstack: Thumbnail](https://www.vidstack.io/docs/player/components/display/thumbnail)
- [Media Chrome: Preview Thumbnail](https://www.media-chrome.org/docs/en/components/media-preview-thumbnail)

PR #648, opened 2026-02-27
feat(react): add thumbnail component

mihar-22:
Ref: https://github.com/videojs/v10/issues/501
Stacked on: #643

## Summary

`<Thumbnail>` React component.

### `<Thumbnail>`

| | |
|---|---|
| Props | `time?: number`, `className`, `style`, `render` (via `UIComponentProps`) |
| Renders | `<div>` (overflow hidden) + `` |
| Sizing | `max-width` / `max-height` via `style` prop or CSS → auto-scales proportionally |
| Data attrs | `data-loading`, `data-error`, `data-hidden` |
| Ref | `ForwardedRef<HTMLDivElement>` |

```tsx
<Thumbnail time={hoverTime} style={{ maxWidth: 200, maxHeight: 150 }} />
```

### Namespace

```ts
Thumbnail.Props  // ThumbnailProps
Thumbnail.State  // ThumbnailCore.State
```

### DIFF
--- a/packages/react/src/ui/thumbnail/thumbnail.tsx
+++ b/packages/react/src/ui/thumbnail/thumbnail.tsx
@@ -1,113 +1,95 @@
 'use client';
 
-import { mapCuesToThumbnails, type ThumbnailConstraints, ThumbnailCore, ThumbnailDataAttrs } from '@videojs/core';
-import { selectTextTrack } from '@videojs/core/dom';
-import type { ForwardedRef } from 'react';
-import { forwardRef, useCallback, useLayoutEffect, useMemo, useRef, useState } from 'react';
+import {
+  mapCuesToThumbnails,
+  ThumbnailCore,
+  ThumbnailDataAttrs,
+  type ThumbnailFetchPriority,
+  type ThumbnailImage,
+} from '@videojs/core';
+import { createThumbnail, selectTextTrack } from '@videojs/core/dom';
+import type { CSSProperties } from 'react';
+import { forwardRef, useEffect, useMemo, useRef, useState } from 'react';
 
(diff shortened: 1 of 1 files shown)


packages/react/src/ui/thumbnail/thumbnail.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+export interface ThumbnailProps extends UIComponentProps<'div', ThumbnailCore.State>, ThumbnailCore.Props {
+  /** Pre-parsed thumbnail images — bypasses the automatic `<track>` detection. */
+  thumbnails?: ThumbnailImage[] | undefined;
+}
+
+export const Thumbnail = forwardRef(function Thumbnail(
+  componentProps: ThumbnailProps,
+  forwardedRef: ForwardedRef<HTMLDivElement>
+) {
```
sampotts: nit:
```
const Thumbnail = forwardRef<HTMLDivElement, ThumbnailProps>(
  function Thumbnail(props, ref) {
    // ...
  }
);
```

The above is the pattern I've always been exposed to but here's ChatGPTs explanation on why:

Pros
	•	Best inference + best consumer type: you reliably get ref correctly typed when someone uses <Thumbnail ref={...} />.
	•	Cleaner function signature (no repeated type noise).
	•	Usually the least-friction with JSX + IDE help.

Cons
	•	Slightly less “local” explicitness (types are up at the callsite, not on params).

mihar-22: Thanks for catching that! Missed it. Fixing...

mihar-22: ye this is awkward, claude not performing today

Issue #220, opened 2025-12-17
UI: Popover Component


mihar-22:
Blocked by: 

- #219
- #221
- #242 

Implement a compound Popover component for Video.js 10, providing the foundation for tooltips, menus, and other floating UI elements. 

The Popover handles positioning, interactions, transitions/animations, safe polygons, and collision avoidance.

## References

- [Base UI: Popover](https://base-ui.com/react/components/popover) — Primary API reference
- [Radix: Popover](https://www.radix-ui.com/primitives/docs/components/popover)
- [Popover Attribute (MDN)](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/popover)
- [CSS Anchor Positioning](https://caniuse.com/css-anchor-positioning)
- [Floating UI](https://floating-ui.com/) — Positioning engine (fallback)
- https://github.com/videojs/v10/discussions/198

PR #650, opened 2026-02-27
fix(packages): fix popover html element types and react focus/positioning

mihar-22:
Refs #220

## Summary

Fix platform-specific bugs in the HTML and React popover implementations discovered during audit of #615.

## Changes

- Fix `modal` attribute type from `String` to `Boolean` on `PopoverElement`
- Reuse `SnapshotController` via `track()` on reconnect to prevent leaking stale controllers
- Remap `onFocusIn`/`onFocusOut` to React's `onFocus`/`onBlur` in trigger and popup components
- Sanitize `useId()` output for valid CSS `<dashed-ident>` tokens in anchor names
- Add RAF-throttled scroll/resize listeners for manual positioning fallback in React

<details>

**SnapshotController leak**: The HTML element created a new `SnapshotController` on every `connectedCallback`, but `ReactiveElement` never removes old controllers. Fixed by storing the controller reference and calling `track()` to switch the tracked state container.

**React focus events**: `createPopover()` emits `onFocusIn`/`onFocusOut` (matching DOM `focusin`/`focusout`), but React maps those to `onFocus`/`onBlur`. Without remapping, focus-based open/close was silently broken.

**Anchor name sanitization**: React's `useId()` returns values like `:r0:` — the colons are invalid in CSS `<dashed-ident>` tokens used by `anchor-name`/`position-anchor`, causing anchor positioning to silently fail.

</details>

## Testing

```bash
pnpm -F @videojs/core test src/dom/ui/popover
pnpm -F @videojs/react test
```

### DIFF
--- a/packages/react/src/ui/popover/popover-popup.tsx
+++ b/packages/react/src/ui/popover/popover-popup.tsx
@@ -71,18 +71,38 @@ export const PopoverPopup = forwardRef(function PopoverPopup(
       return;
     }
 
-    const triggerEl = popover.triggerElement;
-    const popupEl = internalRef.current;
-    if (!triggerEl || !popupEl) return;
+    function measure(): void {
+      const triggerEl = popover.triggerElement;
+      const popupEl = internalRef.current;
+      if (!triggerEl || !popupEl) return;
 
-    const triggerRect = triggerEl.getBoundingClientRect();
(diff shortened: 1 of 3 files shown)


mihar-22:
Splitting into separate PRs for html and react.

Issue #498, opened 2026-02-10
UI: Captions Display Component


mihar-22:
Implement the Captions display component for rendering active text track cues over video. Includes core logic, DOM props builder, and React/HTML implementations.

**Blocked by:**
- #496 (Text Track Store Feature)

## References

- [Vidstack: Captions](https://www.vidstack.io/docs/player/components/display/captions)

PR #692, opened 2026-03-03
feat: add subtitles handling + captions core

### LINKED ISSUE #687: HLS: subtitles handling on source change
Hls.js historically requires extra logic to handle subtitles when changing the source.

luwes:
fix #687, https://github.com/videojs/v10/issues/498, https://github.com/videojs/v10/issues/498

### DIFF
--- a/packages/core/src/dom/store/features/tests/text-track.test.ts
+++ b/packages/core/src/dom/store/features/tests/text-track.test.ts
@@ -159,8 +159,8 @@ describe('textTrackFeature', () => {
       store.attach({ media: video, container: null });
 
       expect(store.state.subtitlesList).toEqual([
-        { kind: 'subtitles', label: 'English', language: 'en', showing: true },
-        { kind: 'captions', label: 'CC', language: 'en', showing: false },
+        { kind: 'subtitles', label: 'English', language: 'en', mode: 'showing' },
+        { kind: 'captions', label: 'CC', language: 'en', mode: 'disabled' },
       ]);
     });
 


packages/core/src/core/media/state.ts
```diff
@@ -197,13 +197,26 @@ export interface MediaTextCue {
   text: string;
 }
 
+export interface MediaSubtitleTrack {
```
mihar-22: (suggestion): should we call this `MediaTextTrack`?

packages/core/src/core/media/state.ts
```diff
@@ -197,13 +197,26 @@ export interface MediaTextCue {
...
   /** Cues from the first `kind="metadata" label="thumbnails"` track. */
   thumbnailCues: MediaTextCue[];
   /** The `<track>` element's `src` for resolving relative cue text URLs. */
   thumbnailTrackSrc: string | null;
+  /** Caption/subtitle tracks that can be selected or toggled. */
+  subtitlesList: MediaSubtitleTrack[];
+  /** Whether captions/subtitles are currently enabled. */
+  subtitlesShowing: boolean;
+  /** Toggle captions/subtitles visibility. Returns the new enabled value. */
+  toggleSubtitles(forceShow?: boolean): boolean;
```
mihar-22, line 243: (suggestion): should we rename these to `textTrackList` (platform) and then `caption*` to align with how we're naming UI etc? E.g.,

- `textTrackList`
- `captionsShowing`
- `toggleCaptions`

mihar-22: (question): We've used the word `visible` as a generic UI word for something showing, would that fit here? E.g., `captionsVisible`, making toggle `toggleCaptionsVisibility`?

packages/core/src/core/media/state.ts
```diff
@@ -197,13 +197,26 @@ export interface MediaTextCue {
   text: string;
 }
 
+export interface MediaSubtitleTrack {
+  kind: 'captions' | 'subtitles';
+  label: string;
+  language: string;
+  showing: boolean;
```
mihar-22: (question): should we align with `mode`?  (https://developer.mozilla.org/en-US/docs/Web/API/TextTrack/mode). I'll admit not the clearer API but aligning with platform would be nice.

packages/core/src/core/ui/captions-button/captions-button-data-attrs.ts
```diff
@@ -0,0 +1,7 @@
+import type { StateAttrMap } from '../types';
+import type { CaptionsButtonState } from './captions-button-core';
+
+export const CaptionsButtonDataAttrs = {
+  /** Present when captions are enabled. */
+  subtitlesShowing: 'data-subtitles-showing',
```
mihar-22: (request): Feels like a long data attribute name and not familiar with others so far, thoughts on `data-visible`? Related comment above.

packages/html/src/ui/captions-button/captions-button-element.ts
mihar-22, line 1: (request): I did a pass recently to dry up buttons to avoid boilerplate, rebasing on main and aligning with latest would be nice. Your feedback there would be appreciated if you have any :)

packages/react/src/ui/captions-button/captions-button.tsx
mihar-22, line 1: (request): React has also been made more dry on main. Rebasing will clean this up.

packages/utils/src/dom/text-track.ts
```diff
@@ -5,3 +5,46 @@ export function findTrackElement(media: HTMLMediaElement, track: TextTrack): HTM
...
+  return ['subtitles', 'captions'].includes(textTrack.kind);
+}
+
+function sortByTextTrackKind(a: TextTrack, b: TextTrack): number {
+  return a.kind >= b.kind ? 1 : -1;
+}
+
+function alwaysTrue(): boolean {
+  return true;
+}
```
mihar-22, line 50: (question): seems unused? maybe I missed it

packages/core/src/dom/media/custom-media-element/index.ts
mihar-22, line 1: (question): Not familiar with this, worth Christian having a look? If you feel confident then no worries.

packages/core/src/core/media/state.ts
luwes: Good calls, in MC we went from the captions naming to the more generic subtitles naming to cover both. 
I think in the UK subtitles covers both, maybe Australia then too?

I don't think we want to align state naming with UI naming. It could, but here we're toggling on / off both captions and subtitles so a more generic term works.

packages/core/src/core/media/state.ts
luwes: yes, agreed. AI did this, wasn't me 😅

packages/core/src/core/ui/captions-button/captions-button-data-attrs.ts
luwes: yes, I wasn't sure if we wanted the data selectors to be more specific or not. 
what is most useful to use in skins? if they're more generic we need to scope with the element class
cc @sampotts

sampotts: We're using `data-visible` elsewhere for elements like the controls, buffering and error dialog so might be good to be consistent here.

Issue #199, opened 2025-11-17
UI: Error Dialog Component


heff:
### Should this be a feature proposal discussion?

- [x] This is not a substantial change

### Is there an existing issue for this?

- [x] I have searched the existing issues and discussions

### Which package(s) does this apply to? Select all that apply

React (@videojs/react), Core (@videojs/core)

### Motivation

Use the player

### Description

When there is a player error, the player should inform the viewer.

With the current example embed code there is a fake URL and when you copy/paste the code it currently fails silently.

### Expected Behavior

The error is displayed.

### Alternatives and Workarounds

asdf

PR #713, opened 2026-03-04
feat(core): add error feature

mihar-22:
Ref #199

## Summary

Add error state management to the player store. This is the data layer for the error dialog — UI components (AlertDialog, ErrorDialog) will follow in separate PRs.

## Changes

- `MediaError` + `MediaErrorState` types (`error`, `dismissError()`)
- `errorFeature` store feature — syncs `error`/`emptied` events from the media element
- `selectError` selector, wired into `videoFeatures` preset (video only)
- Expose `set` in `StateContext` so action closures can mutate state directly

## Testing

`pnpm -F @videojs/store test && pnpm -F @videojs/core test` — 633 tests pass

### DIFF
--- a/packages/core/src/core/media/state.ts
+++ b/packages/core/src/core/media/state.ts
@@ -206,6 +206,32 @@ export interface MediaTextTrackState {
   thumbnailTrackSrc: string | null;
 }
 
+export interface MediaError {
+  /**
+   * The error code (mirrors MediaError.code constants).
+   *
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/MediaError/code
+   */
+  code: number;
+  /**
+   * A human-readable error message.
+   *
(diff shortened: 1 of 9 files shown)


packages/core/src/core/media/state.ts
```diff
@@ -206,6 +206,34 @@ export interface MediaTextTrackState {
   thumbnailTrackSrc: string | null;
 }
 
+export interface MediaPlaybackError {
```
sampotts: nit: Out of interest, why not just `MediaError`? To avoid confusion?

mihar-22: haven't reviewed yet, changing to `MediaError` :)

sampotts: Cool. I agree with your comment on Wes' PR of using the native/platform naming conventions where suitable 👍🏼

PR #748, opened 2026-03-06
fix(html): add destroy

### LINKED ISSUE #689: Bug: StoreError: DESTROYED when using VideoProvider with React StrictMode

React (@videojs/react), Core (@videojs/core)

  - @videojs/react: 10.0.0-alpha.4
  - @videojs/core: 10.0.0-alpha.4
  - @videojs/store: 10.0.0-alpha.4
  - React: 19.0.0
  - Vite: 6.2.1
  - Node: 22.x
  - OS: macOS

### Description

Using any `@videojs/react` Provider component (e.g., `VideoProvider` from the video preset or a custom `createPlayer` Provider) inside React's `<StrictMode>` throws `StoreError: DESTROYED` on mount.

### Reduced test case

_No response_

### Steps to reproduce

1. Create a Vite + React project:
```
  npm create vite@latest vjs-repro -- --template react-ts
  cd vjs-repro
  npm install
  npm install @videojs/react@next
```

  2. Replace src/App.tsx with:
 
```
  import { VideoProvider, Video } from '@videojs/react/video';
  import { Container } from '@videojs/react';

  function App() {
    return (
      <VideoProvider>
        <Container>
          <Video src="https://test-videos.co.uk/vids/bigbuckbunny/mp4/h264/720/Big_Buck_Bunny_720_10s_1MB.mp4" />
        </Container>
      </VideoProvider>
    );
  }

  export default App;
```

  3. Confirm src/main.tsx uses <StrictMode> (the Vite default):
```
  import { StrictMode } from 'react'
  import { createRoot } from 'react-dom/client'
  import App from './App.tsx'

  createRoot(document.getElementById('root')!).render(
    <StrictMode>
      <App />
    </StrictMode>,
  )
```
  4. Run the dev server: `npm run dev`
  5. Open the browser console.

### Current Behavior

The app crashes with an error boundary or blank screen.

### Root Cause

In packages/react/src/player/create-player.tsx, the Provider creates the store via useState and destroys it in a useEffect cleanup:

```
  function Provider({ children }) {
    const [store] = useState(() => createStore()(combine(...config.features)));
    const [media, setMedia] = useState(null);
    useEffect(() => () => store.destroy(), [store]);  // ← problem here
    // ...
  }
```

React StrictMode simulates an unmount/remount cycle in development:

  1. Mount — useState initializer creates a fresh store
  2. Effect runs — registers cleanup: () => store.destroy()
  3. StrictMode simulated unmount — cleanup runs, store.destroy() is called, setting destroyed = true
  4. StrictMode simulated remount — effects re-run, but useState returns the same (now destroyed) store instance
  5. Container effect — calls store.attach(...), which checks if (destroyed) throwDestroyedError() → crash

### Suggested Fix (Claude):

Replace the useState + useEffect pattern with one that can recover from StrictMode's double-mount. For example, using useRef to detect and recreate a destroyed store:

```
  function Provider({ children }) {
    const storeRef = useRef(null);
    if (!storeRef.current || storeRef.current.destroyed) {
      storeRef.current = createStore()(combine(...config.features));
    }
    const store = storeRef.current;
    const [media, setMedia] = useState(null);
    useEffect(() => () => { store.destroy(); }, [store]);
    // ...
  }
```

This way, when `StrictMode` remounts and the effect re-runs, the render phase detects the destroyed store and creates a fresh one before Container tries to `attach()`.

### Workaround:

Remove <StrictMode> from the app's entry point.

### Expected Behavior

  The player renders without errors.

### Errors

Uncaught StoreError: DESTROYED
      at throwDestroyedError (store.js)
      at Object.attach (store.js)
      at context.js
      ...

10.0.0-alpha.4

mihar-22:
Closes #689

## Summary

Prevents `StoreError: DESTROYED` when frameworks (React, Lit, etc.) temporarily disconnect and reconnect custom elements during reconciliation. Elements now defer destruction by 2 rAF ticks after disconnect — if still disconnected, they destroy; if reconnected, the `isConnected` check skips destruction.

## Changes

- `DestroyMixin` in `@videojs/element` — composable mixin on `ReactiveElement` that adds `destroy()`, `destroyCallback()`, `destroyed`, deferred 2-rAF destruction on disconnect, `keep-alive` attribute opt-out, controller tracking with `hostDestroyed()` notification, and `performUpdate` guard
- `MediaElement` simplified to `DestroyMixin(ReactiveElement)` — all destruction logic lives in the mixin
- `HlsVideo` defers `this.destroy()` by 2 rAFs (goes through delegate proxy to HLS engine)
- Store destruction moved from `disconnectedCallback` to `destroyCallback` in `ProviderMixin`
- Imperative API destruction (slider, popover, thumbnail) moved to `destroyCallback`
- `hostDestroyed?()` added to `ReactiveController` interface

<details>

**Why not on `ReactiveElement`?** It's used as a structural type constraint by `SkinMixin` and other composites. Adding public members would break Lit compatibility. The destroy lifecycle lives in `DestroyMixin`; `ReactiveElement` gets a minimal `hostDestroyed?()` hook on the controller interface. See `internal/design/reactive-element-lit-compat.md`.

**Why 2 rAF ticks?** A single rAF can fire before the browser has processed the re-insertion. Two ticks ensure the element has had a full frame to reconnect. The callback checks `isConnected` so reconnected elements survive.

**ContainerMixin unchanged** — It already detaches immediately (lightweight) and re-attaches on reconnect. The store survives because the provider's destruction is deferred.

**HlsVideo uses inline rAFs** — `DestroyMixin` is not applied to `HlsCustomMedia` because it would shadow the delegate-proxied `destroy()` (fixed in #751). Instead, `HlsVideo.disconnectedCallback` schedules the 2-rAF check directly.

</details>

## Testing

```bash
pnpm -F @videojs/element test src/tests/destroy-mixin.test.ts
pnpm -F @videojs/html test
```

13 new tests in `destroy-mixin.test.ts` covering: destroy flag, idempotency, destroyCallback, hostDestroyed notification, connectedCallback guard, performUpdate guard, 2-rAF deferred destruction, reconnect survival, keep-alive attribute.

### DIFF
--- a/internal/design/reactive-element-lit-compat.md
+++ b/internal/design/reactive-element-lit-compat.md
@@ -1,0 +1,52 @@
+---
+status: decided
+date: 2026-03-05
+---
+
+# Destroy Lifecycle: DestroyMixin
+
+## Decision
+
+The deferred destruction lifecycle is implemented as a **composable mixin** (`DestroyMixin`)
+exported from `@videojs/element`, rather than being added directly to `ReactiveElement`.
+
+`ReactiveElement` stays structurally compatible with Lit's `ReactiveElement`.
+
+## Architecture
+
+```
+@videojs/utils/dom   → deferFrames() composable utility
+@videojs/element     → DestroyMixin(HTMLElement) — generic, works with any base
+@videojs/html        → MediaElement = DestroyMixin(ReactiveElement) + hostDestroyed()
+@videojs/core        → HlsCustomMedia = DestroyMixin(HTMLElement) via CustomMediaMixin
+```
+
+### DestroyMixin
+
+Adds `destroyed`, `destroy()`, `destroyCallback()`, deferred 2-rAF scheduling on
+disconnect, cancel on reconnect, and `keep-alive` attribute support. Works with any
+`HTMLElement` subclass.
+
+### MediaElement
+
+Composes `DestroyMixin(ReactiveElement)` and bridges `destroyCallback()` to
+`hostDestroyed()` on reactive controllers. Also guards `performUpdate()` when destroyed.
+
+### HlsCustomMedia
+
+Uses `DestroyMixin(HTMLElement)` as the base for `CustomMediaMixin`. The `DelegateMedia`
+class (from `MediaDelegateMixin`) overrides `destroyCallback()` to call
+`this.#delegate.destroy?.()`, which destroys the HLS engine.
+
+## Why a Mixin
+
+- `ReactiveElement` stays Lit-compatible (no added public surface)
+- Works with `HTMLElement` directly (no `ReactiveElement` required for custom media elements)
+- Composable — any element can opt in
+
+## Alternatives Considered
+
+- **Put everything on `ReactiveElement`** — Simpler (direct `#controllers`
+  access), but breaks Lit structural compat. Rejected.
+- **Separate `MediaElement` class without mixin** — Doesn't compose with
+  `CustomMediaMixin(HTMLElement)` for custom media elements. Rejected.
--- a/packages/element/src/destroy-mixin.ts
+++ b/packages/element/src/destroy-mixin.ts
@@ -1,0 +1,83 @@
+import type { ReactiveElement } from './reactive-element';
+import type { ReactiveController } from './types';
+
+export interface Destroyable {
+  readonly destroyed: boolean;
+  destroy(): void;
+  destroyCallback(): void;
+}
+
+/**
+ * Mixin that adds a deferred destruction lifecycle to a `ReactiveElement`.
+ *
+ * On disconnect, schedules destruction after two animation frames.
+ * If the element reconnects before the frames fire (e.g. DOM shuffling,
+ * framework reconciliation), the `isConnected` check prevents destruction.
+ *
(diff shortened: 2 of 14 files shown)


packages/core/src/core/media/delegate.ts
```diff
@@ -50,6 +51,10 @@ export function MediaDelegateMixin<Base extends Constructor<any>, Delegate exten
       this.#delegate.detach?.();
       super.detach?.();
     }
+
+    destroy(): void {
+      this.#delegate.destroy?.();
+    }
```
luwes: This is not the right way. It was a bug...
See PR https://github.com/videojs/v10/pull/751

luwes (dismissed):

packages/core/src/core/media/delegate.ts
```diff
@@ -3,6 +3,7 @@ import type { Constructor } from '@videojs/utils/types';
 export interface MediaDelegate {
   attach?(target: EventTarget): void;
   detach?(): void;
+  destroy?(): void;
```
luwes: Not needed with fix below

PR #784, opened 2026-03-09
feat(spf): initial push of SPF

cjpillsbury:
VJS media names as `SimpleHlsVideo` (and related) based off of convo with @heff last Friday. Changes outside of SPF should be narrow and backwards compatible.

Testing in sandbox:
1. direct spf playback-engine composition test harness - http://localhost:5173/spf-segment-loading/
2. simple react test page - http://localhost:5173/simple-hls-react/
3. simple html/web component test page - http://localhost:5173/simple-hls-html/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -123,6 +123,9 @@ Pods/
 # Vercel
 .vercel
 
+# Netlify
+.netlify
+
 # -------------------------
 # Package Manager Locks
 # Keep only one when enforcing workspace consistency
--- a/packages/spf/src/utils/index.ts
+++ b/packages/spf/src/utils/index.ts
@@ -1,3 +1,0 @@
-/**
- * Shared SPF utilities
- */


packages/core/src/dom/media/custom-media-element/index.ts
```diff
@@ -434,6 +434,9 @@ export function CustomMediaMixin<T extends Constructor<HTMLElement>>(
 
     connectedCallback(): void {
       this.#init();
+      // Set after #init() so the shadow root exists; safe here since
```
cjpillsbury, line 437: **NOTE:** Not sure if this is still needed after the latest iterations of these abstractions.

packages/core/src/dom/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './feature';
+export * from './media/simple-hls';
```
cjpillsbury: **NOTE:** Looks like this shouldn't be here (I believe we want to use the media-specific exports?), but would be good to confirm.

packages/html/src/store/container-mixin.ts
```diff
@@ -49,7 +49,12 @@ export function createContainerMixin<Store extends PlayerStore>(context: PlayerC
           if (records.some(hasMediaNode)) this.#attachMedia();
         });
 
-        this.#observer.observe(this, { childList: true, subtree: true });
+        this.#observer.observe(this, {
```
cjpillsbury, line 52: **NOTE:** Not sure if this is still needed after the latest iterations of these abstractions. (same as above).

packages/html/src/store/container-mixin.ts
```diff
@@ -87,7 +92,7 @@ export function createContainerMixin<Store extends PlayerStore>(context: PlayerC
         const store = this.#contextStore ?? this.store;
         if (!store) return;
 
-        const media = this.querySelector<HTMLMediaElement>('video, audio') ?? this.#getSlottedMedia();
+        const media = this.querySelector<HTMLMediaElement>('video, audio, [data-media-element]') ?? this.#getSlottedMedia();
```
cjpillsbury, line 95: **NOTE:** Not sure if this is still needed after the latest iterations of these abstractions. (same as above).

packages/html/src/store/container-mixin.ts
```diff
@@ -115,10 +120,15 @@ export function createContainerMixin<Store extends PlayerStore>(context: PlayerC
 }
 
 function isMediaNode(node: Node): boolean {
-  return node instanceof HTMLMediaElement;
+  return node instanceof HTMLMediaElement || (node instanceof Element && node.hasAttribute('data-media-element'));
```
cjpillsbury, line 123: **NOTE:** Not sure if this is still needed after the latest iterations of these abstractions. (same as above).

packages/html/src/store/container-mixin.ts
```diff
@@ -115,10 +120,15 @@ export function createContainerMixin<Store extends PlayerStore>(context: PlayerC
...
 }
 
 function hasMediaNode(record: MutationRecord): boolean {
+  // Attribute mutation: data-media-element was added to a descendant
+  if (record.type === 'attributes' && record.target instanceof Element) {
```
cjpillsbury, line 128: **NOTE:** Not sure if this is still needed after the latest iterations of these abstractions. (same as above).

packages/html/package.json
```diff
@@ -87,6 +87,7 @@
     "clean": "rm -rf dist types"
   },
   "dependencies": {
+    "@videojs/spf": "workspace:*",
```
cjpillsbury, line 90: **NOTE:** Per out of band convo, this was added due to local dev env friction. Theoretically isn't needed/shouldn't be included in a real publish (though mostly harmless if this is complicated?)

packages/core/src/dom/index.ts
cjpillsbury: Removed and looks good locally, so pushing.

.github/workflows/ci.yml
```diff
@@ -140,6 +140,17 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile --prefer-offline
 
+      - name: Cache Playwright browsers
```
cjpillsbury: **NOTE:** We can save this for a followup pass/PR if we don't want to "muddy the waters" of core functionality merging.

.claude/spikes/f1-playlist-resolution.ts
```diff
@@ -0,0 +1,267 @@
+/**
```
cjpillsbury: Removing this.

packages/core/src/dom/media/simple-hls/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+// This is used by the React component.
+export class SimpleHlsMedia extends MediaDelegateMixin(
+  MediaProxyMixin(
+    globalThis.HTMLVideoElement ?? class {},
+    globalThis.HTMLMediaElement ?? class {},
+    globalThis.EventTarget ?? class {}
+  ),
+  SpfMediaDelegate
+) {}
```
luwes, line 20: 👍

packages/html/src/store/container-mixin.ts
luwes: yes, this shouldn't be needed. I have some changes that make this work more like what we have in media-chrome.

packages/sandbox/package.json
```diff
@@ -38,4 +39,4 @@
     "typescript": "^5.0.0",
     "vite": "^6.0.0"
   }
-}
+}
```
luwes:
```suggestion
}

```

packages/html/src/store/container-mixin.ts
cjpillsbury: Ok cool I'll try to back these all out and make sure everything still works as expected. Will keep ya posted.

packages/core/src/dom/media/custom-media-element/index.ts
mihar-22: Fine to keep, we need to clean up this media querying to just use slot media in a future pass.

.github/workflows/ci.yml
mihar-22: Prefer if we do this in a separate follow-up PR? Is that okay?

packages/spf/scripts/measure-size.js
mihar-22, line 1: You might be able to remove this file as CI bundle size reports are capturing SPF already ([link](https://github.com/videojs/v10/pull/784#issuecomment-4027627898))

packages/spf/src/utils/index.ts
mihar-22, line 1: empty file? we generally avoid barrel exports inside the same package for internal exports but not hard requirement

packages/spf/src/core/state/create-state.ts
mihar-22, line 1: Still not 100% sure why SPF can't use `createState` from `@videojs/store`, this looks identical but we can mark as follow-up and have async convo about it.

packages/spf/src/core/reactive/combine-latest.ts
mihar-22, line 1: Observable helpers exported from `@videojs/store/observables` would be pretty awesome. Future convo.

.gitignore
```diff
@@ -123,9 +123,6 @@ Pods/
 # Vercel
 .vercel
 
-# Netlify
-.netlify
-
```
mihar-22, line 128: Should this be deleted?

packages/spf/scripts/measure-size.js
cjpillsbury: Cool let's save that as a fast follow. But awesome that the build tool uses the actual VJS entrypoints! 😎 more accurate.

packages/spf/src/core/state/create-state.ts
cjpillsbury: Yes we should follow up on this one. There are some Big Picture™️ things next cycle and possible points of friction this cycle in some fast follows, depending. TL;DR - Things using createState _**may**_ end up migrating to signals. Need to do a spike on that next cycle.

packages/spf/src/core/reactive/combine-latest.ts
cjpillsbury: Observables **_may_** go away (see prior callout re: createState + signals)

.gitignore
cjpillsbury: Probably not! Thanks for the catch. Will push a re-add.

packages/spf/src/utils/index.ts
cjpillsbury: Shouldn't be a hard requirement. I think it got refactored to nothing over the course of iterations! Thanks for the 🦅 👀 ! ❤️

packages/spf/src/core/features/select-tracks.ts
```diff
@@ -0,0 +1,404 @@
...
+  const switchingSet = videoSet.switchingSets[0];
+  if (!switchingSet || switchingSet.tracks.length === 0) {
+    return undefined;
+  }
+
+  const initialBandwidth = config.initialBandwidth ?? DEFAULT_INITIAL_BANDWIDTH;
+  const safetyMargin = config.safetyMargin ?? DEFAULT_QUALITY_CONFIG.safetyMargin;
+
+  // selectQuality works with both partially resolved and resolved tracks
+  const selected = selectQuality(switchingSet.tracks as any, initialBandwidth, { safetyMargin });
```
luwes, line 137: The default initial bandwidth feels off to me. I would rather either have the first rendition in the m3u8 picked or set an index manually via config (or another option could be to have a slow video startup where enough samples are collected from a "random" chosen rendition and then change to the right index and start playback)

For a dev to choose a default bandwidth feels very fuzzy to me. What do you think?

cjpillsbury: Good feedback but let's save this for a post-main merge!

luwes (approved):
LFGTM 🚀

Issue #504, opened 2026-02-10
Feature: Gestures


mihar-22:
Gesture system for handling user interactions on the player — tapping, clicking, swiping, pinching, and keyboard input. Declarative gesture mapping so skins and users can configure what interactions do (e.g., click-to-play, double-click-fullscreen, swipe-to-seek, pinch-to-zoom).

Includes core logic, DOM props builder, and React/HTML implementations.

## References

- **Vidstack:** [gesture component](https://www.vidstack.io/docs/player/components/display/gesture/)
- **Media Chrome:** [`<media-controller>` gesturesdisabled](https://www.media-chrome.org/docs/en/components/media-controller)

PR #805, opened 2026-03-10
feat: PlayGesture & HotKeys

luwes:
## Summary

Adds **PlayGesture** and **Hotkeys** interaction components.

### PlayGesture

Click/tap the video surface to toggle play/pause — modeled as a declarative gesture component. Built on a `GestureCore` → `PlayGestureCore` abstraction in `@videojs/core` that keeps the logic runtime-agnostic, with DOM binding (`bindGesture`) in `@videojs/core/dom`. Touch events are excluded by default (`type="mouse"`).

### Hotkeys

YouTube-style keyboard shortcuts bound to the media container. `HotkeysCore` maps keys to actions (space/k = play/pause, f = fullscreen, m = mute, arrows = seek/volume, j/l = ±10s seek, `<`/`>` = playback rate). Modifier keys (Ctrl/Meta/Alt) are ignored, and space only fires when the container itself is focused.

### Usage

**HTML** — drop the custom elements inside a skin template:

```html
<media-container>
  <slot name="media"></slot>

  <media-play-gesture></media-play-gesture>
  <media-hotkeys></media-hotkeys>

  <!-- ...controls... -->
</media-container>
```

**React** — render the components inside a `<Container>`:

```tsx
import { Container } from '@videojs/react';
import { PlayGesture } from '@videojs/react/ui/play-gesture';
import { Hotkeys } from '@videojs/react/ui/hotkeys';

function VideoSkin({ children }) {
  return (
    <Container className="my-skin">
      {children}

      <PlayGesture />
      <Hotkeys />

      {/* ...controls... */}
    </Container>
  );
}
```

## References
This PR makes progress towards two sub-issues under the gestures epic (#504)
- #790 — click / tap to play
- #1054 — hotkeys

## Test plan

- [x] `HotkeysCore` unit tests cover all key bindings, modifier exclusion, and edge cases (no media, missing methods)
- [x] `PlayGestureCore` unit tests cover play/pause toggling and ended→play restart
- [ ] Manual: click video surface to toggle playback in HTML and React sandboxes
- [ ] Manual: verify keyboard shortcuts (space, k, f, m, arrows, j, l, <, >) in both sandboxes
- [ ] Manual: confirm touch events do not trigger PlayGesture (mouse-only by default)

### DIFF
--- a/packages/core/src/core/ui/hotkeys/hotkeys-core.ts
+++ b/packages/core/src/core/ui/hotkeys/hotkeys-core.ts
@@ -40,10 +40,10 @@ export class HotkeysCore {
     ArrowRight: () => this.#seekBy(SEEK_SMALL),
     j: () => this.#seekBy(-SEEK_LARGE),
     l: () => this.#seekBy(SEEK_LARGE),
-    ArrowUp: () => this.#setVolumeBy(VOLUME_STEP),
-    ArrowDown: () => this.#setVolumeBy(-VOLUME_STEP),
-    '<': () => this.#setPlaybackRateBy(-RATE_STEP),
-    '>': () => this.#setPlaybackRateBy(RATE_STEP),
+    ArrowUp: () => this.#nudgeVolumeBy(VOLUME_STEP),
+    ArrowDown: () => this.#nudgeVolumeBy(-VOLUME_STEP),
+    '<': () => this.#nudgePlaybackRateBy(-RATE_STEP),
+    '>': () => this.#nudgePlaybackRateBy(RATE_STEP),
   };
 
   #keyupActions: Record<string, () => boolean> = {
@@ -122,7 +122,7 @@ export class HotkeysCore {
     return true;
   }
 
-  #setVolumeBy(delta: number): boolean {
+  #nudgeVolumeBy(delta: number): boolean {
     const { media } = this;
     if (!media?.setVolume || media.volume == null) return false;
     const next = Math.min(1, Math.max(0, media.volume + delta));
@@ -130,7 +130,7 @@ export class HotkeysCore {
     return true;
   }
 
-  #setPlaybackRateBy(delta: number): boolean {
+  #nudgePlaybackRateBy(delta: number): boolean {
     const { media } = this;
     if (!media?.setPlaybackRate || media.playbackRate == null) return false;
     const next = Math.min(RATE_MAX, Math.max(RATE_MIN, media.playbackRate + delta));


packages/core/src/core/ui/gesture/gesture-core.ts
```diff
@@ -0,0 +1,43 @@
+import { defaults } from '@videojs/utils/object';
+
+import type { MediaPlaybackState } from '../../media/state';
+
+export const ALLOWED_TYPES = ['pointerup'] as const;
```
mihar-22: (request): since it's exported from main package entry, recommend `ALLOWED_GESTURE_TYPES` to avoid conflicts or confusion

packages/core/src/core/ui/gesture/gesture-core.ts
```diff
@@ -0,0 +1,43 @@
...
+export const ALLOWED_COMMANDS = ['toggle-paused'] as const;
```
mihar-22: (request): same as earlier, recommend `ALLOWED_GESTURE_COMMANDS`

packages/core/src/core/ui/gesture/gesture-core.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface GestureProps {
+  type: (typeof ALLOWED_TYPES)[number];
```
mihar-22: Great if we extract into reusable type: `type GestureType = (typeof ALLOWED_GESTURE_TYPES)[number]`

packages/core/src/core/ui/gesture/gesture-core.ts
```diff
@@ -0,0 +1,43 @@
...
+  command: (typeof ALLOWED_COMMANDS)[number];
```
mihar-22: Great if we extract into reusable type: type `GestureCommand = (typeof ALLOWED_GESTURE_COMMANDS)[number]`

packages/html/src/media/container-element.ts
```diff
@@ -1,9 +1,38 @@
...
 export class MediaContainerElement extends ContainerMixin(MediaElement) {
   static readonly tagName = 'media-container';
+  static shadowRootOptions = { mode: 'open' } as const;
+  static getTemplateHTML(): string {
+    return /* html */ `
+      <style>
+        slot:not([name]) {
+          border-radius: inherit;
+        }
+      </style>
```
mihar-22: This needed?

packages/html/src/store/container-mixin.ts
mihar-22, line 1: Just noting this will have docs and consumer impact because they have to use shadow DOM for this mixin to work. Not a problem since children will still live in light DOM, just worth calling out.

packages/react/src/ui/gesture/gesture.tsx
```diff
@@ -0,0 +1,58 @@
...
+        gestureCore.activate(playback);
+      },
+      { signal: controller.signal }
+    );
+
+    return () => controller.abort();
+  }, [playback, gestureCore, type]);
+
+  if (!playback) {
+    if (__DEV__) logMissingFeature('Poster', 'playback');
```
mihar-22: 'Poster' => Gesture?

package.json
```diff
@@ -25,7 +25,7 @@
     "dev:site": "turbo run dev --filter=site",
     "dev": "turbo run dev --parallel",
     "dev:packages": "turbo run dev --parallel --filter='./packages/*'",
-    "dev:sandbox": "turbo run dev --filter=@videojs/sandbox",
```
mihar-22: Interesting, what's this change do?

package.json
```diff
@@ -25,7 +25,7 @@
...
+    "dev:sandbox": "turbo watch dev --filter=@videojs/sandbox",
```
mihar-22, line 29: Interesting, what's this change do?

packages/html/src/media/container-element.ts
luwes: it was needed if we used shadow dom in media container because we used border-radius inherit on the media (I think) so this was needed. I have reverted this logic to not use shadow dom...

packages/html/src/store/container-mixin.ts
luwes: reverted and updated

package.json
luwes: this will watch the scripts that are not never ending, in this case the build scripts of the dependencies

packages/core/src/core/ui/gesture/gesture-core.ts
```diff
@@ -0,0 +1,59 @@
+import { defaults } from '@videojs/utils/object';
+
+import type { MediaPlaybackState } from '../../media/state';
+
+export type GestureType = (typeof ALLOWED_GESTURE_TYPES)[number];
+export type GestureCommand = (typeof ALLOWED_GESTURE_COMMANDS)[number];
+
+export interface GestureProps {
+  type: GestureType;
+  command: GestureCommand;
```
mihar-22: (request): missing jsdoc comments, used by API reference tooling for docs.

packages/core/src/core/ui/gesture/gesture-core.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export const ALLOWED_GESTURE_TYPES = ['pointerup'] as const;
+export const ALLOWED_GESTURE_COMMANDS = ['toggle-paused'] as const;
+
+export const PointerTypes = {
+  MOUSE: 'mouse',
+  PEN: 'pen',
+  TOUCH: 'touch',
+} as const;
```
mihar-22: (request): this is handy but more a DOM specific type. Can we move this to `core/ui/types`, or as per suggestion below have `core/dom` layer for gesture and move to `core/dom/ui/event`

packages/core/src/core/ui/gesture/gesture-core.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.#media = media;
+  }
+
+  handleGesture({ pointerType }: { pointerType: string }): void {
+    if (!this.#media) return;
+
+    if (!ALLOWED_GESTURE_TYPES.includes(this.#props.type)) return;
+
+    // TODO: Should `pointerType` be a prop that can be configured?
+    if (pointerType && pointerType === PointerTypes.MOUSE) {
```
mihar-22: Interesting. I like it and if it makes sense for future gestures then good idea! 

(suggestion): if it doesn't make sense as a prop, you could consider moving "pointer type" handling to the DOM layer or framework-layer since it's just a DOM event related guard and not core state management piece.

packages/core/src/core/ui/gesture/gesture-core.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  setProps(props: GestureProps): void {
+    this.#props = defaults(props, GestureCore.defaultProps);
+  }
+
+  setMedia(media: MediaPlaybackState): void {
+    this.#media = media;
+  }
+
+  handleGesture({ pointerType }: { pointerType: string }): void {
```
mihar-22: (request): stricter type `pointerType: PointerType`. 

Requires creating the type:

```ts
type PointerType = PointerTypes[typeof keyof PointerTypes]
```

---

Not required if guarding is moved to `core/dom`

packages/html/src/ui/gesture/gesture-element.ts
```diff
@@ -0,0 +1,71 @@
...
+          (event: PointerEvent) => {
+            const target = event.target as Element;
+            if (target !== container && !target.localName.endsWith('video')) return;
+
+            this.#core.handleGesture(event);
+          },
+          { signal }
+        );
+      }
+    }
```
mihar-22: (suggestion): could move this to `core/dom` to share with html and react. The pointer types could live there too which would make more sense. Claude should be able to do it by referencing other examples.

packages/react/src/ui/gesture/gesture.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export function Gesture({ type, command }: GestureProps) {
+  const [gestureCore] = useState(() => new GestureCore());
+  gestureCore.setProps({ type, command });
+
+  const playback = usePlayer(selectPlayback);
+  if (playback) gestureCore.setMedia(playback);
+  else {
+    if (__DEV__) logMissingFeature('Gesture', 'playback');
+  }
```
mihar-22: (nit): could be `else if (__DEV__)`

PR #827, opened 2026-03-10
feat(html): add CDN bundles and inline template minification

mihar-22:
## Summary

Add self-contained ESM bundles for CDN delivery inside `@videojs/html`. Each entry (video, audio, background, media/*) builds as a single file with zero external imports. Build plugins minify inline CSS and tagged template literals in production builds.

## Changes

- CDN bundles live in `@videojs/html/cdn/` — no separate package
- 7 entry points × 2 modes (dev/prod) = 14 fully self-contained ESM bundles
- Per-entry tsdown configs prevent code-splitting into shared chunks
- `inlineTemplatePlugin` minifies `/*html*/` and `/* css */` tagged template literals via lightningcss
- `inlineCssPlugin` gains `minify` option (default: `true`) for `?inline` CSS imports
- CSS files copied and minified with lightningcss for CDN consumption
- `build:cdn` wired into CI (build job) and CD (release job)
- Site CDN URLs updated to `@videojs/html/cdn/`

<details>

- Each CDN entry is built as a separate tsdown config because rolldown code-splits shared modules when given multiple entries in one config
- CDN entries use relative imports to source modules; the CDN config includes `inlineCssPlugin` and `inlineTemplatePlugin` to handle `?inline` CSS and tagged template minification
- `__DEV__` is replaced via `define` — CDN dev sets `true`, prod sets `false`
- Template literal parser handles `${...}` expressions with proper brace/string/nested-template tracking; CSS expressions are replaced with placeholder tokens before lightningcss runs, then restored

</details>

## Testing

```bash
pnpm build:cdn
# Verify all 14 bundles + 5 CSS files in packages/html/cdn/
# Verify zero import statements in output files
```

### Bundle sizes

| Bundle | Size | Gzip |
|--------|------|------|
| video.js | 97.3 KB | 24.4 KB |
| video-minimal.js | 96.4 KB | 24.2 KB |
| audio.js | 86.7 KB | 22.5 KB |
| audio-minimal.js | 87.2 KB | 23.2 KB |
| background.js | 22.3 KB | 6.8 KB |
| media/hls-video.js | 516.6 KB | 158.2 KB |
| media/simple-hls-video.js | 40.1 KB | 12.4 KB |

### DIFF
--- a/packages/html/tsdown.cdn.config.ts
+++ b/packages/html/tsdown.cdn.config.ts
@@ -42,6 +42,7 @@ for (const mode of buildModes) {
       dts: false,
       minify: isProd,
       noExternal: [/.*/],
+      inlineOnly: false,
       treeshake: {
         moduleSideEffects: [{ test: /\/define\//, sideEffects: true }],
       },


build/plugins/inline-template-plugin.mjs
```diff
@@ -0,0 +1,341 @@
...
+/** Minify the static parts of an HTML template literal. */
+function minifyHtmlQuasis(quasis) {
+  return quasis.map((q, qIdx) => {
+    let s = q;
+
+    // 1. Strip HTML comments.
+    s = s.replace(/<!--[\s\S]*?-->/g, '');
+
+    // 2. Collapse runs of whitespace to a single space.
+    s = s.replace(/\s{2,}/g, ' ');
```
cjpillsbury, line 115: question(non-blocking): since whitespace can affect actual sizing/layout (text under some cases, but also parent/child/sibling DOM element layout behaviors), do we need to worry about this?

mihar-22: Good question! That line is collapsing multiple whitespaces into a single which browsers normally do. I think for our library this is safe. Happy to review later if something funky comes up.

PR #785, opened 2026-03-09
docs(site): React API reference styling sections use correct selectors

decepulis:
## Summary

- Wraps HTML custom element CSS selectors in `<FrameworkCase frameworks={["html"]}>` across 10 API reference pages
- Adds `<FrameworkCase frameworks={["react"]}>` blocks with equivalent CSS using `className`-based selectors
- Fixes the Controls accessibility section which referenced `<media-controls>` / `<media-controls-group>` instead of `Controls.Root` / `Controls.Group`

**Affected pages:** PlayButton, MuteButton, FullscreenButton, PiPButton, Controls, TimeSlider, VolumeSlider, Popover, Poster, Thumbnail

Closes #728

## Test plan

- [ ] Visit React reference pages and confirm styling sections show `className`-based CSS selectors
- [ ] Visit HTML reference pages and confirm they still show custom element selectors
- [ ] Verify site builds without errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #862, opened 2026-03-10
docs(site): add Slider and Tooltip API reference pages

### LINKED ISSUE #859: docs(site): API reference for SliderThumbnail
Add API reference page for the SliderThumbnail component.

### LINKED ISSUE #860: docs(site): API reference for SliderPreview
Add API reference page for the SliderPreview component.

### LINKED ISSUE #861: docs(site): API reference for Tooltip
Add API reference page for the Tooltip component.

decepulis:
## Summary

- Adds API reference page for the base **Slider** components
- Adds API reference page for the **Tooltip** components
- Updates api-reference skill to require `<FrameworkCase>` wrappers on CSS code blocks in Styling sections (encoding #785)
- Adds `partOrder` prop to `<ComponentReference>` so API reference parts render in anatomy order instead of JSON key order

Closes #859, closes #860, closes #861

## Preview links

- [Slider (React)](https://deploy-preview-862--vjs10-site.netlify.app/docs/framework/react/reference/slider/)
- [Slider (HTML)](https://deploy-preview-862--vjs10-site.netlify.app/docs/framework/html/reference/slider/)
- [Tooltip (React)](https://deploy-preview-862--vjs10-site.netlify.app/docs/framework/react/reference/tooltip/)
- [Tooltip (HTML)](https://deploy-preview-862--vjs10-site.netlify.app/docs/framework/html/reference/tooltip/)

### DIFF
--- a/site/src/components/docs/demos/tooltip/html/css/Grouping.html
+++ b/site/src/components/docs/demos/tooltip/html/css/Grouping.html
@@ -1,10 +1,10 @@
 <media-tooltip-group class="html-tooltip-grouping">
-    <button type="button" commandfor="tooltip-bold" class="html-tooltip-grouping__trigger">Bold</button>
-    <media-tooltip id="tooltip-bold" class="html-tooltip-grouping__popup">Toggle bold</media-tooltip>
+    <button type="button" commandfor="tooltip-play" class="html-tooltip-grouping__trigger">Play</button>
+    <media-tooltip id="tooltip-play" class="html-tooltip-grouping__popup">Play video</media-tooltip>
 
-    <button type="button" commandfor="tooltip-italic" class="html-tooltip-grouping__trigger">Italic</button>
-    <media-tooltip id="tooltip-italic" class="html-tooltip-grouping__popup">Toggle italic</media-tooltip>
+    <button type="button" commandfor="tooltip-mute" class="html-tooltip-grouping__trigger">Mute</button>
+    <media-tooltip id="tooltip-mute" class="html-tooltip-grouping__popup">Mute audio</media-tooltip>
 
-    <button type="button" commandfor="tooltip-underline" class="html-tooltip-grouping__trigger">Underline</button>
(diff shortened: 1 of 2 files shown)


mihar-22 (changes requested):
@decepulis thank you for doing this! Some notes:

1. [Slider example](https://deploy-preview-862--vjs10-site.netlify.app/docs/framework/react/reference/slider#examples) has a video which is unrelated to the example. Great to remove it. 

2. Slider _should_ work without player boilerplate (standalone), we should be able to remove all that from snippets.

3. For slider separating into two examples would be nice: basic, with preview

4. [Tooltips](https://deploy-preview-862--vjs10-site.netlify.app/docs/framework/react/reference/tooltip) should be standalone too (no player boilerplate required). 

5. Tooltip anatomy is missing `Provider` in React, and `media-tooltip-group` in HTML. Root element.

6. Tooltips having an additional example for "grouping". I think behaviour section can also discuss "groups"

> Provides a shared delay for multiple tooltips. The grouping logic ensures that once a tooltip becomes visible, the adjacent tooltips will be shown instantly.

mihar-22 (changes requested):
1. In the tooltip group example, the timeout should remain as default value, removing the prop `timeout={1000}`. It should be highlighting faster tooltip displays as part of a group, so recommending something slower like 1s is odd.

2. It would be nice if we could sort the API reference inline with the anatomy (provider -> root -> ...). See screenshots.

<details>
  <summary>🖼️  See screenshots</summary>

Image

Image

</details>

decepulis:
> in the tooltip group example, the timeout should remain as default value, removing the prop timeout={1000}. It should be highlighting faster tooltip displays as part of a group, so recommending something slower like 1s is odd.

oh yeah this reminds me. that was part of an experiment to try and make the html example work. Forgot to remove it. Do html tooltip groups work? Is that example just written wrong?

> It would be nice if we could sort the API reference inline with the anatomy (provider -> root -> ...). See screenshots.

Done. Not by anatomy order but by custom prop for ease of implementation (and maintenance)

PR #890, opened 2026-03-11
feat(spf): stream segment fetches via ReadableStream body

cjpillsbury:
## Summary

Migrates SPF's segment fetching pipeline to use `fetch`'s `response.body` `ReadableStream`, enabling overlapping download and append for video segments while preserving correct behavior for audio and init segments across all browsers.

## Changes

- **`ChunkedStreamIterable`** — new class wrapping a `ReadableStream<Uint8Array>`, yielding chunks at a configurable minimum size (default 128 KB); flushes remainder on stream end
- **`fetchResolvableStream()`** — async generator over `ChunkedStreamIterable`; awaits response headers before yielding body chunks
- **Bandwidth sampling** — `createTrackedFetch` samples throughput per-chunk as data arrives rather than once after a full download
- **`AppendData = ArrayBuffer | AsyncIterable<Uint8Array>`** — `SourceBufferActor` messages accept either; streaming appends emit `partial: true` model state so downstream code never treats an in-flight segment as fully buffered
- **End-to-end streaming** — `fetchBytes` eagerly starts the HTTP connection (TTFB awaited), then streams body chunks directly to the actor for video segments; `minChunkSize: Infinity` collapses the stream to one chunk for init segments
- **`setupSourceBuffers`** — replaces two independent per-track orchestrations with a single function that waits until all selected tracks are resolved before creating any `SourceBuffer`; prevents the Firefox bug where appending video data before the audio `SourceBuffer` exists causes `mozHasAudio` to be permanently `false`
- **Sandbox** — `firefox-mse-repro` harness for manually testing MSE init-segment ordering across Chrome, Firefox, and Safari

## Testing

- All 705 existing SPF tests pass
- 44 new tests covering `ChunkedStreamIterable`, `fetchResolvableStream`, streaming appends, `partial` segment model, `isLastSegmentAppended` with partial segments, and `setupSourceBuffers` coordination
- Smoke tested in Chrome via DevTools MCP: video plays, bandwidth estimation active, `response.body` reads confirmed for video segments
- Firefox `mozHasAudio = false` bug confirmed via sandbox harness and resolved by `setupSourceBuffers` coordination

### DIFF
--- a/packages/sandbox/templates/firefox-mse-repro/index.html
+++ b/packages/sandbox/templates/firefox-mse-repro/index.html
@@ -1,0 +1,107 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Firefox MSE Init Order Repro</title>
+    <style>
+      * { box-sizing: border-box; margin: 0; padding: 0; }
+      body { font-family: monospace; font-size: 13px; background: #1a1a1a; color: #e0e0e0; padding: 16px; }
+      h1 { font-size: 15px; margin-bottom: 4px; color: #fff; }
+      .subtitle { color: #888; font-size: 11px; margin-bottom: 14px; }
+      h2 { font-size: 11px; font-weight: bold; margin: 12px 0 5px; color: #aaa; text-transform: uppercase; letter-spacing: 0.05em; }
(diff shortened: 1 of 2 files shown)


packages/sandbox/templates/firefox-mse-repro/index.html
```diff
@@ -0,0 +1,107 @@
+<!doctype html>
```
cjpillsbury, line 1: may delete this later but keeping for now as a baseline for a likely mozilla bug filing.

PR #904, opened 2026-03-12
fix(spf): propagate byteRange when building segment load tasks

cjpillsbury:
## Summary

Byte-range segments (e.g. Apple fMP4/CMAF streams using `#EXT-X-BYTERANGE`) were being fetched without a `Range` header, causing the entire container file to be downloaded on every segment request. This immediately exhausted the `SourceBuffer` quota (`QuotaExceededError`) for any range-request–based HLS stream.

## Changes

- Both `append-init` and `append-segment` tasks now spread `byteRange` from the parsed segment/init onto the `LoadTask`, so `fetchResolvable` sends the correct `Range` header
- New tests covering `byteRange` propagation for both init and segment tasks

## Testing

`pnpm -F @videojs/spf test src/dom/features/tests/load-segments.test.ts`

### DIFF
--- a/packages/sandbox/app/shared/sources.ts
+++ b/packages/sandbox/app/shared/sources.ts
@@ -1,6 +1,4 @@
-export type SourceId = 'hls-1' | 'hls-2' | 'hls-3' | 'hls-4' | 'hls-5' | 'mp4-1';
-
-export const SOURCES: Record<SourceId, { label: string; url: string; type: 'hls' | 'mp4'; subType?: 'mp4' | 'ts' }> = {
+export const SOURCES = {
   'hls-1': {
     label: 'HLS - Big Buck Bunny',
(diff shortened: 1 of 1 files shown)


packages/sandbox/app/shared/sources.ts
```diff
@@ -1,6 +1,4 @@
-export type SourceId = 'hls-1' | 'hls-2' | 'hls-3' | 'hls-4' | 'hls-5' | 'mp4-1';
-
-export const SOURCES: Record<SourceId, { label: string; url: string; type: 'hls' | 'mp4'; subType?: 'mp4' | 'ts' }> = {
+export const SOURCES = {
```
cjpillsbury, line 1: unrelated but was temporarily adding srcs for testing, so I went ahead and cleaned up the type relationships here to make this less of a pain going forward.

Issue #928, opened 2026-03-13
Review and dogfood skins eject process


mihar-22:
Review the skins eject code and process end-to-end. Dogfood it to find friction points and issues.

Issue #840, opened 2026-03-10
Docs: Add Tailwind as a Supported Style for Documentation


decepulis:
## Summary

The customize-skins guide (`how-to/customize-skins.mdx`) currently only shows CSS variants because Tailwind is not a supported style in the docs framework.

The `ejectedSkins` content collection already has Tailwind variants ready (e.g., `default-video-tailwind`, `default-video-react-tailwind`), but `FRAMEWORK_STYLES` in `site/src/types/docs.ts` only includes `'css'`:

```ts
export const FRAMEWORK_STYLES = {
  react: ['css'],
  html: ['css'],
} as const;
```

## Tasks

- [ ] Add `'tailwind'` to `FRAMEWORK_STYLES` in `site/src/types/docs.ts`
- [ ] Update `STYLE_LABELS` to include a Tailwind label
- [ ] Add `<StyleCase styles={['tailwind']}>` sections to `how-to/customize-skins.mdx` showing Tailwind skin variants
- [ ] Remove the TODO comment in `customize-skins.mdx` referencing this issue

## What makes this tricky.
Once we do this, would users expect ALL the css examples in the api references to have tailwind? Imo yes

PR #958, opened 2026-03-16
docs(skin): add docs on skin styling

### LINKED ISSUE #674: Skins: Styling documentation
Document the styling considerations: 

- Basics around BEM, resets, font stack, scoping, etc
- CSP requirements (verify if we're using inline styles?)
- Tailwind requirements (v4, prefixing coming later as part of eject) 
- Tailwind assumptions (default config, related to prefixing)

sampotts:
## Summary
- Add skin styling guidance covering packaged vs ejected usage, vanilla CSS vs Tailwind, and current styling limitations.
- Document basic skin customization with CSS custom properties and clarify the ejecting workflow for default and minimal skins.
- Add installation guidance for CSP requirements and link the installation flow back to the skin styling docs.

## Testing
- Not run (documentation-only changes).

Closes #674

### DIFF
--- a/site/src/content/docs/concepts/skins.mdx
+++ b/site/src/content/docs/concepts/skins.mdx
@@ -10,6 +10,7 @@ import MinimalFrame from '@/components/frames/Minimal.astro';
 import DocsLink from '@/components/docs/DocsLink.astro';
 import DocsLinkCard from '@/components/docs/DocsLinkCard.astro';
 import Aside from '@/components/Aside.astro';
+import EjectedSkin from '@/components/docs/EjectedSkin.astro';
 
 <Aside type="note" title="New in version 10">
 In prior versions of Video.js, skins were CSS-only themes applied to the same set of default controls. In version 10 the concept of skins has been expanded to include the UI components, so each skin can be completely unique and only include the UI components it needs.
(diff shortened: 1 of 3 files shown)


site/src/content/docs/how-to/installation.mdx
```diff
@@ -96,25 +95,61 @@ Video.js supports a wide range of file types and hosting services. It's easy to
...
+
+```http
+Content-Security-Policy:
+  script-src 'self';
+  style-src 'self' 'unsafe-inline';
+  img-src 'self' https: data: blob:;
+  media-src 'self' https: blob:;
+  connect-src 'self' https:;
+  worker-src 'self' blob:;
+```
```
decepulis, line 145: suggestion(non-blocking): installation is a _very_ tight guide. We should consider moving this CSP section to a new security guide (#966) and linking out.

site/src/content/docs/how-to/installation.mdx
```diff
@@ -96,25 +95,61 @@ Video.js supports a wide range of file types and hosting services. It's easy to
...
+
+## Styling
+
+See the [skins styling](../concepts/skins#styling) section.
```
decepulis: polish(non-blocking): consider using `DocsLinkCard` here, like we do in overview.

Maybe, combined with [my security comment up above](https://github.com/videojs/v10/pull/958/changes#r2940322788), this section could be something like
```mdx
## See also
<DocsLinkCard slug="concepts/skins" anchor="styling" description="Some skins expose CSS custom properties">Skins</DocsLinkCard>
<DocsLinkCard slug="concepts/skins" anchor="styling" description="If you have a content security policy (CSP), you may need to enable additional sources">Security</DocsLinkCard>
```

This would require adding an `anchor` prop to `DocsLinkCard` and the underlying `DocsLink`.

site/src/content/docs/concepts/skins.mdx
```diff
@@ -10,13 +10,14 @@ import MinimalFrame from '@/components/frames/Minimal.astro';
 import DocsLink from '@/components/docs/DocsLink.astro';
 import DocsLinkCard from '@/components/docs/DocsLinkCard.astro';
 import Aside from '@/components/Aside.astro';
-
+import EjectedSkin from '@/components/docs/EjectedSkin.astro';
```
decepulis, line 13: Good call!

site/src/content/docs/how-to/customize-skins.mdx
```diff
@@ -6,13 +6,26 @@ description: Learn how to customize Video.js v10 skins by copying and modifying
...
 
-Video.js v10 comes with pre-built skins like Default and Minimal. If you'd like to customize them you can fully customize them by "ejecting" the code and making it your own.
+Video.js v10 comes with two pre-built skins; Default and Minimal. Basic customization is possible with CSS custom properties, but if you want more control over the design and functionality of your player, you can copy the skin's code into your project and modify it as needed. We call this "ejecting" the skin, since you're taking the internal code that makes up the skin and making it your own.
+
+## Basic customization
+
+| Property Name | Description | Type | Example |
+| ------------- | ----------- | ---- | ------- |
+| `--media-border-radius` | The border radius of the media player | A valid [border-radius value](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/border-radius#values) | `1rem` |
+| `--media-color-primary` | The color of icons and text in media controls | [`<color>`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) | `red` |
```
decepulis, line 16: thought: is every skin going to support every CSS custom property? Maybe not now, but we should consider an api reference for each skin. (I know, eventually this will lead to a bloated sidebar, but, one problem at a time 😅 maybe eventually we'll add a gallery or something)

site/src/content/docs/concepts/skins.mdx
```diff
@@ -73,259 +77,71 @@ When you choose a skin you have two options for how you use it: **packaged** or
...
+- The default font stack includes Inter Variable and Inter (because it looks awesome) but we do not load them, as webfonts for you. If they are not available then system fonts are used.
+
+#### Vanilla CSS
+
+- We use a [BEM](https://en.bem.info/methodology/quick-start/) classname structure and every component classname is scoped with a `media-` prefix.
+
+#### Tailwind
+
+- Currently we're assuming you're using the default configuration and that's all that's supported. With the release of our CLI, this will change, allowing you to specify a custom prefix to the Tailwind classnames. For now, you'll need to edit the ejected skins yourself.
+- We're assuming the latest version, currently 4.2.x.
```
decepulis, line 149: issue(blocking): 

you might've noticed that I haven't actually put the ejected tailwind skins in docs yet. Maybe now's the time we add them to customize-skins. I felt blocked by #840, but maybe I was letting perfect getting in the way of good enough.

We should either
1. add the tailwind skins to the eject docs or
2. remove this paragraph until we do add tailwind to the docs

site/src/content/docs/concepts/skins.mdx
```diff
@@ -73,259 +77,71 @@ When you choose a skin you have two options for how you use it: **packaged** or
...
+</FrameworkCase>
+
+### Current limitations
+
+- In both style systems we assume a 16px root font size and we use rem units for sizing.
+- The default font stack includes Inter Variable and Inter (because it looks awesome) but we do not load them, as webfonts for you. If they are not available then system fonts are used.
+
+#### Vanilla CSS
+
+- We use a [BEM](https://en.bem.info/methodology/quick-start/) classname structure and every component classname is scoped with a `media-` prefix.
```
decepulis, line 144: question:

Is this relevant if our skins use shadow dom, unless you eject?
Or do our skins not use shadow dom?
Or do only our html skins use shadow dom? In which case, this info should be <FrameworkCase frameworks=["react"]> scoped.

I'm still so new to the html custom element world I'm not sure if my question makes sense. Lmk if I can clarify!

site/src/content/docs/how-to/customize-skins.mdx
decepulis, line 1: thought: I'm feeling some tension about both `concepts/skins` and `how-to/customize-skins` existing. 

I actually agree with your choices to put a high-level overview in the concept guide, and then to get into the mechanics in the how-to guide. I'm not actually tossing out any blocking change requests. But I do want to think this through because I fear that the line between the two guides isn't very clear. Like, I can imagine your CSS custom property table existing in either guide. I can even imagine the eject instructions existing in the concept guide, especially once they become a one-liner CLI command. So why did they get their own how-to guide?

What would this look like if we drew the lines in different places? We could...
1. leave things as you have them in this PR. There's a tiny bit duplicate content, with the skins concept focusing more on the high-level and customize-skins focusing more on implementation
2. break up customize-skins into more specific guides. Call this one how-to/eject-skins and put the CSS custom properties in skins/concepts or an [API reference for each skin](https://github.com/videojs/v10/pull/958/changes#r2940405679)
3. consolidate everything into concepts/skins

Let's have this conversation in person. But also, don't let this conversation be blocking. Just something we should clear out eventually. It's not a problem yet.

site/src/content/docs/how-to/installation.mdx
sampotts: Sure, we can add some more details there about `unsafe-inline` if we continue with inline styles. I have an RFC in the works for moving to scoped `<style>` blocks instead.

site/src/content/docs/concepts/skins.mdx
sampotts: I'll go with 2 for now.

site/src/content/docs/how-to/installation.mdx
sampotts: I'll leave out the security page there as I think we need to talk through nonce and `unsafe-inline` stuff first.

site/src/content/docs/how-to/customize-skins.mdx
sampotts: Sure, can break it out into another page but I'll do that in a follow up PR if that's ok.

site/src/content/docs/concepts/skins.mdx
sampotts: Good point. Non-ejected HTML skins don't need a stylesheet since it's in the Shadow DOM but ejected doesn't use the Shadow DOM but it reminds me we should document the stuff we're doing in `SkinMixin` since it injects a `base.css` file that sets some required styles. I'll look into that as part of #928 .

site/src/content/docs/how-to/customize-skins.mdx
sampotts: I'm happy to do option 2 if that makes sense. I'll do it in a follow up PR.

PR #1068, opened 2026-03-20
refactor(react): simplify skin render props with element form

mihar-22:
## Summary

Simplify render prop usage across all 8 React skin files by replacing verbose function wrappers with element form and moving className/children to direct props on headless components.

## Changes

- Replace `render={(props) => <Button {...props} className="...">children</Button>}` with `render={<Button />}` + className and children as direct props on the headless component
- Bake `media-button--icon` into the CSS `Button` component — every usage had it, so it belongs in the component, not repeated at 26 call sites
- Replace `render={(props) => <SliderRoot {...props} />}` with `render={<SliderRoot />}` for all trivial slider passthrough wrappers in Tailwind skins
- Add `SliderBuffer` component to Tailwind skins so `<TimeSlider.Buffer render={<SliderBuffer />} />` replaces the function wrapper needed for `type="buffer"`
- Update component skill docs (polymorphism, react, anti-patterns) to document the element form preference and warn against passing component references directly

<details>

The element form (`render={<Component />}`) is preferred over component references (`render={Component}`) because React's render prop pattern calls functions directly — passing a component reference bypasses React's reconciliation and breaks hooks. This aligns with Base UI's guidance.

The only remaining function render props are `BufferingIndicator` (renders a `<div>`, not a button) and `VolumeSlider.Thumb` (needs `persistent` prop on the render target).

</details>

## Testing

Visual verification — the render output is identical. The element form delegates to `cloneElement` which merges props the same way.

### DIFF
--- a/.claude/skills/component/references/anti-patterns.md
+++ b/.claude/skills/component/references/anti-patterns.md
@@ -65,6 +65,42 @@ import { Button } from '@lib/button';
 
 ---
 
+## Unnecessary Render Function Wrappers
+
+```tsx
+// BAD: Function wrapper just passes props through
+<PlayButton
+  render={(props) => (
+    <Button {...props} className="play-btn">
+      <PlayIcon />
+    </Button>
+  )}
+/>
+
+// BAD: Trivial passthrough
+<TimeSlider.Root render={(props) => <SliderRoot {...props} />} />
+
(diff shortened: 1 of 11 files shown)


sampotts:
Nice!

PR #1076, opened 2026-03-23
docs(rfc): use nonce instead of unsafe-inline

sampotts:


mihar-22:
This doc reads beautifully, thanks for creating it! I agree with almost everything in it, however I'm going to close to avoid merging a stale doc (we already have so many we need to either update/remove) and to avoid having to iterate on it. 

You're more than welcome to follow-up with a PR directly to solve this because I agree we need a solution for CSP. I quite like the [Base UI CSPProvider](https://base-ui.com/react/utils/csp-provider) direction.

PR #1077, opened 2026-03-23
feat(packages): error dialog component

### LINKED ISSUE #1046: Skins: Add error dialogs to HTML skins
Currently the HTML skins have no error handling. We need to add a similar error-dialog component to the HTML skins.

sampotts:
Closes #1046

## Summary

Refactor `ErrorDialog` to mirror the `AlertDialog` compound API — removing `open`/`onOpenChange` props in favor of internal state management driven by the player's error state. The error dialog is now hidden by default and only rendered when an error occurs, preventing flash of fallback content.

## Changes

- `ErrorDialogCore` extends `AlertDialogCore` with `setProps` overridden to no-op (open state is internal)
- New `ErrorDialogElement` (HTML) creates its own `AlertDialogApi`, subscribes to error state via `PlayerController`, and opens/closes the dialog automatically
- New `ErrorDialog` React compound component (`Root`, `Popup`, `Title`, `Description`, `Close`) — `Root` manages the `AlertDialogApi` lifecycle, `Description` falls back to `lastError.message` from context
- Replaced single-component `ErrorDialog` preset with compound components in all 8 React skins
- Removed `<media-alert-dialog>` wrapper from all 8 HTML skin templates — `<media-error-dialog>` is now the root
- Removed duplicate Shadow DOM selectors (`media-error-dialog[data-open]~&`) from CSS and Tailwind — `peer/error` handles both since the error dialog now gets `data-open` directly
- Added `display: none` for closed error dialog state (`[data-open]` controls visibility)
- Description element starts empty in templates; content is set before opening to prevent fallback flash

<details>

Follows the same pattern as `TimeSlider`/`VolumeSlider` extending base `Slider` — `ErrorDialog` doesn't extend `AlertDialogElement` but creates a parallel implementation that manages domain-specific state (error) while reusing `AlertDialogApi` and `AlertDialogContext` for the dialog mechanics.

</details>

## Testing

`pnpm -F @videojs/html test` — 90 tests (4 error-dialog)
`pnpm -F @videojs/react test` — 197 tests
`pnpm -F @videojs/core test` — 673 tests

---

> [!NOTE]
> **Medium Risk**
> Touches error-display UX across `core`, `html`, `react`, and skin styling, changing how dialogs open/close and when they render/are visible. Risk is mainly UI regressions (missing/incorrectly-hidden error dialogs, ARIA/id mismatches), not security or data handling.
> 
> **Overview**
> Adds a first-class `ErrorDialog` implementation across packages: `core` now exports `ErrorDialogCore`/`ErrorDialogDataAttrs`, `html` adds `ErrorDialogElement` (`media-error-dialog`) that auto-opens/closes based on player error state and dismisses errors on close, and `react` introduces a compound `ErrorDialog` API (`Root`/`Popup`/`Title`/`Description`/`Close`) plus context for the last error message.
> 
> Updates all HTML/React audio+video skin templates to render the new `media-error-dialog`/compound components (removing the old React preset and wiring description content to be set before opening), and adjusts skin CSS/Tailwind so the error dialog is **hidden by default** and only shown when `[data-open]` is present; also tweaks safe-id generation to use hyphenated prefixes and reduces minimal video error transition duration.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 07f985ea02952a3554069d47054e71af2c5e81e9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


mihar-22 (changes requested):
Let me know if I'm missing the point of the PR or any of the below is confusing. We can discuss over a call tomorrow if need be! 

Might be best to separate the creation of a new component and the skin update.

---

**If I've understood correctly:**

1. A media `ErrorDialog` is a specific type of `AlertDialog`
2. A media `ErrorDialog` internally controls open state based on the `MediaErrorState`
3. It can provide a better default error title (maybe?) or description 

So if that's all right then I'd roughly expect the following API:

```html
<media-error-dialog> // replacing `media-alert-dialog` as root
// usual alert-dialog parts...
```

```tsx
<ErrorDialog> // React root
// usual AlertDialog parts (re-exports)...
```
> This would be _similar_ to how we tackled time and volume sliders.

**Internal details:**

- `ErrorDialogCore` extends `AlertDialogCore` and controls `open` state
-  New `ErrorDialogElement` which extends `AlertDialogElement` 
   - Removes `open` attribute and makes prop readonly on class 
   - Overrides core to `ErrorDialogCore`
   - If extending fails or doesn't feel right, need composition strategy or mixin.
- New `ErrorDialogRoot` React component
  - Doesn't need controlled prop checks like `AlertDialogRoot`
  - Provides `AlertDialogContextProvider`
  - `ErrorDialog` namespace re-exports all other `AlertDialog` parts.
- A context on both the HTML and React side which the alert dialog title and description can consume from (controlled value passed down from error dialog root).

Issue #222, opened 2025-12-17
Feature: Internationalization


mihar-22:
Implement internationalization (i18n) support for Video.js 10 — translated `aria-label`s, tooltips, time unit strings, and dialog copy, plus locale-aware `Intl` formatting. Works with or without a provider by inheriting the native HTML `lang` attribute, keeps skins presentation-only, and ships CDN-compatible locale modules.

**Design doc:** [`internal/design/i18n.md`](https://github.com/videojs/v10/blob/main/internal/design/i18n.md)

## Goals

- **Opaque camelCase keys** — `play`, `seekForward`, etc. English strings live in `en.ts` as values, not keys
- **Global registry** — `registerI18n(locale, translations)` is the single entry point; locale files are ES modules that self-register
- **Ambient `<html lang>` inheritance** — providers read the native `lang` attribute; zero config for the common case
- **Decoupled from skins** — skins are i18n-unaware; providers distribute context, components auto-forward labels via `I18nController`
- **Typed keys** — `Translator` enforces per-key param shapes via overloaded signatures and `Contains<>` template literals
- **CDN-compatible** — locale files are self-registering ES modules with no global namespace pollution
- **Locale-aware formatting** — native `Intl.DurationFormat`, `Intl.NumberFormat`, `Intl.PluralRules` handle time, percent, and plural selection

## Architecture overview

```
@videojs/core/i18n
  Translations · Translator · createTranslator
  registerI18n · getI18nTranslations · onI18nRegistryChange
  Registry (singleton Map) · en.ts (base layer, always present)
        │
        ├── @videojs/react/i18n
        │     createI18n() → { I18nContext, I18nProvider, useTranslator, useLocale }
        │
        └── @videojs/html/i18n
              createI18n() → { context, I18nController, ProviderMixin, TextMixin }
              <media-i18n-provider lang="es">, <media-text key="play">
              MediaButtonElement uses I18nController → auto-forwards aria-labels
```

## Tier 1 — Foundation

- [ ] #1363 — Core i18n types, translator, and registry

## Tier 2 — Platform layers

- [ ] #1364 — React i18n layer (@videojs/react/i18n)
- [ ] #1365 — HTML i18n layer (@videojs/html/i18n)

## Tier 3 — Component integration

- [ ] #1366 — Opaque translation keys in core UI components
- [ ] #1367 — Wire I18nController into UI elements and move error dialog defaults to core

## Tier 4 — Content and distribution

[...]

PR #1122, opened 2026-03-25
docs(design): i18n

### LINKED ISSUE #1381: Design: Internationalization
Create a design document for how we could implement internationalization, incorporating all feedback and suggestions so far.

sampotts:
Ref: https://github.com/videojs/v10/issues/222

## Summary

Add a design document for the i18n (internationalization) system, capturing the architecture decisions and rationale for how Video.js 10 will handle localization across its monorepo packages.

Closes #1381 

## Changes

- Architecture overview covering the i18n system design and package integration points
- Decision log documenting tradeoffs and the reasoning behind chosen approaches
- Index page tying the docs together

## Testing

Documentation only — no code changes.

---

> [!NOTE]
> **Low Risk**
> Low risk documentation-only addition plus a minor `.gitignore` update; no runtime behavior or APIs are changed in this PR.
> 
> **Overview**
> Adds a comprehensive draft design document (`internal/design/i18n.md`) specifying the proposed Video.js 10 internationalization architecture and APIs (registry-based translation packs, HTML/React provider patterns, locale fallback rules, and CDN-friendly locale modules).
> 
> Updates `.gitignore` to also ignore pnpm’s local store directory (`.pnpm-store/`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc2ca7e7dfef184ef19fb5adad68159ebcf90d27. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/design/i18n.md
+++ b/internal/design/i18n.md
@@ -1,0 +1,1041 @@
+---
+status: draft
+date: 2026-03-25
+---
+
+# Internationalization (i18n)
+
+Opaque-key translation system with a global registry and ES module locale files for CDN, typed hooks for React.
+
+## Problem
+
+All `aria-label`, `aria-valuetext`, tooltip labels, and time unit strings in Video.js 10 are hardcoded English. `PlayButtonCore.getLabel()` returns `'play'`. `formatDuration` returns `'2 minutes, 30 seconds'`. There is no mechanism to supply translated strings without overriding each component's `label` prop individually, and no single language switch for skins.
+
+Requirements:
+
+- Single entry point — one API replaces all English defaults regardless of layer (HTML, React, CDN)
+- Stable keys — renaming an English label must not require updating every locale file
+- Typed keys — TypeScript autocomplete for all keys
+- Decoupled — i18n providers are not embedded in skins; any skin works with any provider
+- Works without skins — standalone VJS components (buttons, sliders, time elements used outside of `<video-skin>`) can consume a provider directly
+- CDN-compatible — ES module locale files self-register with no bundler and no global namespace pollution
+- No side-effect locale imports — bundlers don't accidentally include all languages
+- Framework-agnostic core — the registry and translator live in `@videojs/core/i18n`, re-exported to consumers as `@videojs/html/i18n` and `@videojs/react/i18n`
+- Locale-aware — native `Intl` APIs handle duration, number, and plural formatting
+
+## API
+
+### Registering translations
+
+The single entry point for both HTML and React consumers:
+
+```ts
+// HTML
+import { registerI18n } from '@videojs/html/i18n';
+import es from '@videojs/html/i18n/locales/es';
+
+// React
+import { registerI18n } from '@videojs/react';
+import es from '@videojs/react/i18n/locales/es';
+
+registerI18n('es', es);
+```
+
+`registerI18n` merges (does not replace) — multiple calls for the same language are additive. English defaults (`en.ts`) are pre-registered as the base layer on module import.
+
+### HTML
+
+`<media-i18n-provider>` reads from the nearest ancestor `lang` attribute, matching the inheritance model of the native HTML `lang` attribute. Setting `lang` on `<html>` — or any ancestor — is enough:
+
+```html
+<html lang="es">
+  <body>
+    <video-player>
+      <media-i18n-provider>
+        <video-skin>
+          <video src="..." playsinline></video>
+        </video-skin>
+      </media-i18n-provider>
+    </video-player>
+  </body>
+</html>
+```
+
+An explicit `lang` attribute on the provider overrides the inherited value — useful when a single page hosts players in different languages:
+
+```html
+<html lang="en">
+  <media-i18n-provider lang="es">
+    <video-player>
+      <video-skin>
+        <video src="..." playsinline></video>
+      </video-skin>
+    </video-player>
+  </media-i18n-provider>
+</html>
+```
+
+The element reads from the registry for whichever `lang` is active. No changes to `<video-skin>` or any other skin element.
+
+**CDN**
+
+Locale files are self-registering ES modules. They import `registerI18n` from the same CDN module URL, sharing the same registry instance — no global namespace pollution:
+
+```html
+<script type="module" src="https://cdn.jsdelivr.net/npm/@videojs/html/cdn/video.js"></script>
+<script type="module" src="https://cdn.jsdelivr.net/npm/@videojs/html/cdn/locales/es.js"></script>
+
+<html lang="es">
+  <video-player>
+    <media-i18n-provider>
+      <video-skin>
+        <video src="..." playsinline></video>
+      </video-skin>
+    </media-i18n-provider>
+  </video-player>
+</html>
+```
+
+```js
+// es.js — self-registering, no window touch
+import { registerI18n } from 'https://cdn.jsdelivr.net/npm/@videojs/html/cdn/i18n.js';
+registerI18n('es', { play: 'Reproducir', pause: 'Pausa', /* … */ });
+```
+
+**Custom registration**
+
+```ts
+// Override individual keys
+registerI18n('es', { play: 'Comenzar' });
+
+// Or override a built-in pack entirely
+import myEs from './my-es';
+registerI18n('es', myEs);
+```
+
+### React
+
+`I18nProvider` is a standalone component. Consumers add it explicitly inside the player's `Provider` — the skin itself has no i18n awareness. When `locale` is omitted, it reads from the nearest ancestor `lang` attribute (typically `<html lang="…">`):
+
+```tsx
+import { I18nProvider } from '@videojs/react';
+import { VideoSkin, Video } from '@videojs/react/video';
+
+// locale inherited from <html lang="es">
+<Provider>
+  <I18nProvider>
+    <VideoSkin>
+      <Video src="..." playsInline />
+    </VideoSkin>
+  </I18nProvider>
+</Provider>
+```
+
+Pass `locale` explicitly to override inheritance, or `translations` to bypass the registry entirely:
+
+```tsx
+import es from '@videojs/react/i18n/locales/es';
+
+<Provider>
+  <I18nProvider locale="es" translations={es}>
+    <VideoSkin>
+      <Video src="..." playsInline />
+    </VideoSkin>
+  </I18nProvider>
+</Provider>
+```
+
+**SSR / no flash-of-English** — pass `translations` directly on first render:
+
+```tsx
+// Server component
+const { default: translations } = await import(`@videojs/react/i18n/locales/${locale}`);
+
+<Provider>
+  <I18nProvider locale={locale} translations={translations}>
+    <VideoSkin>
+      <Video src="..." playsInline />
+    </VideoSkin>
+  </I18nProvider>
+</Provider>
+```
+
+**Dynamic switching**
+
+Just flip `locale` — the provider lazy-loads the built-in pack for the new locale:
+
+```tsx
+const [locale, setLocale] = useState('en');
+
+<Provider>
+  <I18nProvider locale={locale}>
+    <VideoSkin>
+      <Video src="..." playsInline />
+    </VideoSkin>
+  </I18nProvider>
+</Provider>
+```
+
+Or for zero-flash switching, pre-import the locale and pass `translations` directly:
+
+```tsx
+const [{ locale, translations }, setLocale] = useState({ locale: 'en', translations: undefined });
+
+async function switchTo(next: string) {
+  const { default: translations } = await import(`@videojs/react/i18n/locales/${next}`);
+  setLocale({ locale: next, translations });
+}
+```
+
+**With next-intl**
+
+```tsx
+import { getLocale } from 'next-intl/server';
+
+const locale = await getLocale();
+const { default: translations } = await import(`@videojs/react/i18n/locales/${locale}`);
+
+<Provider>
+  <I18nProvider locale={locale} translations={translations}>
+    <VideoSkin>
+      <Video src="..." playsInline />
+    </VideoSkin>
+  </I18nProvider>
+</Provider>
+```
+
+### Translation keys
+
+Keys are opaque camelCase identifiers. The English string is the value in `en.ts` — not the key. All keys are optional; missing keys fall back to the English default.
+
+| Key | English default | Params | Used by |
+| --- | --------------- | ------ | ------- |
+| `play` | `'Play'` | — | `PlayButtonCore` |
+| `pause` | `'Pause'` | — | `PlayButtonCore` |
+| `replay` | `'Replay'` | — | `PlayButtonCore` |
+| `mute` | `'Mute'` | — | `MuteButtonCore` |
+| `unmute` | `'Unmute'` | — | `MuteButtonCore` |
+| `seek` | `'Seek'` | — | `TimeSliderCore` (aria-label) |
+| `volume` | `'Volume'` | — | `VolumeSliderCore` (aria-label) |
+| `muted` | `'muted'` | — | `VolumeSliderCore` (aria-valuetext suffix) |
+| `enterFullscreen` | `'Enter fullscreen'` | — | `FullscreenButtonCore` |
+| `exitFullscreen` | `'Exit fullscreen'` | — | `FullscreenButtonCore` |
+| `enableCaptions` | `'Enable captions'` | — | `CaptionsButtonCore` |
+| `disableCaptions` | `'Disable captions'` | — | `CaptionsButtonCore` |
+| `enterPictureInPicture` | `'Enter picture-in-picture'` | — | `PipButtonCore` |
+| `exitPictureInPicture` | `'Exit picture-in-picture'` | — | `PipButtonCore` |
+| `currentTime` | `'Current time'` | — | `TimeCore` |
+| `duration` | `'Duration'` | — | `TimeCore` |
+| `remaining` | `'Remaining'` | — | `TimeCore` |
+| `seekForward` | `'Seek forward {seconds} seconds'` | `{seconds}` | `SeekButtonCore` |
+| `seekBackward` | `'Seek backward {seconds} seconds'` | `{seconds}` | `SeekButtonCore` |
+| `playbackRate` | `'Playback rate {rate}'` | `{rate}` | `PlaybackRateButtonCore` |
+| `timePosition` | `'{current} of {duration}'` | `{current}`, `{duration}` | `TimeSliderCore` (aria-valuetext) |
+| `remaining` | `'remaining'` | — | `formatDuration` (negative time suffix) |
+
+> `timePosition` params are already-formatted time phrases from `Intl.DurationFormat`, not raw numbers.
+
+> `Intl.DurationFormat` handles all duration unit labels; `Intl.NumberFormat` handles percent formatting. Only `muted` and `remaining` are translation keys because `Intl` has no concept of those suffixes.
+
+## Architecture
+
+### Layers
+
+```
+@videojs/core/i18n
+  Translations · Translator · createTranslator
+  registerI18n · getI18nTranslations · onI18nRegistryChange
+  Registry (singleton Map) · en.ts (base layer, always present)
+        │
+        ├── @videojs/react/i18n
+        │     createI18n() → { I18nContext, I18nProvider, useTranslator, useLocale }
+        │     Consumer wraps skin: <I18nProvider locale="es" translations={…}>
+        │
+        └── @videojs/html/i18n
+              createI18n() → { context, I18nController, ProviderMixin, TextMixin }
+              <media-i18n-provider lang="es"> = ProviderMixin(ReactiveElement)
+              <media-text key="play"> = TextMixin(ReactiveElement)
+              MediaButtonElement uses I18nController → t(core.getLabel(state))
+
+@videojs/utils/i18n   pluralize(count, forms, locale?)
+@videojs/utils/time   formatDuration(seconds, options?)
+```
+
+### Core types
+
+```ts
+// @videojs/core/i18n/types.ts
+
+export type BuiltInLocale =
+  | 'ar' | 'de' | 'es' | 'fr' | 'it' | 'ja'
+  | 'ko' | 'nl' | 'pl' | 'pt' | 'ru' | 'tr' | 'zh';
+
+/** Any BCP 47 tag; named built-in locales autocomplete in editors. */
+export type Locale = BuiltInLocale | (string & {});
+
+/** Helper — string type that must contain a given literal substring. */
+type Contains<Needle extends string> = `${string}${Needle}${string}`;
+
+/** Per-key parameter contract — maps each key to the exact params it requires (or `never`). */
+export interface TranslationParams {
+  play: never;
+  pause: never;
+  replay: never;
+  mute: never;
+  unmute: never;
+  seek: never;
+  volume: never;
+  muted: never;
+  enterFullscreen: never;
+  exitFullscreen: never;
+  enableCaptions: never;
+  disableCaptions: never;
+  enterPictureInPicture: never;
+  exitPictureInPicture: never;
+  currentTime: never;
+  duration: never;
+  remaining: never;
+  seekForward: { seconds: string | number };
+  seekBackward: { seconds: string | number };
+  playbackRate: { rate: string | number };
+  timePosition: { current: string | number; duration: string | number };
+}
+
+/** All player translation keys. All keys are optional — missing keys fall back to English.
+ *  Value types enforce the `{param}` placeholders required by `TranslationParams`. */
+export interface Translations {
+  play?: string;
+  pause?: string;
+  replay?: string;
+  mute?: string;
+  unmute?: string;
+  seek?: string;
+  volume?: string;
+  muted?: string;
+  enterFullscreen?: string;
+  exitFullscreen?: string;
+  enableCaptions?: string;
+  disableCaptions?: string;
+  enterPictureInPicture?: string;
+  exitPictureInPicture?: string;
+  currentTime?: string;
+  duration?: string;
+  remaining?: string;
+  seekForward?: Contains<'{seconds}'>;
+  seekBackward?: Contains<'{seconds}'>;
+  playbackRate?: Contains<'{rate}'>;
+  timePosition?: Contains<'{current}'> & Contains<'{duration}'>;
+}
+
+type SimpleKeys = { [K in keyof TranslationParams]: TranslationParams[K] extends never ? K : never }[keyof TranslationParams];
+type ParamKeys = Exclude<keyof TranslationParams, SimpleKeys>;
+
+/** Callable translator. Keys with params require the exact params object; keys without params take no second argument. */
+export type Translator = {
+  (key: SimpleKeys): string;
+  <K extends ParamKeys>(key: K, params: TranslationParams[K]): string;
+  readonly locale?: Locale;
+};
+```
+
+The `Contains<Needle>` helper makes locale files compile-error when they forget a `{param}` placeholder. `Translator`'s overloaded signature catches two classes of mistake at call sites:
+
+```ts
+t('play');                              // ✓ no params
+t('play', { foo: 1 });                  // ✗ TS error — no params accepted
+t('seekForward');                       // ✗ TS error — missing { seconds }
+t('seekForward', { seconds: 10 });      // ✓
+t('seekForward', { second: 10 });       // ✗ TS error — typo, wrong key
+```
+
+### Registry
+
+```ts
+// @videojs/core/i18n/registry.ts
+const registry = new Map<string, Partial<Translations>>();
+const subscribers = new Set<() => void>();
+
+// en.ts is the base layer — pre-registered at module init
+import en from './locales/en';
+registry.set('en', en);
+
+export function registerI18n(locale: string, translations: Partial<Translations>): void {
+  const existing = registry.get(locale) ?? {};
+  registry.set(locale, { ...existing, ...translations });
+  subscribers.forEach(fn => fn());
+}
+
+export function getI18nTranslations(locale: string): Partial<Translations> {
+  const en = registry.get('en') ?? {};
+  // Walk up the BCP 47 subtag chain: zh-Hant-HK → zh-Hant → zh → en
+  const parts = locale.split('-');
+  for (let i = parts.length; i > 0; i--) {
+    const tag = parts.slice(0, i).join('-');
+    const pack = registry.get(tag);
+    if (pack) return { ...en, ...pack };
+  }
+  return { ...en };
+}
+
+export function onI18nRegistryChange(callback: () => void): () => void {
+  subscribers.add(callback);
+  return () => subscribers.delete(callback);
+}
+```
+
+`getI18nTranslations` always returns at least the English defaults (from `en.ts`). No key is ever missing — the per-key fallback is `registry[locale][key] → registry[parent-subtag][key] → … → en[key] → key`. See the [BCP 47 fallback](#bcp-47-fallback) section for the full subtag-truncation rules.
+
+### `createTranslator`
+
+```ts
+export function createTranslator(
+  translations: Partial<Translations> = {},
+  locale?: Locale
+): Translator {
+  // Overload signatures are type-only; implementation uses the loose callable form.
+  function t(key: keyof Translations, params?: Record<string, string | number>): string {
+    let value = (translations[key] ?? key) as string;
+    if (params) {
+      for (const [k, v] of Object.entries(params)) {
+        value = value.replace(`{${k}}`, String(v));
+      }
+    }
+    return value;
+  }
+  t.locale = locale;
+  return t as Translator;
+}
+```
+
+### Locale files
+
+Locale files are TypeScript source files. They use `satisfies Partial<Translations>` so typos in key names are caught at build time — a misspelled key is a compile error, not a silent runtime miss:
+
+```ts
+// locales/en.ts
+import type { Translations } from '../types';
+
+export default {
+  play: 'Play',
+  pause: 'Pause',
+  replay: 'Replay',
+  mute: 'Mute',
+  unmute: 'Unmute',
+  seek: 'Seek',
+  volume: 'Volume',
+  muted: 'muted',
+  enterFullscreen: 'Enter fullscreen',
+  exitFullscreen: 'Exit fullscreen',
+  enableCaptions: 'Enable captions',
+  disableCaptions: 'Disable captions',
+  enterPictureInPicture: 'Enter picture-in-picture',
+  exitPictureInPicture: 'Exit picture-in-picture',
+  currentTime: 'Current time',
+  duration: 'Duration',
+  remaining: 'Remaining',
+  seekForward: 'Seek forward {seconds} seconds',
+  seekBackward: 'Seek backward {seconds} seconds',
+  playbackRate: 'Playback rate {rate}',
+  timePosition: '{current} of {duration}',
+} satisfies Partial<Translations>;
+```
+
+```ts
+// locales/es.ts — partial locale, falls back to en.ts for missing keys
+import type { Translations } from '../types';
+
+export default {
+  // Play controls
+  play: 'Reproducir',
+  pause: 'Pausa',
+  replay: 'Repetir',
+  // Mute
+  mute: 'Silenciar',
+  unmute: 'Activar sonido',
+  // …
+} satisfies Partial<Translations>;
+```
+
+The `satisfies` constraint checks key names and value types without widening the type — `keyof typeof es` remains the narrow set of keys present, not `keyof Translations`. Comments above groups of keys are useful context for translators.
+
+For CDN, locale files are self-registering ES modules. The build emits a separate CDN entry point alongside the bundler-importable output:
+
+```ts
+// locales/es.cdn.ts — CDN entry, self-registers
+import { registerI18n } from '../registry';
+import translations from './es';
+
+registerI18n('es', translations);
+```
+
+```js
+// Built output: es.js (CDN)
+import { registerI18n } from 'https://cdn.jsdelivr.net/npm/@videojs/html/cdn/i18n.js';
+registerI18n('es', { play: 'Reproducir', pause: 'Pausa', /* … */ });
+```
+
+### React: `createI18n`
+
+A factory that closes over a React context so `useTranslator()` returns `Translator` with no explicit type annotation at call sites. Called once in `@videojs/react/i18n`; the result is exported directly.
+
+When `locale` is omitted, `I18nProvider` reads `document.documentElement.lang` and subscribes to changes via `MutationObserver`:
+
+```tsx
+export function createI18n() {
+  const I18nContext = createContext<Translator | null>(null);
+
+  function I18nProvider({ locale: explicitLocale, translations, children }: {
+    locale?: Locale;
+    translations?: Partial<Translations>;
+    children: ReactNode;
+  }) {
+    // Subscribe to ambient DOM lang attribute when no explicit prop is given
+    const ambientLocale = useSyncExternalStore(
+      subscribeToDocumentLang,
+      () => document.documentElement.lang || undefined,
+      () => undefined, // SSR: no ambient locale
+    );
+    const locale = explicitLocale ?? ambientLocale;
+
+    // Lazy-load built-in pack when locale is set
+    const [builtIn, setBuiltIn] = useState<Partial<Translations>>({});
+    useEffect(() => {
+      if (!locale) { setBuiltIn({}); return; }
+      const tags = [...new Set([locale, locale.split('-')[0]])];
+      const load = (i = 0): Promise<void> =>
+        import(`@videojs/core/i18n/locales/${tags[i]}`)
+          .then(m => { registerI18n(tags[i], m.default); setBuiltIn(m.default); })
+          .catch(() => i + 1 < tags.length ? load(i + 1) : setBuiltIn({}));
+      load();
+    }, [locale]);
+
+    // Browser Translation API — background fallback, pre-installed model only.
+    // Enumerates the full key set from en.ts (no consumer input needed).
+    const [browserTranslated, setBrowserTranslated] = useState<Partial<Translations>>({});
+    useEffect(() => {
+      if (!locale) { setBrowserTranslated({}); return; }
+      let cancelled = false;
+      getBrowserTranslations(locale).then(result => {
+        if (!cancelled) setBrowserTranslated(result);
+      });
+      return () => { cancelled = true; };
+    }, [locale]);
+
+    // Priority: browser API < registry/built-in < consumer translations
+    const translator = useMemo(
+      () => createTranslator(
+        { ...browserTranslated, ...getI18nTranslations(locale ?? 'en'), ...builtIn, ...translations },
+        locale
+      ),
+      [browserTranslated, builtIn, translations, locale]
+    );
+
+    return <I18nContext.Provider value={translator}>{children}</I18nContext.Provider>;
+  }
+
+  function useTranslator(): Translator {
+    return useContext(I18nContext) ?? createTranslator(getI18nTranslations('en'));
+  }
+
+  function useLocale(): Locale | undefined {
+    return useTranslator().locale;
+  }
+
+  return { I18nContext, I18nProvider, useTranslator, useLocale };
+}
+
+/** Shared MutationObserver subscription for document.documentElement.lang changes. */
+function subscribeToDocumentLang(onChange: () => void): () => void {
+  const observer = new MutationObserver(onChange);
+  observer.observe(document.documentElement, { attributes: true, attributeFilter: ['lang'] });
+  return () => observer.disconnect();
+}
+```
+
+### HTML: `createI18n` factory
+
+Parallel to React's `createI18n`, `createI18n` bundles the context, controller, provider element, and text element into one call. `@videojs/html/i18n` invokes it once and re-exports the result. Consumer-facing API is a single side-effect import that registers `<media-i18n-provider>` / `<media-text>`:
+
+```ts
+// @videojs/html/i18n/create-i18n.ts
+
+export function createI18n() {
+  const context = createContext<Translator>(Symbol('@videojs/i18n'));
+
+  /** Controller for any element that consumes translated strings (e.g. button labels). */
+  class I18nController implements ReactiveController { /* subscribes to context */ }
+
+  /** Mixin: adds lang attribute + DOM lang resolution + ContextProvider for i18nContext. */
+  const ProviderMixin = <Base extends Constructor<ReactiveElement>>(base: Base) =>
+    class extends base { /* see "Provider mixin" below */ };
+
+  /** Mixin: adds key attribute + subscribes to i18nContext + writes textContent. */
+  const TextMixin = <Base extends Constructor<ReactiveElement>>(base: Base) =>
+    class extends base { /* see "Text mixin" below */ };
+
+  return { context, I18nController, ProviderMixin, TextMixin };
+}
+```
+
+Mirrors `createPlayer` — the factory returns building blocks (context, controller, mixins). Define files compose them onto concrete bases and register the custom elements:
+
+```ts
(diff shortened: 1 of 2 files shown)


internal/design/i18n/index.md
```diff
@@ -0,0 +1,238 @@
...
+<VideoSkin locale="es" src={src} />
+```
+
+The pack is loaded via dynamic `import()` so it has zero impact on the default bundle. If no pack exists for the requested locale, the player falls back to English.
+
+Consumer-provided `translations` always take priority over built-in packs:
+
+```tsx
+// Built-in "es" pack loads, then "Play" is overridden by the consumer value
+<VideoSkin locale="es" translations={{ Play: 'Reproducir' }} src={src} />
```
cjpillsbury: suggestion(non-blocking): consider using `lang` for HTML consistency and interop (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang)

internal/design/i18n/index.md
```diff
@@ -0,0 +1,238 @@
...
+### HTML
+
+```ts
+// Set via property (object)
+skinElement.translations = { Play: 'Reproducir', Pause: 'Pausa' };
+```
+
+```html
+<!-- Or via attribute (JSON string) -->
+<video-skin translations='{"Play":"Reproducir","Pause":"Pausa"}'></video-skin>
```
cjpillsbury: question(blocking): Interested in others' thoughts on this. We've _**mostly**_ (though not entirely) tried to avoid JSON in attributes in Media Chrome & Mux Player/Elements (cc: @luwes)

internal/design/i18n/index.md
```diff
@@ -0,0 +1,238 @@
...
+## Anatomy
+
+### React
+
+The consumer API is a `locale` and `translations` prop directly on the skin. The `translations` type is the skin's own `SkinTranslations`, which extends `Translations` — so core keys and skin-specific keys are overridden in one object:
+
+```tsx
+import { VideoSkin } from '@videojs/react/video';
+
+<VideoSkin
```
cjpillsbury: question(blocking): Have there been any discussions on the tradeoffs of including the provider in the skin vs. elsewhere, e.g.: 
* bundled with our MediaProvider, which can still be two distinct providers "under the hood"
* a standalone provider that folks can opt in to use as they see fit
* something else?
Concerned about this breaking our broad pattern of not adding too much "configuration" to a Skin. (BYO Skins, Skins without labels if they want, etc. etc.)

internal/design/i18n/decisions.md
```diff
@@ -0,0 +1,206 @@
...
+
+**Rationale.** English-as-key gives automatic fallback for any missing translation with zero configuration — `t('Play')` returns `'Play'` if the translations object has no `Play` entry. It is self-documenting (`t('Seek forward {seconds} seconds')` is immediately clear). The tradeoff — renaming a key is a breaking change for existing translation files — is acceptable because these are stable UI labels that rarely change. `Translations` as a TypeScript interface catches key mismatches at compile time, reducing the cost of any future rename.
+
+This pattern is established in Video.js v7/8, GNU gettext, and Flutter intl.
+
+---
+
+## One `Translations` type, not two
+
+**Decision.** There is a single `Translations` interface that carries both the index signature (`[key: string]: string | undefined`) and all the named player keys. Skins extend it directly with their own keys. There is no separate structural base type.
```
cjpillsbury: question(non-blocking): I'm wondering if we should make this (in the TypeScript sense), an extendable set of well-defined string literals instead of simply `string`? I suspect @mihar-22 could help you with that.

internal/design/i18n/decisions.md
```diff
@@ -0,0 +1,206 @@
...
+
+**Alternatives considered.**
+
+- **Separate base interface** (e.g. `BaseTranslations` with only the index signature, `Translations extends BaseTranslations` with the named keys). An extra type and export whose only job is to donate an index signature.
+
+**Rationale.** A base interface that exists solely to donate an index signature is needless indirection — the index signature belongs on the type that consumers actually work with. The HTML `i18nContext` previously typed to a structural base (`Translator<BaseTranslations>`) now uses `Translator<Translations>` directly, which is equally assignable from any `Translator<SkinTranslations>` since `SkinTranslations extends Translations`.
```
cjpillsbury: question(non-blocking): My concern here is that there will be no "hints" for folks on whether a particular "key" is valid. This can easily result in dev toil/maintenance difficulties in both using the API and defining a JSON object for a new lang.

internal/design/i18n/decisions.md
```diff
@@ -0,0 +1,206 @@
...
+
+---
+
+## `createTranslator` factory, not a class or singleton
+
+**Decision.** A plain factory function returning a closure. No class, no module-level state.
```
cjpillsbury: Agree. This feels reversible if needed, and a factory feels like the "right" abstraction, even if we eventually need an object (class or POJO) interface or (small s) singleton pattern.

internal/design/i18n/decisions.md
```diff
@@ -0,0 +1,206 @@
...
+- **Global context typed to `Translations`** — typed, but skins cannot extend it without a cast.
+- **Two contexts** — one for core, one for skin. Consumers would need to merge them.
+
+**Rationale.** Generics allow each skin to extend `Translations` with its own keys (e.g., `'Skip Intro'`, `'Cast to {device}'`) while keeping full TypeScript autocomplete in components. Using `Record<string, string>` loses all key safety. The factory pattern mirrors `createPlayer` — skins already call `createPlayer(features)` to produce a typed store; `createI18nHooks<SkinTranslations>()` follows the same shape and mental model.
+
+---
+
+## `{param}` interpolation, not ICU message format
+
+**Decision.** Simple `{key}` replacement, as in `t('Seek forward {seconds} seconds', { seconds: 10 })`.
```
cjpillsbury: I was equivocating on this one, but I think I agree. This is also potentially reversible/extensible if we ever need it. See, e.g. https://react.i18next.com/misc/using-with-icu-format for prior art.

internal/design/i18n/decisions.md
```diff
@@ -0,0 +1,206 @@
...
+- **Template literals via a tagged function** — `t\`Seek forward ${seconds} seconds\``. No library needed but harder to extract keys for translation files.
+- **Positional args** — `t('Seek forward %s seconds', 10)`. Compact but brittle when argument order differs between languages.
+
+**Rationale.** The player has a small, known set of interpolated strings; none require plural rules at the interpolation site. Pluralization for time units is handled by `Intl.PluralRules` via the `pluralize` utility (or automatically by `Intl.DurationFormat`), not at the `t()` call site. `{key}` replacement adds zero runtime weight, is readable, and is the pattern established in Video.js v8/9.
+
+---
+
+## Native `Intl` APIs for locale-aware formatting
+
+**Decision.** Use `Intl.DurationFormat` for time phrases, `Intl.NumberFormat` with `style: 'percent'` for volume values, and `Intl.PluralRules` (via `pluralize`) for plural selection — with translation-key fallbacks for environments that lack `Intl.DurationFormat`.
```
cjpillsbury: Agree. 👌

internal/design/i18n/decisions.md
```diff
@@ -0,0 +1,206 @@
...
+## Translation applied at the UI layer, not inside Core classes
+
+**Decision.** `PlayButtonCore.getLabel()`, `MuteButtonCore.getLabel()`, etc. continue returning English strings. React components and HTML custom elements apply `t()` before setting `aria-label`.
+
+**Alternatives considered.**
+
+- **Pass `Translator` into each Core class** — via constructor or `setProps({ t })`. Core is the single source of truth for all string production.
+- **Add `translate?` param to `getLabel` and `getAttrs`** — Core calls `translate?.('Play') ?? 'Play'` internally.
+
+**Rationale.** Core is runtime-agnostic and framework-independent. Threading a context-derived translator into Core would create implicit coupling to whichever context system the caller uses (React context, DOM context, or a test double). The UI layer already holds the translator and is the natural place to apply it — it is the same layer that decides which attributes to set on the element. Core producing the key and the UI layer calling `t(key)` keeps the layers clean.
```
cjpillsbury: question(non-blocking): I know we've gone down the path of expecting folks to pass in all labels etc, but I think this will create a lot of noise/burden on the user for the "standard use case". I still think these should be treated more like our built in state management, and they can be directly related to identical logic as state changes (e.g. showing both "Play" and the play icon in paused state; showing "Pause" and the pause icon in paused state). We weave that context through core already and have already solved those hard problems. That said, I believe this can be a future effort and there are no inherent "left turns instead of rights", so I don't think this should be a blocker.

internal/design/i18n/decisions.md
```diff
@@ -0,0 +1,206 @@
...
+- **`locale` inside `translations` object** — `translations.$locale = 'es'`. Unusual convention; mixes metadata with content.
+- **Infer locale from the `lang` attribute on `<html>`** — readable on client via `document.documentElement.lang`. Server-rendered HTML sets this correctly, but reading it from the DOM is implicit and fragile if the player is embedded in an iframe or shadow DOM.
+
+**Rationale.** Passing `locale` explicitly matches how React Aria (`<I18nProvider locale="fr-FR">`) and MUI (`adapterLocale="de"`) handle it. It keeps `Intl` API calls deterministic across server and client — the only way to avoid hydration mismatches caused by `Intl` formatting differences. Carrying `locale` on the translator function means components that need it for `formatDuration` (time slider, time display) don't need a separate `useLocale()` call; `useLocale()` is a thin wrapper for callers that prefer named access.
+
+---
+
+## Separate `i18nContext` for HTML, not extending `playerContext`
+
+**Decision.** A new `i18nContext` is created alongside `playerContext`, `mediaContext`, and `containerContext`.
```
cjpillsbury: Agree. This is a different context with different responsibilities. **_However_** (non-blocking), we may want well-defined hooks that "know about" **_both_** contexts. This relates to my earlier comment about state, above, and was related to our earlier conversations on props hooks, react-aria/adobe spectrum points of inspiration, and the like. e.g. something _**like**_ `usePlayButtonProps(state, props)` having reference to the I18N context. Like above, this is non-blocking, since nothing you're proposing precludes or "code away from" this in a future effort.

internal/design/i18n/index.md
```diff
@@ -0,0 +1,238 @@
...
+<VideoSkin locale="es" translations={{ Play: 'Reproducir' }} src={src} />
+```
+
+For SSR or when you want zero flash-of-English on first render, pre-import the JSON and pass it directly:
+
+```tsx
+import es from '@videojs/core/i18n/locales/es.json';
+
+<VideoSkin locale="es" translations={es} src={src} />
+```
```
decepulis: There's a tiny bit of awkwardness, here, having two props that essentially describe the same thing. `locale="es" translations={es}`. I think I see how we got here. Like, we don't know if `translations` is a partial, so we need `locale` to fall back on. But. Just. Calling it out.

internal/design/i18n/decisions.md
```diff
@@ -0,0 +1,206 @@
...
+- **Module-level global translator** — `setGlobalTranslator(createTranslator(translations))`. Simple API, but incompatible with multiple players on the same page using different languages.
+- **Property polling** — child elements call `this.closest('[translations]')?.translator`. Fragile; breaks with shadow DOM.
+
+**Rationale.** `playerContext` is typed by the store's feature set. Adding a translation type would require changes across the entire store and controller infrastructure with no other benefit. A module-level global breaks the multi-player case. A separate context follows the existing pattern (`alertDialogContext`, `sliderContext`, `tooltipContext`) and has zero impact on the store type or the controller API. It is also independently testable.
+
+---
+
+## `I18nMixin` on the skin element, not a separate provider element
+
+**Decision.** `VideoSkinElement.translations` triggers re-provision to all descendants via `I18nMixin`.
```
cjpillsbury: issue(blocking): This I think pretty strongly is a wrong turn, for a few reasons.
- It creates a hard coupling of skin to i18n, meaning you **_need_** to use a skin to use i18n
- It creates a hard coupling of i18n to a UI component, meaning you _**need**_ to use a particular UI component to use (data/state) i18n
- It's a break in our general "compositional" ethos
- It creates an unnecessary disparity between React vs. HTML "for the wrong reasons". While we're okay with divergences, those should be primarily motivated by the idioms of the platform/framework/etc.
- It creates an unnecessary disparity between React vs. HTML that will complicate any future possible compiler work. While this shouldn't be a _**primary**_ motivator for decisions (i.e. our general stance has been that we should make the "right" decisions and deal with any compiler consequences as needed), it's worth surfacing if there are already other concerns on the decision.

internal/design/i18n/decisions.md
```diff
@@ -0,0 +1,206 @@
...
+- **`Intl.BCP47LanguageTag`** — TypeScript's own type for BCP 47 tags. At the time of writing this is aliased to `string`, so it provides no additional safety.
+
+**Rationale.** `BuiltInLocale | (string & {})` is the established TypeScript pattern for "known autocomplete values, open to others" — used in MUI (`palette.mode`), Radix, and CSS-in-JS libraries for property values. The `string & {}` operand prevents TypeScript from simplifying the union to `string`, which would suppress autocomplete for the named members. The result: editors show a dropdown with the 13 built-in locales when the prop is typed, but `'pt-BR'` or any other BCP 47 tag passes without complaint. `BuiltInLocale` is also useful as a standalone export for runtime checks (`value is BuiltInLocale`). `TimeFormatOptions.locale` and `pluralize`'s third argument stay as `string` because those live in `@videojs/utils`, which cannot import from `@videojs/core`.
+
+---
+
+## Built-in locale packs, lazy-loaded
+
+**Decision.** Ship JSON translation files for common languages in `@videojs/core/i18n/locales/`. Load them dynamically when `locale` changes; merge with consumer `translations` (consumer always wins).
```
cjpillsbury: question(non-blocking): Deeply assuming this and only this does have ripple effects that I think are worth talking through, like how locales are written, stored, made available, build tools, https://caniuse.com/mdn-javascript_statements_import_import_attributes_type_json support, etc. etc. I'm interested in others' thoughts here. This also abuts my (out of band) question re: `async translate()` and optional fallbacks on e.g. https://developer.mozilla.org/en-US/docs/Web/API/Translator if/when available in the future.

cjpillsbury (commented):
Although I threw out a few "blocking" comments, I trust the rest of the team to address/discuss sufficiently and I will be out for a long weekend and don't want to slow forward progress, so setting this review as "Comment" only.

internal/design/i18n/architecture.md
```diff
@@ -0,0 +1,781 @@
...
+
+```ts
+const locale = navigator.language; // 'de-DE', 'en-US', etc.
+```
+
+## Locale-based caption selection
+
+### How it works
+
+The store's `textTrackList` already captures `language` (mapped from the DOM `TextTrack.language` / `srclang` attribute) for every track. When `locale` is set, a new store feature compares it against the track list and activates the best match.
```
cjpillsbury: issue(blocking): I think any `textTrackList` should be out of scope for this effort. @heff + @luwes (and also @gkatsev) and I spent a lot of time talking through the difference between (1) "Application" language settings and preferences versus (2) "user" language settings and preferences versus (3) "default stream" language details. What you're working on primarily is (1) and is typically defined by app designers/developers (even if they expose a UI for "locale selection". (2) is generally deferred to things like `navigation.language` and the like, with the possibility of `localStorage` or similar for a dynamic "user preferences" model. (3) is typically defined in a way (per spec) that provides a "default" (e.g. subtitle, audio, etc.) language for a src, but also "assuming no detectable user preferences".  E.g. from the HLS/Pantos/RFC8216 spec:

>        DEFAULT
> 
>       The value is an enumerated-string; valid strings are YES and NO.
>       If the value is YES, then the client SHOULD play this Rendition of
>       the content in the absence of information from the user indicating
>       a different choice.  This attribute is OPTIONAL.  Its absence
>       indicates an implicit value of NO.
> 

I'd just omit this section/effort altogether for now.

cjpillsbury: Also thanks @mister-ben for calling this out on Discord! 👯

internal/design/i18n/decisions.md
```diff
@@ -0,0 +1,206 @@
...
+date: 2026-03-25
+---
+
+# Design Decisions
+
+---
+
+## English strings as translation keys
+
+**Decision.** Translation keys are the English strings themselves. `t('Play')` not `t(TranslationKey.PLAY)`.
```
luwes: We came to the same conclusion in media-chrome. Makes a lot of sense and should also help gzip.
https://github.com/muxinc/media-chrome/pull/1071#discussion_r1978166484

internal/design/i18n/decisions.md
luwes: The component builders should still get those hints right. A base would only be handy for folks creating a new language dict but you can just copy the english dict for example and start changing values.

internal/design/i18n/decisions.md
luwes: Also where we landed at in Media Chrome 👍

internal/design/i18n/architecture.md
```diff
@@ -0,0 +1,781 @@
...
+
+export const { I18nContext, I18nProvider, useTranslator, useLocale } =
+  createI18nHooks<VideoSkinTranslations>();
+```
+
+The skin component exposes `locale` and `translations` (typed as `VideoSkinTranslations`) as top-level props. Internally, `I18nProvider` is nested inside `Provider` so it can call `usePlayer()` for caption selection. Consumers never interact with either provider directly:
+
+```tsx
+// Consumer API — just props on the skin
+<VideoSkin locale="es" translations={{ Play: 'Reproducir', 'Skip Intro': 'Saltar Intro' }} />
```
luwes: Could we consider the `lang` prop / attr here?
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang

We used this in Media Chrome, it can go on every html tag and gives the browsers extra context.

Or is there a reason not to meddle these two?

internal/design/i18n/decisions.md
cjpillsbury: Yeah I remember having similar feelings and convos when we were talking through it that time 😅

internal/design/i18n/decisions.md
cjpillsbury: thought(non-blocking): Yeah this is pretty different from how it typically works in other (non React-y) envs and I expressed similar sentiments when this has come up in the past, but not a hill I need to die on 😅.

Examples where these are well-defined keys:
- ICU CLDR (ignore the deep semantic structure, as that is irrelevant to the core callout) - https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-person-names-full/main/aa-DJ/personNames.json#L56-L59
- Android Applications - https://developer.android.com/guide/topics/resources/localization#managing-strings
- iOS Applications - https://developer.apple.com/documentation/xcode/localizing-and-varying-text-with-a-string-catalog

(there are more, and there are reasons to not conflate presentation with data, esp if we don't want to assume the english phrasing will be "set in stone", as that has p big ripple effects, esp on 3rd party BYO language keys, but I'm happy to let this be y'allz decision in the end)

internal/design/i18n/index.md
sampotts: They're for different uses but maybe the naming in the example is tripping you up. 

- `locale` (or `lang` as folks are pushing for) is used for Intl API calls to format numbers and duration, get plural rules etc.
- `translations` is providing a dictionary of translation overrides. 

Are you maybe suggesting _we_ handle loading the JSON so they don't have to provide the `translations` prop here?

internal/design/i18n/index.md
sampotts: I'm torn on this one. I do tend to favor using native naming where possible but in this case there are two native names used; the HTML attribute is `lang` but the Intl APIs use `locale` as do a lot of other i18n-related libs.

That said, I'm not precious either way and it was something I went back and forth on a couple of times.

internal/design/i18n/index.md
sampotts: Happy to amend this as required. I wasn't sure what best practice was here.

internal/design/i18n/index.md
sampotts: This is another aspect I went back and forth on. Happy to discuss next week (I'll setup a call) but I figured the skins may own the translations it requires. That said given our eject stance, we could just have "core" translations that cover all of our skins requirements and ejected folks can opt to use our i18n or (more likely) their own implementation.

sampotts: What I'm not keen on is:

```tsx
<AProvider>
  <BProvider>
    <CProvider>
      <DProvider>
        {children}
      </DProvider>
    </CProvider>
  </BProvider>
</AProvider>
```

😆

internal/design/i18n/decisions.md
sampotts: If we're not going to both with skin-specific translations, this would be miles easier for sure. I'm all for strongly typed where possible.

internal/design/i18n/decisions.md
sampotts: Sure, can change this. As above, perhaps we don't need to allow our own skins to have their own translations anyway as it just complicates things needlessly. Something to also discuss.

internal/design/i18n/architecture.md
sampotts: Good call. This can come later.

internal/design/i18n/architecture.md
sampotts: Related above:
https://github.com/videojs/v10/pull/1122#discussion_r2998283409

internal/design/i18n/architecture.md
```diff
@@ -0,0 +1,781 @@
...
+ * The locale tags for which Video.js ships built-in translation packs.
+ * Surfaces as autocomplete on the `locale` prop while still accepting any BCP 47 tag.
+ */
+export type BuiltInLocale =
+  | 'ar' | 'de' | 'es' | 'fr' | 'it' | 'ja'
+  | 'ko' | 'nl' | 'pl' | 'pt' | 'ru' | 'tr' | 'zh';
+
+/**
+ * Any BCP 47 locale tag. Known built-in locales autocomplete in editors;
+ * any other tag (e.g. 'pt-BR', 'zh-TW', 'en-AU') is accepted without a cast.
```
mister-ben: BCP 47 tags don't necessarily match the pattern of the examples given. `zh-Hant-TW`, `es-419`, `yor`, `sr-Latn`, `en-GB-scotland`. Will these be accepted too?

sampotts: Yep, that's the plan. I can add some more examples in here.

internal/design/i18n/decisions.md
```diff
@@ -0,0 +1,223 @@
...
+- **Add `translate?` param to `getLabel` and `getAttrs`** — Core calls `translate?.('Play') ?? 'Play'` internally.
+
+**Rationale.** Core is runtime-agnostic and framework-independent. Threading a context-derived translator into Core would create implicit coupling to whichever context system the caller uses (React context, DOM context, or a test double). The UI layer already holds the translator and is the natural place to apply it — it is the same layer that decides which attributes to set on the element. Core producing the key and the UI layer calling `t(key)` keeps the layers clean.
+
+The one exception is `formatDuration` in `@videojs/utils`, which assembles the time phrase internally and cannot expose individual words for the UI layer to translate. It receives an optional generic callback instead (see the `formatDuration` decision below).
+
+---
+
+## Parametric cores return the template key, not the resolved string
+
```
cjpillsbury: question(non-blocking): This feels counter-intuitive and makes a hard requirement for those not using our i18n infrastructure. In other words, we've just lost the value add for the "simple case" and replaced it with a dependency on the "more advanced case". I won't dig my heels in on this, but it feels like some of the complications here are because we're deciding to use the presentation template string as both "key" and (unresolved) value, some of the complications here are because we've moved away from having i18n considerations more deeply integrated into the components (a la react-spectrum inspiration crud we were exploring/aiming toward last year). This also creates a steeper curve as folks start trying to use VJS "more closely to the metal" (which I think has been a theme of some of our decisions in 2026). Food for thought.

internal/design/i18n/index.md
cjpillsbury: Yeah, per usual, I'm more concerned about "fundamentals" and architectural assumptions. I get not wanting ☝️ (although that has become the "accepted" norm in most of React land, which is what I'm assuming is top of mind with your example). I do think we want to keep i18n (and likely other categories of state) as separate providers from the core media state, it's easy enough to define a "Provider" as a thin wrapper around multiple _**Providers**_ if that's the preferred balance between devex vs. composition

internal/design/i18n/index.md
sampotts: We'll go with `lang`.

internal/design/i18n/index.md
sampotts: We'll remove this.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+## Problem
+
+All `aria-label`, `aria-valuetext`, tooltip labels, and time unit strings in Video.js 10 are hardcoded English. `PlayButtonCore.getLabel()` returns `'play'`. `formatDuration` returns `'2 minutes, 30 seconds'`. There is no mechanism to supply translated strings without overriding each component's `label` prop individually, and no single language switch for skins.
+
+Requirements:
+
+- Single entry point — one API replaces all English defaults regardless of layer (HTML, React, CDN)
+- Stable keys — renaming an English label must not require updating every locale file
+- Typed keys — TypeScript autocomplete for all keys
+- Decoupled — i18n providers are not embedded in skins; any skin works with any provider
```
cjpillsbury, line 19: nitpick: I suspect you actually meet this additional proposed/rephrased requirement, but also VJS components should be able to use i18n providers even when they are not used in skins. Would be good to capture that.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+
+`registerI18n` merges (does not replace) — multiple calls for the same language are additive. English defaults (`en.ts`) are pre-registered as the base layer on module import.
+
+### HTML
+
+Wrap any skin with `<media-i18n-provider>`:
+
+```html
+<media-i18n-provider lang="es">
+  <video-skin src="..."></video-skin>
```
cjpillsbury: nitpick: just an example but maybe remove `src` or update to more accurately reflect real code.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+```
+
+`registerI18n` merges (does not replace) — multiple calls for the same language are additive. English defaults (`en.ts`) are pre-registered as the base layer on module import.
+
+### HTML
+
+Wrap any skin with `<media-i18n-provider>`:
+
+```html
+<media-i18n-provider lang="es">
```
cjpillsbury: question(non-blocking): My expectation was that `lang` would be (at least optionally) asserted+scoped at the component level (including a skin component). This is almost definitely fine for the vast majority of our use cases though, so we can always revisit if/when needed.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+  return () => subscribers.delete(callback);
+}
+```
+
+`getTranslations` always returns at least the English defaults (from `en.ts`). No key is ever missing — the fallback chain is `registry[lang][key] → registry[base][key] → en[key] → key`.
+
+### `createTranslator`
+
+```ts
+export function createTranslator(
```
cjpillsbury, line 391: thought(non-blocking): It would be cool if we eventually supported this at either build time or run time for perf reasons. Def doesn't need to be in scope for this though.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+    return value;
+  }
+  t.locale = locale;
+  return t;
+}
+```
+
+### Locale files
+
+Locale files are TypeScript source files. They use `satisfies Partial<Translations>` so typos in key names are caught at build time — a misspelled key is a compile error, not a silent runtime miss:
```
cjpillsbury, line 412: praise: 😎

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+
+| System | Key type | Distribution | Registry? |
+| --- | --- | --- | --- |
+| VJS v7/8 | English string | Built-in JSON + consumer | No — set via `videojs.addLanguage` |
+| i18next | Opaque string | Consumer JSON | Yes — `i18next.addResourceBundle` |
+| FormatJS / react-intl | Opaque string | Consumer JSON | No — prop-based |
+| GNU gettext | English string | `.po` files | No |
+| Android `strings.xml` | Opaque XML ID | Resource files | No |
+| iOS `NSLocalizedString` | Opaque string | `.strings` files | No |
+| Media Chrome | None (no i18n) | — | — |
```
cjpillsbury: nitpick: this is inaccurate. We support i18n in Media Chrome.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+
+VJS v8's `addLanguage` is the closest precedent. This design replaces English-as-key (VJS v7/8) with opaque keys (consistent with i18next/Android/iOS) while keeping the imperative registration API.
+
+## Edge Cases
+
+**Multiple players, different languages.** Each `<media-i18n-provider>` or `<I18nProvider>` scopes its lang independently. The registry holds all registered langs simultaneously — providers read from it without interfering with each other.
+
+**`registerI18n` called after element mounts.** `<media-i18n-provider>` subscribes to `onRegistryChange`. Calling `registerI18n` after mount triggers a re-render of all descendant elements.
+
+**BCP 47 regional variants without a dedicated pack.** `getTranslations('pt-BR')` falls back to `getTranslations('pt')` then `'en'`. `<media-i18n-provider>` tries to lazy-load `pt-BR.ts`, then `pt.ts`.
```
cjpillsbury: suggestion(blocking): I don't think this should be a blocker, but I wouldn't tuck our `getTranslation()` rules and BCP-47 relationships away as an edge case callout. For example, if folks happen to pull in `es-MX` but no `es`, they should know that we won't (at least inferred from what I've read) "fall back" to `es-MX` for `es-AR`. Also, given that @mister-ben called out "fancier"/"more complex" BCP-47, knowing how defaulting would work for, e.g. `"zh-Hant-HK"` or other script subtags will be good to identify, and these subtags get even more complex (see, e.g. https://datatracker.ietf.org/doc/html/rfc5646#appendix-A). Even if our initial release is "not ideal", making our behaviors and support explicit would be good. See https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag for some high level discussion

cjpillsbury (changes requested):
Only one documentation-centric callout. Otherwise I feel like this addresses all of my major concerns. Looking good.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+
+### Registering translations
+
+The single entry point for both HTML and React consumers:
+
+```ts
+import { registerI18n } from '@videojs/core/i18n';
+import es from '@videojs/core/i18n/locales/es';
+
+registerI18n('es', es);
```
mihar-22: Should we consider a separate `locales` package? We don't currently recommend `core` publicly

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+
+The element reads from the registry for the given `lang`. No changes to `<video-skin>` or any other skin element.
+
+**CDN**
+
+Locale files are self-registering ES modules. They import `registerI18n` from the same CDN module URL, sharing the same registry instance — no global namespace pollution:
+
+```html
+<script type="module" src="https://cdn.videojs.com/videojs.js"></script>
+<script type="module" src="https://cdn.videojs.com/locales/es.js"></script>
```
mihar-22: CDN urls are incorrect:

```html
<script type="module" src="https://cdn.jsdelivr.net/npm/@videojs/html/cdn/video.js"></script>
<script type="module" src="https://cdn.jsdelivr.net/npm/@videojs/html/cdn/locales/es.js"></script>
```

internal/design/i18n.md
mihar-22: We don't publicly recommend core package atm. Not against this but wonder if we should just keep it as re-exports from `html` and `react` packages.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+import myEs from './my-es';
+registerI18n('es', myEs);
+```
+
+### React
+
+`I18nProvider` is a standalone component. Consumers add it explicitly — the skin itself has no i18n awareness:
+
+```tsx
+import { I18nProvider } from '@videojs/react/i18n';
```
mihar-22: This could just be exported from `@videojs/react`, right?

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+import { VideoSkin } from '@videojs/react/video';
+import es from '@videojs/core/i18n/locales/es';
+
+<I18nProvider lang="es" translations={es}>
```
mihar-22: I feel like `locale` makes more sense here in React than `lang` - aligning with Web APIs.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+</I18nProvider>
+```
+
+Omit `translations` to lazy-load the built-in pack automatically:
+
+```tsx
+<I18nProvider lang="es">
+  <VideoSkin src={src} />
+</I18nProvider>
+```
```
mihar-22: Neat DX 👍

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+
+async function changeLang(next: string) {
+  const { default: t } = await import(`@videojs/core/i18n/locales/${next}`);
+  registerI18n(next, t);
+  setLang(next);
+}
+
+<I18nProvider lang={lang}>
+  <VideoSkin src={src} />
+</I18nProvider>
```
mihar-22: This looks weird, why can't we just pass to `translations` prop? I think registering might be a HTML specific feature to play nicely with custom elements and CDN.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+```tsx
+import { getLocale } from 'next-intl/server';
+
+const locale = await getLocale();
+const { default: translations } = await import(`@videojs/core/i18n/locales/${locale}`);
+
+<I18nProvider lang={locale} translations={translations}>
+  <VideoSkin src={src} />
+</I18nProvider>
+```
```
mihar-22, line 205: Amazing, I was wondering how this would feel with other popular libraries in the ecosystem. Love to see more examples like this!

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+  seekForward?: string;
+  seekBackward?: string;
+  playbackRate?: string;
+  timePosition?: string;
+}
+
+export type Translator = {
+  (key: keyof Translations, params?: Record<string, string | number>): string;
+  readonly locale?: Locale;
+};
```
mihar-22, line 338: We can strongly type all of this actually!

<details>
  <summary>Click to see types</summary>

```ts
// Helper: string must contain `{tag}` somewhere
type Contains<Tag extends string> = `${string}{${Tag}}${string}`;

// Map each key to its required params (or never for no params)
interface TranslationParams {
  play: never;
  pause: never;
  replay: never;
  mute: never;
  unmute: never;
  seek: never;
  volume: never;
  muted: never;
  enterFullscreen: never;
  exitFullscreen: never;
  enableCaptions: never;
  disableCaptions: never;
  enterPictureInPicture: never;
  exitPictureInPicture: never;
  currentTime: never;
  duration: never;
  remaining: never;
  seekForward: { seconds: string | number };
  seekBackward: { seconds: string | number };
  playbackRate: { rate: string | number };
  timePosition: { current: string | number; duration: string | number };
}

export interface Translations {
  play?: string;
  pause?: string;
  replay?: string;
  mute?: string;
  unmute?: string;
  seek?: string;
  volume?: string;
  muted?: string;
  enterFullscreen?: string;
  exitFullscreen?: string;
  enableCaptions?: string;
  disableCaptions?: string;
  enterPictureInPicture?: string;
  exitPictureInPicture?: string;
  currentTime?: string;
  duration?: string;
  remaining?: string;
  seekForward?: Contains<"seconds">;
  seekBackward?: Contains<"seconds">;
  playbackRate?: Contains<"rate">;
  timePosition?: Contains<"current"> & Contains<"duration">;
}

// Keys that have no params
type SimpleTranslationKeys = {
  [K in keyof TranslationParams]: TranslationParams[K] extends never ? K : never;
}[keyof TranslationParams];

// Keys that have params
type ComplexTranslationKeys = Exclude<keyof TranslationParams, SimpleKeys>;

export type Translator = {
  (key: SimpleTranslationKeys): string;
  // Has params → second arg required with exact shape
  <K extends ComplexTranslationKeys>(key: K, params: TranslationParams[K]): string;

  readonly locale?: Locale;
};
```

</details>

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+}
+
+export function getTranslations(lang: string): Partial<Translations> {
+  const en = registry.get('en') ?? {};
+  const base = lang.split('-')[0];
+  const pack = registry.get(lang) ?? registry.get(base) ?? {};
+  return { ...en, ...pack };
+}
+
+export function onRegistryChange(callback: () => void): () => void {
```
mihar-22: Needs a more specific name like `onI18nRegistryChange` or something.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+import en from './locales/en';
+registry.set('en', en);
+
+export function registerI18n(lang: string, translations: Partial<Translations>): void {
+  const existing = registry.get(lang) ?? {};
+  registry.set(lang, { ...existing, ...translations });
+  subscribers.forEach(fn => fn());
+}
+
+export function getTranslations(lang: string): Partial<Translations> {
```
mihar-22: Maybe more specific name like `getI18nTranslations`

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+
+`getTranslations` always returns at least the English defaults (from `en.ts`). No key is ever missing — the fallback chain is `registry[lang][key] → registry[base][key] → en[key] → key`.
+
+### `createTranslator`
+
+```ts
+export function createTranslator(
+  translations: Partial<Translations> = {},
+  locale?: Locale
+): Translator {
```
mihar-22, line 394: I think `Translator` would be a generic and more strongly typed.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+import { registerI18n } from 'https://cdn.videojs.com/@videojs/core/i18n.js';
+registerI18n('es', { play: 'Reproducir', pause: 'Pausa', /* … */ });
+```
+
+### React: `createI18nHooks`
+
+A factory that closes over a React context so `useTranslator()` returns `Translator` with no explicit type annotation at call sites. Called once in `@videojs/react/i18n`; the result is exported directly:
+
+```tsx
+export function createI18nHooks() {
```
mihar-22: This would be generic I assume and more strongly typed based on `Translations` type provided.

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+  const I18nContext = createContext<Translator | null>(null);
+
+  function I18nProvider({ lang, translations, translationKeys, children }: {
+    lang?: Locale;
+    translations?: Partial<Translations>;
+    translationKeys?: ReadonlyArray<keyof Translations>;
```
mihar-22: What's this for?

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+    const translator = useMemo(
+      () => createTranslator(
+        { ...browserTranslated, ...getTranslations(lang ?? 'en'), ...builtIn, ...translations },
+        lang
+      ),
+      [browserTranslated, builtIn, translations, lang]
+    );
+
+    const player = usePlayer({ optional: true });
+    useEffect(() => { player?.selectCaptionsByLocale(lang); }, [player, lang]);
```
mihar-22: Out of scope I assume

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+
+Used in skin templates to replace hardcoded English strings:
+
+```html
+<media-play-button commandfor="play-tooltip">
+  ${renderIcon('restart')} ${renderIcon('play')} ${renderIcon('pause')}
+</media-play-button>
+<media-tooltip id="play-tooltip">
+  <media-text key="play"></media-text>
+</media-tooltip>
```
mihar-22: We auto-forward labels now so this wouldn't be needed for this example, but fine for showcasing `media-text`

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+  return { I18nContext, I18nProvider, useTranslator, useLocale };
+}
+```
+
+### HTML: `<media-i18n-provider>`
+
+Reads from the registry, provides via `i18nContext`. Subscribes to registry changes so live elements re-render when `registerI18n` is called after mount:
+
+```ts
+export class MediaI18nProviderElement extends ReactiveElement {
```
mihar-22: I think similar to the i18n factory in React, we would likely have one here for html as well for provider, context, controller, text element, etc. Similar to `createPlayer`.

internal/design/i18n.md
mihar-22: Probably just call it `createI18n`?

internal/design/i18n.md
```diff
@@ -0,0 +1,799 @@
...
+
+User overrides (`localeSelectionOverridden` flag) suppress auto-selection until the media source changes.
+
+### CDN build strategy
+
+The CDN bundle is an ES module. Locale files are separate self-registering ES modules — they import `registerI18n` from the same CDN URL, sharing the singleton registry instance with no global namespace touch:
+
+```html
+<script type="module" src="https://cdn.videojs.com/videojs.js"></script>
+<script type="module" src="https://cdn.videojs.com/locales/es.js"></script>
```
mihar-22: Incorrect URLs - see above

mihar-22 (dismissed):
Awesome 😎 I've left a bunch of comments but overall looking really good. In large I was just pressing on where this should live and stronger types.

I might've missed it but I didn't catch how this will all tie into updating primitive component language (say label on a play button). I can imagine few obvious ways off the top of my head so not concerned.

internal/design/i18n.md
sampotts: Will amend!

internal/design/i18n.md
sampotts: Removed!

internal/design/i18n.md
sampotts: Makes sense! I'll amend.

internal/design/i18n.md
sampotts: Yep!

internal/design/i18n.md
sampotts: Will amend.

internal/design/i18n.md
sampotts: Totally. Will amend.

internal/design/i18n.md
sampotts: Updated!

internal/design/i18n.md
sampotts: Updated!

internal/design/i18n.md
sampotts: Updated!

internal/design/i18n.md
sampotts: Updated!

internal/design/i18n.md
sampotts: Cool, will update anyway.

internal/design/i18n.md
sampotts: Updated!

internal/design/i18n.md
sampotts: Updated!

sampotts:
I've created tickets based on the current state of the document but I can incorporate any further feedback as required.

internal/design/i18n/architecture.md
sampotts: Updated!

internal/design/i18n.md
```diff
@@ -0,0 +1,1041 @@
...
+
+The single entry point for both HTML and React consumers:
+
+```ts
+// HTML
+import { registerI18n } from '@videojs/html/i18n';
+import es from '@videojs/html/i18n/locales/es';
+
+// React
+import { registerI18n } from '@videojs/react';
```
cjpillsbury, line 38: nitpick: import path inconsistencies between react vs. html (assuming a typo)

internal/design/i18n.md
```diff
@@ -0,0 +1,1041 @@
...
+</Provider>
+```
+
+Pass `locale` explicitly to override inheritance, or `translations` to bypass the registry entirely:
+
+```tsx
+import es from '@videojs/react/i18n/locales/es';
+
+<Provider>
+  <I18nProvider locale="es" translations={es}>
```
cjpillsbury, line 140: question(non-blocking): Is there a reason for the `lang` vs. `locale` discrepancy?

internal/design/i18n.md
```diff
@@ -0,0 +1,1041 @@
...
+t('seekForward');                       // ✗ TS error — missing { seconds }
+t('seekForward', { seconds: 10 });      // ✓
+t('seekForward', { second: 10 });       // ✗ TS error — typo, wrong key
+```
+
+### Registry
+
+```ts
+// @videojs/core/i18n/registry.ts
+const registry = new Map<string, Partial<Translations>>();
```
cjpillsbury, line 355: suggestion(non-blocking): If we're going to have stateful/mutative modules (roughly equivalent to a "capital S Singleton"), let's make sure minimally we can easily remove registered languages for testing?

internal/design/i18n.md
```diff
@@ -0,0 +1,1041 @@
...
+    );
+    const locale = explicitLocale ?? ambientLocale;
+
+    // Lazy-load built-in pack when locale is set
+    const [builtIn, setBuiltIn] = useState<Partial<Translations>>({});
+    useEffect(() => {
+      if (!locale) { setBuiltIn({}); return; }
+      const tags = [...new Set([locale, locale.split('-')[0]])];
+      const load = (i = 0): Promise<void> =>
+        import(`@videojs/core/i18n/locales/${tags[i]}`)
```
cjpillsbury, line 506: thought(non-blocking): We may need to do some validation of this strategy in real world app build setups, but I don't see any problem starting with this hope/dream. Let's just make sure we do enough smoke testing of this that we don't think it will break folks doing normal crud with normal frameworks/build tools in the wild.

internal/design/i18n.md
```diff
@@ -0,0 +1,1041 @@
...
+      initialValue: createTranslator(getI18nTranslations('en')),
+    });
+
+    #unsubscribeRegistry: (() => void) | null = null;
+    #langObserver: MutationObserver | null = null;
+
+    /** Resolve effective locale: explicit lang attribute → nearest ancestor[lang] → documentElement.lang → undefined. */
+    get #effectiveLocale(): Locale | undefined {
+      if (this.lang) return this.lang;
+      const ancestor = this.parentElement?.closest('[lang]');
```
cjpillsbury, line 623: thought(non-blocking): Can't remember if we need to account for shadow dom boundaries with this, but we have prior art for that kinda stuff, so we can always update accordingly if/when needed.

cjpillsbury (approved):
This all looks great. Some nb callouts, but I'm comfortable with moving forward here. One high level callout, re: implementation - we may be able to pull out some of this functionality on the react and html side into shareable code (e.g. the MutationObserver stuff). I don't think that needs to be updated in the RFC, but food for thought on maintainability and/or bundle size considerations (though, for the later, it's unlikely there will be a mix and match of html + react usages in real world examples in the wild).

ronald-urbina:
I left this [issue](https://github.com/muxinc/elements/issues/497) of elements that might be related to this discussion.

internal/design/i18n.md
sampotts: As per our discussions, the `lang` attribute makes more sense in HTML but in the React world, locale seems to be more used. I'm happy to make it `lang` to be consistent though if you'd prefer? I don't feel super strongly, Claude just suggested it was more idiomatic for React devs.

internal/design/i18n.md
sampotts: Will update during the build.

PR #1153, opened 2026-03-28
feat(site): rotate poster time

heff:
So many great shots in the demo video. Let's randomize them as the poster.

---

> [!NOTE]
> **Low Risk**
> Low risk UI-only change that affects which poster frame is preloaded/rendered; main concern is minor cache/preload variability between page loads.
> 
> **Overview**
> The home page hero video poster is now randomized by selecting from a small set of predefined timestamps instead of always using `time=0`, and the poster preload link follows this chosen frame.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9330e2cadf0d73d41d05ef807d38a5905ae2843f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/site/src/components/home/Header.astro
+++ b/site/src/components/home/Header.astro
@@ -5,7 +5,9 @@ import HeroVideo from '@/components/home/HeroVideo';
 import SkinControl from '@/components/home/SkinControl';
 import { MUX_URL, VJS10_DEMO_VIDEO } from '@/consts';
 
-const poster = `${VJS10_DEMO_VIDEO.poster}?time=0`;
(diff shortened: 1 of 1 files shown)


heff:
It is cached? Every time I reload the page in the dev site version I get a new thumbnail. If so how do we change it to be on reload on prod?

heff:
I think it would be a lot cooler if it changed every reload, but I tried with React and it would flash the SSR'd poster before switching to the random one. I didn't see a way around that and I don't think it's worth making it more complicated or have a slower perceived poster load. If anyone feels like it's worth pushing on, go for it!

PR #1155, opened 2026-03-30
Fix HTML SSR safety and sandbox skin chunking

sampotts:
## Summary
- Make HTML define modules SSR-safe by guarding custom element registration and stylesheet creation
- Update skin definitions to use safeDefine and add SSR coverage for video skin and simple-hls-video imports
- Remove the redundant poster define import from the html-video sandbox template so skin chunks do not pull in the plain html-video entry in production

## Testing
- pnpm -F @videojs/html test
- pnpm -F @videojs/html build

---

> [!NOTE]
> **Medium Risk**
> Changes custom element registration and shadow-root style application to guard missing browser globals, which could affect element initialization and styling in some environments. Adds tests to cover SSR-like globals and adjusts sandbox imports, but still touches core define/skin infrastructure.
> 
> **Overview**
> **SSR safety hardening for HTML define modules.** Replaces direct `customElements.define(...)` calls across audio/video skins, `captions-button`, and `simple-hls-video` with `safeDefine(...)`, and updates `safeDefine` to no-op when `customElements` is unavailable (e.g. SSR).
> 
> **Skin style application is now resilient without constructable stylesheets.** `SkinMixin` now guards `document` access, allows `createStyles()` to return raw CSS when `CSSStyleSheet` is missing, and falls back to injecting `<style>` tags into the shadow root when `adoptedStyleSheets`/constructable sheets aren’t supported.
> 
> **Test + sandbox adjustments.** Expands `safe-define` tests, adds an SSR-import smoke test for `video/skin` and `simple-hls-video`, and removes a redundant `@videojs/html/ui/poster` import from the `html-video` sandbox template to avoid pulling the plain entry into skin chunks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d93cc7e4b7a600c90b81e337d292c88edcb0a743. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/html/src/define/skin-mixin.ts
+++ b/packages/html/src/define/skin-mixin.ts
@@ -4,17 +4,45 @@ import rootStyles from './base.css?inline';
 import sharedStyles from './shared.css?inline';
 
 const STYLES_ID = '__media-styles';
+type SkinStyles = CSSStyleSheet | string;
 
 function ensureRootStyles(): void {
-  if (document.getElementById(STYLES_ID)) return;
-  const style = document.createElement('style');
+  const doc = globalThis.document;
+  if (!doc || doc.getElementById(STYLES_ID)) return;
+
+  const style = doc.createElement('style');
   style.id = STYLES_ID;
   style.textContent = rootStyles;
-  document.head.appendChild(style);
+  doc.head.appendChild(style);
 }
 
-const sharedSheet = new CSSStyleSheet();
-sharedSheet.replaceSync(sharedStyles);
+function isConstructableStyleSheet(value: SkinStyles): value is CSSStyleSheet {
+  return typeof globalThis.CSSStyleSheet !== 'undefined' && value instanceof globalThis.CSSStyleSheet;
+}
+
+function getStyleText(style: SkinStyles): string {
+  if (typeof style === 'string') return style;
+
+  return Array.from(style.cssRules)
+    .map((rule) => rule.cssText)
+    .join('\n');
+}
+
+function applyShadowStyles(shadowRoot: ShadowRoot, styles: SkinStyles[]): void {
+  if (styles.every(isConstructableStyleSheet) && 'adoptedStyleSheets' in shadowRoot) {
+    shadowRoot.adoptedStyleSheets = styles;
+    return;
+  }
+
+  const doc = shadowRoot.ownerDocument;
+  for (const styleText of styles.map(getStyleText)) {
+    const style = doc.createElement('style');
+    style.textContent = styleText;
+    shadowRoot.appendChild(style);
+  }
+}
+
+const sharedSheet = createStyles(sharedStyles);
 
 /**
  * Mixin for skin elements that renders the template from a static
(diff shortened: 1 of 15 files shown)


packages/html/src/define/skin-mixin.ts
```diff
@@ -4,17 +4,45 @@ import rootStyles from './base.css?inline';
...
+    return;
+  }
+
+  const doc = shadowRoot.ownerDocument;
+  for (const styleText of styles.map(getStyleText)) {
+    const style = doc.createElement('style');
+    style.textContent = styleText;
+    shadowRoot.appendChild(style);
+  }
 }
```
mihar-22, line 43: The code here is overly specific to a "Skins" but the logic is largely generic. Let's generalize and abstract them out in `utils` package please.

packages/html/src/define/skin-mixin.ts
```diff
@@ -26,10 +54,10 @@ sharedSheet.replaceSync(sharedStyles);
  */
 export function SkinMixin<Base extends Constructor<ReactiveElement>>(
```
mihar-22, line 55: (recommendation: non-blocking)

I’d probably avoid a mixin here. Generally, mixins are great for composing into unrelated element types. Things of the same kind are generally a base class. I can see in the futre separating shadow styles and rendering into mixins but only if you needed to compose with other style or render flows (which is very unlikely). Mixins are also harder to read and type, no need for that complexity if not used. It can also trip up first-time readers because a mixin assumes deeper composition story.

So overall, simplifying... this feels more like core element behavior than an optional capability. I’d recommend modeling this as a base element instead and treating shadow root setup as part of the element’s lifecycle. Good reference is [`LitElement`](https://lit.dev/docs/v1/api/lit-element/LitElement/#LitElement) and [`createRenderRoot`](https://lit.dev/docs/v1/api/lit-element/LitElement/#LitElement.createRenderRoot).

---

Separate note: if you're building on top of our `ReactiveElement`, it's missing a `render` lifecycle. It could be a nice addition at some point if you need more dynamic control over the skins (same as how `LitElement` adds `render` over their `ReactiveElement`). Only part that sucks is it would need some kind of DOM patching lib (e.g., nanodom adaptation).

packages/html/src/define/skin-mixin.ts
```diff
@@ -40,25 +68,29 @@ export function SkinMixin<Base extends Constructor<ReactiveElement>>(
...
+/** Create a constructable stylesheet when available, otherwise return raw CSS. */
+export function createStyles(css: string): SkinStyles {
+  if (typeof globalThis.CSSStyleSheet === 'undefined') {
+    return css;
+  }
+
+  const sheet = new globalThis.CSSStyleSheet();
   sheet.replaceSync(css);
   return sheet;
 }
```
mihar-22, line 96: Looks like generic utility that should go in `utils`. With all just check that a similar implementation doesn't exist which can be extended.

Issue #1185, opened 2026-04-01
Docs: Make Installation Page LLM-Friendly


decepulis:
The installation page (`site/src/content/docs/how-to/installation.mdx`) relies heavily on interactive UI components — framework picker, use case picker, skin picker, renderer picker, and tabbed code blocks — that don't translate to plain text. The LLM markdown integration (`llms-markdown.ts`) converts HTML to `.md` after build, but the interactive elements degrade poorly:

- **Pickers** render as bare labels ("React", "HTML", "Video", "Audio") with no surrounding context.
- **Tabs** become flat bullet lists followed by all code blocks, with no indication of which tab maps to which block.
- **Upload/URL widgets** produce non-functional text like "Drop a video — or — Select a file".

The result is that LLMs and other plain-text consumers (e.g., `llms.txt`) see only a confusing single path through what is actually a multi-path interactive guide. The per-framework split via `FrameworkCase` works fine since the build produces separate HTML/React `.md` files, but the inner interactive elements lose their meaning.

Thanks to @esbie for reporting this.

Issue #247, opened 2025-12-17
CLI Package


mihar-22:
Command-line interface for scaffolding Video.js 10 projects and adding components.

### `init`

Scaffolds a new project. Detects framework from `package.json` or prompts.

```bash
npx @videojs/cli init
```

- Detects React / HTML (Prompt if unknown)
- Detects styling preference (CSS / Tailwind) (Prompt if unknown)
- Prompts for skin and whether to have it "ejected"
- Adds markup + styles to your project
- Installs dependencies

PR #1205, opened 2026-04-02
Design: CLI for LLM-friendly Video.js installation

decepulis:
## Summary

Design doc for `@videojs/cli create` — a CLI that generates Video.js installation code, solving the problem of [docs/installation.md](https://videojs.org/docs/framework/react/how-to/installation.md) being bad for LLM consumers (videojs/v10#1185). MVP scope; the same package will later support skin ejection.

An incremental step towards #247. Mostly concerned with closing #1215 and #1185 though.

Implementation: #1215

### DIFF
--- a/internal/design/site/cli-llm-friendly-installation.md
+++ b/internal/design/site/cli-llm-friendly-installation.md
@@ -28,7 +28,7 @@ Two changes:
 ### Interactive (humans, chatbot-assisted)
 
 ```bash
-npx @videojs/cli
+npx @videojs/cli create
 
 # ? Framework: (html / react)
 # ? Use case: (video / audio / background-video)
@@ -40,7 +40,7 @@ npx @videojs/cli
 ### Flags (agentic LLMs)
 
 ```bash
-npx @videojs/cli --framework react --use-case video --renderer hls
+npx @videojs/cli create --framework react --use-case video --renderer hls
 ```
 
 ### LLMCase in MDX
@@ -57,13 +57,13 @@ npx @videojs/cli --framework react --use-case video --renderer hls
 To generate installation code tailored to your project, run:
 
 ```bash
-npx @videojs/cli
+npx @videojs/cli create
 ```
 
 If you already know your choices, pass them as flags:
 
 ```bash
-npx @videojs/cli --framework react --use-case video --skin default --renderer hls
+npx @videojs/cli create --framework react --use-case video --skin default --renderer hls
 ```
 </LLMCase>
 ```
@@ -73,7 +73,7 @@ npx @videojs/cli --framework react --use-case video --skin default --renderer hl
 ### CLI
 
 ```
-npx @videojs/cli [flags]
+npx @videojs/cli create [flags]
 
 Flags:
   --framework <html|react>
--- a/internal/design/site/cli-llm-friendly-installation.md
+++ b/internal/design/site/cli-llm-friendly-installation.md
@@ -3,31 +3,31 @@ status: draft
 date: 2026-04-02
 ---
 
-# CLI for LLM-friendly installation
+# CLI for LLM-friendly installation docs
 
-Generate installation code from the command line. It's docs/installation.md, but without the interactive UI that breaks in plain text.
+Generate installation code from the command line. It's docs/installation.md, but without the interactive React UI that breaks in plain text.
 
 This means... it's finally time for `@videojs/cli`. The same package will later support skin ejection and other workflows.
 
 ## Problem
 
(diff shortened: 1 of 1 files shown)


internal/design/site/cli-llm-friendly-installation.md
```diff
@@ -0,0 +1,64 @@
...
+**`HumanCase` / `LLMCase` MDX components** — Astro components that show different content to browsers and the LLM markdown pipeline. installation.mdx wraps interactive pickers in `HumanCase` and CLI instructions in `LLMCase`. Same file, both audiences. Three consumer types are covered: humans still have their react-powered interactive web page, agentic LLMs run the CLI directly, chat LLMs recommend the CLI to the user.
+
+## API
+
+```
+npx @videojs/cli create [flags]
+
+Flags:
+  --framework <html|react>                                    (required)
+  --use-case <video|audio|background-video>                   (default: video)
```
mihar-22: suggestion: `--preset`

internal/design/site/cli-llm-friendly-installation.md
```diff
@@ -0,0 +1,64 @@
...
+  --skin <default|minimal>                                    (default: default)
+  --renderer <html5-video|html5-audio|hls|background-video>   (default: per use case)
```
mihar-22: suggestion: `--media`

internal/design/site/cli-llm-friendly-installation.md
```diff
@@ -0,0 +1,64 @@
...
+
+**`HumanCase` / `LLMCase` MDX components** — Astro components that show different content to browsers and the LLM markdown pipeline. installation.mdx wraps interactive pickers in `HumanCase` and CLI instructions in `LLMCase`. Same file, both audiences. Three consumer types are covered: humans still have their react-powered interactive web page, agentic LLMs run the CLI directly, chat LLMs recommend the CLI to the user.
+
+## API
+
+```
+npx @videojs/cli create [flags]
+
+Flags:
+  --framework <html|react>                                    (required)
```
mihar-22: 1. Should framework refer to `html` or `react`, or specifically application framework like next, vite, start, react-router, astro, nuxt, etc. Thinking the latter because it'll determine how we scaffold. IF we're letting LLM ultimately decide what to do with the player code then we can ignore this for now, but for users not using an LLM it'll come back up in the future. Want to be careful what `--framework` means. Also, if "frameworks" is broader then we might still need a option to determine html/react if you're targeting something agnostic like astro. 

3. ~Think we're missing `--css` option for vanilla or tailwind~ I saw our installation page doesn't have ejected option or expose any styling so I guess css option doesn't make sense yet.

References:

- https://ui.shadcn.com/docs/installation
- https://github.com/withastro/astro/blob/main/packages/create-astro/README.md

internal/design/site/cli-llm-friendly-installation.md
```diff
@@ -0,0 +1,64 @@
...
+
+- **Separate LLM guide** — Write a purpose-built markdown page for LLMs. Two documents to maintain, guaranteed drift.
+
+- **Expand variants in the markdown pipeline** — Teach `llms-markdown.ts` to understand the picker components and render every combination under structured headers. The pipeline would need to understand component semantics it currently ignores, and the output would be long.
+
+The CLI avoids the combinatorial problem entirely — it lets the consumer narrow their own path.
+
+## Open questions for now
+
+- **Prompt library** — `@inquirer/prompts` vs `@clack/prompts` vs something else?
```
mihar-22: I like clack! there's tons of options so whatever your heart desires

mihar-22 (commented):
This looks awesome, few small suggestions! Ideally the LLM can just look at the project and recommend the best options to run. 

Providing LLM a sample src/poster depending on media would be great.

internal/design/site/cli-llm-friendly-installation.md
decepulis: Great suggestion.

decepulis: I might even go one step further and suggest that we should ditch "use case" on the installation page and teach folks the word "preset" even earlier.

internal/design/site/cli-llm-friendly-installation.md
decepulis: not opposed

internal/design/site/cli-llm-friendly-installation.md
decepulis: Interesting idea. One that could be extended to our installation page, too. Kind of like what you do in vidstack. Where folks select "sveltekit" but they're still given the html custom element options. 

I remember Heff being opposed to this but I don't remember why (it was in the pre-beta rush) so I'm going to treat this as out-of-scope and make a note to revisit it.

internal/design/site/cli-llm-friendly-installation.md
decepulis: I don't know the first thing about these options. I'll just use clack since you like it, and we can throw it out if we decide on something else later!

Issue #505, opened 2026-02-10
[26.2] feat: hotkeys


mihar-22:
Configurable keyboard shortcut system — map key bindings to player actions. Declarative and customizable so skins and users can define their own mappings. Default bindings overridable by consumers, with an extension point for custom actions.

## Prior art

- **Vidstack:** [keyboard API](https://vidstack.io/docs/player/api/keyboard/)
- **Media Chrome:** [keyboard shortcuts](https://www.media-chrome.org/docs/en/keyboard-shortcuts)
- **Plyr:** [keyboard shortcuts](https://github.com/sampotts/plyr)

PR #1290, opened 2026-04-08
fix(packages): remove redundant Shift modifier from playback rate hotkeys

### LINKED ISSUE #1288: Bug: Playback Rate Hotkeys Use Redundant Shift Modifier, Breaking European Layouts
## Bug

The media hotkeys for playback rate (`Shift+>` and `Shift+<` for `speedUp` / `speedDown`) include a redundant `Shift` modifier.

On English keyboards, `>` and `<` are the shifted characters of `.` and `,` — you can't produce them without Shift already held. The `Shift+` prefix adds nothing.

On many European keyboard layouts (French AZERTY, German QWERTZ, Italian, etc.), `<` and `>` live on the same physical key as unshifted/shifted variants. Requiring `shiftKey: true` alongside `key: '<'` makes the binding **impossible to trigger** on these layouts since `{key: '<', shiftKey: true}` can never be produced.

## Fix

Two-part fix:

1. **Bindings**: Define the bindings as simply `>` and `<` without the `Shift` modifier in all skins, design docs, and examples.

2. **Matcher**: Update `matchesHotkeyEvent` to handle **implicit Shift** for non-letter character keys. Characters like `>`, `<`, `?`, `!` require Shift on some layouts but not others — Shift was used to *produce* the character, not as a deliberate modifier. The matcher now ignores `shiftKey` for single non-letter character keys while keeping strict matching for letters (a-z) and named keys (ArrowLeft, Tab, etc.).

## Reported By

Ben on Discord

mihar-22:
Closes #1288
Reference #505 

---

> [!NOTE]
> **Medium Risk**
> Changes core hotkey modifier-matching semantics by ignoring `shiftKey` for single-character non-letter keys unless `Shift` is explicitly part of the binding, which could affect existing shortcut behavior across layouts. Risk is mitigated by expanded unit tests and mostly impacts keyboard handling.
> 
> **Overview**
> Updates playback-rate shortcuts to bind `>`/`<` directly (removing the redundant `Shift+` prefix) across the HTML/React default skins and the hotkey design docs.
> 
> Adjusts `matchesHotkeyEvent` to treat Shift as *implicit* for single non-letter character keys (e.g. `>`, `<`, `?`), so bindings like `>` match regardless of whether Shift was required by the user’s keyboard layout, while explicit bindings like `Shift+>` still require Shift; tests and ARIA formatting expectations are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1926e27126b11be69c5df507377e4651f88a7525. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/dom/hotkey/hotkey.ts
+++ b/packages/core/src/dom/hotkey/hotkey.ts
@@ -31,8 +31,8 @@ const MODIFIER_KEYS = new Set(['shift', 'ctrl', 'alt', 'meta']);
  *
  * @example
  * ```ts
- * parseHotkeyPattern('Shift+>');
- * // [{ modifiers: Set('shift'), key: '>', originalKey: '>' }]
+ * parseHotkeyPattern('>');
+ * // [{ modifiers: Set(), key: '>', originalKey: '>' }]
  *
  * parseHotkeyPattern('0-9');
  * // 10 bindings, one per digit
--- a/packages/core/src/dom/hotkey/hotkey.ts
+++ b/packages/core/src/dom/hotkey/hotkey.ts
@@ -70,6 +70,16 @@ export function parseHotkeyPattern(pattern: string): ParsedHotkeyBinding[] {
   return [{ modifiers, key, originalKey: rawKey }];
 }
 
+/**
+ * Single non-letter character — Shift was used to produce the character itself,
+ * not as a deliberate modifier (e.g. Shift+. → ">", Shift+/ → "?").
+ * Letters excluded because Shift changes case intentionally (k vs K).
+ * Named keys excluded because event.key.length > 1 (ArrowLeft, Tab, etc.).
+ */
+function isImplicitShift(key: string): boolean {
+  return key.length === 1 && !/[a-z]/i.test(key);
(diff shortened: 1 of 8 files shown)


packages/core/src/dom/hotkey/hotkey.ts
```diff
@@ -70,6 +70,16 @@ export function parseHotkeyPattern(pattern: string): ParsedHotkeyBinding[] {
   return [{ modifiers, key, originalKey: rawKey }];
 }
 
+/**
+ * Single non-letter character — Shift was used to produce the character itself,
+ * not as a deliberate modifier (e.g. Shift+. → ">", Shift+/ → "?").
+ * Letters excluded because Shift changes case intentionally (k vs K).
+ * Named keys excluded because event.key.length > 1 (ArrowLeft, Tab, etc.).
+ */
+function isImplicitShift(key: string): boolean {
```
cjpillsbury, line 79: praise: 😎

PR #1292, opened 2026-04-08
refactor(packages)!: replace DelegateMixin & ProxyMixin with MediaHost base classes

luwes:
Implements #1297

## TODO

- [x] Handle `<track>` and `<source>` children in `CustomMediaElement`
- [x] Set attributes when certain props are set on the CE, like `muted`, `autoplay`, `preload`, `src`, etc.

## Summary

Replace `DelegateMixin` and proxy-based `CustomMediaElement` with composable `MediaHost` classes that handle event forwarding and property delegation to native media elements. Remove `playbackId`/`customDomain` from the Mux media API in favor of direct `src` usage.

## Changes

- Delete `DelegateMixin` and `ProxyMixin` — introduce `HTMLMediaElementHost`, `HTMLVideoElementHost`, and `HTMLAudioElementHost` as the new base classes for media engines
- Rewrite `CustomMediaElement` to discover host properties at define-time and wire attribute observation automatically, replacing the old proxy-based approach
- Each media engine (HLS, Dash, Mux, NativeHLS, SimpleHLS, SPF) now extends the host classes and chains lifecycle via `super.attach()`/`super.detach()`
- Use `Mixin<Host, Props>` type util with typed host interfaces for type-safe `super` calls
- Extract Mux Data SDK integration into `MuxDataMediaMixin`
- Remove `playbackId` and `customDomain` props from Mux media elements — consumers pass full `src` URLs
- Remove `DelegateMixin` re-export from `@videojs/html` and `@videojs/react`
- Add `CustomMediaElement` tests covering video attribute forwarding, MediaHost property delegation, event forwarding, and lifecycle management

<details>
<summary>Architecture</summary>

Before: `DelegateMixin(ProxyMixin(HTMLVideoElement), HlsMediaDelegate)` — creates an internal delegate instance, bridges events, intercepts property access via `get`/`set`/`call`.

After: `CustomMediaElement('video', HlsMedia)` — `HlsMedia` extends `HTMLVideoElementHost` which manages the native element lifecycle. `CustomMediaElement` walks the host prototype at define-time to create property descriptors and attribute observers on the custom element, forwarding everything to the host instance.

Each host class extends `EventTarget` and lazily subscribes to native events, forwarding them to consumers. Setters are discovered via `attrToProp` mapping so attributes like `current-time` route through the host property.

</details>

## Testing

`pnpm -F @videojs/core test` — all existing tests updated and passing, plus new `CustomMediaElement` test suite (38 tests) covering shadow DOM setup, attribute forwarding, property delegation, event forwarding, and lifecycle.

**BREAKING CHANGE:** `DelegateMixin` removed. `playbackId` and `customDomain` removed from Mux media elements — use `src` directly.

---

> [!NOTE]
> **High Risk**
> Large refactor of the media abstraction layer (event forwarding, property/attribute delegation, engine lifecycle) and removal of Mux `playbackId` support, which may introduce subtle regressions across HLS/DASH/Mux/native engines despite added tests.
> 
> **Overview**
> **Replaces the `DelegateMixin`/`ProxyMixin` media architecture with `HTMLMediaElementHost`-based classes** that own attach/detach/destroy, proxy common media APIs, and forward native events.
> 
> Rewrites `CustomMediaElement` to auto-generate property accessors and `observedAttributes` from a host class (including forwarding standard media attributes and syncing slotted `<track>`/`<source>` children), and updates all DOM/HTML/React media components (HLS, DASH, native-HLS, simple-HLS, Mux) to use the new host classes.
> 
> **Breaking change:** removes Mux `playbackId`/`customDomain` and updates sandbox/templates and Mux media APIs to accept direct `src` URLs; also moves Mux Data integration into a new `MuxDataMediaMixin`. Adds extensive new unit tests for `CustomMediaElement`, plus new engine tests (HLS/DASH) and capability guards in store features via new media predicate helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a61d196313d34174ac3c1a1d0e1af444fced2443. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/spf/src/dom/playback-engine/adapter.ts
```diff
@@ -1,154 +1,178 @@
...
+    #engine: PlaybackEngine;
+    #config: PlaybackEngineConfig;
+    #preload: '' | 'none' | 'metadata' | 'auto' = '';
+    #loadstartListener: (() => void) | null = null;
+
+    constructor(...args: any[]) {
+      super(...args);
+      this.#config = {};
+      this.#engine = createPlaybackEngine({});
+    }
```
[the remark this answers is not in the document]
luwes: fine for now

packages/core/src/core/media/tests/delegate.test.ts
mihar-22, line 1: Test file name needs to be updated I assume since delegates were removed

packages/core/src/core/media/types.ts
```diff
@@ -11,6 +11,6 @@ type SettableKeys<T> = {
 
 type ExcludeInternal<K> = K extends `_${string}` ? never : K;
 
-export type InferDelegateProps<D extends abstract new (...args: any[]) => any> = Partial<
+export type InferClassProps<D extends abstract new (...args: any[]) => any> = Partial<
   Pick<InstanceType<D>, ExcludeInternal<SettableKeys<InstanceType<D>>>>
 >;
```
mihar-22: Might be a candidate for utility type and not in media/types, not seeing anything media specific.

packages/core/src/dom/media/dash/index.ts
```diff
@@ -1,47 +1,67 @@
...
-export class DashMediaDelegate implements Delegate {
-  #engine: dashjs.MediaPlayerClass;
-  #src: string = '';
+export interface DashMediaProps {
+  readonly engine: dashjs.MediaPlayerClass;
+  src: string;
+  attach(target: EventTarget): void;
+  detach(): void;
+  destroy(): void;
+}
```
mihar-22: It looks like there's a `MediaEngineHost` contract in here which can be reused:

```ts
interface MediaEngineHost<Target extends EventTarget, Engine> {
  readonly engine: Engine;
  src: string;
  attach(target: Target): void;
  detach(): void;
  destroy(): void;
}
```

packages/core/src/dom/media/dash/index.ts
```diff
@@ -1,47 +1,67 @@
...
 import { CustomVideoElement } from '../custom-media-element';
 import { VideoProxy } from '../proxy';
 
-export class DashMediaDelegate implements Delegate {
-  #engine: dashjs.MediaPlayerClass;
-  #src: string = '';
+export interface DashMediaProps {
+  readonly engine: dashjs.MediaPlayerClass;
+  src: string;
+  attach(target: EventTarget): void;
```
mihar-22: This type of `target` is misleading and dangerous at runtime because it's being type cast and not validated. This seems to me it should be `HTMLVideoElement`. Was there any reason?

As an aside, feels a little strange these are "Props" and mixed in. I'd probably expect these to just be implemented by something. Maybe I missed how this is being used more widely.

packages/core/src/dom/media/dash/index.ts
```diff
@@ -1,47 +1,67 @@
...
+}
 
-  constructor() {
-    this.#engine = dashjs.MediaPlayer().create();
-    this.#engine.initialize(undefined, undefined, false);
-  }
+interface DashMediaHost extends EventTarget {
+  attach?(target: EventTarget): void;
+  detach?(): void;
+}
```
mihar-22: Similar to comments above, risky/misleading target type. Seems to also be a subset of the expected engine host contract, some inconsistency.

packages/core/src/dom/media/dash/index.ts
```diff
@@ -1,47 +1,67 @@
...
+  return DashMediaImpl as any;
+};
+
+export class DashMediaBase extends DashMediaMixin(EventTarget) {}
+
+export class DashCustomMedia extends DashMediaMixin(CustomVideoElement) {}
 
-// This is used by the React component.
-export class DashMedia extends DelegateMixin(VideoProxy, DashMediaDelegate) {}
+export class DashMedia extends DashMediaMixin(VideoProxy) {}
```
mihar-22: I was having trouble understanding the difference and when to use each of these, or when we rely them on internally. I can see from the comment `DashMedia` is for React, jumped between `CustomVideoElement` and `VideoProxy` and I wasn't sure. The difference or reason for both existing is a little tricky to figure out. If they're all valid then I think some comment blocks or renaming could help a lot.

packages/core/src/dom/media/hls/hlsjs.ts
```diff
@@ -55,6 +54,6 @@ class HlsJsMediaDelegateBase extends EventTarget implements Delegate {
   }
 }
 
-export class HlsJsMediaDelegate extends HlsJsMediaPreloadMixin(
-  HlsJsMediaMetadataTracksMixin(HlsJsMediaTextTracksMixin(HlsJsMediaErrorsMixin(HlsJsMediaDelegateBase)))
+export class HlsJsMediaBase extends HlsJsMediaPreloadMixin(
+  HlsJsMediaMetadataTracksMixin(HlsJsMediaTextTracksMixin(HlsJsMediaErrorsMixin(HlsJsMediaBaseImpl)))
```
mihar-22: No issues with the mixin approach but curious why it was preferred over having this logic just living in the base implementation?

packages/core/src/dom/media/hls/index.ts
```diff
@@ -24,165 +24,188 @@ export const SourceTypes = {
...
+  type: SourceType | undefined;
+  preferPlayback: PlaybackType | undefined;
+  config: Record<string, any>;
+  debug: boolean;
+  preload: PreloadType;
+  attach(target: HTMLMediaElement): void;
+  detach(): void;
+  load(): void;
+  destroy(): void;
+}
```
mihar-22: This looks like a good place to question what props does a custom media expose because it differentiates with the Dash one I saw earlier. Ideally we'd have a consistent API and expect anything else accessed directly on the underlying engine itself. I can see us removing most of these but love to know what you think!

packages/core/src/dom/media/hls/index.ts
```diff
@@ -24,165 +24,188 @@ export const SourceTypes = {
...
 
-  get error() {
-    return this.#delegate?.error ?? null;
-  }
+interface HlsMediaHost extends EventTarget {
+  readonly target?: EventTarget | null;
+  attach?(target: EventTarget): void;
+  detach?(): void;
+}
```
mihar-22: Same comments as the one on the dash host around shared contract and risky target type.

mihar-22 (commented):
Amazing! This is a massive step in the right direction imo, it's already become so much clearer with the delegate concept being dropped. I largely left feedback around DASH and HLS APIs. I think my comments will cover the rest sufficiently as it's largely repeated designs across media.

I still have concerns around around proxies and the contractual inconsistencies that come from dynamically generated properties in our `core` package. Particularly around the "Base" Media classes that we expose to React and contracts these create. My hope is we can simplify it down to a single core class (down from three per media atm), and then leave the proxying and funny element business to the `html` package. There's also things I'm hoping to push together around decoupling DOM and types here overall. Lastly figuring out a engine lifecycle and hooks/callbacks.

Anyway, not in scope for this PR ofc. Stuff I'm hoping to explore and have more discussions together! 

Great work, thank you :)

packages/core/src/dom/media/dash/index.ts
mihar-22: Actually does this contract overlap with `DashMediaProps`?

packages/core/src/dom/media/hls/hlsjs.ts
luwes: To keep these things separated, these files tend to become so large otherwise and hard to follow.

packages/core/src/core/media/tests/delegate.test.ts
luwes: this tests the internal delegates in HlsMedia, primarily the bridgeEvents util.

packages/core/src/core/media/types.ts
```diff
@@ -1,16 +1,242 @@
-// Detects readonly vs writable properties via conditional type identity check.
-type IfEquals<X, Y, A, B> = (<T>() => T extends X ? 1 : 2) extends <T>() => T extends Y ? 1 : 2 ? A : B;
+export interface EventLike<Detail = void> {
+  readonly type: string;
+  readonly timeStamp: number;
+  readonly detail?: Detail;
+}
```
mihar-22: I was thinking we should actually call this `MediaEvent`. It's a specific event type for Media. The "Like" might be doing too much work in the original types design.

packages/core/src/core/media/types.ts
```diff
@@ -1,16 +1,242 @@
...
+
+export interface MediaBufferEvents {
+  progress: EventLike;
+}
+
+export interface TimeRangeLike {
+  readonly length: number;
+  start(index: number): number;
+  end(index: number): number;
+}
```
mihar-22: Similar to reasons above, maybe we just do `TimeRange` or `MediaTimeRange`.

packages/core/src/dom/media/custom-media-element/index.ts
```diff
@@ -140,273 +55,290 @@ function getVideoTemplateHTML(attrs: Record<string, string>): string {
...
-  static shadowRootOptions: ShadowRootInit;
-  static Events: string[];
-  readonly target: HTMLVideoElement;
-  attributeChangedCallback(attrName: string, oldValue?: string | null, newValue?: string | null): void;
-  connectedCallback(): void;
-  disconnectedCallback(): void;
-  init(): void;
-  handleEvent(event: Event): void;
-}
+const excludedProperties = ['attach', 'detach', 'destroy'];
```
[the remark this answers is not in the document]
luwes: this is fine

PR #1297, opened 2026-04-09
design: media contracts

mihar-22:
Exploratory design document for making our media contracts more flexible, robust, and type-safe.

---

> [!NOTE]
> **Low Risk**
> Documentation-only change adding a draft design proposal; no runtime code or APIs are modified.
> 
> **Overview**
> Adds a new draft design document, `internal/design/media.md`, proposing a shift to *capability-based*, DOM-agnostic media contracts with typed event targets.
> 
> The doc outlines a potential future direction including interface-based capabilities, type guards, a split between core DOM hosts and HTML custom elements, and an async engine lifecycle—without changing any current implementation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 04b06b833f551bdc779255080f2adb5fea306e39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/design/media.md
+++ b/internal/design/media.md
@@ -382,9 +382,9 @@ interface Media extends
 }
 ```
 
-That's it — `play()`, `pause()`, `paused`, `readyState`, and optional
-untyped references to engine and native element. A native `HTMLVideoElement`
-satisfies this. So does a bare-bones custom implementation.
+That's it — `play()`, `readyState`, and optional untyped references to
+engine and native element. A native `HTMLVideoElement` satisfies this.
+So does a bare-bones custom implementation.
 
 Everything else is opt-in via capabilities. `Video` and `Audio` are the
 standard full contracts — each extends `EventTargetLike` once with a
--- a/internal/design/media.md
+++ b/internal/design/media.md
@@ -1,0 +1,831 @@
+---
+status: draft
+date: 2026-04-08
+---
+
+# Media Contracts
+
+A capability-based contract system for media. Custom media authors implement
+plain getters/setters against typed interfaces — no mixins, no proxy magic,
+no DOM recreation.
+
+## Problem
+
+The core media contract is tangled up with platform-specific machinery.
+Mixins, proxies, prototype-walking, and event monkey-patching all live at
+the contract level — so integrators who want to build their own media
+element can't separate "what the API is" from "how the DOM wiring works."
+They look at our HLS implementation as a reference and find four
+interlocking abstractions baked into core that are tough to reason through.
+We want a gold standard like HLS that people can read, understand, and
+replicate without absorbing the full DOM stack.
+
+Proxying isn't inherently a problem — but proxying the entire
+`HTMLMediaElement` API at the core contract level is. Types like
+`HTMLMediaElement`, `TextTrackList`, `Event`, and `EventTarget` make it very difficult to
+build implementations for platforms with DOM limitations or no DOM at all
+(React Native, SSR). We don't need to mirror the DOM 1:1 — just a specific
+subset covering what our player actually uses. Right now there isn't really
+a clear contract; it's an all-or-nothing `HTMLMediaElement` type that ends
+up nudging our own store features into reaching deeper into the DOM than
+they should. And because media creation is fully DOM-coupled in core, we
+wind up needing separate class hierarchies for different platforms (one for
+custom elements, one for React/SSR) — which is the kind of leaky
+architecture that the mixin system was trying to solve but ended up
+compounding. Forwarding a well-defined set of contract properties at the
+HTML element level is fine — the issue is doing it for a platform-specific
+extensive API at the core contract level.
+
+A few things we'd like to address:
+
+1. **Dynamic contract in core** — `defineClassPropHooks` dynamically adds
+   properties by walking prototype chains at runtime, and this happens at
+   the core contract level. The API surface ends up being unpredictable:
+   `in` checks behave inconsistently, `Object.keys()` returns different
+   results depending on which delegate is composed, enumerability varies.
+   It's hard to know the media object's shape from reading any single class.
+
+2. **Inherited complexity** — Media implementors inherit four abstractions
+   (`ProxyMixin`, `DelegateMixin`, `defineClassPropHooks`, `bridgeEvents`)
+   that they don't really need to understand but have to work within. When
+   something breaks, tracing where a property value comes from is tricky —
+   `get(prop)` bounces through delegate → proxy → native element via
+   prototype forwarding that doesn't really show up in the debugger. It's
+   hard to follow, harder to debug, and tough to walk new contributors
+   through.
+
+3. **Misleading store type** — `PlayerTarget.media` is typed as
+   `HTMLMediaElement`, so store features naturally treat it as a full DOM
+   element — calling `querySelectorAll`, accessing `shadowRoot`, referencing
+   `HTMLMediaElement.HAVE_FUTURE_DATA`. The type kind of invites this, and
+   new features tend to reach deeper into the DOM because it looks like
+   they can.
+
+4. **No clear capabilities** — Today it's all-or-nothing. There's no good
+   way to express that a media implementation supports play/pause but not
+   text tracks, or volume but not fullscreen. Engine-specific properties
+   (config, debug, preferPlayback) also end up leaking onto the media
+   object via `defineClassPropHooks`, which muddies what "the media API"
+   actually is.
+
+5. **No async lifecycle** — Engine teardown is synchronous. When switching
+   sources, the old engine's MSE cleanup may not have finished before the
+   new engine starts, which can lead to race conditions.
+
+## Solution
+
+Replace the mixin architecture with:
+
+- **No dynamic contracts or proxies in core** — the core contract is
+  static, typed interfaces. Forwarding a well-defined set of properties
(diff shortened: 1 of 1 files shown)


internal/design/media.md
```diff
@@ -0,0 +1,829 @@
...
+  playing: EventLike;
+  waiting: EventLike;
+}
+
+interface MediaPlaybackCapability {
+  play(): Promise<void>;
+  readonly readyState: MediaReadyStateValue;
+}
+
+const MediaReadyState = {
```
heff: Suggest: I'm cautious to include this in Playback. It feels required, and it's not required for simple use cases like GIF, and non-network related use cases, like a local file. Maybe put it in "Network"?

internal/design/media.md
```diff
@@ -0,0 +1,829 @@
...
+  readonly pip?: boolean;
+  getPictureInPictureAvailability?(): Promise<MediaFeatureAvailability>;
+}
+```
+
+Availability methods (`get*Availability`) return
+`Promise<MediaFeatureAvailability>` — `'available'`, `'unavailable'`, or
+`'unsupported'`. They're optional on the capability (not all implementations
+need them) and async so they can probe the platform without blocking. Store
+features call them once at attach time and cache the result.
```
heff, line 366: comment: I like this. 👍 I think Airplay might break the "cache the result" idea with availability changing over time.

internal/design/media.md
```diff
@@ -0,0 +1,832 @@
...
+  MediaSeekCapability,
+  MediaSourceCapability,
+  MediaVolumeCapability,
+  MediaPlaybackRateCapability,
+  MediaBufferCapability,
+  MediaErrorCapability,
+  MediaTextTrackCapability,
+  MediaFullscreenCapability,
+  MediaPictureInPictureCapability,
+  EventTargetLike<VideoEvents> {}
```
luwes, line 411: 👍

internal/design/media.md
```diff
@@ -0,0 +1,832 @@
...
+The minimum viable media is `MediaPlaybackCapability`:
+
+```ts
+interface MediaEvents extends MediaPlaybackEvents {}
+
+interface Media extends
+  MediaPlaybackCapability,
+  EventTargetLike<MediaEvents> {
+  readonly engine?: unknown;
+  readonly nativeElement?: unknown;
```
luwes, line 381: this is called `target` currently, I don't think it needs to change. what should change is that target will include the container for things that were mentioned in the meeting.

internal/design/media.md
```diff
@@ -0,0 +1,832 @@
...
+  hlserror: EventLike<{ fatal: boolean; details: string }>;
+}
+
+class HlsVideo extends HTMLVideoElementHost<HlsVideoEvents>
+  implements MediaEngineHost<Hls, HTMLVideoElement> {
+
+  readonly engine: Hls | null;
+  attachEngine(target: HTMLVideoElement) { /* ... */ }
+  detachEngine() { /* ... */ }
+  destroyEngine(): Promise<void> { /* ... */ }
```
luwes, line 527: let's keep these named attach, detach, destroy like currently is. Or do these have different functions?

internal/design/media.md
```diff
@@ -0,0 +1,832 @@
...
+Set externally via `setEngineCallbacks()`.
+
+```ts
+interface MediaEngineCallbacks<Engine = unknown, Target = unknown> {
+  onEngineConfig(config: Record<string, unknown>): void;
+  onEngineCreate(engine: Engine): void;
+  onEngineAttach(engine: Engine, target: Target): void;
+  onSourceChange(engine: Engine, src: string): void;
+  onEngineDetach(engine: Engine): void;
+  onEngineDestroy(engine: Engine): Promise<void>;
```
luwes, line 605: will need more context on why these would be needed. until now we haven't seem to have needed these.

internal/design/media.md
```diff
@@ -0,0 +1,832 @@
...
+
+### MediaVolumeCapability
+
+```ts
+interface MediaVolumeEvents { volumechange: EventLike; }
+
+interface MediaVolumeCapability {
+  volume: number;
+  muted: boolean;
+  getVolumeAvailability?(): Promise<MediaFeatureAvailability>;
```
luwes, line 247: these shouldn't be instance methods I think because they would crowd the media API. aren't we handling this in the store currently?

internal/design/media.md
```diff
@@ -0,0 +1,832 @@
...
+
+`HlsVideo` is a plain class in `@videojs/core`. The HTML package wraps
+it as a custom element:
+
+```ts
+// packages/html/src/media/hls-video/index.ts
+export class HlsVideoElement extends VideoElement {
+  static readonly tagName = 'hls-video';
+  readonly media = new HlsVideo();
+}
```
luwes, line 805: HlsVideo defines the API of HLSVideoElement so this wouldn't work. Let's say we add a debug prop on HlsVideo. 

HlsVideoElement should extend HlsVideo and HTMLElement for custom element creation.

internal/design/media.md
mihar-22: If we're referring to the `nativeElement` then `target` sounds goood to me!

internal/design/media.md
mihar-22: Sounds good to me! What I have here does read funny in that you're attaching `target` to the engine, and not the engine itself. I guess I was looking for names that avoid collisions, e.g., `destroy` more generally media vs. destroying the engine. Realistically I don't know if that carries any weight.

internal/design/media.md
mihar-22: Generally speaking this is how people will have consistent ways to hook into the lifecycle and configure the engine. Whether that's configuring, attaching headers to source, any setup logic, additional cleanup on destroy, etc. This section definitely needs more thought and input from everyone. 

This is something we should put aside and discuss more, but I think it'll also help with some of the questions around props you've been thinking to expose on media or not. Preferable to hook into specific points imo.

internal/design/media.md
mihar-22: I think we're referring to `getVolumeAvailability`? Not sure I follow, media authors will need a way to tell the store whether this is supported or not. Today we're doing this ad-hoc for native media element only.

internal/design/media.md
mihar-22: If you do that you're then mixing the core media with `HTMLElement`, rather than viewing it as something media can be hosted on (e.g., Custom Elements). 

The idea here is that the core `HTMLMediaElementHost` accepts a native element via `attachElement` and implements the Media interface. This creates a cleaner base to use in frameworks (e.g., React). Avoids muddying the foundation and tangling with the DOM too early before it's needed. I think we see this manifest as separate classes today in the codebase like `HlsMedia`, `HlsCustomMedia`, and `HlsMediaBase`. I don't think that paints a great experience for media authors, or even for us internally needing 2-3 classes per media implementation.

We can push CE integration into the `html` package and proxy the Media on to the Custom Element, rather than the other way around. If you want to expose `debug` on `HlsVideo` then it would come along for the ride on the custom element. You could also decide not expose those types of props at all until there's good reason, or we could think about how we handle logging in our `Media` contract and have consistency between different engines.

internal/design/media.md
mihar-22: Might be worth considering `attachTarget` and `detachTarget`

heff:
This is not necessary for the contract, but how would we feel about a `createMedia` function to make however we architect this easier for people to put together? I'm at least feeling like from the convo that there's actually no "simple" version when we're trying to support many frameworks with one Media, not just web components.

Something like
```js
// assumes <video> under the hood or provided somehow
const MyMediaCore = createMediaCore(videoMediaFeatures, {
  // Lifecycle hooks, function overrides
})

export const MyMediaHTML = createHtmlMedia(MyMediaCore);
export const MyMediaReact = createReactMedia(MyMediaCore);

// creates an <iframe> under the hood
const MyIframeMedia = createIframeMedia(videoMediaFeatures, {
  // Lifecycle hooks, function overrides  
});
```

mihar-22:
@heff in this design doc I covered how you would support multiple frameworks with a single Media from core. We just have to make sure our custom element integration is kept separately in the `html` package. Rather than proxying a custom element on to media, we would proxy the media on to the custom element. This way you're not violating or muddying the **core** Media classes/contracts. We should dictate and provide the API, not the DOM - even if we're conforming to a subset so that a native element works out of the box.

internal/design/media.md
```diff
@@ -0,0 +1,832 @@
...
+
+## HLS — The Reference Implementation
+
+The HLS media is the primary example. It's a plain class in core — no
+`HTMLElement`, no shadow DOM.
+
+### Core class
+
+```ts
+class HlsVideo extends HTMLVideoElementHost<HlsVideoEvents>
```
sampotts, line 712: Mega nit but why don't we use caps here? i.e. `HLSVideo` to match native `HTMLElement` etc.

mihar-22: I've seen so many variations of this in the wild. I simply stuck with what we had. In Vidstack I went with "HLS" and it's what I would vote for as it's also more consistent with web api naming conventions as you called out.

hls.js uses `Hls` as the type name (class `Hls`) so that's probably why. Weird to me but that sets some precedent.

@luwes @heff any opinions?

PR #1362, opened 2026-04-18
refactor(packages): improve webkit presentation API handling and button availability

mihar-22:
## Summary

Refactor WebKit fullscreen and picture-in-picture handling to use the modern presentation mode API (`webkitSetPresentationMode`) instead of deprecated methods. Replace the `available` boolean on buttons with `disabled` and `hidden` states that follow standard ARIA patterns — `aria-disabled` for non-interactive controls and HTML `hidden` for unsupported features.

## Changes

- Use `webkitSetPresentationMode('fullscreen')` instead of `webkitEnterFullscreen` / `webkitExitFullscreen` for iOS Safari
- Convert fullscreen/PiP features from `async`/`await` to sync promise chaining to preserve user-activation tokens on iOS Safari
- Add strict WebKit type guards (`isWebKitVideoElement`, etc.) replacing loose optional-field interfaces
- Move capability predicates (`isMediaPauseCapable`, etc.) to runtime-agnostic `core/media/predicate.ts`
- Add symbol-based identification to media host classes with `resolveHTMLMediaElement` helpers
- **Breaking:** replace `available: boolean` + `data-available` with `disabled` and `hidden` states on cast, fullscreen, and PiP buttons
- `aria-disabled` is now set from button state (disabled by prop or unavailability), not just props
- HTML `hidden` attribute applied when a feature is unsupported (e.g., no fullscreen API)
- React buttons return `null` when `state.hidden` is true
- Skins: replace old `has-data-[availability]:not-data-[available]:hidden` Tailwind classes with `data-[disabled]` styling (both default and minimal skins)
- Skins CSS uses `&[data-disabled]` instead of `&[disabled]`
- Sync PiP state on `play` event for iOS Safari background-to-foreground transitions
- Correct PiP feature event target to listen on the video element for `webkitpresentationmodechanged`
- Chain `requestPictureInPicture` after `exitFullscreen` so PiP activates once fullscreen exits

<details>

The async-to-sync change preserves the user activation token that iOS Safari requires for fullscreen requests. With `await`, the microtask boundary caused Safari to reject requests silently.

The `disabled` / `hidden` split follows the design doc in `internal/design/ui/disabled-hidden.md`: `aria-disabled` keeps the button visible and discoverable (screen readers still announce it) while `hidden` removes it entirely for unsupported features. This is more semantically correct than the single `available` boolean.

The PiP feature now correctly listens on the video element for `webkitpresentationmodechanged` and chains `requestPictureInPicture` after `exitFullscreen` when transitioning from fullscreen. It also syncs on `play` to handle iOS Safari returning from background while in PiP mode.

</details>

## Testing

1. `pnpm -F @videojs/core test` — unit tests for button cores and feature state
2. `pnpm -F @videojs/react test` — React button integration
3. Manual: verify fullscreen/PiP buttons show `aria-disabled` when unavailable and `hidden` when unsupported
4. Manual: iOS Safari — verify PiP state updates correctly after backgrounding/foregrounding
5. Visual: confirm disabled buttons render with reduced opacity/grayscale in both default and minimal skins

---

> [!NOTE]
> **Medium Risk**
> Touches core presentation (fullscreen/PiP/cast) flows and changes button rendering/attributes across HTML, React, skins, and tests; behavior differs per browser and now propagates errors instead of swallowing them.
> 
> **Overview**
> Refactors fullscreen and picture-in-picture support detection and transitions to use WebKit’s presentation mode API (via new `presentation/webkit.ts`) and new `resolveHTML*` helpers/symbol-based host guards, updating cast/fullscreen/PiP features and tests to match.
> 
> Updates Cast/Fullscreen/PiP buttons to model **`disabled` vs `hidden`** explicitly: `aria-disabled` now reflects both prop-based disabling and feature unavailability, unsupported features set native `hidden` (and React buttons return `null` via `isSupported`), and styling shifts from `[disabled]`/`data-availability` hiding to `[data-disabled]` with docs/e2e/tests adjusted accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4cdc2d7ebfb59d406a2d009e6f24874ff771857. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/core/src/dom/media/predicate.ts
```diff
@@ -1,81 +1,89 @@
...
-): value is {
-  querySelectorAll: (selectors: T) => NodeListOf<HTMLElementTagNameMap[Extract<T, keyof HTMLElementTagNameMap>]>;
-} {
-  return (
-    isObject(value) && 'querySelectorAll' in value && isFunction((value as Record<string, unknown>).querySelectorAll)
-  );
+export function resolveHTMLAudioElement(media: Media): HTMLAudioElement | null {
+  if (media instanceof HTMLAudioElement) return media;
+  if (isHTMLAudioElementHost(media)) return media.target;
+  return null;
```
luwes, line 88: This feels like it's going against the uniform media API design. We shouldn't have to do this. I might need more context.

luwes (changes requested):

packages/core/src/dom/media/predicate.ts
mihar-22: No problem, happy to discuss more! I get where that sense is coming from, valid. I'll do my best to lay my thoughts out below.

As you know, we have an interesting and unique challenge of wanting to support native browser media out of the box (e.g., `<video>`) and also have a flexible enough `Media` interface that anyone can implement easily. The "easy" part comes from the last wave of changes we made to untangle assumptions about what subset of DOM-like APIs we support, and how any "Media" can satisfy all, or some, of the complete interface through "capabilities." 

Based on our definition of "Media", we can't _assume_ how it's implemented. At the end of the day, how someone satisfies the interface is up to the them. This means we don't actually know contractually if it's a `HTMLMediaElement` that's been directly given to the store, something of the shape `{ target: HTMLMediaElement }`,  rendering inside an `<iframe>`, rendering to a `<canvas>`, or however they've decided to build it out depending on the platform, environment, or tech they want to support. As another example, [moq.dev](https://doc.moq.dev/) is built on `WebTransport`, `WebCodecs`, and `WebAudio`. Another cool library that someone might build upon is [Mediabunny](https://mediabunny.dev/). React Native won't have a DOM.

Basically this all means, we need a way to resolve DOM media elements if they're needed in our store DOM features to support certain APIs like fullscreen and pip out of the box. Normally, we would move that logic into the `HTMLVideoElementHost` but because we want to support `<video>` itself for spec-compliance, we simply can't. This would be true for external store feature authors too, they might need access to these specific elements _when available_.

The media type guards like `isHlsMedia` and `isDashMedia` are for lower-level access to media/engine APIs without bundling the media themselves via `instanceof` checks or some other direct reference (keeping it treeshakeable). Important to remember the store is a base/generic `Media` interface, it needs to be narrowed for deeper access. You can imagine ourselves or someone else externally building out a store feature that might need it for something like analytics, logging, debugging, testing, etc. It could even be an odd case where we need a small workaround in a store feature for a specific media. This happens all the time.

In summary, the resolvers and media type guards don't invalidate the uniform Media API design. They're just there to support the flexible contracts we've set up.

---

Something I haven't tackled in this PR is that Media authors should be able to provide their own fullscreen and pip interfaces separately from the DOM. This means notify the store of availability, handle request/exit, dispatch events, and whatever else is needed. This can ultimately be a separate conversation in another PR. Thought it was worth noting here.

mihar-22: I was thinking about this today a bit more, wanted to illustrate the problem and a potential way we could decide to move forward now, or in the future to strength our Media API design.

So the problem I was describing above is because as of today we're expecting:

In HTML via fallback:

```html
<media-container>
  <video />
</media-container>
```

Which will essentially do the following in JS:

```ts
store.attach(video); // video = <video>
```

This is what leads to the awkwardness and expectations in our store features. Technically it's also why we have to design our Media API a certain way. 

Basically it can be the `<video>` element _or_ it could be a target (e.g., `HTMLVideoElementHost`).

---

There's a world where the default expectation could be:

```ts
import { HTMLVideoElementHost } from '@videojs/html';
const media = new HTMLVideoElementHost();
media.attach(video); // video = <video>
store.attach(media);
```

> This is no different to [playback tech in VJS 8](https://legacy.videojs.org/guides/tech/#building-an-api-wrapper)

This means we now control the Media API completely, not partially by the DOM. But it also means in order to do this via HTML we need a custom element that attaches native media so host class is treeshaken out:

```html
<media-container>
  <html-video> <!-- performs the code above -->
    <video />
  </html-video>
</media-container>
```

A bit ugly if someone wants to use native media elements, but it does mean we can move native fullscreen/pip handling out of store features and into media layer (where it belongs imo). Why I like this:

1. Media API becomes solely responsible for media-related behaviour - it's not snuck into store features or living in potentially multiple places. It also means the store features are further disconnected from the DOM.
2. Maybe `document` fullscreen and pip handling could be composed in via classes or mixins so others can take advantage of it too. Clearer boundary there too between what is specifically native media handling vs. document. 
3. Consistent with our other media in terms of how context is used for discovery and attaching. Technically anyone could pass in a `<video>` compliant element in and it would work.

Anyway, all for food for thought :) I just wanted to make sure I illustrated some of the challenges we have. It all kind of hinges on how we strongly we believe passing in `<video>` with no adapters should work and more importantly why.

heff: There's some fundamental misalignment happening here that is hard to nail down and we should talk IRL, but I think it might boil down to the following. And I can't fully tell if we're disagreed on something or if we didn't implement something as planned.

> Basically this all means, we need a way to resolve DOM media elements if they're needed in our store DOM features to support certain APIs like fullscreen and pip out of the box. Normally, we would move that logic into the `HTMLVideoElementHost` but because we want to support `<video>` itself for spec-compliance, we simply can't. This would be true for external store feature authors too, they might need access to these specific elements when available.

Specifically "need a way to resolve DOM media elements if they're needed in our store"

If the store needs access to the internal DOM media element then we're not doing the Media API well. Nothing should have to reach through the Media API contract and grab the internal media. We might do that for debugging, but any normal feature operation that needs to reach through actually needs an extension of the Media API.

> This is no different to playback tech in VJS 8

I intentionally moved away from techs when building Media Chrome because it was completely redundant. We had a `<video>` API (vjs player) wrapping a `<video>` API (tech) wrapping a `<video>`API (`<video>`). We definitely haven't missed techs in that codebase. It does add needed intentionally in extending the media API. But I would rather get more aggressive in extending the API than going back to Techs (i.e. `HTMLVideoElementHost`). We should be intentional, naturally, because it's an API devs will work with directly. But at the same time we shouldn't see it as heavy thing we can't touch and have to work around. We're in control of our ecosystem.

@mihar-22 This is where I think I'm feeling some hesitancy from you, at least from the suggestion of requiring a `<html-video>` wrapper for `<video>`.

> but it does mean we can move native fullscreen/pip handling out of store features and into media layer

This is where I get confused where the misalignment is happening, because I'd agree with this and assumed this was already the case. When the store triggers PiP and `<hls-video>` is being used, does it call `requestPictureInPicture` on `<hls-video>` or does it reach deeper and call it directly on the internal `<video>`. If it's the former, that's intended, if it's the latter it goes against the Media API contract. 

So help me understand if you can see where the misalignment is happening more clearly.

PR #1395, opened 2026-04-21
docs(design): add design doc for live presets

luwes:
## Summary

Records the decision to ship live playback as dedicated, **live-only** `live-video` / `live-audio` presets rather than branching the existing `video` / `audio` skins on `streamType`. Live skins render live UI unconditionally — no in-skin VOD fallback, no `unknown`-state placeholder. VOD-only apps keep their current store shape and bundle; live-only apps get a live-dedicated fork template with no dead branches.

## Changes

- New design doc `internal/design/ui/live-presets.md` covering the decision, alternatives, rationale, trade-offs, consequences, open questions, and prior art.

## Key points captured in the doc

- **Two presets, two jobs.** Base presets play VOD, live presets play live. A preset is a commitment to one mode, not a runtime toggle between both.
- **Live skin is live-only.** Renders a live indicator, jump-to-live-edge button, and (where applicable) a DVR-window slider unconditionally. Does not read `streamType` and does not attempt to recover to a VOD layout.
- **Feature arrays start minimal.** `liveVideoFeatures` / `liveAudioFeatures` initially alias `videoFeatures` / `audioFeatures` — `streamTypeFeature` is intentionally **not** included yet. It's an additive, non-breaking change to append it once a live-only affordance actually needs stream-type state.
- **Base skins stay VOD-only.** No change to `VideoSkin` / `AudioSkin` or to the derived store types.
- **Primary rationale:** VOD players don't pay for live; explicit author opt-in for live capability; no unknown-state flash on first paint; clean fork templates in both directions; easy to layer dual-mode support on later, hard to retract once shipped.
- **Explicitly deferred:** runtime mode switching within a single preset, and live-replay / post-stream-VOD flows — revisit when a concrete product need surfaces.

## Testing

Docs-only; no code changes.

---

> [!NOTE]
> **Low Risk**
> Docs-only change that adds a design decision record; no runtime, API, or behavior changes.
> 
> **Overview**
> Adds a new design doc `internal/design/ui/live-presets.md` that records the decision to ship **separate, live-only** `live-video`/`live-audio` presets (rather than branching existing VOD skins on `streamType`), outlines the intended feature/skin shape at a high level, and documents trade-offs, alternatives, and open questions for future implementation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 463a462ad4bd2f41e030975ac8739e3bc5bd22d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/design/ui/live-presets.md
+++ b/internal/design/ui/live-presets.md
@@ -18,22 +18,27 @@ Two presets, two jobs: base presets play VOD, live presets play live. A preset i
 
 ## Context
 
-`streamTypeFeature` exposes `streamType: 'on-demand' | 'live' | 'unknown'` in the player store. Live playback needs visibly different UI from VOD:
+Live playback needs visibly different UI from VOD:
 
 - The duration display and standard scrubbing affordances disappear or change
 - A live indicator / "jump to live edge" button takes their place
 - The time slider, if shown at all, represents a DVR window, not `[0, duration]`
 - Live-only affordances are likely to grow (latency badge, "behind live" state, chat slot, etc.)
 
-Not every app needs this. Most VOD-only integrations don't want to pay for live UI code. And apps that do play live content almost always know that at embed time — an event page, a channel page, a DVR UI. "Live-aware" and "VOD-only" are two meaningfully different products, and authors pick one at integration time.
+`streamTypeFeature` exposes `streamType: 'on-demand' | 'live' | 'unknown'` and could drive this at runtime inside one shared skin. We're choosing not to.
 
-This doc records both the split (live vs. VOD as separate presets) and the narrower commitment inside the live presets (live-only, no in-skin VOD fallback).
+Authors almost always know at embed time whether their product plays live or VOD — an event page, a channel page, a DVR UI. "Live-aware" and "VOD-only" are two meaningfully different products, and the vast majority of `video` / `audio` integrations are VOD-only.
+
+This doc records two decisions, with separate rationales below:
+
+1. **Split** — live and VOD ship as separate presets.
+2. **Live-only** — the live preset's skin renders live UI unconditionally, with no in-skin VOD fallback.
 
 ## Shape
 
 ### Feature arrays
 
-Base presets stay as today. Live presets start identical in capability and will diverge as live-only needs emerge:
+Base presets stay as today. Live presets start close in capability and will diverge further as live-only needs emerge. `playbackRateFeature` is dropped — speed controls aren't meaningful for live playback, where the user is always pinned to (or chasing) the live edge.
 
 ```ts
 // packages/core/src/dom/store/features/presets.ts
@@ -52,7 +57,19 @@ export const videoFeatures: VideoFeatures = [
   errorFeature,
 ];
 
-export const liveVideoFeatures: VideoFeatures = videoFeatures;
+export const liveVideoFeatures: VideoFeatures = [
+  playbackFeature,
+  volumeFeature,
+  timeFeature,
+  sourceFeature,
+  bufferFeature,
+  fullscreenFeature,
+  pipFeature,
+  remotePlaybackFeature,
+  controlsFeature,
+  textTrackFeature,
+  errorFeature,
+];
 
 export const audioFeatures: AudioFeatures = [
   playbackFeature,
@@ -64,10 +81,21 @@ export const audioFeatures: AudioFeatures = [
   errorFeature,
 ];
 
-export const liveAudioFeatures: AudioFeatures = audioFeatures;
+export const liveAudioFeatures: AudioFeatures = [
+  playbackFeature,
+  volumeFeature,
+  timeFeature,
+  sourceFeature,
+  bufferFeature,
+  errorFeature,
+];
 ```
 
-`streamTypeFeature` is intentionally **not** included yet. The live skin assumes live, so it doesn't need runtime detection. When a live-only affordance actually needs stream-type state (e.g., surface an error if a VOD source is loaded, or drive a "behind live" badge derived from seekable + streamType), it's an additive, non-breaking change to append it to `liveVideoFeatures` / `liveAudioFeatures`.
+`streamTypeFeature` is intentionally **not** included. The live skin assumes live, so it doesn't need runtime detection. When a live-only affordance actually needs stream-type state (e.g., surface an error if a VOD source is loaded, or drive a "behind live" badge derived from seekable + streamType), it's an additive, non-breaking change to append it to `liveVideoFeatures` / `liveAudioFeatures`.
+
+No live-specific time feature either. `timeFeature` already resolves `duration` to `seekable.end(last)` when the native duration is `Infinity`, which is all the live skin needs for now. If `currentTime` or other time values need live-specific handling later, that's also additive.
+
+Beyond `playbackRateFeature`, the live feature arrays start as near-copies of their VOD counterparts. That's a conservative starting point, not a commitment that live needs every remaining feature. Further bundle wins for live-only are expected to come primarily from the skin (smaller DOM, fewer subscribers — see below). Trimming additional features from the live arrays is a follow-up once the skins land and we can measure.
 
 ### Preset packages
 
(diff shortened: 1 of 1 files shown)


internal/design/ui/live-presets.md
```diff
@@ -0,0 +1,182 @@
...
+
+Base presets stay as today. Live presets start identical in capability and will diverge as live-only needs emerge:
+
+```ts
+// packages/core/src/dom/store/features/presets.ts
+export const videoFeatures: VideoFeatures = [
+  playbackFeature,
+  playbackRateFeature,
+  volumeFeature,
+  timeFeature,
```
heff, line 49: Question: Do we not need live-specific time handling capability?

internal/design/ui/live-presets.md
```diff
@@ -0,0 +1,182 @@
...
+`streamTypeFeature` exposes `streamType: 'on-demand' | 'live' | 'unknown'` in the player store. Live playback needs visibly different UI from VOD:
+
+- The duration display and standard scrubbing affordances disappear or change
+- A live indicator / "jump to live edge" button takes their place
+- The time slider, if shown at all, represents a DVR window, not `[0, duration]`
+- Live-only affordances are likely to grow (latency badge, "behind live" state, chat slot, etc.)
+
+Not every app needs this. Most VOD-only integrations don't want to pay for live UI code. And apps that do play live content almost always know that at embed time — an event page, a channel page, a DVR UI. "Live-aware" and "VOD-only" are two meaningfully different products, and authors pick one at integration time.
+
+This doc records both the split (live vs. VOD as separate presets) and the narrower commitment inside the live presets (live-only, no in-skin VOD fallback).
```
heff: Suggest: This is a key detail, so we should make sure we're capturing the reasoning for both clearly and separately, so if we're looking back we can quickly remember why.

For the split, I'm not seeing a mention of filesize being a driving factor. We could mention the "Every byte must be earned" principle and that the vast majority of uses of `video` won't use live.

For the no VOD fallback on the live side, we're still talking through that but let's make sure to capture the specific reasoning. There's a bit of AI-verbosity in this doc and it'd be good to just say the reasons succinctly somewhere.

internal/design/ui/live-presets.md
```diff
@@ -0,0 +1,182 @@
...
+  bufferFeature,
+  fullscreenFeature,
+  pipFeature,
+  remotePlaybackFeature,
+  controlsFeature,
+  textTrackFeature,
+  errorFeature,
+];
+
+export const liveVideoFeatures: VideoFeatures = videoFeatures;
```
heff: This doesn't make a strong argument for live not supporting VOD also. In chat you said:

> I do feel like if it is actually live only (non-dvr) for starters it is a nice opportunity to start with a smaller bundle in UI. No time slider, no thumb previews, no seek buttons, etc.

Should that be reflected in the state composition better? Are there sub features of these features that might be split out into their own thing?

internal/design/ui/live-presets.md
```diff
@@ -0,0 +1,182 @@
...
+- **Branch the existing `video` / `audio` skins on `streamType`.** One preset per medium, one skin that covers both modes by reading `streamType`. Pattern already used for `volumeAvailability === 'unsupported'` (see `VolumePopover` in `packages/react/src/presets/video/skin.tsx:60`).
+- **Single preset, runtime-selected skin.** Keep one `video` / `audio` preset but have it mount a different skin tree based on `streamType`. Hides the split from the author but retains the unknown-state and dead-code problems.
+
+## Rationale
+
+Live-only presets win on four axes:
+
+1. **VOD players don't pay for live.** The dominant case is on-demand. Keeping `streamTypeFeature` and the live UI tree out of `videoFeatures` / `audioFeatures` means the VOD store has no `streamType` slice, the VOD skin has no live-branch code, and bundles for VOD-only apps omit both. A shared-skin approach forces every player to carry the stream-type event plumbing and the live-UI tree, even when the source is known to be VOD.
```
heff: This is where we could specifically mention file size. It's not a strong argument without answering another level of "why does this matter?". "carry the stream-type event plumbing" doesn't go far enough.

internal/design/ui/live-presets.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Alternatives Considered
+
+- **Live skin branches on `streamType` to cover VOD too.** Ship `liveVideoFeatures = [...videoFeatures, streamTypeFeature]` and render live vs. on-demand time controls inside the live skin based on detection, with a neutral placeholder while `streamType === 'unknown'`. Supports runtime source switches between live and VOD within one preset.
+- **Branch the existing `video` / `audio` skins on `streamType`.** One preset per medium, one skin that covers both modes by reading `streamType`. Pattern already used for `volumeAvailability === 'unsupported'` (see `VolumePopover` in `packages/react/src/presets/video/skin.tsx:60`).
+- **Single preset, runtime-selected skin.** Keep one `video` / `audio` preset but have it mount a different skin tree based on `streamType`. Hides the split from the author but retains the unknown-state and dead-code problems.
+
+## Rationale
```
heff: This section is combining the rational for both decisions and I'd love to see that split out.

internal/design/ui/live-presets.md
luwes: No, I tried to keep it simpler downstream for now and updated the duration to seekable-end for live. 
We can do the same for current time if needed. In media-chrome we had this separately and handled this in the UI itself but I'm don't see the need for it yet. Let me know if you can think of edge cases.

internal/design/ui/live-presets.md
luwes: The only thing for now that I could see was the `playbackRateFeature` that could be removed for live and also don't forget `streamTypeFeature`.

Added an open question to split up `textTracksFeature` since it holds subtitles, chapters and storyboards...
https://github.com/videojs/v10/pull/1395/changes#diff-8c4183f7c04c0f6fd650eb76f2b53ba4a351fc449b2ee9437db48e6fe43f39a5R212

internal/design/ui/live-presets.md
luwes: https://github.com/videojs/v10/pull/1395/changes#diff-8c4183f7c04c0f6fd650eb76f2b53ba4a351fc449b2ee9437db48e6fe43f39a5R147-R177

## Rationale: Split live from VOD

The dominant case is VOD — the vast majority of `video` / `audio` integrations never play live. Every byte the VOD preset carries for a capability it doesn't use is a byte the dominant case pays for something it doesn't get.

A shared `video` preset covering both modes forces every VOD player to ship:

- The `streamType` state slice and the events, subscribers, and predicate helpers that maintain it.
- The live UI tree — live indicator, jump-to-live-edge button, DVR slider — rendered conditionally but present in the bundle.
- The branching logic itself (detection, unknown-state handling, transitions).

A separate `live-video` preset keeps that cost with the apps that actually use live. The VOD store has no `streamType` slice; `VideoSkin` has no live branches; nothing in the default import path references live code. The bundle-size report already separates `/video/skin` from a future `/live-video/skin` entry — the split is visible and measurable.

Secondary wins:

- **Explicit opt-in matches author intent.** Picking `live-video` is a one-word signal that live is the supported path — clearer than "use `video` and set some prop" and more discoverable than "`video` happens to work if the source is live."
- **Smaller type surface in the common store.** `VideoPlayerStore` / `AudioPlayerStore` keep their current state shape. Adding `streamType` to the base ripples into every consumer of those types.
- **Targetable visual regression tests.** Each skin snapshots independently; the VOD snapshot stays stable without a `streamType` harness.

## Rationale: Live preset renders live UI only

Given we're shipping a dedicated live preset, it could still support VOD internally — branch the live skin on `streamType` and fall back to VOD controls when the source isn't live. We're not doing that either.

1. **Smaller live bundle.** Live-only (especially non-DVR) unlocks a meaningfully smaller UI: no time slider, no thumbnail previews, no seek buttons, no remaining-time display. A dual-mode live skin has to ship all of those for the VOD branch, plus the branching. Live-only gets to be genuinely smaller, not just differently shaped.

2. **No unknown-state flash.** A dual-mode live skin starts every load in `streamType === 'unknown'` and resolves to `live` (or `on-demand`) after manifest detection. Either we render a neutral placeholder during that window — a third UI state to design and test — or we pick a default and flip visibly once detection completes. A live-only skin has one shape from first paint.

3. **Fork-template clarity.** Skins are reference implementations authors copy and mutate. A live-only skin gives live-app authors a starting point with only live concepts in the tree — no conditional helpers, no `unknown` placeholder, no dead VOD branches to delete. VOD authors forking `VideoSkin` get the same treatment from the other side.

4. **Headroom for divergence.** Live UI tends to accumulate bespoke affordances — latency indicator, "behind live" badge, DVR-aware scrubber interactions, live chat/reactions slot. Each lands naturally inside `LiveVideoSkin` without leaking concepts into the dual-mode branching.

5. **Easy to add, hard to remove.** If we later find authors commonly need one preset that handles live → VOD replay or cross-mode source swaps, we can add `streamTypeFeature` and branch the skin — additive. Shipping a dual-mode skin and later deciding the unknown-state handling and dead code aren't worth it is a breaking change for anyone on the VOD path.

heff (approved):

PR #1447, opened 2026-04-23
docs(design): draft composable media design

luwes:
## Good to know

No new methods on the host: extensions install themselves via `installExtension`, layers attach via `addLayer`, and consumers reach them through `getExtensions(media)` — all imported functions. The host also keeps a single `target` getter/setter instead of attach/detach/getTarget, shrinking the public surface. Happy to discuss.

## Summary

Draft design doc for replacing the static media-mixin chain with a runtime-composable model. Two roles, one host:

- **Media Extension** — installs itself on a host, owns its lifecycle and state, and may push one or more layers (or only observe). Usually produced by a factory like `googleCast({ … })`.
- **Media Layer** — anything pushed onto the layer stack via `addLayer(media, layer)`.

## Problem

The current `MuxDataMediaMixin(GoogleCastMixin(HlsMedia))` pattern doesn't scale:

- Inheritance forces independent feature axes onto one linear chain — combining `N` needs a class per combination.
- A mixin is bound to its base, so each feature must be re-applied per media implementation.
- A mixin yields a new subclass, so you can't augment an instance you already hold.
- The class is fixed at module load — no code-splitting, lazy loading, or runtime add/remove/swap.

## Solution

A host and its layers form a singly-linked list rooted at the host with `HTMLMediaElement` at the tail. `addLayer` inserts new layers right below the host and returns a destroy that pops them. Layers forward through the chain via `super` — `host.play()` cascades down until it reaches the target.

Extensions are deduplicated per host (keyed by factory), looked up via `getExtensions(media)`, and torn down through `extension.destroy()` or host destruction. Strict by default: extensions can't extend the host's API surface — subclass the host when you need to.

## What's in the doc

- **Quick Start** — imperative, dynamic import, factory style, and a forward-looking React `<GoogleCast>` sketch.
- **API Surface** — `MediaExtension` + `installExtension` / `getExtensions`, `MediaLayer` (`root` / `next` / `target`) + `addLayer`, and the chain-anatomy diagram.
- **Bundle cost** noted inline (brotli: extension ~151 B, layer ~543 B).

## Testing

Docs-only — no code paths touched.

---

> [!NOTE]
> **Low Risk**
> Documentation-only change with no runtime or API behavior modified.
> 
> **Overview**
> Adds a **draft design doc** (`internal/design/media-features.md`) proposing **composable media**: replace static mixin inheritance with **media extensions** (install/destroy, deduped per host via factory-keyed registry) and **media layers** (stacked under the host via `addLayer`, singly-linked chain to the `HTMLMediaElement`).
> 
> The doc frames **why** mixins fail (combinatorial classes, no runtime add/remove, no code-splitting) and documents **how** consumers would compose features—imperative install, dynamic import, factory `createHlsMedia([…])`, and a sketch React child pattern—plus API shapes for `MediaExtension`, `installExtension` / `getExtensions`, `MediaLayer` (`root` / `next` / `target`), and call forwarding through the layer stack.
> 
> **No implementation** in this PR; it is documentation only ahead of a future refactor away from patterns like `MuxDataMediaMixin(GoogleCastMixin(HlsMedia))`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a947ef8cef3dd04f837a0830f28aecd348fc5957. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/design/media-features.md
+++ b/internal/design/media-features.md
@@ -27,8 +27,6 @@ and progressing to the underlying primitive.
    instance composition (`new HlsMedia({ features })` / `.use(f)`).
 4. Add class-time composition (`HlsMedia.with(...)`) for custom-element
    authors.
-5. Optional: side-effect import entry points (`@videojs/html/auto`) as
-   sugar only — **not** the only path.
 
 Keep existing mixins usable internally; migrate leaf-first (`live`,
 `errors`, `streamType`) where the lifecycle is already clean, and then
@@ -264,7 +262,7 @@ or use the existing `containerContext` / `playerContext` pattern).
 
 ## Option 2 — `defineMediaFeature` primitive
 
-Underlying primitive that Option 1's child elements (and Options 5–9)
+Underlying primitive that Option 1's child elements (and Options 3–5)
 are built on. Mirrors `defineSlice` on the media side.
 
 ```ts
@@ -273,7 +271,24 @@ export interface MediaFeatureContext<Host> {
   engine?: unknown;
   target: HTMLMediaElement | null;
   signal: AbortSignal;
+  // Install a *new* member on the host (e.g. `liveEdgeStart`).
   define: (name: string, desc: PropertyDescriptor) => void;
+  // Wrap an *existing* method on the host (e.g. `play`, `pause`,
+  // `load`). The wrapper receives the previous implementation and must
+  // return a replacement; wrappers compose LIFO and are restored on
+  // `signal` abort.
+  override: <K extends FunctionKeys<Host>>(
+    name: K,
+    wrap: (original: Host[K]) => Host[K]
+  ) => void;
+  // Same idea for getters/setters (e.g. `currentTime`, `volume`,
+  // `paused`). The wrapper receives the previous descriptor and
+  // returns a replacement descriptor.
+  overrideAccessor: <K extends keyof Host>(
+    name: K,
+    wrap: (prev: { get?: () => Host[K]; set?: (v: Host[K]) => void }) =>
+      { get?: () => Host[K]; set?: (v: Host[K]) => void }
+  ) => void;
   emit: (type: string) => void;
 }
 
@@ -294,7 +309,7 @@ across delegates, types via `UnionToIntersection<InferAPI<F[number]>>`
 **Cons:** reliance on `Object.defineProperty` for per-feature props
 (already used across the codebase).
 
-### Worked example: `live` on `HlsMedia`
+### Worked example: `live` on `HlsJsMedia`
 
 Today's equivalent is a mixin that does a constructor-time
 `engine.on(...)`, holds private state, and has no standard teardown
@@ -418,9 +433,56 @@ What the shape buys us:
   plus an `AbortController`, drive it, then `controller.abort()`. No
   class instantiation, no mixin composition.
 
+### Overriding host members
+
+`define()` installs *new* members; `override()` and `overrideAccessor()`
+wrap *existing* ones. This is the answer to "how does a feature swap
+behaviour on `play` / `pause` / `currentTime` when casting?"
+
+Under the hood each helper does the same thing as a classical mixin
+override, just scoped to a feature + `AbortSignal`:
+
+```ts
+// packages/core/src/dom/feature.ts (sketch — inside the host registry)
+
+function override<K extends FunctionKeys<Host>>(name: K, wrap: Wrapper) {
+  const proto = Object.getPrototypeOf(host);
+  const prevDesc = findDescriptor(proto, name);
+  const original = prevDesc.value as Host[K];
+  const replacement = wrap(original);
+  Object.defineProperty(host, name, { ...prevDesc, value: replacement });
+  signal.addEventListener('abort', () => {
+    Object.defineProperty(host, name, prevDesc);
+  });
+}
+
+function overrideAccessor<K extends keyof Host>(name: K, wrap: AccessorWrapper) {
+  const prevDesc = findDescriptor(Object.getPrototypeOf(host), name);
+  const next = wrap({ get: prevDesc.get, set: prevDesc.set });
+  Object.defineProperty(host, name, { ...prevDesc, ...next });
+  signal.addEventListener('abort', () => {
+    Object.defineProperty(host, name, prevDesc);
+  });
+}
+```
+
+Two guarantees:
+
+- **Wrappers see the previous implementation.** A second feature
+  overriding the same member gets *this feature's* wrapped version as
+  its `original`, so multi-feature overrides compose LIFO — identical
+  to a mixin chain, but declarative.
+- **Cleanup is automatic.** `signal.abort()` (detach / destroy /
+  element removal) restores the exact descriptor present before this
+  feature installed its wrapper. No bookkeeping on the host.
+
 ### Sketch: `googleCastFeature`
 
-Same primitive, wrapping the side-effecty Cast framework:
+Same primitive, wrapping the side-effecty Cast framework. The current
+`GoogleCastMixin` overrides ~14 members on the host (`play`, `pause`,
+`load`, `currentTime`, `volume`, `muted`, `paused`, `ended`, `seeking`,
+`readyState`, `duration`, `playbackRate`, `remote`). Each is the same
+shape: "if casting, defer to `provider`, else `super`". As a feature:
 
 ```ts
 // packages/core/src/dom/media/google-cast/feature.ts
@@ -438,7 +500,7 @@ export function googleCastFeature(config: {
   return defineMediaFeature<GoogleCastMediaHost>({
     name: 'google-cast',
     supports: () => requiresCastFramework(),
-    setup({ host, signal, define }) {
+    setup({ host, signal, define, override, overrideAccessor }) {
       if (!host.disableRemotePlayback) loadCastFramework();
 
       const provider = new GoogleCastProvider(host, {
@@ -449,9 +511,32 @@ export function googleCastFeature(config: {
         customData: () => config.customData,
       });
 
-      // Only `remote` needs to be on the host — the rest live in `config`
-      // and reach the Cast framework through the provider above. No more
-      // `castSrc` / `castReceiver` / `castContentType` on the element.
+      // Method overrides — same shape as today's mixin, minus the
+      // `super.*` boilerplate.
+      override('play',  (base) => () => provider.isCasting ? provider.play()  : base.call(host));
+      override('pause', (base) => () => provider.isCasting ? provider.pause() : base.call(host));
+      override('load',  (base) => () => provider.isCasting ? provider.load()  : base.call(host));
+
+      // Accessor overrides — read-only state.
+      for (const key of ['paused', 'ended', 'seeking', 'readyState', 'duration'] as const) {
+        overrideAccessor(key, ({ get }) => ({
+          get: () => (provider.isCasting ? provider[key] : get!.call(host)),
+        }));
+      }
+
+      // Accessor overrides — read/write state.
+      for (const key of ['currentTime', 'volume', 'muted', 'playbackRate'] as const) {
+        overrideAccessor(key, ({ get, set }) => ({
+          get: () => (provider.isCasting ? provider[key] : get!.call(host)),
+          set: (v) => {
+            if (provider.isCasting) (provider as any)[key] = v;
+            else set!.call(host, v);
+          },
+        }));
+      }
+
+      // `remote` is a *new* member that cast adds on top of the host
+      // surface — no prior descriptor to wrap.
       define('remote', { get: () => provider.remote });
 
       signal.addEventListener('abort', () => provider.destroy());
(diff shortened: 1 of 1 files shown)


internal/design/media-features.md
```diff
@@ -0,0 +1,628 @@
...
+  install(host: HTMLMediaElementHost<HTMLMediaElement, any>) {
+    const off = new AbortController();
+    host.addEventListener('load',   () => initMuxDataSdk(host, this.options),  { signal: off.signal });
+    host.addEventListener('detach', () => host.target?.mux?.destroy(),         { signal: off.signal });
+    return () => off.abort();
+  }
+}
+```
+
+### HLS — engine selection as an extension
```
luwes: this example is actually not fully correct. listening to the load event doesn't make sense.

internal/design/media-features.md
```diff
@@ -0,0 +1,628 @@
...
+  implements MuxMediaProps {}
+
+// MuxMediaProps surfaces the union of every mixin's props on the host:
+type MuxMediaProps = HlsMediaProps & GoogleCastMediaProps & MuxDataMediaProps;
+// e.g. castSrc, castContentType, castReceiverApplicationId, muxEnvKey, …
+```
+
+That pattern has costs we want to remove:
+
+1. **Static composition.** The class is baked at module load. Cast can't be
```
cjpillsbury: question(non-blocking) - Are there actual+plausible scenarios we're concerned about here? Is this about "on the fly" dynamic loading or more about "composing at the time of usage"? Just making sure we're accurately capturing the actual concerns we're attempting to address here and why mixin-style class definition extensions are insufficient to account for them. Maybe one or two realistic example scenarios would help here.

internal/design/media-features.md
```diff
@@ -0,0 +1,628 @@
...
+   code-split or lazy loaded. Once mixed in, behavior can't be added,
+   removed, or swapped at runtime. Cast doesn't have a way to truly turn
+   itself off.
+2. **Muddled media API.** Mixins add config props to the media API surface
```
cjpillsbury: question(non-blocking): This isn't inherently the case, right? A mixin-style class definition extension could still "namespace" things to extension objects, right?

internal/design/media-features.md
```diff
@@ -0,0 +1,628 @@
...
+
+We want a model where Cast, Ads, and future extensions are independent
+modules that attach to a host at runtime, and where redirecting playback
+(Cast, ads, ...) is a single instance swap.
+
+## Solution
+
+Two roles, one host:
+
+- **Extension** — A small class registered via `host.use(extension)`. Owns its
```
cjpillsbury: question(non-blocking): Here and likely elsewhere, I think it's worthwhile to try to distinguish aspects of this API proposal that are:
1. used + assumed by the media store and related -vs-
2. used + assumed by our core / shared media abstractions (base classes, our extension definitions themselves, our react / web component implementations, etc) -vs-
3. helpful and formalized but not "deeply assumed" extensions to our core / shared media abstractions

Take a concrete example: Let's say someone already has a feature-rich playback engine/player that already includes things like google cast and custom Mux Data integration (e.g. shaka player, an in house player/playback engine for an M&E company, etc.).
1. Is there anything from the media store or "above" the media API that would require this person to rely on `host.use(extensionButNotReallyBcItIsBuiltIn)`?
2. Will this person still have code that we publish that will help them conform to VJS expectations if they don't need to use `host.use(extension)`? Core classes they can reach for? Code (e.g. easy to use hooks) to make React integration easy and doesn't assume they're using our implementation of extensions?

Describing the contracts based on what assumptions are where would be helpful.

internal/design/media-features.md
```diff
@@ -0,0 +1,628 @@
...
+  own lifecycle (`install` → uninstaller) and any state it needs. May install
+  a strategy. May only observe.
+- **Strategy** — A class extending `BaseMediaStrategy` that handles the host's
+  full media surface while it's active. Inherits the "do the native thing"
+  behavior for accessors it doesn't override.
+
+The host owns **one active strategy at a time** (`host.strategy`). All accessors and
```
cjpillsbury: issue(blocking): This heavily constrains what is possible and how things need to be written for extensions. Let's say I have an extension for ads and an extension for chromecast. Both will need need to be "active" at different times, and some of those times will be the same time (an ad playing while also initiating remote playback via cast SDK). What would this look like?

internal/design/media-features.md
```diff
@@ -0,0 +1,628 @@
...
+
+```ts
+import { createHlsMedia } from '@videojs/core/dom/media/hls';
+import { GoogleCast } from '@videojs/core/dom/media/google-cast';
+import { MuxData } from '@videojs/core/dom/media/mux';
+
+const media = createHlsMedia();
+media.attach(videoElement);
+
+media.use(new GoogleCast({ receiverApplicationId: '…' }));
```
cjpillsbury: suggestion(non-blocking): Since the factory is just a thin wrapper around the class instantiation, consider only documenting or potentially even only exporting (in npm sense) one. If so, my S1 would be the factory since it works for both classes and other conventions equally.

internal/design/media-features.md
```diff
@@ -0,0 +1,628 @@
...
+```
+
+`createHlsMedia` itself is a thin wrapper around the imperative API:
+
+```ts
+export function createHlsMedia(
+  extensions: ReadonlyArray<Extension<HTMLVideoElementHost<any, any>>> = []
+): HTMLVideoElementHost {
+  const media = new HTMLVideoElementHost();
+  media.use(hls()); // built-in: routes between HlsJsMedia and NativeHlsMedia
```
cjpillsbury: question(blocking): What happens if `media.use(extension)` is invoked e.g. mid-playback?

internal/design/media-features.md
```diff
@@ -0,0 +1,628 @@
...
+```
+
+Every accessor and method on the host (`paused`, `currentTime`, `play`, …) is
+a one-liner that forwards to the currently active strategy.
+
+| Member | Purpose |
+| - | - |
+| `use(extension)` | Register an extension. Returns a disposer that detaches it. Calling `use()` twice with the same `id` is a no-op. |
+| `route(strategy)` | Route the host's media surface to a strategy. Returns a disposer that restores the previous routing *only if* this strategy is still the active one when called — so an extension never accidentally clears a strategy that another extension has since installed. Pass `null` to clear the routed slot. |
+| `route(strategy, { base: true })` | Same shape, different slot: replaces the host's always-on **base** strategy instead of layering on top of it. Used by extensions that want to extend defaults rather than take over (e.g. cast adds a `remote` accessor to the base so it's available even when no session is connected). The active routed strategy still overlays the base; clearing the base disposer restores the host's built-in default. |
```
cjpillsbury: question(blocking): I think we need more discussion/definition on what "active" means, what replacing strategies means, etc. I don't have a clear sense of what is being added where, what exists (and where) before vs. after invoking `host.route(strategy, { base: true })`. I don't have a clear sense of what an extension that both needs to partially extend/override default behaviors and partially add/decorate additional properties/methods/events and partially add optional configuration-style properties that are only relevant to particular extensions.

internal/design/media-features.md
```diff
@@ -0,0 +1,628 @@
...
+The extension owns the install lifecycle and toggles routing:
+
+```ts
+export class GoogleCast implements Extension<HTMLMediaElementHost<HTMLMediaElement, any>> {
+  readonly id = 'google-cast';
+
+  constructor(private options: GoogleCastOptions) {}
+
+  install(host: HTMLMediaElementHost<HTMLMediaElement, any>) {
+    const provider = new GoogleCastProvider(host, this.options);
```
cjpillsbury: From reading below, it looks like this is also a "strategy"?

internal/design/media-features.md
```diff
@@ -0,0 +1,628 @@
...
+      provider.destroy();
+    };
+  }
+}
+```
+
+Two distinct routing slots are at work:
+
+- **Base** (`{ base: true }`) — the always-on fallback. The cast extension
+  uses this to shadow only `remote` while leaving every other accessor
```
cjpillsbury: So what happens if a "base strategy" adds a new property/method?

internal/design/media-features.md
```diff
@@ -0,0 +1,628 @@
...
+
+That's the entire new surface. The base strategy (the direct-to-`target`
+`BaseMediaStrategy`) and the currently routed strategy are both internal
+state, accessed via the protected `strategy` getter.
+
+### Extension
+
+```ts
+interface Extension<HostType = unknown> {
+  readonly id: string;
```
cjpillsbury: nitpick(blocking): Consider calling this `name` or similar. `id` was confusing to me because I wasn't sure what level of uniqueness this was identifying. If it's solely for deduplication (which I'm assuming means unintentional multiple invocations of `host.use(extensionInstanceA)`, we may want to consider solving that elsewhere/in a different way to avoid deeply assuming that all extensions will effectively be "lowercase s singletons" for a given media "host".

cjpillsbury (commented):
I think we'll need to discuss this more. A high level callout: none of the examples show what properties/methods/events live where "from the outside". My callout on trying to separate out different aspects of this API proposal would be helpful, at least to me - e.g. extension "registration" type concerns/considerations vs. what a media "looks like" that has been extended vs. extension-specific configuration type concerns/considerations. And there are some things in the background of order effects/assumptions, uniqueness effects/assumptions, "stacking" effects/assumptions that I'd like to drill into here.

internal/design/media-features.md
luwes: > Is this about "on the fly" dynamic loading or more about "composing at the time of usage"?

I think both are concerns I have with mixins, they don't support dynamic loading and every media class has to be prebaked with the mixin compared to the composable pattern which is more flexible.

internal/design/media-features.md
luwes: how do you mean?

internal/design/media-features.md
luwes: good call. there is indeed not a requirement to use the host.use(), it's there just for convenience.
the media API defined by the types is the real requirement, how you get there is up to you.

only when you start using our built extensions you'll have to be required to extend from our Host class so the extensions can hook in.

internal/design/media-features.md
luwes: hmm, I don't think that should be the case. While your casting and an ad plays it will be handled by the casting strategy which should support ads if needed because the ad is playing in the receiver. The events and controlling ads etc should be from the receiver at that time https://developers.google.com/cast/docs/web_receiver/ad_breaks

internal/design/media-features.md
luwes: that's up to us :)

internal/design/media-features.md
luwes: not much yet because it can't be reached from the host interface. 
it's only useful to override media API props currently.

internal/design/media-features.md
luwes: yes, it is. I wasn't sure to rename it or keep what we have currently in code. 
It does more than just being a strategy so it felt ok.

internal/design/media-features.md
cjpillsbury: Hmmm not sure that's quite true, but we can get into it in our discussion. TL;DR - Just like this convention, the mixin could be applied later and its source could be loaded lazily, as long as this is all done before using the resultant (now mixed) media class definition.

internal/design/media-features.md
cjpillsbury: A mixin can still add sub-objects. I think you have a mental model of some of these things that are not fully clear to me from this documentation, so I'm plausibly missing something that you have in mind, but let's pretend TextTracks are an extension for the sake of discussion (they very well could be for folks using WebCodecs + Canvas instead of HTMLMediaElement). The text-track-specific API is "bucketed" to a narrow extension of Media API - `textTracks` and `addTextTrack()`, and, in a dream world future/a media extension we may want to support, `removeTextTrack()`. Those could very easily be added via a mixin, but the bulk of the "extended api" "lives in" `textTracks` itself. This same point could hold true for other ways of "bucketing"/"namespacing" things.

internal/design/media-features.md
cjpillsbury: Might be a bad example (would have to consider the "cast with ads" vs. "cast without ads" code splitting or bundling, etc. etc.). My larger concern is that this idea of a single active strategy at a time could very plausibly create a lot of pain. I'm happy to brainstorm on other examples, or go down the list of "plugins"/"extensions" from VJS v8 and pressure test this, but I'm definitely worried about the general idea of cross-cutting extensions.

internal/design/media-features.md
cjpillsbury: I think my point is that this document completely under-defines expectations here, but most of these examples would probably be especially complex if extensions are added/removed later (e.g. mid-playback). Couple this with my questions at the top ("what problems/use cases are we actually trying to solve"), and it's not obvious to me that we'd want extensions added/removed dynamically and without any constraints. I'm trying to understand the architecture in relation to what problems we're trying to solve with it or useful possibilities we're trying to unlock with it.

internal/design/media-features.md
luwes: Yes, technically it is possible but if you need to wait until the resultant media class definition that defeats the purpose. There couldn't be any basic playback until that resultant media class is loaded unless you first make an instance without mixins, then later one with mixins and swap out the whole instance which has issues with keeping state between the two.

sampotts:
It looks good to me but my knowledge of this area is limited at best. The only question I had was in relation to the actual extensions; I'm assuming the intention here that all these extensions are internal? i.e. there's no way to pass in custom extensions via `createPlayer`? Is this something we should consider in this design or something for later? I realize it's a can of worms in terms of how we support UI/skin extension so I get it if we want to address it later.

internal/design/media-features.md
luwes: This is fixed and supported in the new design 🎉

luwes:
> It looks good to me but my knowledge of this area is limited at best. The only question I had was in relation to the actual extensions; I'm assuming the intention here that all these extensions are internal? i.e. there's no way to pass in custom extensions via `createPlayer`? Is this something we should consider in this design or something for later? I realize it's a can of worms in terms of how we support UI/skin extension so I get it if we want to address it later.

custom extensions are def possible and with this design it's even more possible to have them be added via createPlayer because they support runtime adding / removing. We'll have to think about this more though because UI/store is decoupled enough that this might be a bad practice.

PR #1469, opened 2026-04-25
refactor(core)!: unify fullscreen and pip on media capabilities

luwes:
## Summary

Lift fullscreen and picture-in-picture state onto the media capability interfaces so the host (rather than the presentation utilities) owns "am I in this mode?", and collapse the WebKit branches onto a single `webkitSetPresentationMode` path.

## Changes

- `MediaFullscreenCapability` gains `isFullscreen` + `exitFullscreen`; `MediaPictureInPictureCapability` gains `isPictureInPicture` + `exitPictureInPicture`. `requestPictureInPicture` is now `Promise<void>`.
- `HTMLVideoElementHost` implements the new getters/methods, deriving state from `document.fullscreenElement` / `document.pictureInPictureElement` and `webkitPresentationMode`.
- Drop `webkitDisplayingFullscreen`, `webkitEnterFullscreen`, `webkitExitFullscreen` from the WebKit type and all call sites — `webkitSetPresentationMode('fullscreen' | 'inline' | 'picture-in-picture')` covers everything.
- Rename `isFullscreenElement` → `isFullscreen` and `isPictureInPictureElement` → `isPictureInPicture`; both now consult the capability getters as a final fallback so non-`HTMLMediaElement` hosts work.
- Detect fullscreen via the `:fullscreen` pseudo-class so requests issued on a descendant inside a shadow tree (e.g. native controls fullscreening the inner `<video>`) are still recognized.
- `exitFullscreen` now requires a `media` argument so the WebKit and capability fallbacks have a target.

<details>
<summary>Why</summary>

Presentation utilities previously assumed `media` was an `HTMLMediaElement` and reached for vendor APIs directly. Pushing the "is this thing in fullscreen / PiP" decision into the host lets custom media targets (mse-shim, react-native, etc.) participate without the utilities special-casing them, and removes the iOS-vs-everything-else fork inside `requestFullscreen` / `exitFullscreen`.

</details>

## Testing

- New `HTMLVideoElementHost` unit tests cover `isFullscreen` / `isPictureInPicture` across DOM and WebKit paths.
- Existing fullscreen and pip feature tests updated for the renamed helpers, the new shadow-tree/`:fullscreen` path, and the host-driven flow.
- `pnpm -F @videojs/core test`

---

> [!NOTE]
> **Medium Risk**
> Medium risk because fullscreen/PiP detection and enter/exit behavior is refactored across multiple presentation utilities and player features, with new capability requirements and Safari/WebKit path changes that could affect edge-case platform behavior.
> 
> **Overview**
> Refactors fullscreen and picture-in-picture to be *capability-driven*: `MediaFullscreenCapability`/`MediaPictureInPictureCapability` now expose `isFullscreen`/`isPictureInPicture` plus `exitFullscreen`/`exitPictureInPicture`, and `HTMLVideoElementHost` implements these by checking `document.*Element` and WebKit `webkitPresentationMode`.
> 
> Presentation helpers are updated to use the new capabilities as a fallback for non-`HTMLMediaElement` media, and Safari handling is collapsed onto `webkitSetPresentationMode` (removing `webkitEnterFullscreen`/`webkitExitFullscreen`/`webkitDisplayingFullscreen`). Fullscreen detection is also expanded to use the `:fullscreen` pseudo-class to handle shadow DOM/descendant fullscreen scenarios.
> 
> Player store features (`fullscreen`, `pip`, `remote-playback`) are updated to the renamed helpers (`isFullscreen`, `isPictureInPicture`) and to pass `media` into `exitFullscreen`, with extensive new/updated tests covering host-based media and the new detection paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4219547a2dd868a460a7b668c66051b748e702d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/core/media/types.ts
+++ b/packages/core/src/core/media/types.ts
@@ -174,11 +174,15 @@ export interface MediaTextTrackCapability {
 }
 
 export interface MediaFullscreenCapability {
+  readonly isFullscreen: boolean;
   requestFullscreen(): Promise<void>;
+  exitFullscreen(): Promise<void>;
 }
 
 export interface MediaPictureInPictureCapability {
-  requestPictureInPicture(): Promise<unknown>;
+  readonly isPictureInPicture: boolean;
+  requestPictureInPicture(): Promise<void>;
+  exitPictureInPicture(): Promise<void>;
 }
 
 export interface MediaRemotePlaybackCapability {
--- a/packages/core/src/core/media/types.ts
+++ b/packages/core/src/core/media/types.ts
@@ -175,14 +175,14 @@ export interface MediaTextTrackCapability {
 
 export interface MediaFullscreenCapability {
   readonly isFullscreen: boolean;
-  requestFullscreen(): Promise<void>;
-  exitFullscreen(): Promise<void>;
+  requestFullscreen(): Promise<unknown>;
+  exitFullscreen(): Promise<unknown>;
 }
 
 export interface MediaPictureInPictureCapability {
   readonly isPictureInPicture: boolean;
-  requestPictureInPicture(): Promise<void>;
-  exitPictureInPicture(): Promise<void>;
+  requestPictureInPicture(): Promise<unknown>;
+  exitPictureInPicture(): Promise<unknown>;
 }
 
 export interface MediaRemotePlaybackCapability {
--- a/packages/core/src/dom/media/video-host.ts
+++ b/packages/core/src/dom/media/video-host.ts
@@ -1,3 +1,4 @@
+import { isFunction } from '@videojs/utils/predicate';
 import type { Video, VideoEvents } from '../../core/media/types';
 import type { WebKitDocument, WebKitPresentationMode, WebKitVideoElement } from '../presentation/types';
 import { HTMLMediaElementHost } from './media-host';
@@ -18,39 +19,40 @@ export class HTMLVideoElementHost extends HTMLMediaElementHost<HTMLVideoElement,
   get webkitSetPresentationMode(): ((mode: WebKitPresentationMode) => void) | undefined {
     const target = this.target as unknown as WebKitVideoElement | null;
     const fn = target?.webkitSetPresentationMode;
-    return typeof fn === 'function' ? fn.bind(target) : undefined;
+    return isFunction(fn) ? fn.bind(target) : undefined;
   }
 
   get isPictureInPicture(): boolean {
     return (
-      (!!this.target && document.pictureInPictureElement === this.target) ||
+      (!!this.target && globalThis.document?.pictureInPictureElement === this.target) ||
       this.webkitPresentationMode === 'picture-in-picture'
     );
   }
 
   get isFullscreen(): boolean {
-    if (this.webkitPresentationMode === 'fullscreen') return true;
     if (!this.target) return false;
-    const doc = document as WebKitDocument;
-    return doc.fullscreenElement === this.target || doc.webkitFullscreenElement === this.target;
+    if (this.webkitPresentationMode === 'fullscreen') return true;
+    const doc = globalThis.document as WebKitDocument;
+    return doc?.fullscreenElement === this.target || doc?.webkitFullscreenElement === this.target;
   }
 
-  async requestPictureInPicture(): Promise<void> {
+  async requestPictureInPicture() {
     if (!this.target) return Promise.reject();
-    await this.target.requestPictureInPicture();
+    return this.target.requestPictureInPicture();
   }
 
-  async exitPictureInPicture(): Promise<void> {
-    if (document.pictureInPictureElement === this.target) {
-      await document.exitPictureInPicture();
-    }
+  async exitPictureInPicture() {
+    if (!this.target) return Promise.reject();
+    return globalThis.document?.exitPictureInPicture();
   }
 
-  requestFullscreen(): Promise<void> {
-    return this.target?.requestFullscreen() ?? Promise.reject();
+  requestFullscreen() {
+    if (!this.target) return Promise.reject();
+    return this.target.requestFullscreen();
   }
 
-  exitFullscreen(): Promise<void> {
-    return document.exitFullscreen();
+  exitFullscreen() {
+    if (!this.target) return Promise.reject();
+    return globalThis.document?.exitFullscreen();
   }
 }
--- a/packages/core/src/dom/media/video-host.ts
+++ b/packages/core/src/dom/media/video-host.ts
@@ -1,5 +1,5 @@
 import type { Video, VideoEvents } from '../../core/media/types';
-import type { WebKitPresentationMode, WebKitVideoElement } from '../presentation/types';
+import type { WebKitDocument, WebKitPresentationMode, WebKitVideoElement } from '../presentation/types';
 import { HTMLMediaElementHost } from './media-host';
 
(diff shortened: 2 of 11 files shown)


packages/core/src/dom/presentation/fullscreen.ts
```diff
@@ -89,4 +96,11 @@ export async function exitFullscreen(media?: EventTarget): Promise<void> {
   if (isFunction(doc.webkitExitFullscreen)) {
     return doc.webkitExitFullscreen();
   }
+
+  const video = media as unknown as MediaFullscreenCapability;
+  if (isFunction(video.exitFullscreen)) {
+    return video.exitFullscreen();
+  }
+
+  throw new DOMException('Fullscreen not supported', 'NotSupportedError');
```
mihar-22: Does exiting fullscreen natively no-op when not supported or throw an error? I believe it's no-op. Just wondering what would make sense for us to do. I like that error is clear it's not supported, but is it necessary?

packages/core/src/dom/media/video-host.ts
```diff
@@ -11,31 +11,46 @@ export class HTMLVideoElementHost extends HTMLMediaElementHost<HTMLVideoElement,
...
+    const target = this.target as unknown as WebKitVideoElement | null;
+    const fn = target?.webkitSetPresentationMode;
+    return typeof fn === 'function' ? fn.bind(target) : undefined;
   }
 
-  requestPictureInPicture() {
-    return this.target?.requestPictureInPicture() ?? Promise.reject();
+  get isPictureInPicture(): boolean {
+    return (
+      (!!this.target && document.pictureInPictureElement === this.target) ||
```
mihar-22: I think this can just be `document.pictureInPictureElement === this.target`?

packages/core/src/core/media/types.ts
```diff
@@ -174,11 +174,15 @@ export interface MediaTextTrackCapability {
 }
 
 export interface MediaFullscreenCapability {
+  readonly isFullscreen: boolean;
   requestFullscreen(): Promise<void>;
+  exitFullscreen(): Promise<void>;
```
mihar-22: prefer if the return types of request/exit was `unknown` to keep it flexible incase some Media decides to return something

packages/core/src/core/media/types.ts
```diff
@@ -174,11 +174,15 @@ export interface MediaTextTrackCapability {
...
 }
 
 export interface MediaPictureInPictureCapability {
-  requestPictureInPicture(): Promise<unknown>;
+  readonly isPictureInPicture: boolean;
+  requestPictureInPicture(): Promise<void>;
+  exitPictureInPicture(): Promise<void>;
```
mihar-22: prefer if the return types were `unknown` for same reasons as fullscreen, example could be `PictureInPictureWindow` that's natively returned

packages/core/src/dom/media/video-host.ts
```diff
@@ -11,31 +11,46 @@ export class HTMLVideoElementHost extends HTMLMediaElementHost<HTMLVideoElement,
...
+      this.webkitPresentationMode === 'picture-in-picture'
+    );
   }
 
-  requestFullscreen() {
-    return this.target?.requestFullscreen() ?? Promise.reject();
+  get isFullscreen(): boolean {
+    if (this.webkitPresentationMode === 'fullscreen') return true;
+    if (!this.target) return false;
```
mihar-22: we should probably flip this, if there's no `target` then it can't be fullscreen no matter what I assume

packages/core/src/dom/media/video-host.ts
```diff
@@ -11,31 +11,46 @@ export class HTMLVideoElementHost extends HTMLMediaElementHost<HTMLVideoElement,
...
+    const doc = document as WebKitDocument;
+    return doc.fullscreenElement === this.target || doc.webkitFullscreenElement === this.target;
   }
 
-  webkitEnterFullscreen() {
-    return (this.target as unknown as WebKitVideoElement | null)?.webkitEnterFullscreen?.();
+  async requestPictureInPicture(): Promise<void> {
+    if (!this.target) return Promise.reject();
+    await this.target.requestPictureInPicture();
   }
```
mihar-22: Let's change return type to `Promise<PictureInPictureWindow>` and `return this.target.requestPictureInPIcture()`?

Setting `unknown` on base media interfaces will allow it.

packages/core/src/dom/media/video-host.ts
```diff
@@ -11,31 +11,46 @@ export class HTMLVideoElementHost extends HTMLMediaElementHost<HTMLVideoElement,
...
 
-  webkitExitFullscreen() {
-    return (this.target as unknown as WebKitVideoElement | null)?.webkitExitFullscreen?.();
+  async exitPictureInPicture(): Promise<void> {
+    if (document.pictureInPictureElement === this.target) {
+      await document.exitPictureInPicture();
+    }
+  }
```
mihar-22: This does change the behaviour from native slightly since it will throw if [`document.pictureInPictureElement` is `null`](https://developer.mozilla.org/en-US/docs/Web/API/Document/exitPictureInPicture#exceptions) and this guards against it. Not sure how faithful we want to be.

packages/core/src/dom/media/video-host.ts
```diff
@@ -11,31 +11,46 @@ export class HTMLVideoElementHost extends HTMLMediaElementHost<HTMLVideoElement,
...
+  get webkitPresentationMode() {
+    return (this.target as WebKitVideoElement | null)?.webkitPresentationMode;
   }
 
-  get webkitPresentationMode() {
-    return (this.target as unknown as WebKitVideoElement | null)?.webkitPresentationMode ?? 'inline';
+  get webkitSetPresentationMode(): ((mode: WebKitPresentationMode) => void) | undefined {
+    const target = this.target as unknown as WebKitVideoElement | null;
+    const fn = target?.webkitSetPresentationMode;
+    return typeof fn === 'function' ? fn.bind(target) : undefined;
```
mihar-22: cool if we can use `isFunction` utility

packages/core/src/dom/presentation/pip.ts
```diff
@@ -18,56 +13,52 @@ export function isPictureInPictureEnabled(): boolean {
...
   if (isFunction(document.exitPictureInPicture)) {
     return document.exitPictureInPicture();
   }
+
+  const video = media as unknown as MediaPictureInPictureCapability;
+  if (isFunction(video.exitPictureInPicture)) {
+    return video.exitPictureInPicture();
+  }
+
+  throw new DOMException('Picture-in-Picture not supported', 'NotSupportedError');
```
mihar-22: Same question as fullscreen, should this throw?

mihar-22 (dismissed):
Perfect! Thank you for doing this ❤️ 

Small recommendations and questions below.

packages/core/src/dom/presentation/fullscreen.ts
luwes: yes fair point, let's remove it for a smaller bundle

packages/core/src/dom/media/video-host.ts
luwes: what if they're both null?

packages/core/src/core/media/types.ts
luwes: I'll try but got some type issues before with unknown.

PR #1489, opened 2026-04-28
fix(icons): use icon exports in ejected skins

### LINKED ISSUE #1478: Feature: Refactor Ejected Skins to Remove Inlined SVGs
Currently we inline the `<svg>`s for icon in the ejected skins but that makes them noisy. Instead we should re-export the icons via `@videojs/html` and `@videojs/react` and use `media-icon` and `<name>Icon` respectively.

sampotts:
## Summary
- Adds public icon re-export entrypoints through `@videojs/html/icons`, `@videojs/html/icons/*`, `@videojs/react/icons`, and `@videojs/react/icons/*` so generated/ejected code no longer imports from private `@videojs/icons/*` paths.
- Generates ejected HTML skins with `<media-icon>` elements instead of inline SVG markup, while React ejected skins keep component icon imports from public `@videojs/react/icons` paths.
- Splits HTML SVG string exports from media-icon registration: `@videojs/html/icons` is export-only, and `@videojs/html/icons/element` is the side-effectful custom-element registration path.
- Adds default and minimal media-icon element entrypoints so consumers and CDN skins can eagerly register one icon family when that is preferable.
- Updates the private `@videojs/icons` element build to support lazy family loading through loaders, pending-load de-duping, instance re-rendering, and SSR-safe custom-element registration guards.
- Bundles the relevant icon family into the CDN UI skin entrypoints instead of loading icon definitions from a separate CDN file.
- Keeps React package skin source on local `@/icons` imports and rewrites those imports to public `@videojs/react/icons` paths when ejected snippets are generated.
- Updates sandbox and e2e generated pages to use the new public icon entrypoints.
- Updates package side-effect metadata and tsdown `moduleSideEffects` so `define/*` modules and `icons/(dist/)?element/*` registration modules are preserved while plain icon string exports remain tree-shakeable.
- Updates icon package docs and adds focused HTML icon tests covering export-only imports, lazy media-icon loading, unknown families, missing icons, and family-specific registration.
- Keeps the site Vite optimizer from pre-bundling `@resvg/resvg-js`, which preserves the existing server-only OG image route behavior during this workspace update.

Closes #1478

## Validation
- `pnpm -F @videojs/icons build`
- `pnpm -F @videojs/html test src/icons/tests/index.test.ts`
- `pnpm -F @videojs/html build`
- `pnpm -F @videojs/html build:cdn`
- `pnpm -F site ejected-skins`
- `pnpm typecheck`
- `pnpm check:workspace`
- `pnpm lint:fix:file <changed files>`
- `git diff --check`

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 4aa342ec489038d8b64ca1249fec442abfc59579. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/e2e/scripts/generate-pages.ts
+++ b/apps/e2e/scripts/generate-pages.ts
@@ -296,7 +296,8 @@ function ejectedHtmlPage(): string {
   // Path from pages/ to the site content
   const jsonPath = '../../../../../../site/src/content/ejected-skins.json';
 
-  return `import '@videojs/html/video/ui';
+  return `import '@videojs/html/icons/element';
+import '@videojs/html/video/ui';
 import ejectedSkins from '${jsonPath}';
 
 interface EjectedSkinEntry {
--- a/apps/sandbox/templates/simple-hls-html/main.ts
+++ b/apps/sandbox/templates/simple-hls-html/main.ts
@@ -14,7 +14,7 @@ import '@videojs/html/media/container';
 import '@videojs/html/media/simple-hls-video';
 import '@videojs/html/ui/play-button';
 import '@videojs/html/ui/mute-button';
-import { pauseIcon, playIcon, restartIcon, volumeHighIcon, volumeOffIcon } from '@videojs/icons/html';
+import { pauseIcon, playIcon, restartIcon, volumeHighIcon, volumeOffIcon } from '@videojs/html/icons';
 
 const html = String.raw;
 
--- a/apps/sandbox/templates/simple-hls-react/main.tsx
+++ b/apps/sandbox/templates/simple-hls-react/main.tsx
@@ -1,4 +1,5 @@
 import '@app/styles.css';
+
 // SimpleHlsVideo + Video.js integration sandbox — React
 // http://localhost:5173/simple-hls-react/
 //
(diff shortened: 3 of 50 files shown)


mihar-22 (commented):
Awesome, few missing pieces:

1. The `media-icon` element that's built would need to support lazy loading of families/collections. It currently just loads all icons for both default/minimal. Screenshot below.
2. I'd probably remove loading icons from the CDN directly for now. We can figure this out at a later time when we're more confident with how we're packaging/shipping them.

<details>
  <summary>See screenshot</summary>
  
  Image
  
</details>

mihar-22 (approved):
That all seems to look good to me! I guess only question which we briefly touched on in Slack is whether we want to use the `media-icon` in the HTML skins to be consistent with eject (not a blocker).

sampotts:
> That all seems to look good to me! I guess only question which we briefly touched on in Slack is whether we want to use the `media-icon` in the HTML skins to be consistent with eject (not a blocker).

Happy to do that if we prefer? I can do it in a follow-up PR.

PR #1523, opened 2026-05-06
fix(html): set min dimensions on background-video host

### LINKED ISSUE #1497: Background Video Installation Snippet Renders Empty Without a Sized Parent
The HTML installation snippet for the `background-video` preset (generated on `/docs/framework/html/how-to/installation`) does not render visibly when copy-pasted into an empty page.

`background-video-player` is `display: contents` and `background-video-skin` is `width: 100%; height: 100%`, so the rendered size is dictated by the skin's containing block. Pasting the snippet into a fresh `<body>` (default `height: auto`) collapses the skin to 0×0. The other presets hide this because `<video>` has intrinsic dimensions; the background renderer suppresses that, so the parent must size it.

A friction-log copy-paste should produce a visible result. Three possible approaches, ordered lightest to heaviest:

1. **Docs-only note next to the snippet.** Add a short aside when `preset=background-video` explaining that an ancestor must be sized, with two example layouts (full-bleed `position: fixed; inset: 0` and aspect-ratio container).
2. **Make the generated snippet self-contained.** Wrap the markup in a sized container (e.g. `<div style="position: fixed; inset: 0">`) inside `site/src/utils/installation/codegen.ts`. Solves copy-paste but commits the docs to one layout (full-bleed) which may not match the user's intent (inline hero, card, etc.).
3. **Ship a sensible default in `background/skin.css`.** E.g. give the player `display: block; position: relative; width: 100%; aspect-ratio: 16/9`. Most invasive — a breaking change for existing consumers who already wrap with their own sizing.

Looking for opinions on which direction to take. My current lean is (1) plus a small "two layouts" snippet block, since (2) over-commits the example and (3) is a behavior change.

## Context

- [Installation guide](https://videojs.org/docs/framework/html/how-to/installation) - Select HTML and Background video.
- [installation.mdx](https://github.com/videojs/v10/blob/main/site/src/content/docs/how-to/installation.mdx)
- [codegen.ts](https://github.com/videojs/v10/blob/main/site/src/utils/installation/codegen.ts) — `generateHTMLMarkup`
- [background/skin.css](https://github.com/videojs/v10/blob/main/packages/html/src/define/background/skin.css)

spuppo-mux:
Refs #1497

## Summary

Give `background-video-skin` a default `min-width: 300px` / `min-height: 150px` so the installation snippet renders visibly when copy-pasted into an empty page. Mirrors the CSS spec's intrinsic 300×150 default for replaced elements with no natural dimensions.

## Changes

- Add `min-width: 300px; min-height: 150px` to `background-video-skin` in `packages/html/src/define/background/skin.css`. The skin is the in-flow element (`display: block; width: 100%; height: 100%`), so the floor reserves layout space when the parent is unsized.
- The previous iteration of this PR put the floor on `:host` of `<background-video>`, but the skin absolutely-positions that element (`position: absolute; inset: 0`), so the min-* did not reserve space and could leave the parent collapsed while the video overflowed. Moving it onto the in-flow skin element addresses that review comment.

## Testing

No existing snapshot covers `background-video-skin`. Manual: pasting the generated install snippet into a fresh page now renders the player at 300×150 instead of collapsing to 0×0. Adding a `background-skin.spec.ts` visual test is a reasonable follow-up — happy to fold one into this PR if preferred.

## Risk

Low. CSS-only; existing consumers can override with `min-width: 0; min-height: 0` or by setting explicit `width` / `height` / `aspect-ratio` on a parent. The behavior change is limited to cases that previously rendered as 0×0.

### DIFF
--- a/packages/html/src/define/background/skin.css
+++ b/packages/html/src/define/background/skin.css
@@ -7,7 +7,9 @@ background-video-skin {
   position: relative;
   display: block;
   width: 100%;
+  min-width: 300px;
   height: 100%;
+  min-height: 150px;
   object-fit: var(--media-object-fit);
 }
 
--- a/packages/html/src/media/background-video/index.ts
+++ b/packages/html/src/media/background-video/index.ts
@@ -7,8 +7,6 @@ function getTemplateHTML(attrs: Record<string, string>) {
     <style>
       :host {
         position: relative;
-        min-width: 300px;
-        min-height: 150px;
       }
 
       video {
--- a/packages/html/src/media/background-video/index.ts
+++ b/packages/html/src/media/background-video/index.ts
@@ -7,6 +7,8 @@ function getTemplateHTML(attrs: Record<string, string>) {
     <style>
       :host {
         position: relative;
+        min-width: 300px;
+        min-height: 150px;
       }
 
       video {


luwes (approved):
👍

PR #1524, opened 2026-05-07
docs(site): add ejected skin section to installation guide

### LINKED ISSUE #1385: Docs: Add Ejected Skins to the Install Guide
Parent sub-epic: #2184

> [!NOTE]

### Which documentation could use some work?

general framework documentation

### Is this a discrepancy or an improvement?

improvement

### 🖊 Describe what you'd like to see change in the documentation

Make the ejected version of the skin available in the install guide, not just on the homepage, or instead of the homepage to remove duplication.

At the same time make "no skin" an option.

### 💡 Describe how you think this change will help the documentation

It'll make it more gooder.

ronald-urbina:
Closes [1385](https://github.com/videojs/v10/issues/1385)

Adds a new "Ejecting a skin" section to the installation guide so users
can access the ejected skin workflow without going back to the homepage.

## Changes
- New `EjectedSkinToggle` component with a default/minimal radio switcher
  that shows the correct ejected-skin install steps per variant
- `installation.mdx` wired up with the toggle and a teaser link near the
  skin picker section

---

> [!NOTE]
> **Medium Risk**
> Touches shared installation codegen used by the site and CLI bundle; incorrect ejected JSON or placeholder resolution would break copy-paste install paths, but behavior is heavily tested and background-video is explicitly guarded.
> 
> **Overview**
> Adds **packaged** vs **ejected** as a first-class installation choice so generated snippets either use pre-built skins or copy ejected skin source into the project.
> 
> The installation guide gets an **embed method** picker (hidden for background-video), and HTML/React code blocks switch to multi-file output (`index.html` / `index.ts` / `skin.css` or `MyPlayer.tsx` + `skin.css`) when ejected is selected. Shared codegen in `codegen.ts` branches on `embedMethod`, loading snippets via new `ejected.ts` from `ejected-skins.json`; background-video always stays packaged.
> 
> The ejected-skins build now emits **placeholder** HTML (`{{SRC}}`, `{{POSTER}}`, `{{MEDIA_TAG}}`) with **separate** `cdnScript` / `cdnStylesheet` fields instead of baking CDN tags into the template. Docs `EjectedSkin.astro` resolves placeholders for display.
> 
> `@videojs/cli docs how-to/installation` gains `--embed-method`, interactive prompting for embed method, and inference of background-video from `--media` without extra prompts. CLI bundles alias `ejected.ts` and formats ejected output consistently with the site.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5dce5fa411064e8e495eb2ae07f982254c67fe23. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


site/src/components/installation/EjectedSkinToggleClient.tsx
```diff
@@ -0,0 +1,31 @@
...
+  { value: 'default', label: 'Default', image: <Sparkles size={32} /> },
+  { value: 'minimal', label: 'Minimal', image: <Minus size={32} /> },
+];
+
+interface Props {
+  children: React.ReactNode;
+}
+
+export default function EjectedSkinToggleClient({ children }: Props) {
+  const [variant, setVariant] = useState<Variant>('default');
```
decepulis: ~imo these should listen to the same nanostore that powers the skin toggle higher up the page.
If you select minimal at the top, you should be viewing the minimal code example at the bottom.~

Not relevant in light of the comment below

decepulis (changes requested):
In addition to the comments I made below, it's important for this guide and `npx @videojs/cli docs how-to/installation` to stay in sync. You should port the changes you made here to that guide as well, so that when agentic readers hit this guide with `--skin minimal`, they're served the minimal ejected skin as well.

~(Though before you spend too much time on that, I'm also starting to wonder if we should link to the eject guide instead of putting eject instructions here... I asked heff about that [here](https://github.com/videojs/v10/issues/1385#issuecomment-4444836702). Sorry for bringing up a design conversation like this up after implementation has already started!)~

Update: comment below

decepulis:
@ronald-urbina we talked about this in a meeting today. We're imagining a slightly different shape for this flow.
https://github.com/videojs/v10/issues/1385#issuecomment-4445614826
let me know if you have any questions or just want to talk this through!

packages/cli/src/commands/docs.ts
```diff
@@ -109,7 +119,8 @@ Installation flags (for docs how-to/installation):
   --skin <default|minimal>
   --source-url <url>
   --media <html5-video|html5-audio|hls|background-video>
-  --install-method <cdn|npm|pnpm|yarn|bun>`;
+  --install-method <cdn|npm|pnpm|yarn|bun>
+  --embed-method <packaged|ejected>`;
```
decepulis, line 123: What happens if a user does `--embed-method ejected --media background-video`? Do we error invalid combos?

packages/cli/src/utils/format.ts
```diff
@@ -1,3 +1,4 @@
+import ejectedSkins from '@/content/ejected-skins.json';
```
decepulis: Why does the CLI need to import the generated `ejected-skins.json` directly? The existing installation CLI path treats `packages/cli/src/utils/format.ts` as a thin markdown wrapper around the shared site installation generators in `site/src/utils/installation/codegen.ts`.

This adds a second path for installation codegen where the CLI knows about generated content IDs, the JSON shape, and CSS-style filtering. Could we move the ejected-skin lookup/generation behind the shared installation codegen boundary instead, so the CLI remains a wrapper and the site/CLI do not diverge on how ejected snippets are selected?

This will probably help clear up the typecheck/test errors we're seeing, too.

site/src/components/installation/EmbedMethodPicker.tsx
decepulis, line 1: Looks great

site/src/content/docs/how-to/installation.mdx
```diff
@@ -43,7 +46,8 @@ npx @videojs/cli docs how-to/installation \
   --skin <default|minimal> \
   --media <html5-video|html5-audio|hls|background-video> \
   --source-url <url> \
-  --install-method <cdn|npm|pnpm|yarn|bun>
+  --install-method <cdn|npm|pnpm|yarn|bun> \
+  --embed-method <packaged|ejected>
```
decepulis, line 48: Once you merge in main, you may have to make a small change to this section here

site/src/content/docs/how-to/installation.mdx
```diff
@@ -120,6 +126,10 @@ Video.js supports a wide range of file types and hosting services. It's easy to
   </TabsRoot>
 </FrameworkCase>
 
+<EmbedMethodSection>
+
+<div data-embed-variant="packaged" data-cli-replace="installation">
```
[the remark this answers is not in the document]
decepulis: Yeah. This is correct. Keep the `data-cli-replace` div separate from `data-embed-variant` (if you choose to keep that after your next round of revisions) and make sure it still enables the experience we want in the CLI!

site/src/content/docs/how-to/installation.mdx
```diff
@@ -120,6 +126,10 @@ Video.js supports a wide range of file types and hosting services. It's easy to
   </TabsRoot>
 </FrameworkCase>
 
+<EmbedMethodSection>
```
decepulis: Hmm I'd like to propose a pretty big UX change.
1. Currently, when you select "Eject", the "create your player [react]" and "use your player" headers disappear. We should bring them back.
2. Currently, when you select "Eject"...
    - in React, the ejected skin currently _replaces_ "create your player" and "use your player". Instead of that, it might be neater if Skin.tsx and skin.css were additional files in "create your player". So, keep the current `./components/player/index.tsx`, but add `./components/player/skin.tsx` and `./components/player/skin.css`. This will require a rethink of how `codegen.ts > generateReactCreateCode` works, so it returns multiple files and the front-end shows those multiple files (tabs) and the cli shows those multiple files one-after-the-other 
        React mockup
    - in HTML, this could be more simple. Just take the ejected skin you have, and replace the code block under "use your player". Again, the logic should live in codegen.ts, not in format.ts

decepulis (changes requested):
Still some work to do!
1. a change to the UI, putting ejected skins in the existing header hierarchy and code blocks rather than replacing it
2. some broken tests and typechecking
3. some suggestions about refactoring; more logic should live in codegen.ts

site/src/content/docs/how-to/installation.mdx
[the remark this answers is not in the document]
ronald-urbina: done

packages/cli/src/commands/docs.ts
ronald-urbina: is ignored as it happens if skin is added into a background-video command

packages/cli/src/utils/format.ts
ronald-urbina: done

site/src/content/docs/how-to/installation.mdx
ronald-urbina: done

packages/cli/src/commands/docs.ts
decepulis: I'd prefer to have an explicit error to help users understand what's going on!

site/src/utils/installation/codegen.ts
decepulis, line 1: I'm worried the string replacement approach you're taking here is brittle. If the generated code changes too much, then these utils will change, too. I'd sugest instead maybe making `build-ejected-skins` output some sort of parameterizable output. Stable strings like `{{ SRC }}` that can be easily replaced wherever it's needed -- EjectedSkin.astro and codegen -- independently

decepulis (changes requested):
This is a lot closer! Thanks for addressing my feedback.

Here's a few more items for this round 🙂

In addition to my one review comment... two more things!

### General comment
Make sure your tests pass before the next round of review!

### UX improvement
Image

This is a bit of an awkward experience. 
1. We should use filenames for all three. `index.html`, `skin.css`, and maybe `index.ts`. 
2. They should be in the same tab container, not two separate ones, at this point. If possible, we should only combine them into the same tab container in the ejected case. In the packaged case, it still makes sense to have them as two separate code blocks

let me know if I can clarify!

site/src/utils/installation/codegen.ts
ronald-urbina: Already done, now the build script outputs stable placeholder tokens and the consumers just swap them in. No more brittle regex.

PR #1534, opened 2026-05-12
docs(site): add autoplay how-to guide

### LINKED ISSUE #1039: Feature: Autoplay with Muted Fallback
Smart autoplay support, a la Media Chrome's approach. Try autoplay unmuted → fall back to muted → expose failure state for fallback UI (e.g., "click to play" button).

## Prior art
- **VJS v8:** [autoplay option](https://legacy.videojs.org/guides/options), [best practices](https://videojs.com/blog/autoplay-best-practices-with-video-js/)
- **Mux Player:** [autoplay guide](https://www.mux.com/docs/guides/web-autoplay-your-videos)
- **Vidstack:** [autoplay API](https://vidstack.io/docs/player/api/autoplay/)
- **Plyr:** [GitHub](https://github.com/sampotts/plyr)
- **Media Chrome:** [docs](https://www.media-chrome.org)

### LINKED ISSUE #1854: Docs: Autoplay How-To Guide
Write a guide explaining browser autoplay policies and how to work around them in Video.js 10: why autoplay gets blocked, how to detect success vs. failure, and how to fall back to muted playback with a "click to play" recovery UI. Roughly modeled on [Mux's autoplay guide](https://www.mux.com/docs/guides/web-autoplay-your-videos).

decepulis:
Adds an autoplay how-to guide (and rounds out some problems Rahim and I saw with `usePlayer` along the way)

Closes #1854 

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime or API behavior impact.
> 
> **Overview**
> Adds a new **How to → Autoplay** guide that explains browser autoplay constraints, recommends `muted`/`playsInline`, and documents calling **`play()`** manually (instead of relying on the `autoplay` attribute) so apps can handle promise rejections— including React via `usePlayer` and HTML via the player store—with a table of common `DOMException` names and notes on default component behavior when playback is blocked.
> 
> The **`usePlayer`** reference page now opens with a small example contrasting direct store access vs selector subscriptions.
> 
> The guide is registered in **`docs.config.ts`** under the How-to section.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e98b91d6121c0c348f5e29fa8a7eaec615140467. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+import { createPlayer } from '@videojs/react';
+import { Video, videoFeatures } from '@videojs/react/video';
+
+const Player = createPlayer({ features: videoFeatures });
+
+function App() {
+  return (
+    <Player.Provider>
+      <Player.Container>
+        <Video src="hero.mp4" autoPlay muted playsInline />
```
mihar-22: I recommend removing the noise from the code snippet and just focusing on line 40. The reader only needs to understand which props are set and where. You could leave the import of `<Video>` just to make it clear where that's coming from.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+A few gates worth knowing about:
+
+- **Chrome's Media Engagement Index** — a per-origin score Chrome updates as viewers interact with media on your site. High scores get unmuted autoplay; fresh origins and incognito sessions do not.
+- **Safari Auto-Play settings** — Safari lets viewers configure autoplay per site. Some sites get audio, some get only muted video, some get nothing.
+- **iOS Low Power Mode** — disables autoplay entirely until the viewer taps to play.
+- **Mobile browsers and embedded WebViews** — autoplay generally requires `playsinline`, and WebViews can apply additional restrictions. See <DocsLink slug="concepts/browser-support">Browser support</DocsLink> for our notes on WebViews and PWAs.
+
+For more, see MDN's [Autoplay guide for media and Web Audio APIs](https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Autoplay).
+
+## The simplest path: muted autoplay
```
mihar-22: I'd consider pulling this section "The simplest path: muted autoplay" up into a "Usage" section (maybe different name?) and then following up with heavy details on how browsers handle autoplay.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+description: How browser autoplay policies work, why they block playback, and how to recover when they do.
+---
+
+import FrameworkCase from '@/components/docs/FrameworkCase.astro';
+import DocsLink from '@/components/docs/DocsLink.astro';
+import Aside from '@/components/Aside.astro';
+
+Browser autoplay policies decide whether your player can start playing on its own, or whether the viewer has to press play first. The rules differ by browser, device, and prior engagement, and they change. This guide explains how those rules work and shows two patterns for handling them: a declarative one for when you don't need to react to failure, and a programmatic one for when you do.
+
+## How browsers gate autoplay
```
mihar-22: Not sure about this title "How browsers gate autoplay"... on [MDN](https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Autoplay#autoplay_and_autoplay_blocking) it's referred to as "Autoplay Blocking" and in the [Chrome Blog](https://developer.chrome.com/blog/autoplay) as "Autoplay Policies". I'm thinking the latter because it's what it technically is about.

The description of the page is using the word "policies" as well.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+
+Every modern browser allows muted autoplay. Most also allow unmuted autoplay once the viewer has interacted with your site, or when they've engaged with your content before. The exact thresholds are browser-specific heuristics, and there's no API to ask upfront whether autoplay will succeed.
+
+A few gates worth knowing about:
```
mihar-22: "A few gates worth knowing about:" -> "A few policies worth knowing about"?

site/src/content/docs/how-to/autoplay.mdx
mihar-22: [Vidstack Example](https://vidstack.io/docs/player/api/autoplay/?styling=default-theme)

site/src/content/docs/how-to/autoplay.mdx
mihar-22, line 1: Think I'm starting to feel the confusion of "Concepts" vs. "How to". This feels like a central "concept". Don't have to do it now but it might be worth discussing whether we should simplify down to just concepts. We also  need to move "Getting Started > Overview" to "Concepts > Introduction".

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
+---
+title: Autoplay videos
```
mihar-22: If this happens to move into "Concepts" it would just be "Autoplay"

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+description: How browser autoplay policies work, why they block playback, and how to recover when they do.
```
mihar-22: Maybe this document is more about how to autoplay in general and configure it, policies/recovery likely lower-level details.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+---
+
+import FrameworkCase from '@/components/docs/FrameworkCase.astro';
+import DocsLink from '@/components/docs/DocsLink.astro';
+import Aside from '@/components/Aside.astro';
+
+Browser autoplay policies decide whether your player can start playing on its own, or whether the viewer has to press play first. The rules differ by browser, device, and prior engagement, and they change. This guide explains how those rules work and shows two patterns for handling them: a declarative one for when you don't need to react to failure, and a programmatic one for when you do.
```
mihar-22: I'm not sure that this should be the intro to this page, it's explaining autoplay policies which is a subsection. I think you can just dive straight into usage.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+
+<FrameworkCase frameworks={["html"]}>
+```html title="index.html"
+<script type="module">
+  import '@videojs/html/video/player';
+</script>
+
+<video-player>
+  <media-container>
+    <video src="hero.mp4" autoplay muted playsinline></video>
```
mihar-22: ditto, same as react snippet comment

site/src/content/docs/how-to/autoplay.mdx
mihar-22: Read of this page is that it's too AI-generated. The overall flow and how it's organized are not landing well.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+<DocsLink slug="reference/use-player">`usePlayer`</DocsLink> with <DocsLink slug="reference/feature-playback">`selectPlayback`</DocsLink> gives you a `play()` action that returns the same promise, and <DocsLink slug="reference/feature-volume">`selectVolume`</DocsLink> lets you toggle mute on rejection. Wrap that up as a hook:
+
+```tsx title="useAutoplay.ts"
+import { selectPlayback, selectVolume, usePlayer } from '@videojs/react';
+import { useEffect, useRef, useState } from 'react';
+
+export type AutoplayStatus = 'playing' | 'muted' | 'blocked';
+
+export function useAutoplay(): AutoplayStatus | undefined {
+  const store = usePlayer();
```
mihar-22: this won't work if it's used top-level outside of the `Provider` component, important detail. 

```tsx
// hook won't work out here - needs store or to be inside child component
<Provider>
```

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+The `autoplay` attribute is a fire-and-forget API: the browser either plays your video or quietly does nothing. If you need to know which happened — for analytics, a retry, or a "click to play" overlay — call `play()` yourself instead. It returns a promise that resolves on success and rejects when autoplay is blocked.
+
+```js
+// The browser-level contract this all builds on:
+video.play().catch(() => {
+  // Autoplay was blocked.
+});
+```
+
+In v10, route the same logic through the player store so it works for whatever media element you're using — a plain `<video>`, an HLS adapter, or any custom media element.
```
mihar-22: Feels like specific detail that could be condensed into a simple snippet that's easier to understand. Very quick example (crude):

```ts
store.play(); // calls play() on current media (e.g., <video>)
```

I think I would merge this comment/detail into the usage code snippet for this section.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+
+<FrameworkCase frameworks={["react"]}>
+For pre-packaged decorative video — autoplay, mute, and loop baked in, with no controls — use the `@videojs/react/background` preset. See <DocsLink slug="concepts/presets">Presets</DocsLink>.
+</FrameworkCase>
+
+<FrameworkCase frameworks={["html"]}>
+For pre-packaged decorative video — autoplay, mute, and loop baked in, with no controls — use the `@videojs/html/background` preset. See <DocsLink slug="concepts/presets">Presets</DocsLink>.
+</FrameworkCase>
+
+## Detect failure and recover
```
mihar-22: There's a relationship between "policies" and "failure", linking up and maybe naturally following would be great.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+<FrameworkCase frameworks={["react"]}>
+<DocsLink slug="reference/use-player">`usePlayer`</DocsLink> with <DocsLink slug="reference/feature-playback">`selectPlayback`</DocsLink> gives you a `play()` action that returns the same promise, and <DocsLink slug="reference/feature-volume">`selectVolume`</DocsLink> lets you toggle mute on rejection. Wrap that up as a hook:
+
+```tsx title="useAutoplay.ts"
+import { selectPlayback, selectVolume, usePlayer } from '@videojs/react';
+import { useEffect, useRef, useState } from 'react';
+
+export type AutoplayStatus = 'playing' | 'muted' | 'blocked';
+
+export function useAutoplay(): AutoplayStatus | undefined {
```
mihar-22: This hook is hard to reason about with the status and nested play calls. It's should be a single play call and single then/catch. 

1. `usePlayer` would come from the `createPlayer` function and would be typed - no need for selectors
2. to be more robust, should wait for playback to be ready before calling play

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+For pre-packaged decorative video — autoplay, mute, and loop baked in, with no controls — use the `@videojs/react/background` preset. See <DocsLink slug="concepts/presets">Presets</DocsLink>.
+</FrameworkCase>
+
+<FrameworkCase frameworks={["html"]}>
+For pre-packaged decorative video — autoplay, mute, and loop baked in, with no controls — use the `@videojs/html/background` preset. See <DocsLink slug="concepts/presets">Presets</DocsLink>.
+</FrameworkCase>
+
+## Detect failure and recover
+
+The `autoplay` attribute is a fire-and-forget API: the browser either plays your video or quietly does nothing. If you need to know which happened — for analytics, a retry, or a "click to play" overlay — call `play()` yourself instead. It returns a promise that resolves on success and rejects when autoplay is blocked.
```
mihar-22: hard to read, em dashes, can be simplified

site/src/content/docs/how-to/autoplay.mdx
mihar-22: Not sure about the word "recover". More specifically about handling autoplay manually and handling failures.

site/src/content/docs/how-to/autoplay.mdx
mihar-22: I see, code example is disconnected from how it's actually used below. Likely someone will copy/paste this and it won't work then open a issue without seeing below.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+In v10, route the same logic through the player store so it works for whatever media element you're using — a plain `<video>`, an HLS adapter, or any custom media element.
+
+<FrameworkCase frameworks={["react"]}>
+<DocsLink slug="reference/use-player">`usePlayer`</DocsLink> with <DocsLink slug="reference/feature-playback">`selectPlayback`</DocsLink> gives you a `play()` action that returns the same promise, and <DocsLink slug="reference/feature-volume">`selectVolume`</DocsLink> lets you toggle mute on rejection. Wrap that up as a hook:
+
+```tsx title="useAutoplay.ts"
+import { selectPlayback, selectVolume, usePlayer } from '@videojs/react';
+import { useEffect, useRef, useState } from 'react';
+
+export type AutoplayStatus = 'playing' | 'muted' | 'blocked';
```
mihar-22: `muted`? maybe it's an object `{ playing: false, muted: true, blocked: false }|`

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+  }, [store]);
+
+  return status;
+}
+```
+
+Call `useAutoplay()` inside any component nested in `<Player.Provider>`, condition your fallback UI on the returned status, and drop the `autoPlay` attribute from `<Video>` — the hook handles `play()` itself.
+
+```tsx title="HeroPlayer.tsx"
+function HeroPlayer() {
```
mihar-22: "HeroPlayer" -> "HeroLayout"?

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+          if (volume && !volume.muted) volume.toggleMuted();
+          playback.play().then(
+            () => setStatus('muted'),
+            () => setStatus('blocked'),
+          );
+        },
+      );
+    };
+    tryNow();
+    return store.subscribe(tryNow);
```
mihar-22: lol, this would be called on every single player store update including time updates. We need to add and track `canPlay` state in the store

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+      },
+    );
+    return true;
+  };
+
+  if (!tryPlay()) {
+    const unsub = player.store.subscribe(() => { if (tryPlay()) unsub(); });
+  }
+</script>
+```
```
mihar-22: This is a CDN-specific example and again complicated. For devs using bundlers, they would have a typed `usePlayer` hook from `createPlayer`. Similar comments to React hook.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+
+Drop the `autoplay` attribute from your media when you do this — combining the declarative attribute with a manual `play()` call leads to duplicate attempts and inconsistent state.
```
mihar-22: Seems like a worthy diff code snippet at the start section rather than the end.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+</FrameworkCase>
+
+<Aside type="note">
+A rejected `play()` promise is a [`DOMException`](https://developer.mozilla.org/en-US/docs/Web/API/DOMException) whose `name` distinguishes the cause. `NotAllowedError` is the autoplay policy blocking playback, `NotSupportedError` is a media-decoding failure, and `AbortError` is an unrelated playback interruption. Most autoplay handling can treat all rejections the same; if you only want to log true autoplay failures, branch on `error.name === 'NotAllowedError'`.
+</Aside>
+
+## Handle blocked playback
```
mihar-22: This title vs. "Detect failure and handle recover" both sound like they're talking about the same thing. Seems like a list of best practises below when inevitably autoplay fails, so maybe just "Best practises" as a H3

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+
+When even muted autoplay fails, autoplay isn't going to happen without the viewer's input. Treat it as expected, not as an error:
```
mihar-22: This line is worded strangely, sentence doesn't flow well

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,203 @@
...
+
+- **Keep the poster visible.** The player isn't loading — it's waiting. Don't show a buffering spinner.
+- **Expose a clear play affordance.** A play button overlaid on the poster is the conventional pattern. The player's own controls already work for this — make sure they render in the blocked state.
+- **Don't retry on every render.** The browser's decision is based on engagement state that won't change without viewer input. Retrying in a loop wastes work and inflates analytics.
+- **Don't log autoplay rejection as a fatal error.** A blocked autoplay is the browser doing its job. If your analytics treats every `play()` rejection as a playback failure, your error rates will spike on first-visit sessions where the policy hasn't been earned yet. Distinguish a `NotAllowedError` on the initial autoplay attempt from real playback errors.
+
+Wire your UI to whichever signal the handler above exposes, and show the fallback overlay only when status reaches `'blocked'`.
```
mihar-22: need to link to section being discussed on this line

mihar-22 (commented):
I've left a lot of my general thoughts going over this doc below. I think in large I'm pointing out how to turn this into a "Concept" document instead of "How to" which feels wordy and focuses more on explaining rather than action.

But part of it for me also feels like worth stepping back and exploring if we can implement a `autoplay` store feature to reduce complexity. I was wrong in the other PR. Ultimately, I think the right approach is we handle autoplay so we can avoid edge cases and give devs an easy way to handle failure. This shouldn't be a consumer's responsibility.

site/src/content/docs/how-to/autoplay.mdx
decepulis: Good call

site/src/content/docs/how-to/autoplay.mdx
decepulis: Makes sense. Folks don't need to know the weeds of the problem if they just want it to work and they're ok with muted.

site/src/content/docs/how-to/autoplay.mdx
decepulis: Like I mentioned in person, yeah, this was a draft pr for a reason! You peeked too early.

We've got more to talk about re: concepts / how-to. But, just to give you my preamble, I've got a fair bit of conviction about that [diataxis-founded](https://diataxis.fr) divide. "Concepts" are the compositional lego blocks you can use anywhere. "How-to" are the manuals on how to assemble those lego blocks into specific builds. 

It's painfully true that a lot of things we want to say can be framed as both. Consider the Mux Docs; in that world, we've historically made _every_ concept into a how-to-framed guide. (Though that approach obviously has its limitations; like, where are you supposed to learn what an "asset" is? We didn't have an answer to that at Mux until we added some [getting started guides](https://www.mux.com/docs/core/mux-fundamentals)).  

In this case...? This guide teaches you foundational autoplay concepts _and_ it takes you to a specific outcome that won't be relevant across the codebase -- just "_how to_ make sure your video autoplays". I think it's in the right place, but I see the gray area.

Like I said. More to talk about. Just sharing where my head's at. I like the divide, but it's tricky.

decepulis:
@mihar-22 skipping the rest of your comments for now. This is still a draft and a _lot_ of that is going to change. If we even pursue this guide; sounds like we're reopening the "fix this in code instead of guides" conversation again.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,75 @@
...
+</FrameworkCase>
+<FrameworkCase frameworks={["html"]}>
+```html
+<video autoplay muted playsInline>
+```
+</FrameworkCase>
+
+## Detect and recover from blocked autoplay
+
+The `autoplay` attribute is fire-and-forget: the browser plays your video or quietly does nothing. To know which happened so you can retry muted, to log the outcome, or to show a "click to play" overlay, call `play()` yourself. It returns a promise that resolves on success and rejects when the browser blocks playback.
```
heff: Second sentence reads a little cumbersome.

suggest:

> The autoplay attribute is fire-and-forget: the browser plays your video or quietly does nothing. **Instead**, call play() so you can retry muted, log the outcome, or show a “click to play” overlay. It returns a promise that resolves on success and rejects when the browser blocks playback.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,75 @@
...
+    // do something if you want to handle or track this error
+  });
+}, [play])
+```
+</FrameworkCase>
+
+<FrameworkCase frameworks={["html"]}>
+```js
+const player = document.querySelector('video-player');
+player.store.state.play().then(() => {
```
heff: note: `player.store.state.play()` feels pretty verbose, and `state.play()` feels a little off. Would love to revisit this at some point. I think it'd be part of the HTML API conversation we didn't make it to last week.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,75 @@
...
+  // autoplay was successful!
+}).catch((error) => {
+  // do something if you want to handle or track this error
+});
+```
+</FrameworkCase>
+
+Drop the `autoplay` attribute from your media when you do this. Combining the declarative attribute with a manual `play()` call leads to duplicate attempts and inconsistent state.
```
heff: suggest: Move this above the code snippet so it doesn't get missed.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+<Aside type="note">
+A rejected `play()` promise is a [`DOMException`](https://developer.mozilla.org/en-US/docs/Web/API/DOMException) whose `name` tells you the cause. Two are standardized: `NotAllowedError`, the autoplay policy blocking playback, and `NotSupportedError`, when the source isn't a supported media format. Some browsers also reject with `AbortError` when an unrelated call interrupts playback. Most handling can treat every rejection the same; to log only true autoplay failures, branch on `error.name === 'NotAllowedError'`.
```
heff: suggest: Make this part of the example code. Check for NotAllowedError specifically.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,75 @@
...
+</Aside>
+
+## Handle the blocked state
+
+When muted autoplay fails too, playback won't start without the viewer. Treat that as expected, not as an error:
+
+- **Keep the poster visible.** The player isn't loading, it's waiting. Don't show a buffering spinner.
```
heff: This isn't the reader's job, it's Video.js's.

suggest: Verify video.js handles this well when using the code above, and then tell the reader that's video.js will do for them.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,75 @@
...
+- **Expose a clear play affordance.** A play button over the poster is the conventional pattern. The player's own controls already do this, so make sure they render in the blocked state.
```
heff: > play affordance

suggest: less fancy wordses. If we had a reading-level target/checker on these docs I'd expect this to fail.

> A play button over the poster is the conventional pattern.

suggest: Also frame this as what the default player skin should do in this case rather than giving them a chore to figure out

note: We should have more to say about background videos specifically, probably in the bg video doc, and understand clearly how the background video media works in each of the problem scenarios.

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,75 @@
...
+- **Don't log a blocked autoplay as a fatal error.** It's the browser doing its job. If your analytics counts every `play()` rejection as a playback failure, your error rates will spike on first visits where the policy hasn't been earned yet. Distinguish a `NotAllowedError` on the initial attempt from real playback errors.
```
heff: I'm not quite sure where this would happen. The error doesn't bubble automatically? But I feel like we should give them a full example code that does this for them.

> Distinguish a `NotAllowedError` on the initial attempt

I don't think the "initial attempt" is special compared to later attempts firing this error?

site/src/content/docs/how-to/autoplay.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+Wire your fallback UI to the status the handler exposes, and show the overlay only once it reaches `'blocked'`.
```
heff: > reaches `'blocked'`

Is this a formal state somewhere or a made up one?

> Wire your fallback UI

This part makes it feel like a half solution. I feel like we could either show them an example UI for this purpose, or assume/verify that the existing player UI handles it fine already when autoplay is blocked, i.e. shows the controls.

heff (changes requested):

site/src/content/docs/how-to/autoplay.mdx
decepulis: Good call. Pared this way back and made clear that our player ui components show the controls for them

site/src/content/docs/how-to/autoplay.mdx
decepulis: iirc the intention here was to discourage people from logging an error if it's just `NotAllowedError`. Instead, they should try again muted or offer a button. Made that more clear.

site/src/content/docs/how-to/autoplay.mdx
decepulis: Love me some affordances. Hate me some jargon. Not sure how that slipped in, thanks.

In the case of background videos... by default the media is `autoplay muted`. Which _should_ work in every scenario... but... what if it doesn't? Hmm...

site/src/content/docs/how-to/autoplay.mdx
decepulis: Done!

site/src/content/docs/how-to/autoplay.mdx
decepulis: Got rid of the aside and reorganized it into something more tolerable, including moving NotAllowedError into the example code

site/src/content/docs/how-to/autoplay.mdx
decepulis: Done

site/src/content/docs/how-to/autoplay.mdx
decepulis: Reworded

decepulis:
Closing in favor of #1945 ("docs: add QA-verified how-to guides for player capabilities"), which adds its own `how-to/autoplay.mdx` following the how-to template that PR introduces, backed by real demo files instead of inline snippets. #1854 is now tracked there.

Two things from this PR that #1945's version doesn't cover, worth folding in over there:

- The `DOMException` name table (`NotAllowedError` / `NotSupportedError` / `AbortError`) for readers who call `play()` themselves and want to distinguish a blocked autoplay from a bad source.
- The warning not to combine the `autoplay` attribute with a manual `play()` call, which leads to duplicate attempts and inconsistent state.

Also still open from @mihar-22's review here: whether autoplay should be a store feature so consumers don't handle blocked playback themselves. That's a player question rather than a docs one, and it outlives this PR.

PR #1556, opened 2026-05-18
fix(core): prevent mobile controls flash on first tap after auto-hide

### LINKED ISSUE #1446: Bug: mobile controls flash and immediately hide on first tap after auto-hide

HTML Example

Android 16; 24129PN74C Build/BP2A.250605.031.A3
Chrome Beta 148.0.7778.28

### Description

On mobile devices, after playback starts and the controls auto-hide, the first tap on the video area only makes the controls flash briefly and disappear immediately.

The second tap works as expected and keeps the controls visible for the normal timeout.

I can reproduce this on the official Video.js homepage demo as well, so it does not seem specific to my integration.

### Reduced test case

https://videojs.org/

### Steps to reproduce

1. Open the demo on a android mobile device
2. Start playback
3. Wait for controls to auto-hide
4. Tap once on the blank video area
5. Controls flash briefly and disappear
6. Tap again
7. Controls now stay visible normally

### Current Behavior

The first tap only causes a brief flash.  
The second tap behaves correctly.

### Expected Behavior

The first tap after auto-hide should restore controls and keep them visible for the normal timeout.

### Errors

No console errors.

v10

R-Delfino95:
Closes #1446

## Summary

On Android Chrome, after controls auto-hide, the first tap on the video area made the controls flash briefly and immediately disappear. The second tap behaved correctly. The bug reoccurred every cycle (interact with any button → wait for auto-hide → tap → flash).

### Root cause

Three things interact:

1. The skin registers a `tap` gesture with `action="toggleControls"` (touch) alongside `doubletap` gestures (`seekStep` ×2, `toggleFullscreen`). Because doubletap bindings exist, `TapRecognizer.handleUp` defers the tap action by `DOUBLETAP_WINDOW = 200 ms` to disambiguate, and the deferred callback re-reads live state at fire time.

2. `<media-container>`'s own `pointerup` listener calls `this.focus({ preventScroll: true })` so keyboard events reach the hotkey coordinator. That call fires `focusin` synchronously.

3. The container's `pointerup` listener is registered in `connectedCallback`, before `controlsFeature.attach` registers its own. So on every `pointerup`, the container's listener runs first.

The result, on the first tap after auto-hide (`controlsVisible=false`):

- `pointerdown` (touch) → records `pointerDownTime`.
- `pointerup` (touch) fires. Container's listener runs first → `this.focus()` → `focusin` fires synchronously.
- The `focusin` handler in `controlsFeature` calls `setActive()` → flips `controlsVisible` to `true`.
- The controls feature's own `pointerup` listener runs next (records the touch timestamp), then the gesture coordinator defers the tap action by 200 ms.
- 200 ms later the deferred callback fires `toggleControls()`, reads `controlsVisible=true`, calls `setInactive()` → controls hide. That's the flash.

The second tap worked because focus was already inside the container, so `this.focus()` was skipped and no synchronous `focusin` fired.

A secondary issue: Android Chrome dispatches a synthetic `mouseleave` after `touchend`, which independently called `setInactive()`.

### Fix

In `packages/core/src/dom/store/features/controls.ts`:

- Track touch timestamps on `pointerdown` as well as `pointerup` (renamed `lastTouchUpAt` → `lastTouchAt`). Recording on `pointerdown` is essential: the container's `pointerup` listener runs before ours, so without an earlier timestamp the guard below would always see `lastTouchAt=0` and not short-circuit.
- Guard `focusin` against synthetic focus within 500 ms of any touch event — mirrors the existing `mouseleave` guard.
- On touch `pointermove`, only refresh the idle timer when already active; don't flip visibility. Prevents the same race if a `pointermove` fires mid-tap on devices that drift slightly during touch.
- Guard `mouseleave` against the synthetic event Android Chrome dispatches after `touchend`.

Keyboard-driven focus still activates controls — `lastTouchAt` is only set on touch events, so non-touch `focusin` (e.g. tabbing into the container) is unaffected.

## Test plan

- [x] `pnpm -F @videojs/core test src/dom/store/features/tests/controls.test.ts` — 34 tests pass, including new coverage for: `focusin` fired between touch `pointerdown` and `pointerup` (the actual race) is ignored, `focusin` after the guard window still activates, synthetic `mouseleave` within 500 ms of touch `pointerup` is ignored, touch `pointermove` does not flip visibility mid-gesture.
- [x] `pnpm typecheck`
- [x] `pnpm -F @videojs/core build`
- [x] `pnpm lint:fix:file` on both touched files
- [x] Manual: Android Chrome — play → wait for auto-hide → tap blank area → controls show and remain for the full idle timeout. Verified across multiple cycles.
- [x] Manual: desktop Chrome — keyboard focus on the container still shows controls (`focusin` guard only suppresses inside the 500 ms touch window).
- [x] Manual: iOS Safari — first tap after auto-hide still shows controls without flash.

---

> [!NOTE]
> **Medium Risk**
> Touches core player input/visibility logic on all platforms, but changes are narrowly scoped to touch settle guards and pointermove behavior with strong test coverage.
> 
> **Overview**
> Fixes the **Android Chrome “first tap flash”** after controls auto-hide by preventing synthetic pointer/focus side effects from fighting the deferred `toggleControls` tap gesture.
> 
> `controlsFeature` now tracks **`lastTouchAt`** on touch `pointerdown` and `pointerup` and treats events within **`TOUCH_SETTLE_DELAY` (500 ms)** as part of an in-progress touch interaction. **`focusin`** and **`mouseleave`** handlers ignore activity during that window so container `focus()` and post-touchend synthetic `mouseleave` no longer call `setActive`/`setInactive` mid-toggle. Touch **`pointermove`** only refreshes the idle timer when already active instead of always calling `setActive`, avoiding visibility flips during a tap.
> 
> Tests add coverage for touch `pointermove`, ignored synthetic `focusin`/`mouseleave`, and keyboard `focusin` after the guard expires.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8096e80d4f41fabe1786e1c61fd75c944add680c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/core/src/dom/store/features/controls.ts
```diff
@@ -117,15 +137,42 @@ export const controlsFeature = definePlayerFeature({
...
     listen(container, 'pointerdown', onPointerDown, { signal });
     listen(container, 'pointerup', onPointerUp, { signal });
     listen(container, 'keyup', setActive, { signal });
-    listen(container, 'focusin', setActive, { signal });
+    listen(
+      container,
+      'focusin',
+      () => {
+        // Ignore focusin from the container's own pointerup focus grab.
+        if (lastTouchAt > 0 && Date.now() - lastTouchAt < 500) return;
```
sampotts: nit(non-blocking): is it worth extracting this into a function given we use the same logic below for `mouseleave`?

sampotts (approved):

packages/core/src/dom/store/features/controls.ts
R-Delfino95: Good catch! I pushed that change, thanks!

PR #1559, opened 2026-05-18
docs(site): add Security concept page

### LINKED ISSUE #966: Docs: Security
Documentation covering Video.js 10 content security and safe defaults.

decepulis:
Closes #966

## Summary

Move Content Security Policy guidance out of the installation page and into a dedicated Security concept page so installation stays focused on setting up Video.js.

## Changes

- Add a Security concept page with the existing CSP requirements and example policy.
- Replace the installation page's CSP section with a link to the Security guide.
- Add Security to the Concepts sidebar for HTML and React documentation.

## Testing

- `env -u SENTRY_AUTH_TOKEN pnpm build:site` (14/14 tasks passed; HTML and React Security routes prerendered)

---

> [!NOTE]
> **Low Risk**
> Documentation-only content move and sidebar link; no runtime or application code changes.
> 
> **Overview**
> **Moves Content Security Policy (CSP) guidance off the installation guide** into a new **Security** concept page (`concepts/security`), so installation stays focused on setup.
> 
> The new page carries the same CSP directives (`media-src`, `img-src`, `connect-src`, `blob:` / `worker-src` for HLS variants, `style-src 'unsafe-inline'`) and example policy, with short notes on CDN `script-src` and tightening broad `https:` sources. The installation page **drops its CSP section** and adds a **See also** link to Security. **`docs.config.ts`** registers **Security** under the Concepts sidebar for HTML and React docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb40a5244ca92f2d33097b1c3e641296e5d44b14. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


decepulis (commented):
Codex review (comment-only): useful structure, but I would not treat this as fully reliable yet. The page hits the right buckets for #966, but I see one material overclaim and a few docs-accuracy issues worth tightening.

1. [P1] The safe-default paragraph overclaims URL handling. The new page says users do not need to scrub URLs before passing them to the player, and the PR body says source URLs are forwarded through `setAttribute`. That is true for later attribute changes, but not for initial custom-media construction: initial attributes are copied into an HTML string and assigned through `shadowRoot.innerHTML`, and `serializeAttributes` does not escape quote characters. A value like `src='x" onerror="...'` becomes a separate `onerror` attribute in the generated inner `<video>`. Either the docs should avoid promising this, or the implementation should escape/DOM-create the initial media element. Relevant code: `packages/core/src/dom/media/custom-media-element/index.ts` around the template/constructor path and `packages/utils/src/dom/attributes.ts` around `serializeAttributes`.

2. [P2] The CSP table blurs how captions are loaded. Browser `<track>` loads are covered by `media-src` (MDN lists `<track>` under `media-src`), while HLS manifests/segments and any JS/fetch/XHR paths need `connect-src`. I would split this so readers do not incorrectly put ordinary caption files only under `connect-src`.

3. [P2] The nonce/hash guidance for `style-src` sounds more actionable than the current implementation supports. The player injects internal `<style>` tags / shadow style fallbacks without an obvious user nonce plumbing path. Direct `element.style.foo = ...` is not blocked by `style-src` per MDN, but `<style>` tags and style attributes are. I would phrase this as "`unsafe-inline` is currently the practical requirement for Video.js injected style tags" rather than implying a user can wire nonce/hash support today.

4. [P3] The SRI/package-manager sentence is directionally right but too broad. Safer wording: SRI is for browser-loaded CDN scripts; package-manager installs rely on lockfile/registry integrity instead.

Overall: useful guide shape, not hallucinated as a whole, but the safe-defaults sentence is the risky part I would not ship as written.

site/src/content/docs/concepts/security.mdx
```diff
@@ -0,0 +1,97 @@
...
+  connect-src 'self' https:;
+  worker-src 'self' blob:;
+  style-src 'self' 'unsafe-inline';
+```
+
+Replace the origins to match where your media, posters, and captions are hosted.
+
+### Inline styles
+
+The player uses inline styles for layout and state transitions. Add `'unsafe-inline'` to `style-src` to allow them, or use a [nonce or hash](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Security-Policy/style-src) if your build can generate them. A nonce is the tightest option: generate one per request, and apply it to both the response header and the player's served styles.
```
sampotts: We don't have any way to facilitate a `nonce` yet. We have an RFC open for it but it's on the back burner for now. I'm not sure if we want to include it just yet?

decepulis: Yeah sorry about that. Claude hallucinated that and I didn't catch it; you and Codex did.

sampotts: All good! It's pretty common for libraries to have it so I get why Claude added it. I think we'll want to add support for it at some point, especially if we want larger, more security-conscious orgs to adopt it - it can be painful to get security approval.

site/src/content/docs/concepts/security.mdx
```diff
@@ -0,0 +1,91 @@
...
+  media-src 'self' https: blob:;
+  connect-src 'self' https:;
+  worker-src 'self' blob:;
+```
+
+Replace the origins to match where your media, posters, and captions are hosted.
+
+## `crossorigin` attribute
+
+When the player loads a video, caption file, or poster from another domain, the browser needs to know whether to send cookies and whether your page is allowed to read the response. The [`crossorigin`](https://developer.mozilla.org/docs/Web/HTML/Reference/Attributes/crossorigin) attribute tells the browser how to handle these requests.
```
luwes: "When the player" more accurately would be "When the html video element"

Because the crossorigin attribute doesn't affect the loading of Hls.js via xhr or fetch, that's handled by that request itself unless Hls.js uses that value in its requests but I'm not aware of that.

site/src/content/docs/concepts/security.mdx
```diff
@@ -0,0 +1,31 @@
...
+Content Security Policy (CSP) limits which resources a page can load. Video.js uses the URLs you configure for media, posters, thumbnails, and captions. Some HLS renderers also create blob URLs or web workers, so your policy must allow those resources too.
+
+## Allow player resources
+
+- `media-src` must allow your media URLs.
+- `img-src` must allow any poster or thumbnail image URLs.
+- `connect-src` must allow HLS manifests, playlists, captions, and segment requests when using HLS playback.
+- `media-src blob:` is required when using the HLS player variants, which use MSE-backed playback.
+- `worker-src blob:` is required when using the `hls.js` player variants.
+- `style-src 'unsafe-inline'` is currently required for some player UI and HTML player styling behavior.
```
sampotts: Should we remove `currently` just to remove the expectation that this might change in future?

sampotts (approved):

sampotts (approved):

site/src/content/docs/concepts/security.mdx
decepulis: looks like claude got it. I didn't even know Claude was watching this thread 😂

PR #1568, opened 2026-05-20
feat(site): hide footer color bars as a scroll-discovered easter egg

decepulis:
## Summary

The footer color bars were always meant to be an easter egg, but they've been sitting visible at the bottom of every page. This PR makes them actually hidden — discovered by reaching the end of a page and pausing, then playfully bounced away the moment a user scrolls into them.

The component splits into two:

- **`ColorBars.tsx`** — pure presentational primitive. Used statically by the blog index header, where the bars are a deliberate design element, not an easter egg.
- **`FooterEasterEgg.tsx`** — React island (`client:idle`). Wraps `ColorBars` with sentinel `IntersectionObserver`s, debounced dwell timers, and a lazy-loaded motion spring for the retreat. Used by every footer placement (home, blog post layout, docs layout, error page, markdown layout / privacy page — a new placement).

The old `FooterEasterEgg.astro` is deleted.

## How it works

1. **Hidden until armed.** On default load, the bars wrapper is `display: none` (`hidden motion-reduce:block`). No bars in layout, none in the a11y tree.
2. **Arm trigger.** When the user fires a real input event (`wheel` / `touchmove` / `keydown` / `pointerdown`) *and* the bottom sentinel intersects the viewport *and* scrolling stops for 200ms — `setShow(true)` toggles the wrapper to `block`. The bars now sit just below the viewport. Programmatic scrolls (e.g., docs scroll-restoration) don't count as input, so the egg doesn't pre-arm on browser-back.
3. **Repeating retreat.** Scrolling into the bars (top sentinel intersects, even one pixel) + 300ms pause → `await import('motion')` + a spring animates `window.scrollY` to `barsTop - innerHeight - 16` (8–16px buffer below viewport, so a 1px rounding wobble can't re-trigger).
4. **User input cancels the spring.** Any `wheel`/`touchmove`/`keydown`/`pointerdown` during the animation aborts it and re-arms.
5. **Reduced motion.** `prefers-reduced-motion: reduce` users get the bars statically via the `motion-reduce:block` CSS variant; the controller short-circuits at init, no listeners attached.

Spring: motion `{ type: 'spring', stiffness: 300, damping: 22, mass: 1 }` — F2 (snappy with light overshoot). All tunables are module-level constants for hand-tuning.

## Test plan

- [ ] On the homepage, scroll all the way to the bottom and pause for ~200ms. The page should silently grow — bars are now reachable below the fold.
- [ ] Scroll down to peek at the bars; pause. The page springs you back so the bars are just off-screen.
- [ ] Do it again — the retreat repeats every time. Bars never let you dwell.
- [ ] Mid-spring, start scrolling (wheel, touch, arrow keys). The spring cancels immediately and respects your input.
- [ ] Same gesture on `/privacy/`, `/blog/`, a docs page, and a 404. Docs uses the `short` (h-25) variant; the rest use default (h-52).
- [ ] Visit `/blog/`. The header (above the post list) still has the small color bars as a static design element — they don't disappear and don't bounce.
- [ ] Enable `prefers-reduced-motion: reduce` in DevTools (Rendering panel). Reload any page with the footer egg. Bars appear statically at the bottom from initial paint; no scroll-driven behavior.
- [ ] Tab through the page with the keyboard. The bars are not focusable; SR users skip the subtree (`aria-hidden="true"`).
- [ ] Page navigation resets the egg — visiting a new page requires re-arming.

---
_Generated by [Claude Code](https://claude.ai/code/session_01H4oU4yFjU1vWwBWsvRLqrF)_


sampotts:
This could be a massive nit pick and it could be just me but I find it a little odd that I reveal the footer, start reading something and then it scrolls itself to hide the footer again. I don't mind the footer being static to be honest - I know some folks think it's too much though.

decepulis:
@sampotts heard. And I don’t think I like this experiment very much either. I’ll play a bit more with the parameters, but, I don’t think I’m keeping it!

PR #1573, opened 2026-05-20
feat(core,html,react): add YouTube provider, refactor IframeMediaHost and fix stale TextTrackList for iframe providers

ronald-urbina:
### Summary

Adds a full YouTube embed provider — `YouTubeMedia` (core), `YouTubeVideo` custom element (html), and `YoutubeVideo` React component (react) — and extracts a shared `IframeMediaHost` base class that both YouTube and Vimeo now extend.

Also fixes a bug where iframe providers created a new `TextTrackList` on every `src` change, leaving stale listeners on the old list.

---

#### New: YouTube provider
- **`packages/core/src/dom/media/youtube/index.ts`** — `YouTubeMedia`: full `MediaEngineHost` implementation over the YouTube IFrame Player API
  - Parses video IDs, watch URLs, Shorts URLs, playlist URLs, and live stream URLs
  - CSS crop trick to hide YouTube title/branding bars when native controls are disabled
  - Polling-based `timeupdate` (250 ms) — the YouTube SDK doesn't emit these natively
  - Seek synthesis with 600 ms fallback for buffered positions
  - Synthetic caption tracks via `getOption('captions', 'tracklist')`
  - YouTube error code → `MediaError` mapping
  - Supported props: `src`, `autoplay`, `controls`, `loop`, `muted`, `playsinline`, `nocookie`, `start`
- **`packages/core/src/dom/media/youtube/api.ts`** — lightweight IFrame API loader with module-level caching and multi-player callback chaining
- **`packages/html/src/media/youtube-video/`** — `YouTubeVideo` custom element (Shadow DOM, observed attributes)
- **`packages/react/src/media/youtube-video/`** — `YoutubeVideo` React component (`useMediaInstance`, `useSyncProps`, `useComposedRefs`)
- **`packages/core/src/dom/media/youtube/tests/youtube-media.test.ts`** — ~690-line test suite

#### Refactor: shared `IframeMediaHost`
- **`packages/core/src/dom/media/iframe-host.ts`** — abstract base class extracted from the Vimeo provider
  - Manages playback state cache (paused, ended, duration, currentTime, buffered, readyState, volume, muted, playbackRate, seeking, pip, error)
  - Overlay div pattern to prevent iframe pointer events from blocking controls
  - PiP support via `requestPictureInPicture` postMessage for Safari
  - Synthetic `TextTrackList` management
  - Lifecycle: `attach` / `detach` / `destroy`
- **`packages/core/src/dom/media/vimeo/index.ts`** — simplified ~320 lines by extending `IframeMediaHost`
- **`packages/html/src/media/vimeo-video/`** — new HTML custom element for Vimeo (parallel to YouTube)

#### Bug fix: stale `TextTrackList` for iframe providers
- **`packages/core/src/dom/store/features/text-track.ts`** — re-subscribes to track list events (`addtrack`, `removetrack`, `change`) on every `loadstart`, using an `AbortController` to clean up the previous subscription before attaching to the new list

#### Sandbox / demos
- `html-youtube-video` and `react-youtube-video` sandbox templates
- `html-vimeo-video` and `react-vimeo-video` sandbox templates (Vimeo was missing React)
- Navbar and source picker updated to include YouTube entries

---

### Architecture notes

- **Embed-only props** (`autoplay`, `controls`, `loop`, `nocookie`, `start`) require an iframe remount on change; runtime props (`muted`, `volume`, `playbackRate`) delegate to the SDK directly
- YouTube's `loop` param requires `playlist` to be set to the same video ID — handled automatically
- `nocookie` defaults to `false` (opt-in) so URLs stay consistent with what users provide
- The CSS crop technique offsets and clips the iframe to hide YouTube's title bar and bottom branding without stretching

### DIFF
--- a/apps/sandbox/app/shell/navbar.tsx
+++ b/apps/sandbox/app/shell/navbar.tsx
@@ -29,6 +29,7 @@ type NavbarProps = {
   isMuxVideo: boolean;
   isMuxAudio: boolean;
   isVimeoVideo: boolean;
+  isYoutubeVideo: boolean;
   platforms: readonly Platform[];
   stylings: readonly Styling[];
   presets: readonly Preset[];
@@ -82,6 +83,7 @@ export function Navbar({
   isMuxVideo,
   isMuxAudio,
   isVimeoVideo,
+  isYoutubeVideo,
   platforms,
   stylings,
   presets,
--- a/apps/sandbox/app/shell/navbar.tsx
+++ b/apps/sandbox/app/shell/navbar.tsx
@@ -52,6 +52,7 @@ const PRESET_LABELS: Record<Preset, string> = {
   'simple-hls-video': 'Simple HLS Video',
   'dash-video': 'DASH Video',
   'vimeo-video': 'Vimeo Video',
+  'youtube-video': 'Youtube Video',
   audio: 'Audio',
   'background-video': 'Background Video',
 };
--- a/packages/core/src/dom/media/youtube/index.ts
+++ b/packages/core/src/dom/media/youtube/index.ts
@@ -1,0 +1,590 @@
+import { IframeMediaHost } from '../iframe-host';
+import { loadYouTubeApi, type YTCaptionTrack, type YTNamespace, type YTPlayer } from './api';
+
+// -- Embed URL helpers --
+
+const EMBED_BASE = 'https://www.youtube.com/embed';
+const EMBED_BASE_NOCOOKIE = 'https://www.youtube-nocookie.com/embed';
+
+const VIDEO_RE =
+  /(?:youtu\.be\/|youtube(?:-nocookie)?\.com\/(?:embed\/|v\/|watch\?v=|watch\?.+&v=|shorts\/|live\/))((\w|-){11})/;
+const PLAYLIST_RE = /(?:youtu\.be\/|youtube(?:-nocookie)?\.com\/.*?[?&]list=)([\w_-]+)/;
+
+// YouTube player state values (mirrors YT.PlayerState).
+const YTState = {
+  UNSTARTED: -1,
+  ENDED: 0,
+  PLAYING: 1,
+  PAUSED: 2,
+  BUFFERING: 3,
+  CUED: 5,
+} as const;
+
+function parseStartTime(src: string): number | undefined {
+  const match = src.match(/[?&]t=([\dms]+)/i);
+  if (!match?.[1]) return undefined;
+  const val = match[1].toLowerCase();
+  let total = 0;
+  let found = false;
+  const m = val.match(/(\d+)m/);
+  if (m?.[1]) {
+    total += parseInt(m[1], 10) * 60;
+    found = true;
+  }
+  const s = val.match(/(\d+)s?$/);
+  if (s?.[1]) {
+    total += parseInt(s[1], 10);
+    found = true;
+  }
+  return found ? total : undefined;
+}
+
+interface EmbedOptions {
+  autoplay: boolean;
+  controls: boolean;
+  loop: boolean;
+  muted: boolean;
+  playsinline: boolean;
+  start: number;
+}
+
+function buildEmbedUrl(src: string, nocookie: boolean, opts: EmbedOptions): string {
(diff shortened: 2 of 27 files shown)


apps/sandbox/app/shell/navbar.tsx
```diff
@@ -52,6 +53,7 @@ const PRESET_LABELS: Record<Preset, string> = {
   'simple-hls-video': 'Simple HLS Video',
   'dash-video': 'DASH Video',
   'vimeo-video': 'Vimeo Video',
+  'youtube-video': 'Youtube Video',
```
sampotts, line 56: nit: You**T**ube Video

packages/skins/src/default/tailwind/video.tailwind.ts
```diff
@@ -36,6 +36,11 @@ export const root = (isShadowDOM: boolean) =>
       '[&_.media-vimeo]:block [&_.media-vimeo]:w-full [&_.media-vimeo]:h-full [&_.media-vimeo]:overflow-hidden [&_.media-vimeo]:rounded-(--media-video-border-radius)':
         !isShadowDOM,
     },
+    // YouTube embed element
+    {
+      '[&_.media-youtube]:block [&_.media-youtube]:w-full [&_.media-youtube]:h-full [&_.media-youtube]:overflow-hidden [&_.media-youtube]:rounded-(--media-video-border-radius)':
```
sampotts, line 41: Can we not apply the classNames directly onto the component (it looks like you're passing className down to the rendered `<div>` where `.media-youtube` is) and share it with the Vimeo media? Maybe an `embeds.ts` - I'll leave naming up to you! It just feels a little odd to use class names like `.media-youtube` in the Tailwind styles and target them this way.

packages/skins/src/minimal/tailwind/video.tailwind.ts
```diff
@@ -33,6 +33,11 @@ export const root = (isShadowDOM: boolean) =>
       '[&_.media-vimeo]:block [&_.media-vimeo]:w-full [&_.media-vimeo]:h-full [&_.media-vimeo]:overflow-hidden [&_.media-vimeo]:rounded-(--media-video-border-radius)':
         !isShadowDOM,
     },
+    // YouTube embed element
+    {
+      '[&_.media-youtube]:block [&_.media-youtube]:w-full [&_.media-youtube]:h-full [&_.media-youtube]:overflow-hidden [&_.media-youtube]:rounded-(--media-video-border-radius)':
```
sampotts, line 38: As above.

packages/core/src/dom/media/youtube/index.ts
```diff
@@ -0,0 +1,590 @@
...
+      muted: this.muted,
+      playsinline: this.#playsinline,
+      start: this.#start,
+    });
+    iframe.allow = 'accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture';
+    iframe.setAttribute('allowfullscreen', '');
+    iframe.style.border = 'none';
+
+    if (!this.#controls) {
+      // CSS crop trick — hides the YouTube title bar (top) and branding bar (bottom)
```
sampotts, line 319: Are we sure we're not breaching YouTube terms of service by doing this? We'll want to check we're not causing issues for users with this. I got a cease and desist from YouTube for hiding their chrome. I appreciate it doesn't look as nice though.

sampotts (commented):
I realized it's draft after diving in so just a few minor comments. Cheers!

sampotts:
I also built the sandbox and noticed a few things:
- The styling doesn't seem to work - I think we're missing `import '@app/styles.css';` in the main.tsx/main.ts for the sandboxes. I had to hack in some styles to get it working.
- PIP doesn't work (tested on Safari 26.5)
- Toggling captions shows a message in the frame related to settings. 
- There's a bit of a clash between our input feedback and the YouTube central controls. Perhaps we need to hide our play/pause feedback? 
- Related to the above but hotkeys seem to flow to YouTube but gestures use ours so you end up with a UI mismatch. Should we disable the YouTube hotkeys and rely on ours only?

PR #1591, opened 2026-05-25
feat(html): i18n

### LINKED ISSUE #1365: Feature: HTML i18n Layer (@videojs/html/i18n)
HTML-facing i18n entry point. Adds `createI18n()` factory returning mixins, controller, and context, plus `<media-i18n-provider>` and `<media-text>` custom elements.

**Blocked by:** #1363
### `createI18n()` factory (`@videojs/html/i18n/create-i18n.ts`)

Returns `{ context, I18nController, ProviderMixin, TextMixin }`, mirroring the `createPlayer` pattern — the factory produces building blocks (mixins + controller + context), and define files compose them onto base classes.

### `ProviderMixin`

- Higher-order function that extends any `ReactiveElement` base
- Adds `lang` attribute (reflected), `ContextProvider` for i18n context
- `#effectiveLocale` getter: explicit `lang` → `parentElement.closest('[lang]')` → `document.documentElement.lang`
- Subscribes to registry changes (`onI18nRegistryChange`) and document-wide `lang` attribute mutations (`MutationObserver`)
- Lazy-loads built-in pack on locale change with BCP 47 fallback
- Triggers `selectCaptionsByLocale` via nearest `[store]` ancestor

### `TextMixin`

- Extends `ReactiveElement` with `key` attribute
- Uses `I18nController` to subscribe to context
- Updates `textContent` on key or locale change

### `I18nController`

- `ReactiveController` consuming the factory's context
- Exposes `.value` returning the active `Translator` (falls back to English when no provider)
- Used by `TextMixin` and `MediaButtonElement`

### Define files

- `@videojs/html/src/define/media-i18n-provider.ts` — `ProviderMixin(ReactiveElement)` + `safeDefine('media-i18n-provider', …)`
- `@videojs/html/src/define/media-text.ts` — `TextMixin(ReactiveElement)` + `safeDefine('media-text', …)`

### Re-exports (`@videojs/html/i18n/index.ts`)

- `registerI18n`, etc. from `@videojs/core/i18n`
- Top-level re-export from `@videojs/html`
- Side-effect import registers both custom elements

## Tests

- `<media-i18n-provider>` explicit `lang` attribute, ambient `<html lang>` inheritance, dynamic updates
- `<media-text key="…">` rendering and live updates on locale change
- `I18nController` fallback translator when no provider in scope

## References

- Design doc: `internal/design/i18n.md` — HTML section, `createI18n` factory, mixins

### LINKED ISSUE #1367: Feature: Wire I18nController into UI Elements and Move Error Dialog Defaults to Core
Two coupled pieces that together finish the "skins are i18n-unaware" goal:

1. Wire `I18nController` into `MediaButtonElement` and `MediaUIElement` so `aria-label` auto-forwards from the i18n registry.
2. Move the error dialog's default English strings (title, description, retry button) out of HTML skins and into core, matching the pattern used for buttons and other core UI components.

**Blocked by:** #1363, #1365, #1366
### `packages/html/src/ui/media-button-element.ts`

- Add `I18nController` instance
- In `update()`: `this.setAttribute('aria-label', t(core.getLabel(state), core.getLabelParams?.(state)))`
- Same wiring for `MediaUIElement` where applicable

### `packages/core/src/core/ui/error-dialog/` (new)

- Move default strings (`errorTitle`, `errorDescription`, `retry`, etc.) from HTML skins into core as opaque translation keys
- Add to `Translations` interface and `en.ts`
- Expose via a core contract the dialog element consumes (e.g. `getErrorDialogContent(state)` or similar)

### `packages/html/src/define/video/skin.ts` / `audio/skin.ts` / `minimal-skin.ts`

- Remove hardcoded English title/description/button strings for the error dialog
- Consume from the core contract, translated via `I18nController`

## Rationale

The error dialog is currently the last place hardcoded English lives inside HTML skins. Moving it into core means skins stay presentation-only — they render components and elements but never own copy, mirroring the approach for buttons, tooltips, etc. This eliminates any need for i18n logic to live in the skin package.

## Tests

- `MediaButtonElement` applies translated `aria-label` and updates on locale change
- Error dialog shows translated strings in Spanish when `<html lang="es">` + `es` pack registered
- Ejected skin templates do not contain hardcoded English for the error dialog

## References

- Design doc: `internal/design/i18n.md` — HTML: `I18nController`, Decisions (Standalone providers — skins are i18n-unaware)

sampotts:
Refs #222
Closes #1365
Closes #1367

## Summary

Adds the HTML i18n layer with ambient `<html lang>` inheritance, browser translation fallback, default `loadLocale`, and translated labels across web component controls.

## Changes

- `createI18n()` — `I18nController`, `ProviderMixin`, `TextMixin`, `<media-i18n-provider>`, `<media-text>`
- Controls auto-forward translated `aria-label`s and tooltips via `I18nController`
- Error dialog uses core i18n helpers; CDN i18n entry point and build plugin
- Package exports for `@videojs/html/i18n` and `@videojs/html/cdn/i18n`

<details>
<summary>Stack info</summary>

**PR 3 of 5** — stacks on locale packs PR.

</details>

## Testing

```bash
pnpm -F @videojs/html build
pnpm -F @videojs/html test src/i18n
```

---

> [!NOTE]
> **Medium Risk**
> Touches many player UI and accessibility strings across skins and controls; async locale/browser loading adds edge cases, though covered by new tests.
> 
> **Overview**
> Adds an HTML **i18n layer** on top of `@videojs/core/i18n`: `createI18n()` exposes Lit context, `I18nController`, provider/text mixins, and registers `<media-i18n-provider>` / `<media-text>`. Locale resolution follows explicit `lang`, ambient `<html lang>`, lazy shipped locale packs, registry overlays, and optional browser translation (with sequence guards on async loads).
> 
> **Players** and **`SkinElement`** are wrapped with `I18nProviderMixin` so skins and controls share a translator. **UI elements** consume `I18nController` and use `resolveControlAttrs` / `resolveControlLabel` for translated `aria-label`s, tooltips, error-dialog copy, time remaining phrases, and input-indicator labels. Skin templates drop hardcoded error-dialog strings in favor of runtime-filled elements.
> 
> **Packaging & CDN:** new `@videojs/html/i18n` and locale export paths, `build:cdn` runs locale generation plus `cdnI18nExternalPlugin` (shared `i18n.js` on jsDelivr in prod, relative in dev), and the video CDN entry pulls in `define/i18n`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85e56ff65408681e0b1244d89fe950e5026fe42d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/html/src/define/audio/minimal-skin.tailwind.ts
```diff
@@ -36,11 +36,11 @@ function getTemplateHTML() {
       <media-error-dialog class="${error.root}">
         <div class="${error.dialog}">
           <div class="${error.content}">
-            <media-alert-dialog-title class="${error.title}">Something went wrong.</media-alert-dialog-title>
+            <media-alert-dialog-title class="${error.title}"></media-alert-dialog-title>
```
cjpillsbury, line 40: thought(non-blocking): If desired/valuable, we could in the future have a "why not both" support that uses the text contents as a fallback if/when i18n providers aren't available.

packages/html/src/define/tests/registration.test.ts
```diff
@@ -134,6 +134,16 @@ describe('composite define registration', () => {
     });
   });
 
+  describe('video/ui', () => {
```
cjpillsbury, line 137: praise: 👌

packages/html/src/define/tests/skin-element.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    context,
+    initialValue: {
+      translator: ((key: string) => (key === 'menuSettings' ? 'Ancestor settings' : key)) as Translator,
+      locale: 'xx' as Locale,
+    },
+  });
+
+  override connectedCallback(): void {
+    super.connectedCallback();
+    void this.#provider;
```
cjpillsbury: question(non-blocking): What is this unyielded read doing? Shouldn't `#provider` get created on instantiation? Not pressing since this is a test, but would be good to clarify via code comment why this line is here or remove if unneeded.

packages/html/src/i18n/tests/provider.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  customElements.define(MediaTextElement.tagName, MediaTextElement);
+}
+
+afterEach(() => {
+  resetI18nRegistryForTesting();
+  document.body.innerHTML = '';
+});
+
+describe('SkinElement', () => {
+  it('uses its own translator for shadow labels', async () => {
```
cjpillsbury: suggestion(non-blocking): Calling this test file and these test cases `skin-element` is a bit confusing, since it seems to be only validating i18n stuff. I'd recommend either renaming or making this an actual generic skin test.

cjpillsbury: suggestion(non-blocking): For the i18n tests, consider adding additional test permutations (undefined key, language changes, etc. etc.)

packages/html/src/define/video/minimal-skin.tailwind.ts
```diff
@@ -138,39 +138,40 @@ function getTemplateHTML() {
               </media-volume-slider>
             </media-popover>
 
-            <button commandfor="settings-menu" aria-label="Settings" class="${cn(button.base, button.subtle, button.icon, menu.settingsTrigger, 'media-button--settings')}">
+            <button commandfor="settings-menu" aria-labelledby="settings-label" class="${cn(button.base, button.subtle, button.icon, menu.settingsTrigger, 'media-button--settings')}">
```
cjpillsbury, line 141: thought(non-blocking): I don't know if our "eject" story is intended to include skins on the light DOM, but, if so, this pattern may get a bit more complicated if there are 2+ skins sharing the same `id`s on the same light DOM. Nothing we need to change now, since this is the right pattern to establish and we shouldn't over-complicate parameterized ids unless/until needed, but thought it was worth calling out as a possible/plausible future concern.

packages/html/src/define/video/minimal-skin.tailwind.ts
```diff
@@ -138,39 +138,40 @@ function getTemplateHTML() {
...
               ${renderIcon('gear', { class: cn(icon, menu.settingsIcon) })}
+              <media-text id="settings-label" key="menuSettings" class="media-sr-only"></media-text>
```
cjpillsbury: question(non-blocking): Is there a reason we "baked in" the keys and maybe `<media-text>` usages in some components (e.g. in the error dialog case) but stuck with our standard, BaseUI-style, fully exposed usages here? My S1 is the explicitly composed approach here should be preferred unless there is a compelling reason not to. Can be a followup though.

packages/html/src/define/video/ui.ts
```diff
@@ -34,6 +34,7 @@ import {
   defineSliders,
   defineTime,
 } from '../ui/compounds';
+import '../i18n';
```
cjpillsbury: suggestion(non-blocking): After looking into the code, it looks like the `safeDefine(VideoPlayerElement)` + `safeDefine(MediaContainerElement)` is redundant and `import { VideoPlayerElement } from './player';` could have similarly instead simply been `import './player';`. Still, we should probably keep the patterns consistent between the two, so consider changing one or the other (my S1 is change yours to be similarly redundant)

packages/html/src/define/global.css
```diff
@@ -3,7 +3,8 @@
 /* -------------------------------------------------------------------------- */
 
 video-player,
-live-video-player {
+live-video-player,
+media-i18n-provider {
```
cjpillsbury: praise: 👌

packages/html/src/define/i18n.ts
```diff
@@ -0,0 +1,12 @@
+import { MediaI18nProviderElement, MediaTextElement } from '../i18n/elements';
+import { safeDefine } from './safe-define';
+
+safeDefine(MediaI18nProviderElement);
+safeDefine(MediaTextElement);
+
+declare global {
```
cjpillsbury, line 8: issue(blocking): This doesn't conform to the conventions of all of the other components, where:
1. the tag name is defined in this module (so that we don't deeply assume the element definition has a specific tag name)
2. separate modules, one per component.

this relates to my comment above for the video's define as well.

packages/html/src/i18n/base.ts
```diff
@@ -0,0 +1,99 @@
...
+export type ReactiveElementMixinBase = Constructor<ReactiveElement> & Pick<typeof ReactiveElement, 'properties'>;
+
+export type I18nControllerConstructor = new (
+  host: ReactiveControllerHost & HTMLElement
+) => ReactiveController & {
+  readonly value: Translator;
+  readonly locale: Locale;
+};
+
+export type I18nTextMixin = <Base extends ReactiveElementMixinBase>(Base: Base) => Constructor<ReactiveElement> & Base;
```
cjpillsbury: praise: 😎

packages/html/src/i18n/base.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export interface I18nBase {
+  context: I18nLitContext;
+  fallbackTranslator: Translator;
+  I18nController: I18nControllerConstructor;
+  TextMixin: I18nTextMixin;
+}
+
+export function createI18nBase(): I18nBase {
+  const fallbackTranslator = createTranslator(getI18nTranslations('en'), 'en');
```
cjpillsbury: suggestion(non-blocking): not sure if this is feasible within the larger architecture, but consider passing `'en'` in as a default but overridable param of `createI18nBase()` (so something like `function createI18nBase(lang = 'en')`

packages/html/src/i18n/base.ts
```diff
@@ -0,0 +1,99 @@
...
+      });
+      host.addController(this);
+    }
+
+    get value(): Translator {
+      return this.#consumer.value?.translator ?? fallbackTranslator;
+    }
+
+    get locale(): Locale {
+      return this.#consumer.value?.locale ?? 'en';
```
cjpillsbury: nitpick(blocking): even if you don't make this a param, we should replace inline `'en'` with a variable for maintainability.

packages/html/src/i18n/elements.ts
```diff
@@ -0,0 +1,12 @@
+import { ReactiveElement } from '@videojs/element';
+
+import { I18nTextMixin } from './instance';
+import { I18nProviderMixin } from './provider';
+
+export class MediaI18nProviderElement extends I18nProviderMixin(ReactiveElement) {
```
cjpillsbury: issue(blocking): See above comments, re: `tagName` and related.

packages/html/src/i18n/instance.ts
```diff
@@ -0,0 +1,7 @@
+import { createI18nBase } from './base';
+
+export const i18nBase = createI18nBase();
+
+export const context = i18nBase.context;
```
cjpillsbury: question(blocking): Is this a pattern used elsewhere? My Spidey Sense is tingling here.

packages/html/src/i18n/base.ts
```diff
@@ -0,0 +1,99 @@
+import { createTranslator, getI18nTranslations, type Locale, type Translator } from '@videojs/core/i18n/base';
+import type { PropertyValues, ReactiveController, ReactiveControllerHost, ReactiveElement } from '@videojs/element';
+import { type Context, ContextConsumer, createContext } from '@videojs/element/context';
+import type { Constructor } from '@videojs/utils/types';
+
+/** Reflected i18n keys are untyped strings; the runtime translator accepts any key. */
+function translateReflectedKey(translator: Translator, key: string): string {
+  const translateLoose = translator as (k: string, params?: unknown) => string;
```
cjpillsbury: nitpick(non-blocking): `params` is unused. Consider removing from type. Also, I've noticed a few places where we're going from weaker lang types to stronger ones via casting. As a followup, I suspect we can improve this either through rethinking/redefining our types or pushing the loose/string lang types to the "outer boundaries" better.

packages/html/src/i18n/locale.ts
```diff
@@ -0,0 +1,22 @@
+import type { Locale } from '@videojs/core/i18n';
+import { localeFromDomLang as domLangFromString, effectiveLocale, nearestLang } from '@videojs/utils/dom';
+
+/** DOM `lang` values are untyped strings; align with core {@link Locale} at the boundary. */
+export function localeFromDomLang(raw: string | undefined): Locale | undefined {
+  const tag = domLangFromString(raw);
+  return tag === undefined ? undefined : (tag as Locale);
```
cjpillsbury: suggestion(non-blocking): `return (domLangFromString(raw) as Locale) ?? undefined;`

Also, this looks like another one of those weaker -> stricter type conversions (same as `resolvePlayerLocale`, below)

cjpillsbury (changes requested):
Broad issue - I'm noticing a lot of slight but potentially impactful deviations from the equivalent classes, higher order functions, etc. in the store/player provider implementations, and it's not clear to me why. May need to discuss this in a bit more detail (possibly out of band) before merging. Overall this is feeling like the right direction and this will likely be more just moving some bits around.

packages/html/src/define/i18n.ts
cjpillsbury: Resolving in a slightly different way (more in line with the ui components pattern and how it relates to the ui barrel module)

packages/html/src/define/video/minimal-skin.tailwind.ts
sampotts: Yeah this is totally valid. Need a nicer way to do this.

packages/html/src/define/video/minimal-skin.tailwind.ts
sampotts: Are you saying we should use `<media-text>` everywhere instead? I'm down, might be a little noisier but certainly clearer what's going on.

packages/html/src/i18n/tests/provider.test.ts
sampotts: Renamed!

sampotts: Added some more tests 👍🏼

packages/html/src/i18n/instance.ts
sampotts: Removed 👍🏼

packages/html/src/i18n/base.ts
sampotts: Yeah, it's not great I admit that.

packages/html/src/i18n/locale.ts
sampotts: Done.

packages/html/src/i18n/elements.ts
sampotts: Ditched this.

packages/html/src/define/audio/minimal-skin.tailwind.ts
sampotts: Yeah, I'd be open to this. I've got it sat in a stash so basically:

```html
<media-alert-dialog-title>Doh!</media-alert-dialog-title>
```
Would render "Doh!" if there's no provider wrapper element, otherwise fall back to default (English). Similarly, perhaps we'd also want: 

```html
<media-text key="nonExistentKey">Test</media-text>
```
Would render "Test" as key is non existent. Or...  

```html
<media-text key="play">Test</media-text>
```

Would render "Test" if there's no provider wrapper element.

packages/html/src/define/video/minimal-skin.tailwind.ts
cjpillsbury: Yeah, that was my thought anyway. This could be a followup / definitely shouldn't be blocking imo, but it does feel more in line with our general approach for these things? Also would be interested in others' thoughts here.

packages/html/src/i18n/create-i18n.ts
```diff
@@ -0,0 +1,34 @@
+import { createContext } from '@videojs/element/context';
```
cjpillsbury: issue(blocking): Unless I missed something, it looks like this is not needed/used anymore? (except in its own tests). If so, please delete it + corresponding test file.

cjpillsbury (approved):
I have one minor blocking callout but am preemptively approving so you can easily update and mash that merge. This is looking 😎.

One last high level (non-blocking) thought - In this current approach, we are definitionally including i18n in all of our media preset component definitions. This should either be anticipated as one of our eject stories or some other opt-in/opt-out story.

packages/html/src/define/audio/minimal-skin.tailwind.ts
sampotts: ### Option 2

```html
<media-alert-dialog-title>Doh!</media-alert-dialog-title>
```

Should _always_ render `Doh!`. Why? They're purposefully overriding the default. 

```html
<media-text key="meh">Test</media-text>
```

Should only render `Test` if the provider is missing or the key isn't found. The difference here is they are opting into i18n with `<media-text>`.

----

### Option 3

Require the use of `<media-text>` everywhere instead of baking it in. Downside there is it'll render an extra `<span>` but maybe that's worth it to keep it more obvious that you either use static text content or `<media-text>`. It would also align better with the React side with `<media-text key="key">` being the sorta equivalent of `{t('key')}`. 

```html
<media-alert-dialog-title>
  <media-text key="close">Go away</media-text>
</media-alert-dialog-title>
```

It would only render `Go away` if the provider was omitted which is slightly different to how React would handle it - there we'd always fall back to English for `t('...')` calls.

packages/html/src/define/skin-element.ts
```diff
@@ -17,7 +18,7 @@ const sharedSheet = createShadowStyle(sharedStyles);
  * `static template` into it, and applies shared + per-skin styles
  * via `adoptedStyleSheets` (or `<style>` fallback).
  */
-export class SkinElement extends ReactiveElement {
+export class SkinElement extends I18nProviderMixin(ReactiveElement) {
```
mihar-22: Rough thought but this is a possible smell to me in that it needs to be baked in and not an optional higher-level concern. Largely related to my comments on #1708. 

In other words, I think `i18n` should be composed in rather than provided out of the box but this circles back to conversation around keys because a default english translation is required since keys have no meaning on their own.

packages/html/src/i18n/controller.ts
```diff
@@ -0,0 +1,61 @@
...
+  fallbackTranslator ??= createTranslator(getI18nTranslations(DEFAULT_LOCALE), DEFAULT_LOCALE);
+  return fallbackTranslator;
+}
+
+export class I18nController implements ReactiveController {
+  readonly #host: ReactiveControllerHost & HTMLElement;
+  readonly #consumer: ContextConsumer<I18nContext, ReactiveControllerHost & HTMLElement>;
+  #unsubscribeRegistry: (() => void) | undefined;
+
+  constructor(host: ReactiveControllerHost & HTMLElement, context: I18nContext) {
```
mihar-22: (suggestion): pull `ReactiveControllerHost & HTMLElement` up into a `I18nControllerHost` type

packages/html/src/i18n/controller.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  hostDisconnected(): void {
+    this.#unsubscribeRegistry?.();
+    this.#unsubscribeRegistry = undefined;
+  }
+}
+
+export namespace I18nController {
+  export type Constructor = new (host: ReactiveControllerHost & HTMLElement, context: I18nContext) => I18nController;
+}
```
mihar-22, line 65: I think this can be `Constructor = typeof I18nController`, also good to add `Host = I18nControllerHost` to the namespace

packages/html/src/i18n/create-i18n.ts
```diff
@@ -0,0 +1,32 @@
...
+  });
+  const TextMixin = createTextMixin({ i18nContext });
+
+  return {
+    context: i18nContext,
+    I18nController,
+    ProviderMixin,
+    TextMixin,
+  };
+}
```
mihar-22, line 32: beautiful 🖼️

packages/html/src/i18n/locale.ts
```diff
@@ -0,0 +1,21 @@
+import type { Locale } from '@videojs/core/i18n';
+import { localeFromDomLang as domLangFromString, effectiveLocale, nearestLang } from '@videojs/utils/dom';
+
+/** DOM `lang` values are untyped strings; align with core {@link Locale} at the boundary. */
+export function localeFromDomLang(raw: string | undefined): Locale | undefined {
+  return (domLangFromString(raw) as Locale) ?? undefined;
+}
```
mihar-22: we could make the return type generic on `domLangFromString`, or we could move it out of from utils for now if it's only required in the `html` package.

packages/html/src/i18n/locale.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/** Delegates to {@link effectiveLocale}; result is typed as {@link Locale} for player UI. */
+export function resolvePlayerLocale(explicit: Locale | undefined, inherited: Locale | undefined): Locale {
+  return effectiveLocale(explicit, inherited) as Locale;
+}
```
mihar-22, line 7: same comments as `localeFromDomLang`

packages/html/src/i18n/provider-mixin.ts
```diff
@@ -0,0 +1,163 @@
...
+
+import type { I18nContext, I18nContextValue } from './context';
+import { getFallbackTranslator } from './controller';
+import { resolveProviderLocale } from './locale';
+import type { LocaleLoader, ReactiveElementMixinBase } from './types';
+
+export interface I18nProviderConfig {
+  i18nContext: I18nContext;
+  /** Override lazy loading of shipped locale packs (tests or custom loaders). */
+  loadLocale?: LocaleLoader | undefined;
```
mihar-22: maybe `loader` or `load`

packages/html/src/i18n/provider-mixin.ts
```diff
@@ -0,0 +1,163 @@
...
+import { mergeLocaleOverlays, subscribeAmbientLang } from '@videojs/utils/dom';
+import type { Constructor } from '@videojs/utils/types';
+
+import type { I18nContext, I18nContextValue } from './context';
+import { getFallbackTranslator } from './controller';
+import { resolveProviderLocale } from './locale';
+import type { LocaleLoader, ReactiveElementMixinBase } from './types';
+
+export interface I18nProviderConfig {
+  i18nContext: I18nContext;
```
mihar-22: probably just `context`

packages/html/src/i18n/text-mixin.ts
```diff
@@ -0,0 +1,54 @@
...
+import type { PropertyValues, ReactiveElement } from '@videojs/element';
+import type { Constructor } from '@videojs/utils/types';
+import type { I18nContext } from './context';
+import { I18nController } from './controller';
+import type { ReactiveElementMixinBase } from './types';
+
+export type I18nTextMixin = <Base extends ReactiveElementMixinBase>(Base: Base) => Constructor<ReactiveElement> & Base;
+
+export interface TextMixinConfig {
+  i18nContext: I18nContext;
```
mihar-22: probably just `context` is fine

packages/html/src/ui/captions-radio-group/captions-radio-group-element.ts
mihar-22, line 1: Maybe it'd help if we had a base radio group element that the rest can extend and share basic logic on

mihar-22 (approved):

packages/html/src/define/video/minimal-skin.ts
```diff
@@ -116,39 +116,40 @@ function getTemplateHTML() {
...
               ${renderIcon('gear', { class: 'media-icon media-icon--settings' })}
+              <media-text id="settings-label" key="Settings" class="media-sr-only"></media-text>
             </button>
             <media-menu id="settings-menu" side="top" align="center" class="media-popover media-menu media-menu--settings">
               <media-menu-view class="media-menu__panel">
                 <div class="media-menu__group">
                   <media-menu-item commandfor="settings-quality-menu" type="quality" data-setting="quality" class="media-menu__item media-menu__item--submenu">
                     ${renderIcon('switches', { class: 'media-icon' })}
-                    <span>Quality</span>
+                    <media-text key="Quality"></media-text>
```
mihar-22: I think it would be nice if we let the text content live in the DOM and we pull from there into the translation function. It's naturally where the text should live, easier to read/review, no need for a new attr/prop, key would conflict in some frameworks like React, and importantly if for some reason i18n/translation fails the text is still in the DOM. It's also confusing because it's called "key" but is the actual text.

mihar-22: Library I've been looking at that which seems really nice has this too: https://generaltranslation.com/en-US/docs/react/introduction#the-t-component

mihar-22 (approved):

packages/html/src/define/video/minimal-skin.ts
sampotts: Totally agree - `key` makes no sense now. I thought about this after the change to use English but totally forgot to actually implement it 🤦🏼

PR #1596, opened 2026-05-25
feat(packages): add pauseOnDrag to time slider

### LINKED ISSUE #1440: Feature: Pause While Scrubbing
Optionally pause playback while the user is actively dragging the time slider thumb, resuming (if it was playing before) when they release. Reduces the jittery feel during long scrubs and avoids audio glitches.

Scope is narrow — core smooth-seeking behavior (throttled commits, raw precision during drag, final commit on release) is already shipped via `commitThrottle` on the time-slider element. This is the remaining slice: an optional pause-on-drag mode.

## Prior art

**Has it:**
- **Vidstack** — opt-in `pauseWhileDragging` prop on `<TimeSlider>` (default `false`). Closest match to the shape we'd want here. [source](https://github.com/vidstack/player/blob/main/packages/vidstack/src/components/ui/sliders/time-slider/time-slider.ts#L49) · [docs](https://vidstack.io/docs/player/components/sliders/time-slider/)
- **Plyr** — always-on, no opt-out. Binds `mousedown/mouseup/keydown/keyup/touchstart/touchend` on the seek input; tracks a `play-on-seeked` attribute to resume after release. [source](https://github.com/sampotts/plyr/blob/master/src/js/listeners.js#L650-L676)
- **Video.js v8** — always-on with platform disables (`disableSeekWhileScrubbingOnMobile`, `disableSeekWhileScrubbingOnSTV`). `SeekBar.handleMouseDown` stores `videoWasPlaying`, pauses, resumes in `handleMouseUp`. The platform flags suggest mobile/STV caveats worth thinking about (iOS audio glitching during scrub is the usual reason to skip it). [pause](https://github.com/videojs/video.js/blob/main/src/js/control-bar/progress-control/seek-bar.js#L297-L313) · [resume](https://github.com/videojs/video.js/blob/main/src/js/control-bar/progress-control/seek-bar.js#L417-L441)

**Doesn't have it:**
- **Media Chrome** — `<media-time-range>` tracks pointer-down and suppresses range updates during drag, but only dispatches `MEDIA_SEEK_REQUEST`; no pause behavior.
- **Mux Player** — built on Media Chrome, so inherits the same behavior.

R-Delfino95:
Closes #1440 

### What it does

Adds opt-in `pauseWhileDragging` mirroring Vidstack's API:

- **React:** `<TimeSliderRoot pauseWhileDragging />` (default `false`)
- **HTML:** `<media-time-slider pause-while-dragging>` (boolean attribute)

Wraps the existing `onDragStart` / `onDragEnd` callbacks — user-provided callbacks still fire alongside. Reads playback state via the existing `selectPlayback` selector + `PlayerController` / `useLatestRef`, mirroring how `time` and `buffer` are already wired.

Tracks `wasPlayingBeforeDrag` internally so a player paused before the drag doesn't get accidentally auto-played on release.

### Scope choice

Opt-in default (`false`), matching Vidstack's `pauseWhileDragging` prop. v8 and Plyr are always-on (with platform disables for mobile/STV iOS audio glitching). I went opt-in because:

1. It's a behavior change consumers should consent to (subtle iOS audio glitch on some streams during scrub-pause-resume)
2. Vidstack's prior art is the closest match to v10's API shape
3. Easy follow-up to flip the default if maintainers prefer

`play()` rejection on resume (autoplay policy, ad-paused state) is swallowed here — the existing error feature surfaces it via the player error state.

### Verified locally

- `pnpm -F @videojs/react test` — **224/224 pass** (+4 new behavioral tests: default no-op, pause/resume when playing, no-resume when already paused, user-callback forwarding)
- `pnpm -F @videojs/html test` — **120/120 pass** (+1 new: attribute reflection)
- `pnpm lint` / `pnpm typecheck` / `pnpm check:workspace` — all clean

### Open question

The React test uses a hoisted-ref pattern to capture `onDragStart` / `onDragEnd` from the createSlider mock. If you'd prefer not extending the slider mock that way, happy to drop the React behavioral tests and lean on Playwright E2E for slider behavior (matches the existing convention in the HTML element tests).

---

> [!NOTE]
> **Medium Risk**
> Touches playback pause/play during seek UI interactions; default-off limits blast radius, but resume and mid-drag teardown paths affect player state.
> 
> **Overview**
> Adds an opt-in **`pauseOnDrag`** option (default `false`) on the time slider so playback can pause while scrubbing and resume on release only if it was playing when the drag started.
> 
> **`TimeSliderCore`** gains `startDrag` / `endDrag` using `MediaPlaybackState`, with `#wasPlayingBeforeDrag` so already-paused media is not auto-played; resume still runs if `pauseOnDrag` is turned off mid-drag. Failed `play()` on resume is swallowed.
> 
> **React** (`TimeSliderRoot`) and **HTML** (`media-time-slider` / `pause-on-drag`) wire drag lifecycle through the core via `selectPlayback`, still calling user `onDragStart` / `onDragEnd`. Both call `endDrag` on unmount/disconnect/destroy so a mid-drag teardown does not leave the player paused.
> 
> Tests cover core drag behavior, HTML attribute reflection, and React integration (including unmount resume and callback forwarding).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71ef8a8092ffd08e5b537b178f1ea871fdd5fb95. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/html/src/ui/time-slider/time-slider-element.ts
+++ b/packages/html/src/ui/time-slider/time-slider-element.ts
@@ -116,21 +116,26 @@ export class TimeSliderElement extends MediaElement {
   }
 
   override disconnectedCallback(): void {
-    // Resume playback if disconnected mid-drag — createSlider's destroy() does not
-    // fire onDragEnd, so without this the player would stay paused. Read playback
-    // before super so the PlayerController is still attached.
-    if (this.#wasPlayingBeforeDrag) {
-      this.#playbackState.value?.play().catch(() => {});
-      this.#wasPlayingBeforeDrag = false;
-    }
+    this.#resumeIfDragPaused();
     super.disconnectedCallback();
     this.#disconnect?.abort();
     this.#disconnect = null;
   }
 
   override destroyCallback(): void {
+    this.#resumeIfDragPaused();
     this.#slider?.destroy();
     super.destroyCallback();
+  }
+
+  // createSlider's destroy() does not fire onDragEnd, so a teardown mid-drag
+  // would leave playback paused. Called from both disconnect and destroy paths
+  // before super so the PlayerController is still attached.
+  #resumeIfDragPaused(): void {
+    if (this.#wasPlayingBeforeDrag) {
+      this.#playbackState.value?.play().catch(() => {});
+      this.#wasPlayingBeforeDrag = false;
+    }
   }
 
   protected override willUpdate(_changed: PropertyValues): void {
--- a/packages/html/src/ui/time-slider/time-slider-element.ts
+++ b/packages/html/src/ui/time-slider/time-slider-element.ts
@@ -7,6 +7,7 @@ import {
   logMissingFeature,
   type SliderApi,
   selectBuffer,
+  selectPlayback,
   selectTime,
 } from '@videojs/core/dom';
 import type { PropertyDeclarationMap, PropertyValues } from '@videojs/element';
@@ -30,7 +31,10 @@ export class TimeSliderElement extends MediaElement {
     orientation: { type: String },
     disabled: { type: Boolean },
     thumbAlignment: { type: String, attribute: 'thumb-alignment' },
-  } satisfies PropertyDeclarationMap<Exclude<keyof TimeSliderCore.Props, 'value' | 'min' | 'max'>>;
+    pauseWhileDragging: { type: Boolean, attribute: 'pause-while-dragging' },
+  } satisfies PropertyDeclarationMap<Exclude<keyof TimeSliderCore.Props, 'value' | 'min' | 'max'>> & {
+    pauseWhileDragging: { type: BooleanConstructor; attribute: string };
+  };
 
   label = TimeSliderCore.defaultProps.label;
   changeThrottle = TimeSliderCore.defaultProps.changeThrottle;
@@ -39,11 +43,18 @@ export class TimeSliderElement extends MediaElement {
   orientation = TimeSliderCore.defaultProps.orientation;
   disabled = TimeSliderCore.defaultProps.disabled;
   thumbAlignment = TimeSliderCore.defaultProps.thumbAlignment;
+  /**
+   * When true, pause playback while the user is dragging the thumb,
+   * resuming on release if it was playing before. Default `false`.
+   */
+  pauseWhileDragging = false;
 
   readonly #core = new TimeSliderCore();
   readonly #provider = new ContextProvider(this, { context: sliderContext });
   readonly #timeState = new PlayerController(this, playerContext, selectTime);
   readonly #bufferState = new PlayerController(this, playerContext, selectBuffer);
+  readonly #playbackState = new PlayerController(this, playerContext, selectPlayback);
+  #wasPlayingBeforeDrag = false;
 
   #slider: SliderApi | null = null;
   #disconnect: AbortController | null = null;
@@ -74,9 +85,21 @@ export class TimeSliderElement extends MediaElement {
       },
       changeThrottle: this.changeThrottle,
       onDragStart: () => {
+        this.#wasPlayingBeforeDrag = false;
+        const playback = this.#playbackState.value;
+        if (this.pauseWhileDragging && playback && !playback.paused) {
+          this.#wasPlayingBeforeDrag = true;
+          playback.pause();
+        }
         this.dispatchEvent(new CustomEvent('drag-start', { bubbles: true }));
       },
       onDragEnd: () => {
+        if (this.pauseWhileDragging && this.#wasPlayingBeforeDrag) {
+          this.#playbackState.value?.play().catch(() => {
+            // Resume play() can reject (autoplay policy, etc.) — surface via existing error feature.
+          });
+        }
+        this.#wasPlayingBeforeDrag = false;
         this.dispatchEvent(new CustomEvent('drag-end', { bubbles: true }));
       },
       adjustPercent: (raw, thumbSize, trackSize) => this.#core.adjustPercentForAlignment(raw, thumbSize, trackSize),
--- a/packages/html/src/ui/time-slider/time-slider-element.ts
+++ b/packages/html/src/ui/time-slider/time-slider-element.ts
@@ -31,7 +31,7 @@ export class TimeSliderElement extends MediaElement {
     orientation: { type: String },
     disabled: { type: Boolean },
     thumbAlignment: { type: String, attribute: 'thumb-alignment' },
-    pauseWhileDragging: { type: Boolean, attribute: 'pause-while-dragging' },
+    pauseOnDrag: { type: Boolean, attribute: 'pause-on-drag' },
   } satisfies PropertyDeclarationMap<Exclude<keyof TimeSliderCore.Props, 'value' | 'min' | 'max'>>;
 
   label = TimeSliderCore.defaultProps.label;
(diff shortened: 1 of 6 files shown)


R-Delfino95:
Hi @shiminshen! Thanks for your contribution 🙌

 I created the PR with your changes, as external PRs are currently restricted.

I gave it a try and it looks correct to me. Would you mind rebasing on main and giving it another quick test, just to make
sure nothing has changed in the meantime?

Also, Cursor Bot detected some potential issues; would you mind reviewing those as well?

Thanks again!

shiminshen:
Thanks @R-Delfino95! Rebased on `main` (no conflicts in the slider files) and both Cursor Bot findings addressed in 80d201d2:

- **Resume gated on current flag**: the drag-end guard now depends only on the captured `wasPlayingBeforeDrag` flag. That flag is already a strict subset (only set when `pauseWhileDragging` is true at drag start *and* playback is playing), so dropping the redundant prop check means a mid-drag toggle no longer leaves the player paused. Symmetry: "if we paused it, we resume it."
- **Unmount mid-drag skips resume**: added a `useEffect` cleanup (React) and a top-of-`disconnectedCallback` check (HTML) that resume playback if the slider goes away during an active drag. The HTML version reads playback state *before* `super.disconnectedCallback()` so the `PlayerController` is still attached.

Two new React tests cover both edge cases. Existing tests still pass; lint and the React/HTML time-slider typecheck are clean.

packages/react/src/ui/time-slider/time-slider-root.tsx
```diff
@@ -34,17 +39,41 @@ export const TimeSliderRoot = forwardRef<HTMLDivElement, TimeSliderRootProps>(
...
+      }
+      onDragStart?.();
+    };
+
+    const handleDragEnd = () => {
+      if (pauseWhileDragging && wasPlayingRef.current) {
+        playbackRef.current?.play().catch(() => {
+          // Resume play() can reject (autoplay policy, etc.) — surface via existing error feature.
+        });
+      }
```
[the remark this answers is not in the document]
shiminshen: Fixed in 80d201d2 — the drag-end guard now relies only on the captured `wasPlayingBeforeDrag` flag, which is already a strict subset (it can only be `true` when `pauseWhileDragging` was true at drag start *and* playback was playing). Dropping the redundant prop check means the captured intent ("we paused it") survives the prop being toggled off between `dragStart` and `dragEnd`. New test `resumes on drag-end even if pauseWhileDragging is turned off mid-drag` covers this.

packages/react/src/ui/time-slider/time-slider-root.tsx
```diff
@@ -34,17 +39,41 @@ export const TimeSliderRoot = forwardRef<HTMLDivElement, TimeSliderRootProps>(
...
+      wasPlayingRef.current = false;
+      onDragEnd?.();
```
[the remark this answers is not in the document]
shiminshen: Fixed in 80d201d2. React: a `useEffect` cleanup checks `wasPlayingRef.current` on unmount and calls `play()` if a drag had paused playback. HTML: `disconnectedCallback` reads `#playbackState.value` and resumes *before* `super.disconnectedCallback()` so the `PlayerController` is still attached to the player context. New test `resumes on unmount if a drag paused playback` covers the React path.

packages/html/src/ui/time-slider/time-slider-element.ts
```diff
@@ -93,6 +116,13 @@ export class TimeSliderElement extends MediaElement {
   }
 
   override disconnectedCallback(): void {
+    // Resume playback if disconnected mid-drag — createSlider's destroy() does not
+    // fire onDragEnd, so without this the player would stay paused. Read playback
+    // before super so the PlayerController is still attached.
+    if (this.#wasPlayingBeforeDrag) {
+      this.#playbackState.value?.play().catch(() => {});
+      this.#wasPlayingBeforeDrag = false;
+    }
```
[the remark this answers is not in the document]
shiminshen: Fixed in 59d88a03 — extracted the existing `disconnectedCallback` resume into a `#resumeIfDragPaused()` helper and called it from `destroyCallback` too (before `super` so the `PlayerController` is still attached). Covers the manual-`destroy()` and `keep-alive` paths where `disconnectedCallback` never runs.

No new unit test on the HTML side — the existing time-slider element tests don't have a drag-simulation harness (the behavioural drag coverage lives in the React tests via mocked `createSlider`), and the HTML change is a literal mirror of the resume logic already exercised by that path. Happy to add an HTML drag harness if you'd prefer direct coverage here.

packages/react/src/ui/time-slider/time-slider-root.tsx
```diff
@@ -79,8 +119,8 @@ export const TimeSliderRoot = forwardRef<HTMLDivElement, TimeSliderRootProps>(
         const media = mediaRef.current;
         if (media) media.seek(core.rawValueFromPercent(percent));
       },
-      onDragStart,
-      onDragEnd,
+      onDragStart: handleDragStart,
+      onDragEnd: handleDragEnd,
```
mihar-22: We can inline the handlers here to avoid naming/defining them above and separating from config.

packages/html/src/ui/time-slider/time-slider-element.ts
```diff
@@ -30,7 +31,10 @@ export class TimeSliderElement extends MediaElement {
     orientation: { type: String },
     disabled: { type: Boolean },
     thumbAlignment: { type: String, attribute: 'thumb-alignment' },
-  } satisfies PropertyDeclarationMap<Exclude<keyof TimeSliderCore.Props, 'value' | 'min' | 'max'>>;
+    pauseWhileDragging: { type: Boolean, attribute: 'pause-while-dragging' },
+  } satisfies PropertyDeclarationMap<Exclude<keyof TimeSliderCore.Props, 'value' | 'min' | 'max'>> & {
+    pauseWhileDragging: { type: BooleanConstructor; attribute: string };
+  };
```
mihar-22: This union `&` of types shouldn't happen, types come from `TimeSliderCore.Props`

packages/html/src/ui/time-slider/time-slider-element.ts
```diff
@@ -39,11 +43,18 @@ export class TimeSliderElement extends MediaElement {
   orientation = TimeSliderCore.defaultProps.orientation;
   disabled = TimeSliderCore.defaultProps.disabled;
   thumbAlignment = TimeSliderCore.defaultProps.thumbAlignment;
+  /**
+   * When true, pause playback while the user is dragging the thumb,
+   * resuming on release if it was playing before. Default `false`.
+   */
+  pauseWhileDragging = false;
```
mihar-22: Descriptions don't belong here, they should be derived from implementing `TimeSliderCore.Props`. See how above none of them have it. Default value comes from core too.

mihar-22 (changes requested):
Thank you, Renzo. Before a deeper review I'd like to see more conformance with our core design.

1. Lets rename this to `pauseOnDrag`. Little more terse. 
2. Move core state logic and props to the `TimeSliderCore` class. Please see other core classes and interfaces as reference. Make sure they're documented. 
3. The PR title `feat(html,react)` contains an invalid scope. Here it'd be `feat(packages)`

packages/react/src/ui/time-slider/time-slider-root.tsx
damonshen17: Done in 2ae34cc8 — and after 71ef8a80 they just call core.startDrag/endDrag.

packages/html/src/ui/time-slider/time-slider-element.ts
damonshen17: Fixed in 2ae34cc8 — types come from TimeSliderCore.Props now.

packages/html/src/ui/time-slider/time-slider-element.ts
damonshen17: Removed in 2ae34cc8 — description and default come from core now.

damonshen17:
@mihar-22 done in 71ef8a80:

1. Renamed to `pauseOnDrag` (`pause-on-drag` attribute on the element).
2. Moved the pause/resume logic into `TimeSliderCore` as documented `startDrag`/`endDrag` methods (same shape as `PlayButtonCore.toggle`) — html/react just delegate now. Added core unit tests for it.
3. On the title — I cannot edit it since @R-Delfino95 opened the PR. Renzo, mind changing it to `feat(packages): add pauseOnDrag to time slider`?

Tests, typecheck and lint all pass.

mihar-22 (approved):
ty!!

Issue #873, opened 2026-03-10
Feature: Create mux-background-video Element


decepulis:
Create a `<mux-background-video>` custom element in Video.js 10.

Inspired by the [existing mux-background-video element](https://github.com/muxinc/mux-background-video), but built on Video.js architecture with its own API. Not all capabilities from the original element will necessarily carry over — the goal is a Video.js-native Mux background video element, not a 1:1 port.

This issue may be divided into multiple efforts as the scope becomes clearer.

spuppo-mux:
Extending the description based on a recent meeting with Christian and changing this to an Epic.

## Scope:
This epic covers everything needed to ship a usable <mux-background-video> element in Video.js 10, built on SPF:

- An SPF engine composition for the background-looping-video use case (HLS multivariant playlist, silent, autoplay, single rendition, seamless loop)
- An adapter exposing an HTML-media-element-shaped API for Video.js consumption
- Optional feature bundles layered via a functional decorator pattern (`withAudio()`, `withPreload()`)
- Full Video.js component integration, including the web-component surface and whatever tree-shaking / bundle strategy we land on
- Extra features: maxResolution

## Phases
Tracked as separate issues under this epic. Phase 1 is the foundation everything else builds on.
- **Phase 1 - SPF composition + adapter (basic case)**
   - Issue: https://github.com/videojs/v10/issues/1586
   - The most common use case only: video-only, autoplay, looping, single rendition, no audio, no preload awareness, no ABR, no multi-CDN. Includes an example page.
- **Phase 2 - withAudio() decoration**
   - Issue: https://github.com/videojs/v10/issues/1613 
   - Functional decorator that spreads audio-related behaviors into the base composition. First proof point for the "include only what you need" bundling pattern.
-  **Phase 3 - withPreload() decoration**
   - Functional decorator that adds preload behaviors and unsets the loadActivated: true initial state. Second proof point for the decorator pattern.
- **Phase 4 - Feature and example cleanup**
   - Includes details that come up during implementation and features that were forgotten on earlier steps (maxResolution)
   - Also includes cleaning up the example page.
- **Phase 5 - Video.js component integration**
   - Full <mux-background-video> component surface in Video.js 10, including how the decorations map to the web-component layer and how tree-shaking is exposed to consumers. Needs broader team alignment


## Explicitly out of scope (for this epic)

[...]

PR #1602, opened 2026-05-26
feat(spf): background looping video (phase 1)

### LINKED ISSUE #1586: Feature: [SPF] Background Looping Video — Phase 1: Composition + Adapter
## Motivation
Phase 1 of #873. Demonstrates SPF's composability by removing behaviors from the standard playback engine rather than adding new logic — the composition is the standard HLS engine minus the behaviors that don't apply, plus an initial-state override and a video-track picker. Intentionally narrow: video-only, autoplay, looping, single rendition, HLS multivariant playlist. Audio, preload-aware variants, and the Video.js component integration are later phases under the parent epic.

## Description
Build the SPF foundation for the new `<mux-background-video>` Video.js 10 element: an engine composition for background-looping-video, plus an adapter that exposes an HTML-media-element-shaped API. Validated locally via `apps/sandbox` wired into Video.js.

**In scope:** HLS multivariant playlist input, the engine composition, the adapter (following the existing default-playback-engine adapter pattern), sandbox validation.

**Out of scope**, later phases under #873: `withAudio()` (Phase 2), `withPreload()` (Phase 3), the full Video.js component / web component surface (Phase 4).

## Behavior
Plays an HLS multivariant playlist immediately on load, ignoring audio and subtitles even when present. Picks a single rendition up front (no ABR) and loops it. Seeking is disabled by design. No multi-CDN logic, no preload-state monitoring.

## Composition
Starting from the standard HLS playback engine, remove: audio track selection, text track selection, preload monitoring, play/seek event monitoring, ABR (the complicated track-selection version), and back-buffer flushing.

**Keep:** duration tracking (drives downstream updates), MSE management, loop handling (so engagement behaves correctly across loops), segment management, and playback position monitoring (needed by forward-buffer management).

**Add:** the existing `SelectVideoTrack` behavior — unused in the standard composition — wired with a config-provided picker function, defaulting to max resolution. It watches the presentation and picks the single video track on initial load.

**Initial state override:** set `loadActivated: true` in the composition's initial state. Every downstream behavior that would otherwise wait for preload/play/seek signals just sees "we're loaded and going" from the start.

## Adapter
Build alongside the engine, following the patternused for the existing default playback engine adapter (mixin-based). The adapter hides SPF internals and exposes an HTML-media-element-shaped public API — methods, properties, public/private boundary. This is the foundation for the Phase 4 Video.js component. The media renderer API architecture is being restructured right now, so the exact shape may shift; build to the current pattern and we'll adapt.

## Wiring the engine to a media element
Either `engine.context.mediaElement.set(<mediaElement>)` after construction, or `initialContext: { mediaElement }` passed to `createComposition`. The behaviors handle both orders, so timing doesn't matter — pick whichever is cleaner in the sandbox. Sandbox templates published to source may need updating; the sandbox setup is a bit tangled today.

## Acceptance Criteria
- [ ] HLS multivariant playlist plays end-to-end: autoplay, silent, single rendition, looping
- [ ] SelectVideoTrack picks the max-resolution track on initial load
- [ ] Adapter exposes an HTML-media-element-shaped public surface
- [ ] Sandbox example at apps/sandbox runs the engine wired into Video.js
- [ ] No regressions in the standard playback engine

## Examples in other libraries
https://github.com/muxinc/mux-background-video

spuppo-mux:
## Summary

Phase 1 of `<mux-background-video>` (#873, Phase 1 #1586): a silent,
autoplay-looping, single-rendition playback engine built by **subtracting**
behaviors from the standard HLS engine rather than adding logic — a
composability demonstration for SPF.

Includes:
- **`pickMaxResolutionVideoTrack`** primitive (next to `pickFirstTrackId`) —
  picks the highest-resolution rendition, bandwidth as fallback.
- **`createBackgroundLoopingVideoEngine`** — subtracts audio, text, ABR,
  preload-monitoring, and play/seek triggers from `createSimpleHlsEngine`;
  adds `selectVideoTrack` with the max-res picker; seeds `loadActivated: true`.
- **`BackgroundLoopingVideoMediaElement`** — adapter parallel to
  `SimpleHlsMediaElement` (`@videojs/spf/background-looping-video`).
- Use-case design doc + a sandbox demo (`/spf-background-looping-video/`) that
  surfaces the engine internals for review.

## Design decision

The adapter is a **separate** component parallel to `SimpleHlsMediaElement`,
not a unification. The three adapters (this, simple-hls, forthcoming
mode-override) share the `SimpleHlsMediaMixin` / `shareSignals` pattern — cost
is config surface, not architecture. Open to revisiting if a fourth converges.

## Review focus

1. **Subtraction-list correctness** — are the right behaviors removed, nothing
   missing or wrongly dropped? (`engine.ts`)
2. **`loadActivated: true` mechanism** — seeding into post-preload-gate state
   from frame 0; first use of the pattern.
3. **Picker primitive shape** — `pickMaxResolutionVideoTrack` API + bandwidth
   fallback as a reusable primitive.

## Deferred / future work

- **MPEG-TS sources don't play — fMP4/CMAF only.** Shared SPF HLS-pipeline
  limitation (parser hardcodes `video/mp4`, no transmux), not specific to this
  variant. Mux delivers CMAF, so the use case is unaffected. **Decision
  deferred to the team**: fail-loud detection (small) vs. a transmuxer (large,
  own RFC) vs. document-and-stay-fMP4. Sandbox filters to fMP4 sources for now.
- **`loop` / `muted` setters are noops in Phase 1** (adapter pins `loop=true` /
  `muted=true`). 3 adapter tests are `it.skip` with an unskip-when-implemented
  note.
- **Sampling-strip `loadVideoSegments`** — bandwidth sampling is wasted work
  without ABR; Phase 3 candidate.
- Rendition switching exists in the **sandbox only** (rebuilds the engine via
  the `picker` config); the engine pins one rendition per session by design.

## Test plan

- [ ] Unit: engine composition, adapter lifecycle, `pickMaxResolutionVideoTrack`
  (24 pass, 3 skipped).
- [ ] Manual: `/spf-background-looping-video/` — video autoplays muted + loops;
  diag shows `loadActivated: true`, max-res rendition, no audio-side context
  keys; rendition picker swaps quality; source switch works.
- [ ] Confirm no MPEG-TS source in the picker (fMP4 only).

---

> [!NOTE]
> **Medium Risk**
> Adds a new public playback engine and MSE pipeline variant (load-always-on semantics and video-only composition); existing HLS engine paths are unchanged but reviewers should validate subtraction completeness and autoplay/MSE edge cases.
> 
> **Overview**
> Adds **Phase 1** SPF support for silent, autoplay-looping background video: a **subtractive** HLS engine variant plus a parallel media adapter, aimed at `<mux-background-video>`.
> 
> **`pickMaxResolutionVideoTrack`** is a new track-selection primitive (resolution first, bandwidth as tiebreaker), with unit tests. **`createBackgroundLoopingVideoEngine`** composes a slimmer pipeline than `createSimpleHlsEngine`—no audio/text, ABR (`switchVideoQuality`), or preload/play-seek load triggers—and instead uses **`selectVideoTrack`** with that picker (overridable) and **`initialState.loadActivated: true`** so loading is not gated on preload/play.
> 
> **`BackgroundLoopingVideoMediaElement`** (mixin + standalone class) mirrors a subset of the HTML media API, defaulting **loop/muted/autoplay** to on and applying them on `attach`; Phase 1 leaves several setters as no-ops. The feature ships as **`@videojs/spf/background-looping-video`** (package `exports` + `tsdown` entry).
> 
> Also included: a **design doc** for the use case, a **sandbox** page that exercises the adapter and shows engine diagnostics, and **engine/adapter tests** (some loop/muted setter tests skipped until implemented).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc1a71cea0a75795aee5bf791b768f6382cb2c42. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/sandbox/templates/spf-background-looping-video/index.html
+++ b/apps/sandbox/templates/spf-background-looping-video/index.html
@@ -1,0 +1,294 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Sandbox — SPF Background Looping Video</title>
+    <style>
+      html,
+      body {
+        height: 100%;
+        padding: 0;
+        margin: 0;
+        font-family:
+          'Inter', system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, sans-serif;
+        color: #fff;
+        background: #000;
+      }
+
+      .stage {
+        position: relative;
+        width: 100vw;
+        height: 100vh;
+        overflow: hidden;
+      }
+
+      #bg-video {
+        position: absolute;
+        inset: 0;
+        display: block;
+        width: 100%;
+        height: 100%;
+        object-fit: cover;
+      }
+
+      .scrim {
+        position: absolute;
+        inset: 0;
+        pointer-events: none;
+        background: linear-gradient(180deg, rgba(0, 0, 0, 0.55) 0%, rgba(0, 0, 0, 0.2) 35%, rgba(0, 0, 0, 0.7) 100%);
+      }
+
+      .hero {
+        position: absolute;
+        inset: 0;
+        display: flex;
+        flex-direction: column;
+        justify-content: center;
+        max-width: 720px;
+        padding: 0 8vw;
+      }
+
+      .hero .eyebrow {
+        margin: 0 0 12px;
+        font-size: 12px;
+        text-transform: uppercase;
+        letter-spacing: 0.18em;
+        opacity: 0.7;
+      }
+
+      .hero h1 {
+        margin: 0 0 16px;
+        font-size: clamp(36px, 6vw, 72px);
+        font-weight: 600;
+        line-height: 1.05;
+        letter-spacing: -0.02em;
+      }
+
+      .hero p {
+        max-width: 540px;
+        margin: 0 0 24px;
+        font-size: 17px;
+        line-height: 1.5;
+        opacity: 0.88;
+      }
+
+      .hero .links {
+        font-size: 13px;
+        opacity: 0.75;
+      }
+
+      .hero .links a {
+        padding-bottom: 1px;
+        margin-right: 18px;
+        color: inherit;
+        text-decoration: none;
+        border-bottom: 1px solid rgba(255, 255, 255, 0.4);
+      }
+
+      .hero .links a:hover {
+        border-bottom-color: rgba(255, 255, 255, 0.9);
+        opacity: 1;
+      }
+
+      /* Top-right diagnostic strip — surfaces the three Phase 1 review-focus
+         signals: loadActivated, the picker result, and the absence of audio-
+         side context. */
+      .diag {
+        position: absolute;
+        top: 16px;
+        right: 16px;
+        min-width: 240px;
+        padding: 12px 14px;
+        font-family: ui-monospace, SFMono-Regular, Menlo, monospace;
+        font-size: 12px;
+        line-height: 1.6;
+        background: rgba(0, 0, 0, 0.55);
+        border: 1px solid rgba(255, 255, 255, 0.1);
+        border-radius: 8px;
+        -webkit-backdrop-filter: blur(8px);
+        backdrop-filter: blur(8px);
+      }
+
+      .diag .diag-title {
+        margin-bottom: 8px;
+        font-size: 10px;
+        text-transform: uppercase;
+        letter-spacing: 0.12em;
+        opacity: 0.55;
+      }
+
+      .diag .row {
+        display: flex;
+        gap: 12px;
+        justify-content: space-between;
+      }
+
+      .diag .row .key {
+        opacity: 0.65;
+      }
+
+      .diag .row .val {
+        text-align: right;
+      }
+
+      .diag .row .val.ok {
+        color: #7ee787;
+      }
+
+      .diag .row .val.absent {
+        color: #ffa657;
+      }
+
+      /* Bottom-right control card — reviewer affordances + roadmap. */
+      .controls {
+        position: absolute;
+        right: 16px;
+        bottom: 16px;
+        min-width: 260px;
+        padding: 14px 16px;
+        font-size: 13px;
+        background: rgba(0, 0, 0, 0.55);
+        border: 1px solid rgba(255, 255, 255, 0.1);
+        border-radius: 8px;
+        -webkit-backdrop-filter: blur(8px);
+        backdrop-filter: blur(8px);
+      }
+
+      .controls label {
+        display: block;
+        margin-bottom: 6px;
+        font-size: 10px;
+        text-transform: uppercase;
+        letter-spacing: 0.12em;
+        opacity: 0.6;
+      }
+
+      .controls select {
+        width: 100%;
+        padding: 6px 8px;
+        font: inherit;
+        color: inherit;
+        background: rgba(255, 255, 255, 0.08);
+        border: 1px solid rgba(255, 255, 255, 0.15);
+        border-radius: 4px;
+      }
+
+      .controls .picker-section {
+        margin-top: 14px;
+      }
+
+      .rendition-buttons {
+        display: flex;
+        flex-direction: column;
+        gap: 4px;
+      }
+
+      .rendition-buttons button {
+        width: 100%;
+        padding: 6px 8px;
+        font: inherit;
+        font-size: 12px;
+        color: inherit;
+        text-align: left;
+        cursor: pointer;
+        background: rgba(255, 255, 255, 0.06);
+        border: 1px solid rgba(255, 255, 255, 0.12);
+        border-radius: 4px;
+      }
+
+      .rendition-buttons button:hover {
+        background: rgba(255, 255, 255, 0.12);
+      }
+
+      .rendition-buttons button.selected {
+        color: #7ee787;
+        background: rgba(126, 231, 135, 0.12);
+        border-color: rgba(126, 231, 135, 0.4);
+      }
+
+      .controls .roadmap {
+        padding-top: 12px;
+        margin-top: 14px;
+        border-top: 1px solid rgba(255, 255, 255, 0.1);
+      }
+
+      .controls .roadmap-title {
+        margin-bottom: 8px;
+        font-size: 10px;
+        text-transform: uppercase;
+        letter-spacing: 0.12em;
+        opacity: 0.5;
+      }
+
+      .controls .roadmap-item {
+        display: flex;
+        gap: 8px;
+        align-items: baseline;
+        font-size: 12px;
+        line-height: 1.6;
+        opacity: 0.55;
+      }
+
+      .controls .roadmap-item .phase {
+        flex-shrink: 0;
+        padding: 1px 5px;
+        font-family: ui-monospace, SFMono-Regular, Menlo, monospace;
+        font-size: 10px;
+        background: rgba(255, 255, 255, 0.08);
+        border-radius: 3px;
+      }
+    </style>
+  </head>
+  <body>
+    <div class="stage">
+      <video id="bg-video" playsinline muted></video>
+      <div class="scrim"></div>
+
+      <div class="hero">
+        <p class="eyebrow">SPF · Phase 1</p>
+        <h1>Background Looping Video</h1>
+        <p>
+          Subtractive engine variant — audio, text tracks, ABR, and preload monitoring are composed out; a max-resolution
+          picker is composed in. The SPF foundation for <code>&lt;mux-background-video&gt;</code>.
+        </p>
+        <p class="links">
+          <a href="https://github.com/videojs/v10/issues/1586" target="_blank" rel="noopener">#1586 (Phase 1)</a>
+          <a href="https://github.com/videojs/v10/issues/873" target="_blank" rel="noopener">#873 (parent epic)</a>
+          <a
+            href="https://github.com/videojs/v10/blob/main/internal/design/spf/use-cases/background-looping-video.md"
+            target="_blank"
+            rel="noopener"
+            >use-case doc</a
+          >
+        </p>
+      </div>
+
+      <div class="diag" id="diag">
+        <div class="diag-title">Engine state</div>
+        <div class="row"><span class="key">loadActivated</span><span class="val" id="diag-load">—</span></div>
+        <div class="row"><span class="key">selected rendition</span><span class="val" id="diag-rendition">—</span></div>
+        <div class="row"><span class="key">context keys</span><span class="val" id="diag-context">—</span></div>
+        <div class="row"><span class="key">audio side</span><span class="val absent">subtracted</span></div>
+      </div>
+
+      <div class="controls">
+        <label for="source-select">Source</label>
+        <select id="source-select"></select>
+
+        <div class="picker-section">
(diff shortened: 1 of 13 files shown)


packages/spf/src/playback/engines/background-looping-video/adapter.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    attach(mediaElement: HTMLMediaElement): void {
+      super.attach?.(mediaElement);
+      // Apply adapter-owned native props before the engine takes over —
+      // the underlying element needs `loop` / `muted` set for the use case's
+      // autoplay-looping semantics.
+      mediaElement.loop = this.#loop;
+      mediaElement.muted = this.#muted;
+      this.#signals.context.mediaElement.set(mediaElement);
+      this.play();
```
cjpillsbury: nitpick(blocking): unless there were issues here, use `autoplay` here and follow the same convention you did for the other props.

packages/spf/src/playback/engines/background-looping-video/adapter.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    async play(): Promise<void> {
+      const mediaElement = this.#signals.context.mediaElement.get();
+      if (!mediaElement) {
+        return Promise.reject(new Error('BackgroundLoopingVideoMediaElement: no media element attached'));
+      }
+
+      // TODO: (To be resolved for the PR).
+      // This could be needed since one could modify loadActivated
+      // this.#signals.state.loadActivated.set(true);
```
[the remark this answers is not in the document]
cjpillsbury: nitpick(non-blocking): Since the purpose of the adapter is to provide the "public API" for the composition, you can remove this todo and commented out code.

packages/spf/src/playback/engines/background-looping-video/adapter.ts
```diff
@@ -0,0 +1,238 @@
...
+    // Each assignment destroys the current engine and starts a fresh one,
+    // matching the browser's load algorithm reset on src change.
+    // -------------------------------------------------------------------------
+
+    get src(): string {
+      return this.#signals.state.presentation.get()?.url ?? '';
+    }
+
+    set src(value: string) {
+      const prevMediaElement = this.#signals.context.mediaElement.get();
```
cjpillsbury: question(blocking): do we need to destroy and recreate the engine or re-set the media element here? This is a legitimate question, since I recently did a pass of the codebase to attempt to ensure that we can "recycle" playback engine compositions across setting `src`.

packages/spf/src/playback/engines/background-looping-video/engine.ts
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * await engine.destroy();
+ * ```
+ */
+export function createBackgroundLoopingVideoEngine(
+  config: BackgroundLoopingVideoEngineConfig = {}
+): Composition<BackgroundLoopingVideoEngineState, BackgroundLoopingVideoEngineContext> {
+  const finalConfig = {
+    ...config,
+    picker: config.picker ?? pickMaxResolutionVideoTrack,
```
cjpillsbury, line 135: question(non-blocking): For background looping, do you think we'd want to allow someone to override these config values? My thought is no, but I'm interested in your perspective here.

apps/sandbox/templates/spf-background-looping-video/index.html
```diff
@@ -0,0 +1,294 @@
...
+
+      <div class="hero">
+        <p class="eyebrow">SPF · Phase 1</p>
+        <h1>Background Looping Video</h1>
+        <p>
+          Subtractive engine variant — audio, text tracks, ABR, and preload monitoring are composed out; a max-resolution
+          picker is composed in. The SPF foundation for <code>&lt;mux-background-video&gt;</code>.
+        </p>
+        <p class="links">
+          <a href="https://github.com/videojs/v10/issues/1586" target="_blank" rel="noopener">#1586 (Phase 1)</a>
```
cjpillsbury: nitpick(non-blocking): I'd lean towards dropping the github issue links.

cjpillsbury (changes requested):
A few tiny callouts but overall this is looking good!

packages/spf/src/media/primitives/select-tracks.ts
```diff
@@ -178,6 +178,34 @@ export function pickVideoTrack(presentation: Presentation, config: VideoSelectio
...
+ */
+export function pickMaxResolutionVideoTrack(presentation: MaybeResolvedPresentation): string | undefined {
+  const videoSet = presentation.selectionSets?.find((set) => set.type === 'video') as VideoSelectionSet | undefined;
+  const tracks = videoSet?.switchingSets[0]?.tracks;
+  if (!tracks?.length) return undefined;
+
+  let bestId: string | undefined;
+  let bestArea = -1;
+  let bestBandwidth = -1;
+  for (const track of tracks) {
```
cjpillsbury, line 197: thought(non-blocking): Unrelated, we should probably do some mild refactors of these picker functions and also move them over to `/media` somewhere since they aren't tied to the SPF `/core` architecture

packages/spf/src/playback/engines/background-looping-video/adapter.ts
spuppo-mux: Not sure how to validate whether it's _needed_, as far as I'm aware behavior seems ok after switching `src`s and with the memory inspector I don't see any weird stuff (filtering by background and buffer).

What I tested was: opening my example (http://localhost:5173/spf-background-looping-video/), capturing a memory snapshot, switching src many times, and then after a second mem snapshot, verifying that there are no excess elements created.

Open to suggestions on other ways to test this!

packages/spf/src/playback/engines/background-looping-video/engine.ts
spuppo-mux: Picker I think it's nice to be able to choose another track if needed, parsePresentation I understand would be useful if we wanted to use dash for example.
resolveDuration I think makes sense to just be set to getResolvedSelectedTrackDuration.

apps/sandbox/templates/spf-background-looping-video/index.html
spuppo-mux: True. Was thinking of tidying/discarding the example on the final phase, but it may be less of a headache to maintain tidy on each iteration.

packages/spf/src/playback/engines/background-looping-video/adapter.ts
spuppo-mux: Actually, I forgot I was rebuilding the adapter for each src change so disregard the last comment.

spuppo-mux: Also realized that I misunderstood your question. 

Confirmed again by changing my example so that I don't rebuild the adapter on src change. Commenting the engine destroy and rebuild lead to the same results

cjpillsbury: memory profiling is actually probably overkill for this. Just confirming that playback behavior works as expected as you switch across `src` and as you switch across `mediaElement` is more than sufficient. We _**should**_ be able to `createEngine()` once and then simplify a lot of the adapter code to just do e.g. `this.#signals.state.presentation.set({ url: value });` and let the engine behave as expected.

packages/spf/src/playback/engines/background-looping-video/adapter.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    set src(value: string) {
+      this.#cancelPendingPlay();
+
+      if (value) {
+        this.#signals.state.presentation.set({ url: value });
+      } else {
+        this.#signals.state.presentation.set(undefined);
+      }
+    }
```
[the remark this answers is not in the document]
spuppo-mux: Will be addressed as a separate ticket

packages/spf/src/playback/engines/background-looping-video/engine.ts
spuppo-mux: Update: picker to be modified to implement maxResolution

cjpillsbury (approved):
Out of band callout that we'll need to make sure the "final form" of this supports "max-resolution" capping. Pre-approving on the assumption that this will be added at a later stage and to allow for forward progress.

packages/spf/package.json
```diff
@@ -32,6 +32,11 @@
       "types": "./dist/dev/hls.d.ts",
       "development": "./dist/dev/hls.js",
       "default": "./dist/default/hls.js"
+    },
+    "./background-looping-video": {
+      "types": "./dist/dev/background-looping-video.d.ts",
+      "development": "./dist/dev/background-looping-video.js",
+      "default": "./dist/default/background-looping-video.js"
```
mihar-22, line 39: We likely want to call this just `/background-video`? Our preset is just called `/background` atm, not sure I agree with it but prior decision (could be updated too after asking team).

packages/spf/src/playback/engines/background-looping-video/engine.ts
mihar-22, line 1: We could update the types here to be more terse and readble, e.g.,

- `BackgroundLoopingVideoEngineState` -> `BackgroundVideoState`
- `BackgroundLoopingVideoEngineContext` -> `BackgroundVideoContext`
- `createBackgroundLoopingVideoEngine` -> `createBackgroundVideoEngine`
- etc...

packages/spf/src/playback/engines/background-looping-video/adapter.ts
mihar-22, line 1: Similar to other file, naming could be more terse and readable.

packages/spf/src/playback/engines/background-looping-video/adapter.ts
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * @example
+ * class BackgroundLoopingVideoMedia extends BackgroundLoopingVideoMediaMixin(HTMLVideoElementHost) {}
+ *
+ * const media = new BackgroundLoopingVideoMedia();
+ * media.attach(document.querySelector('video'));
+ * media.src = 'https://stream.mux.com/abc123.m3u8';
+ * media.play();
+ */
+export function BackgroundLoopingVideoMediaMixin<Base extends Constructor<any>>(BaseClass: Base) {
```
mihar-22, line 65: `BackgroundLoopingVideoMediaMixin` reminding me of my Java days 😆 

Maybe `BackgroundVideoMixin`, similar for other names in this file?

mihar-22 (commented):
Great work @spuppo-mux, some naming feedback to consider with Christian

packages/spf/src/playback/engines/background-looping-video/engine.ts
spuppo-mux: Felt similar about the _Java-like_ names, was also wondering whether I should use `namespace` like the core modules (eg. BufferingIndicatorCore)

```ts
export namespace BackgroundVideo {
  export type State = BackgroundVideoState;
  export type Context = BackgroundVideoContext;
}
```

packages/spf/package.json
cjpillsbury: Defer to others (maybe loop in @heff ?) on the naming. Happy to move forward with whatever the teams prefers there.

heff: Yeah, "looping" is redundant in the context of the use case IMO. `background-video` for consistency with the preset feels best.

Issue #791, opened 2026-03-09
The title hover issue


decepulis:
Hovering over the player shows the native browser title tooltip which degrades UX.

<img width="1580" height="982" alt="Image" src="https://github.com/user-attachments/assets/69858478-cb1e-4bc3-bd51-dd9bd1ca7897" />

PR #1603, opened 2026-05-26
feat: add title display and media metadata state

### LINKED ISSUE #1123: Feature: Title Display
Video players commonly display a title overlay — typically at the top of the player — showing the name of the current media.

## Prior art

- **Vidstack** — Provides a [Title](https://vidstack.io/docs/player/components/display/title/) component. The title is sourced from player state (`providedTitle` or `inferredTitle`).
- **Mux Player** (built on Media Chrome) — Supports a `title` attribute on `<mux-player>`. Media Chrome itself provides a generic [`<media-text-display>`](https://www.media-chrome.org/docs/en/get-started) element that can be slotted into any chrome region.
- **Plyr** — Accepts a `title` [option](https://github.com/sampotts/plyr#options) in its constructor config. Currently used for accessibility (`aria-label` on play buttons, `title` on embed iframes) and browser MediaSession metadata rather than as a visible UI element.

## Related

- #791 — The tooltip side-effect of setting a native `title` attribute

Jerricho93:
## Summary

- Add `metadataFeature` to `packages/core` with `title` state and `setTitle` — syncs to `navigator.mediaSession.metadata` when available
- Add `TitleCore`, `TitleDataAttrs`, and `selectMetadata` selector for runtime-agnostic title logic
- Add `<media-title>` custom element (HTML) and `Title.Value` component (React) that read from the metadata store
- Add `media-title` to default skin templates (HTML and React, CSS and Tailwind variants) — fades in when the player is paused
- Default skin overlay gains a conditional top gradient that activates only when a title is visible (paused + title present)
- Sandbox derives title from source label; accepts `media-title` attribute / `title` prop on all player elements — never uses the native `title=` attribute, avoiding the browser tooltip issue (#791)

> Minimal skin title display is scoped to a follow-up PR.

## Test plan

- [ ] `pnpm -F @videojs/core test src/core/ui/title`
- [ ] `pnpm -F @videojs/html test src/ui/title`
- [ ] `pnpm -F @videojs/react test src/ui/title`
- [ ] `pnpm typecheck`
- [ ] `pnpm lint`
- [ ] Sandbox: set `media-title` on HTML player → title fades in when paused, hidden on error, no native browser tooltip on hover
- [ ] Sandbox: React skin with `title` prop → same behavior
- [ ] Verify `navigator.mediaSession.metadata.title` is set in Chrome
- [ ] `pnpm check:workspace`

Closes #1123

---

> [!NOTE]
> **Medium Risk**
> Touches core store presets and player attach lifecycle (MediaSession cleanup); broad surface area across HTML/React/skins but behavior is additive and covered by new tests.
> 
> **Overview**
> Adds **metadata** player state (`title` / `setTitle`) via a new `metadataFeature`, wired into all video/audio feature presets and `selectMetadata`. When supported, title changes sync to **`navigator.mediaSession.metadata`** on attach, detach, and destroy.
> 
> **HTML:** player elements accept **`media-title`** (not native `title`); new **`<media-title>`** reads store state through `TitleCore`. **React:** `Provider` accepts a **`title`** prop (layout + attach sync); **`Title.Value`** is exported.
> 
> **Default video/live skins** (CSS + Tailwind) add a top title overlay that shows when controls are visible and playback is **paused**, with overlay gradient tweaks when a title is present; hidden on error. Sandbox templates derive title from source labels (`useTitle` / `renderPlayerAttrs`). Includes unit tests and a metadata feature doc page.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 92b03b809f0822fc4ca88d4199d5e3d70079c41b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/sandbox/app/shared/sources.ts
```diff
@@ -3,53 +3,62 @@ import { getMuxAssetId } from './mux';
 export const SOURCES = {
   'hls-1': {
     label: 'HLS - Big Buck Bunny',
+    title: 'Big Buck Bunny',
```
sampotts: A nit and not a blocker but we could probably derive the label from `{this.type.toUpperCase()} {this.title}` to save the double up.

sampotts: Or inversely, just do `label.split(' - ')[1]` to get the title.

packages/react/src/player/create-player.tsx
```diff
@@ -69,19 +71,30 @@ export function createPlayer<const Features extends AnyPlayerFeature[]>(
...
     useDestroy(store);
 
+    const titleRef = useRef(title);
+    titleRef.current = title;
+
     useEffect(() => {
       if (!media) return;
-      return store.attach({ media, container });
+      const cleanup = store.attach({ media, container });
+      const s = store.state as { setTitle?: (t: string | null) => void };
```
sampotts: nit: can we use `state` here instead of `s`? It's a personal thing but I'm not a fan of the single letter variable names (e.g. `e` for error or event) unless it's inside a simple `.map()` or `.filter()` since it makes it a bit harder to understand.

packages/skins/src/default/css/components/overlay.css
```diff
@@ -6,7 +6,9 @@
   position: absolute;
   inset: 0;
   pointer-events: none;
-  background-image: linear-gradient(to top, oklch(0 0 0 / 0.5), oklch(0 0 0 / 0.3) 25%, oklch(0 0 0 / 0));
+  background-image:
+    linear-gradient(to bottom, oklch(0 0 0 / 0.5), oklch(0 0 0 / 0.3) 25%, oklch(0 0 0 / 0)),
+    linear-gradient(to top, oklch(0 0 0 / 0.5), oklch(0 0 0 / 0.3) 25%, oklch(0 0 0 / 0));
```
sampotts: Couldn't we combine these and just add another stop in? Or we could just go with `oklch(0 0 0 / 0.5), oklch(0 0 0 / 0), oklch(0 0 0 / 0.5)`

Also, how do we handle if there's no title/metadata? It might look a little weird to have the extra gradient if there's no title/metadata.

packages/skins/src/default/tailwind/components/overlay.ts
```diff
@@ -6,7 +6,7 @@ export const overlay = cn(
   'pointer-events-none rounded-[inherit]',
   // Default: hidden
   'opacity-0',
-  'bg-linear-to-t from-black/50 via-black/30 via-25% to-transparent',
+  '[background-image:linear-gradient(to_top,oklch(0_0_0_/_0.5),oklch(0_0_0_/_0.3)_25%,oklch(0_0_0_/_0)),linear-gradient(to_bottom,oklch(0_0_0_/_0.5),oklch(0_0_0_/_0.3)_25%,oklch(0_0_0_/_0))]',
```
sampotts: If we simplify as above, we can revert to using the native Tailwind classnames.

packages/skins/src/default/css/components/title.css
```diff
@@ -0,0 +1,29 @@
+.media-title {
+  position: absolute;
+  inset-inline: 0;
+  top: 0;
+  padding: 0.75rem 1rem;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  font-size: 1rem;
+  font-weight: 500;
+  color: var(--media-color-primary, oklch(1 0 0));
```
sampotts, line 10: Did you try without this? I think the currentColor would be --media-color-primary but I could be wrong!

packages/skins/src/default/tailwind/components/title.ts
```diff
@@ -0,0 +1,17 @@
+import { cn } from '@videojs/utils/style';
+
+export const title = cn(
+  'absolute top-0 inset-x-0',
+  'py-3 px-4',
+  'pointer-events-none',
+  'text-base font-medium',
+  '[color:var(--media-color-primary,oklch(1_0_0))]',
```
sampotts, line 8: As above re: whether we need this.

packages/skins/src/minimal/tailwind/components/overlay.ts
```diff
@@ -6,7 +6,7 @@ export const overlay = cn(
   'pointer-events-none rounded-[inherit]',
   // Default: hidden
   'opacity-0',
-  'bg-linear-to-t from-black/70 via-black/50 via-[7.5rem] to-transparent',
+  '[background-image:linear-gradient(to_top,oklch(0_0_0_/_0.5),oklch(0_0_0_/_0.3)_25%,oklch(0_0_0_/_0)),linear-gradient(to_bottom,oklch(0_0_0_/_0.5),oklch(0_0_0_/_0.3)_25%,oklch(0_0_0_/_0))]',
```
sampotts: As with the other skin but I guess this crept in?

sampotts (commented):
Looking good! Cheers for doing this. We probably want Rahim or Wes to give a once over too but I've left a few comments. 

Also, I built and tried the sandbox but couldn't see the title:
https://v10-sandbox-git-fork-jerricho93-pr-title-core-default-mux.vercel.app/?platform=react&styling=css&preset=hls-video&skin=default&source=hls-4&autoplay=0&muted=0&loop=0&preload=metadata

packages/skins/src/default/css/components/title.css
Jerricho93: `.media-title` and `.media-controls` are siblings, so we can't inherit from that. What could be done is defining color in `.media-default-skin--video` and then both components can inherit from that. Would that be a better solution?

sampotts:
It's looking good. I think we'll want some responsive styles for the font size and padding. On a smaller screen let's go with:

```css
font-size: 0.875rem;
padding: 1rem 1.5rem;
```

then using an container query for `@container media-root (width > 42rem)` to pump up the font size and padding:

```css
font-size: 1rem;
padding: 1.5rem 2rem;
```

I'd also like to add a text-shadow to the title to improve the contrast a touch. We have a shadow color based on the current color that we should be able to use here. It might need the color definition moved to the skin though as you suggested. We'd just need to make sure that `--media-color-primary` still works. 

I also reckon the title should only show when the player is paused rather than every time the controls are visible? What do you think?

Jerricho93:
@sampotts sure, I can push those changes. One question regarding displaying the title: if we're going to just display it when the video is paused, should the top gradient also be displayed only when paused? Otherwise it may cause the player to look too "dark" on the top side for no specific reason. What do you think?

Also, we held off on moving `--media-color-primary` into the skin. If we define it on `.media-default-skin--video`, the class selector would win over a consumer's `:root` override due to specificity, and we would be then ignoring their customization. This pattern keeps the variable "unowned" by the skin so the cascade stays clean. Does that make sense, or were you thinking of a different approach?

packages/skins/src/default/css/components/title.css
sampotts: Sorry for the slow response. I think this can go in as-is and we'll come back to tidying up customization and CSS custom properties. I've already added an issue for it.

apps/sandbox/app/shared/html/sandbox-state.ts
```diff
@@ -36,6 +36,13 @@ export function createHtmlSandboxState(audioOnly?: boolean): HtmlSandboxState {
   };
 }
 
+/** Render player-level attributes (media-title) as an HTML attribute string. */
+export function renderPlayerAttrs(state: HtmlSandboxState): string {
+  const { label } = SOURCES[state.source];
+  const title = label.split(' - ').slice(1).join(' - ');
+  return `media-title="${title}"`;
```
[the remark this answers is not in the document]
sampotts: Probably worth fixing this if we can to ensure we avoid any XSS issues.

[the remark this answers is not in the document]
Jerricho93: Will do. Didn't add it initially since this is the sandbox and the title comes from a fixed set of typed constants, but happy to add those checks for hygiene! Will add it to both renderPlayerAttrs and renderMediaAttrs for consistency.

mihar-22:
Closing as we're at least temporarily moving this back into discussion/design! Thank you Manuel 🙏

PR #1605, opened 2026-05-26
feat(spf): multi-track audio + skills building features and behaviors

cjpillsbury:
## TL;DR

Ships multi-language audio for the HLS engine: language-aware default selection, programmatic track selection, and same-codec mid-stream language switches without stalls or visible gaps. Also unifies the previously-separate video-quality-switching and audio-track-switching helpers into one variant-agnostic helper, so a future audio-bitrate-switching variant can drop in against the same shape.

Internal-only — no public API change.

### Bucket 1 — Runtime (focus here)

**Read fully:** `track-switching.ts` (new) — variant-agnostic body shared by video-ABR (bandwidth-driven) and audio-pin variants; shape leaves room for a future audio-bitrate variant to slot in.

**Targeted careful read:**
- `segment-loader.ts` (+90) — `staleRange` anchored at the segment containing the playhead, not the next boundary; cross-rendition switch path distinguished from same-rendition bitrate change; new-track segments survive the bandwidth-aware dedup. Stall-bug fix [`cb6cd8b2`](https://github.com/videojs/v10/pull/1605/commits/cb6cd8b2) — the same class of bug can re-emerge if `planTasks` ordering changes.
- `source-buffer.ts` (+22) — `initTrackLanguage` captured *before* the append so the segment-loader can detect cross-rendition transitions correctly.
- `select-tracks.ts` — single-writer invariant on `selectedAudioTrackId`; language-aware default selection; filter short-circuit when only one track remains.

**Skim structure only:** `quality-selection.ts` (single caller, no shim debt), `setup-buffer-actors.ts` (lifecycle revert; no stray flush effect), sandbox demo (mirrors the existing Video Renditions panel).

**Skim file:** Tests (4 files, +628 total) — assertion shape across cross-rendition switches, stall regression, default-selection tiebreaks.

### Bucket 2 — Feature-registry doc cascades (skim)

**Skim file:** `multi-language-audio.md` (+178) — what shipped this iteration vs. what's deferred (persistence, sub-segment A/V sync).
**Skim structure only:** `audio-abr.md`, `audio-playback.md` — cross-refs to the new track-switching helper match the landed code.
**Skim or skip:** `clusters.md` (+20), `audio-only-mode-override.md` — narrow cluster + use-case updates.

### Bucket 3 — Agent skills, no runtime impact (skim or skip)

`spf-implement-feature/SKILL.md` (+522 net) — frontmatter intact; new disciplines from [`0027c495`](https://github.com/videojs/v10/pull/1605/commits/0027c495) and [`effa6b3d`](https://github.com/videojs/v10/pull/1605/commits/effa6b3d) extend existing rules rather than contradicting them.

## Smoke test

**Sandbox:** `/spf-segment-loading/?muted=true&preload=auto&autoplay=true&src=https://stream.mux.com/s41JYeqIpBMBzE4OzxDyGR2yrp2hD1CQ6gJN9SlVGDQ.m3u8`

Load locally (`pnpm dev:sandbox` → paste the path into the running Vite server) or via the PR's deploy preview. The `spf-segment-loading` test page exposes manual audio-track selection; the source is a public Mux HLS stream with multiple audio-language tracks.

**Observe:**

- Default selection on first load picks an expected language (language-aware default).
- The audio-track picker is visible and lists every available language from the manifest.
- Selecting a different audio track produces a near-immediate audible language change (within ~1 second on a normal network) — no visible stall, no buffer-pause gap. The switch lands mid-segment, not at the next boundary.
- Unmute to confirm the audible language matches the picker's selection (the URL starts muted so autoplay works in browsers that block sound-on-load).

## What changed — by surface

**Multi-language audio for HLS.** Default audio-track selection is now language-aware (honoring `preferredAudioLanguage` with a fall-through to a usable track when the preferred language is absent). Programmatic selection lands as a new user-controlled filter slot, sibling to the existing video-track filter. Same-codec mid-stream language switches replan from the segment under the playhead and let MSE's overwrite-on-append carry the swap; cross-codec switches piggy-back on the existing 5.1-surround `changeType()` path. A/V re-sync after the switch lands at the next segment boundary — a starting point with room to tighten (see *Reviewer callouts*).

**Track-switching unified.** Video and audio used to have parallel-but-separate switching helpers. They're now one variant-agnostic helper with per-variant selection functions — video stays bandwidth-driven, audio pins to current. A future audio-bitrate variant can land as a third variant against the same shape, not a fork.

## Notable design decisions

- **MSE overwrite-on-append for cross-rendition switches.** Cross-rendition flushes used to be modeled as an explicit `remove [nextBoundary, ∞)` task ordered before the new track's `appendBuffer`. The refactor drops the explicit `remove` and lets MSE's overwrite-on-append carry the swap, with a new `staleRanges` concept narrowing the planner's view of "what's buffered." Alternative considered: keep the explicit `remove` task. Rejected because the gap between remove completing and the next segment landing produced an observable stall window.

- **`staleRanges` vs `removes` as separate planTasks concepts.** `removes` model content that won't be replaced (forward-buffer past the load window, back-buffer behind the keep window); `staleRanges` model content that *will* be replaced by appends at matching timestamps (cross-rendition switch). Same-rendition bitrate changes need neither — overwrite-on-append plus the actor's time-aligned dedup handles them. Alternative considered: a flag on the `remove` task (`auto-remove` vs `overwrite`). Rejected because the no-remove-task shape models the actual MSE behavior more directly.

- **Flush lives in the slot writer, not the lifecycle setup.** Initial commits parked the cross-rendition flush effect inside `setupAudioBufferActors`. The refactor moved it to `segment-loader`'s `planTasks` (which already owns the per-tick load plan). Alternative considered: leave the flush in the lifecycle behavior. Rejected because (a) it violated the "flush belongs in the slot writer" convention surfaced during the refactor, (b) it forked the audio path's shape from the video sibling, blocking the eventual audio-bitrate variant.

- **[Deferred] Persistence of programmatic audio-track selection across sessions** — selection is in-memory only. Documented in `multi-language-audio.md`.
- **[Author] A/V re-sync after a language switch lands at the next segment boundary.** Refinements (sub-segment alignment, fast-path for same-init-segment switches) are documented but not implemented. Reviewers: flag if you think this iteration should ship with a tighter sync model.
- **[Author] Cross-codec audio switch piggy-backs on the existing 5.1-surround `changeType()` path** — not separately tested in this PR. Reviewers: confirm the piggy-back is acceptable, or flag if you want a dedicated cross-codec test.

## Breaking changes

`switchVideoQuality` → `switchVideoTrack` rename and the `selectQuality` signature change are both internal SPF only; no external consumers.

## Test plan

- [x] `pnpm -F @videojs/spf test src/playback` — all pass
- [ ] E2E reliability sweep on multi-audio sources — deferred to follow-up

---

> [!NOTE]
> **Medium Risk**
> Changes live in segment loading, source-buffer metadata, and engine composition on the playback path; regressions could affect ABR or mid-stream switches, though coverage is mostly unit tests with E2E deferred.
> 
> **Overview**
> Adds **multi-language audio** to the HLS engines: **`pickAudioTrack`** becomes the default picker ( **`preferredAudioLanguage`** → default → first track), a new **`userAudioTrackSelection`** filter slot mirrors video’s constraint pattern, and **`switchAudioTrack`** owns **`selectedAudioTrackId`** with filter reactivity in both **`createSimpleHlsEngine`** and **`createHlsAudioOnlyEngine`** (replacing lifecycle-only **`selectAudioTrack`** there).
> 
> **Mid-stream language switches** are handled in **`segment-loader` `planTasks`**: **`initTrackLanguage`** on the source-buffer actor, a language-mismatch cross-rendition path, and **`staleRanges`** so shared segment IDs don’t block replanning—relying on MSE overwrite-on-append instead of an ahead-of-playhead **`remove`** for those swaps.
> 
> **Track switching is consolidated**: **`quality-switching.ts`** is removed in favor of **`track-switching.ts`** (**`switchVideoTrack`** / **`switchAudioTrack`**), public exports rename **`switchVideoQuality` → `switchVideoTrack`**, and **`selectQuality`** takes a single context object for the shared **`selectOptimal`** contract.
> 
> Sandbox **`spf-segment-loading`** gains a multi-audio Mux source and an audio-track picker; SPF feature/use-case docs and the **`spf-implement-feature`** skill are updated to match what landed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9d18a7eeb81b68b13f9261d8bd363147d3544380. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/spf/src/playback/actors/dom/segment-loader.ts
```diff
@@ -307,23 +317,80 @@ export function createSegmentLoaderActor(
...
+    // Both categories affect `effectiveBuffered` so `getSegmentsToLoad`
+    // re-plans new-track segments inside them.
+    const removes: Array<{ start: number; end: number }> = [];
+    const staleRanges: Array<{ start: number; end: number }> = [];
     if (range) {
+      if (isCrossRenditionSwitch) {
+        // Mark current-segment-start onward as stale. Falls back to the
+        // first buffered segment after currentTime when the playhead sits
+        // in a buffer gap.
+        const currentSeg = actorCtx.segments.find(
```
cjpillsbury, line 353: note: this plus the overwrite below is what allows for "fast switching" when switching languages. We don't flush the currently playing buffered audio segment, but we will overwrite it with a segment from the newly selected language track "under foot".

packages/spf/src/playback/behaviors/track-switching.ts
```diff
@@ -0,0 +1,448 @@
+/**
```
cjpillsbury, line 1: note: this is actually a rename and extension of "quality-switching". Because of the scope of changes (adding audio use cases, refactoring the video "quality switching" to be more generic, comments and renaming of things), git diff isn't showing it as an actual diff unfortunately.

apps/sandbox/templates/spf-segment-loading/main.ts
```diff
@@ -169,6 +174,64 @@ function renderRenditionPicker() {
   }
 }
 
+function renderAudioTrackPicker() {
```
cjpillsbury, line 183: note: currently adding UI for every audio track, so if there are e.g. multiple tracks for different bitrates with the same language, you'll see each of those in the UI.

packages/spf/src/playback/actors/dom/segment-loader.ts
```diff
@@ -307,23 +317,80 @@ export function createSegmentLoaderActor(
...
+    // (c) the languages differ. The buffered range from the current segment
+    // boundary forward is treated as stale (new track's same-timestamp
+    // segments will overwrite it via MSE append-at-same-timestamp); no
+    // explicit `remove` task is emitted for the cross-rendition range.
+    // Computed against the currently-buffered segments (stable reference;
+    // new track's playlist may not be resolved yet at first load).
+    const isCrossRenditionSwitch =
+      actorCtx.initTrackId !== undefined &&
+      actorCtx.initTrackId !== track.id &&
+      actorCtx.initTrackLanguage !== track.language;
```
spuppo-mux, line 332: question(non-blocking): I don't fully understand the reason to also check that the language changed? Shouldn't track id changing be enough?

packages/spf/src/playback/behaviors/select-tracks.ts
```diff
@@ -127,8 +133,15 @@ const VIDEO_TRACK_SELECTION_CONFIG = {
...
-const defaultAudioPicker: TrackPicker = (presentation) => pickFirstTrackId(presentation, 'audio');
+/**
+ * Default audio picker: three-tier per the multi-language-audio Tier 1
+ * default-selection contract — `preferredAudioLanguage` → `DEFAULT=YES` →
+ * first track. Honored only when the picker is invoked; the picker is only
+ * called from inside `'presentation-resolved'`, which gates on
+ * `isResolvedPresentation`, so the cast to `Presentation` is safe.
+ */
+const defaultAudioPicker: TrackPicker<SelectAudioTrackConfig> = (presentation, config) =>
+  pickAudioTrack(presentation as Presentation, { ...config, type: 'audio' });
```
spuppo-mux: > type: 'audio'

thought: Feels like it would be implicit; or is this casting the params?

packages/spf/src/playback/actors/dom/segment-loader.ts
cjpillsbury: In the world of exactly one audio track per language, yes, but not with e.g. ABR/multi-bitrate or multi-CDN. We typically try to avoid "aggressive buffer flushing" unless there are user-facing (e.g. language switch) or technical-constraining (e.g. switching from 5.1 surround sound to stereo) considerations. In this case, we're saying flush the buffer aggressively because we don't want a user to e.g. end up hearing English for 5 seconds when they want to be hearing Spanish.

spuppo-mux: Oh ok got it. So thinking it in CMAF HAM terms, if switching between switching sets we keep the buffer, but switching between selection sets we immediately discard the old (to be removed eventually) and buffer then new track immediately?

cjpillsbury: That's a great reframing but also draws attention to the fact that we're not properly taking advantage of switching set vs. selection set in our HLS -> CMAF-HAM modeling! Let's keep that out of scope but it's something that maybe we can work on together after background looping video! 👯 ❤️

packages/spf/src/playback/behaviors/select-tracks.ts
cjpillsbury: Good catch. Either a dum dum Claude thing or a dum dum TypeScript thing (that resulted from a different dum dum Claude thing). I'll try to task switch over to fix this ASAP.

spuppo-mux:
Last thing, not sure how blocking since it's sandbox, but when testing the example there's a weird behavior with the audio track selector. I think it is the state updating that is triggering re-renders (not so familiar with signals yet to track it down completely). After a while it stops refreshing and the picker works better.

I can attach a recording if the explanation does not make sense

apps/sandbox/templates/spf-segment-loading/main.ts
spuppo-mux: We may want to add bitrate to the name too so the difference is more clear.
image

packages/spf/src/playback/behaviors/track-switching.ts
cjpillsbury, line 1: from @mihar-22 out of band: rename this to `av-track-switching.ts` or similar to make clear that this is only about A/V (and not text)

cjpillsbury:
> Last thing, not sure how blocking since it's sandbox, but when testing the example there's a weird behavior with the audio track selector. I think it is the state updating that is triggering re-renders (not so familiar with signals yet to track it down completely). After a while it stops refreshing and the picker works better.
> 
> I can attach a recording if the explanation does not make sense

Yes please or feel free to share out of band on slack (or both!)

packages/spf/src/playback/behaviors/track-switching.ts
cjpillsbury: Gonna hold off on this rename since text tracks will eventually need to incorporate something similar here (minimally for multi-CDN/content steering) and may use this as a shared behavior.

packages/spf/src/playback/actors/dom/segment-loader.ts
cjpillsbury: Gonna resolve this but it is gesturing at a worthwhile improvement I think as a followup at some point.

spuppo-mux:
Sorry I forgot to send the recording! It was too large to share and forgot to re-record it. I see that you could fix it still, Sandbox audio track selections works much better now. 

I did a bit of testing switching tracks and seems to be working well! Tested seeking back and forwards and it behaves as expected

spuppo-mux (approved):

PR #1570, opened 2026-05-20
docs: JSDoc every public-API export in react, html, and core

decepulis:
Resolves #1558.

> **Split notice:** The skin eject-discoverability portion of this PR — the immediate fix for #1558 — has been split into #1574. That smaller PR is the recommended merge order. This PR keeps the broader rule (JSDoc on every public-API export) and the cross-package sweep for team discussion.

## Why

Per the bench data in #1558, AI coding agents Read 15–27 distinct `.d.ts` files per Video.js install — `node_modules/@videojs/{react,html,core}/dist/dev/**/*.d.ts`. The file is already in the agent's context window when it Reads it. JSDoc on the published `.d.ts` ships documentation alongside the symbol — zero extra round trips. The same comment feeds three downstream channels:

1. The api-docs-builder's reference pages (props tables, util references).
2. IDE hover-docs.
3. AI-agent context.

This PR fills the gap and codifies the rule so future exports don't slip back.

## What changed

**1. `CLAUDE.md` — "JSDoc on Published Exports" rule** (`docs(claude)` commit). Every public-API export carries JSDoc that survives into the shipped `.d.ts`. Voice A: one-sentence, declarative, active, present tense, says what the type doesn't. Inherited fields stay at the base; leaves document only the fields they add. Public class members (non-`#`, non-`@internal`) get their own summaries. Monolithic surfaces (skins + `createPlayer`) get richer prose + a single `@see` link.

Two carve-outs added after review:
- Skip the summary on **leaf wrapper types** that extend a base and add no fields (`interface FooProps extends BarProps {}`).
- Skip JSDoc on **uniform plumbing methods on `*Core` classes** (`setProps`, `getState`, `getAttrs`, `setMedia`). Behavior methods still get JSDoc.

**2. `@videojs/react` sweep** (~80 files post-split). UI primitives, hooks, player factory + context, media adapters, util helpers. `createPlayer` overloads get `@see reference/create-player`. Skin files now ship via #1574.

**3. `@videojs/html` sweep** (~90 files post-split). Custom-element classes, mixins, controllers, context modules, player factory, `*PlayerElement` shells. `createPlayer` overloads get `@see reference/html-create-player`. Skin defines now ship via #1574.

[...]

Issue #1558, opened 2026-05-18
Docs: JSDoc the Published .d.ts Exports in @videojs/react and @videojs/html — AI Agents Read 15-27 Type Files per Install


decepulis:
> [!NOTE]
> **Triage estimate** — cycle priority and story points were assigned during an AI-assisted planning pass and are rough first guesses. Confirm or adjust when you pick this up.

Filing from a video-player AI-coding-agent benchmark I run
(https://github.com/decepulis/ax-bench). Companion to #1557.

## TL;DR

- When AI coding agents (Claude Code) want to know what a prop
  is called or what a function accepts, they `Read` and `Grep`
  `.d.ts` files inside `node_modules/<pkg>/dist/` aggressively,
  before reaching for the docs site.
- In our runs, **every single v10 cell `Read` 15-27 distinct
  `.d.ts` files** across `@videojs/react`, `@videojs/html`, and
  `@videojs/core`. Across all 8 cells: **63 unique `.d.ts`
  paths** the agent treated as authoritative.
- The `.d.ts` files are already in the agent's context window
  when it `Read`s them. JSDoc on the exports means the
  documentation arrives with the symbol — zero extra round
  trips.

## Evidence

Sample tool calls (all from a single rung 1):

```json
{"name":"Read","input":{"file_path":".../node_modules/@videojs/react/dist/dev/presets/video/index.d.ts"}}
{"name":"Read","input":{"file_path":".../node_modules/@videojs/react/dist/dev/player/create-player.d.ts"}}
{"name":"Read","input":{"file_path":".../node_modules/@videojs/react/dist/dev/presets/video/minimal-skin.d.ts"}}
{"name":"Read","input":{"file_path":".../node_modules/@videojs/html/dist/dev/define/media/hls-video.d.ts"}}
{"name":"Read","input":{"file_path":".../node_modules/@videojs/core/dist/dev/core/ui/mute-button/mute-button-core.d.ts"}}
```

Per-cell distinct `.d.ts` file count across our 8 video-js runs:

| Cell | Distinct `.d.ts` files Read |
|---|---:|
| video-js-with-docs_run-0 | 27 |
| video-js-with-docs_run-1 | 23 |
| video-js-with-docs_run-2 | 16 |
| video-js_run-0 | 15 |
| video-js_run-1 | 21 |
| video-js_run-2 | 27 |
| video-js_run-3 | 22 |
| video-js_run-4 | 26 |

[...]

PR #1646, opened 2026-06-02
chore(build): enforce JSDoc presence on public-API exports in check:workspace

### LINKED ISSUE #1571: Chore: Enforce JSDoc Presence on Public-API Exports in pnpm check:workspace
> [!NOTE]

Add a JSDoc-presence check to `pnpm check:workspace` that fails CI when any public-API export in a published package lacks a JSDoc summary. Follow-up to #1558 / #1570 — enforcement was deferred in that PR (D7) so reviewers carry the burden today; this check moves it to CI.

Walk every file behind a `package.json` `"exports"` entry, parse with the TypeScript Compiler API, and assert each exported symbol has a non-empty JSDoc summary. Failure mode is a hard CI failure, not a warning.

## Scope

Initial coverage matches the sweep in #1570:

- `@videojs/react`
- `@videojs/html`
- `@videojs/core`

Expand the coverage list as follow-up sweeps land for `store`, `spf`, `element`, and `utils`.

## Out of Scope

- Per-field enforcement on `*Props` interfaces (Ring 2).
- `@see`-presence on monolithic surfaces (Ring 3, skins + `createPlayer`).

Both need richer heuristics and can layer on later.

R-Delfino95:
# PR Description

## Why

Resolves #1571.

We agreed to require JSDoc for all public API exports back in #1558 and #1570, but up until now, it was just a review-time guideline. This meant reviewers had to catch missing docs manually, and some gaps inevitably slipped through. 

This PR automates that process by adding an automated check to `pnpm check:workspace`. Now, if any public export in a published package is missing a JSDoc summary, the CI build will fail.

The check inspects the source code directly, meaning it runs quickly and doesn't require a full build step to test locally or in CI.

> **Note:** This is blocked by #1570. That PR contains the actual documentation updates. Until #1570 lands, this check will fail on several files in `@videojs/react`, `@videojs/html`, and `@videojs/core`. Recommended merge order: #1570 → this PR.

---

## What's in this PR

We added a **JSDoc presence check** to our workspace tools. It scans public exports for the following initial packages:
- `@videojs/react`
- `@videojs/html`
- `@videojs/core`

It ensures each public export has an actual descriptive summary.

### How it works (High Level)
1. **Finds Entry Points:** It reads `tsdown.config.ts` for each package to find the official public entry files.
2. **Scans Exports:** It uses the TypeScript Compiler API to track down the actual source declarations.
3. **Smart Exclusions:** - It skips exports marked as `@internal`.
   - It skips empty interfaces that just extend another interface.
   - It handles complex re-exports correctly so each component/function is only checked once.

### Why is this located under `site/`?
The script relies on the TypeScript Compiler API and `tsx` to run. Instead of adding these heavy development dependencies to the repository root, we placed the script inside `site/scripts/jsdoc-presence/` because the documentation builder there already uses them.

### How to add more packages later
When we are ready to expand coverage, you just need to add the package name to the `PACKAGES` list inside `site/scripts/jsdoc-presence/cli.ts`.

---

## Test Plan

- [x] **Integration Test:** Verified that running `pnpm check:workspace` on the current `main` branch correctly flags the existing missing JSDocs with clear error messages.
- [x] **Unit Tests:** Added 18 unit tests in `site/scripts/tests/jsdoc-presence.test.ts` to cover edge cases like internal tags, re-exports, and empty extensions.
- [x] **Manual Verification:** Temporarily added documentation to a flagged file and confirmed the checker turned green, then went back to red when removed.
- [x] **Lint & Typechecks:** Ensured `pnpm lint` and `pnpm astro check` pass successfully with the new files.

### DIFF
--- a/build/scripts/check-workspace.mjs
+++ b/build/scripts/check-workspace.mjs
@@ -11,7 +11,10 @@
  * 4. Package metadata — non-private packages have required fields
  * 5. Release-please config — every versioned package is registered
  * 6. Define imports — no bare side-effect imports from relative paths
+ * 7. Bundled docs publishing — html/react ship per-framework docs
+ * 8. JSDoc presence — every public-API export carries a JSDoc summary
  */
+import { execFileSync } from 'node:child_process';
 import { existsSync, readdirSync, readFileSync } from 'node:fs';
 import { dirname, join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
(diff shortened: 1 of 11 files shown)


mihar-22:
I love the thought behind this PR, you noticed a repeated issue and built some guardrails 👌 

I'm going to close this out because when I originally created `check-workspace` it was mostly intended for workspace-level checks but I did (lazily) sneak in the side-effect import checks for the HTML package into here. It's worth redesigning the checks by (a) pushing checks down to their respective packages, and (b) moving this workflow into `turbo` so they can be cached.

I'm currently not sure we want to document _all_ public exported APIs. Sounds reasonable but I would need to test what errors are surfaced from this to determine next steps. Important to understand how this impacts PR reviews and general workflows. Small concern that with AI and lack of guidance via a skill or something that it could just create bloat. 

Again, thank you and likely once we resolve the above we can and should circle back to exploring this check!

PR #1652, opened 2026-06-03
feat(skin): show scrubber preview timestamps

### LINKED ISSUE #1131: Feature: Show Preview Timestamp on Scrubber
Currently we only show the preview time on the scrubber if there's a storyboard due to an issue with nested `backdrop-filter` causing a difference between the tooltips (using popover, rendered in the top layer and don't suffer from the issue) and the preview time tooltip.

sampotts:
## Summary

- Add scrubber pointer timestamp previews for audio and video skins when no storyboard thumbnail is shown
- Split storyboard thumbnail styling from generic slider preview styling across CSS and Tailwind skins
- Port the updated markup and Tailwind tokens through React and HTML presets, including live skin cleanup where applicable

## Screenshots

Since the sandbox always includes a storyboard for videos. It's hard to preview it. The little preview indicator has been added to the storyboard version too though. 

### Default

CleanShot 2026-06-03 at 14 31 06@2x

CleanShot 2026-06-03 at 14 31 41@2x

### Minimal 

CleanShot 2026-06-03 at 14 32 24@2x

CleanShot 2026-06-03 at 14 39 03@2x

## Validation

- pnpm -F @videojs/skins build
- pnpm -F @videojs/react build
- pnpm build:sandbox

Closes #1131

---

> [!NOTE]
> **Low Risk**
> UI-only skin and styling changes across presets; no auth, data, or playback logic. Main review focus is visual parity when thumbnails are present vs absent.
> 
> **Overview**
> Adds **scrubber pointer previews** on time sliders across default/minimal **audio and video** skins (HTML, React, CSS, and Tailwind). Each slider now includes `media-slider-preview` / `TimeSlider.Preview` with a `type="pointer"` value so users see the hover timestamp when no storyboard is shown.
> 
> For **video**, storyboard hover UI is renamed from **preview** to **thumbnail** (`media-thumbnail`, `thumbnail.*` tokens). Positioning and CSS variables move from `--media-preview-*` to `--media-slider-thumbnail-*`. A separate **slider preview** layer sits beside the thumbnail; styling hides the text preview when a visible storyboard image is present.
> 
> **Slider preview** styling (animated timestamp + track indicator on `data-pointing`) is new in shared slider CSS/Tailwind. Seek tooltips in several audio presets drop hard-coded popup text (self-closing `Tooltip.Popup`). Live skins get the same tooltip markup cleanup only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 58c2edc127c41a1705292f9f78b1e0ca0bea7dd5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/skins/src/minimal/css/video.css
```diff
@@ -261,21 +261,27 @@
...
     var(--media-slider-pointer),
-    calc(100% - var(--media-preview-max-width) / 2 - var(--media-preview-padding) + var(--media-preview-inset))
+    calc(
+      100% -
+      var(--media-slider-thumbnail-max-width) /
+      2 -
+      var(--media-slider-thumbnail-padding) +
+      var(--media-slider-thumbnail-inset)
+    )
   );
```
mihar-22, line 285: This calculation looks extremely complicated. I can't tell what it's doing on glance but hopefully in the future we can simplify it down.

mihar-22 (approved):
Looks beautiful 👌

packages/skins/src/minimal/css/video.css
sampotts: This is about as simple as I could get it. Essentially it's to keep the thumbnail popover within the bounds of the player so we're using `cqw` units to calculate the inset and then calculating the `left` value based on that plus taking into account padding and the width. 

We _could_ (and probably should) however bake some of this logic into the `TimeSlider.Preview` component as we've got something kinda similar to calculate left etc but it doesn't take into account the boundary. Perhaps we can use some of the logic we have for popovers which already handles it nicely. I'll open an issue.

PR #1654, opened 2026-06-03
feat(spf): Add maxResolution to Background looping video

### LINKED ISSUE #1635: Feature: [SPF] Background Looping Video — Phase 4: Feature and Example cleanup.
Issue meant to track additional details caught during Epic implementation. As of now:

- Add maxResolution support
- Clean up example:
   - Remove AI slop
   - All relevant controls/config present
- Add emptied listener to track current time - Addressed on https://github.com/videojs/v10/issues/1633
- Mild refactors:
    - Move pickers to `/media` dir https://github.com/videojs/v10/pull/1602#discussion_r3318233932

spuppo-mux:
Related to https://github.com/videojs/v10/issues/1635

Added `maxResolution` config to Background Video in the form of an attribute to be consumed by the picker used on engine creation. The `this.maxResolution` is accessed when the picker is invoked, so we will always access the current value.

Also updated the sandbox example to include a max rendition picker instead of a rendition picker. Removed some no longer relevant text from example. 

**Missing from #1635:** refactoring the pickers into media/ and function renaming deferred to a follow up up PR

---

> [!NOTE]
> **Medium Risk**
> Changes core video track selection and the background-looping adapter’s default picker behavior; well covered by unit tests but affects which HLS variant loads for a given cap.
> 
> **Overview**
> Adds **`maxResolution`** on `BackgroundLoopingVideoMediaElement` (config + getter/setter) so rendition choice is capped by height labels like `720p` or by pixel area, without rebuilding the engine—the adapter’s closure picker reads the live value on each presentation resolve.
> 
> Track selection primitives gain **`maxResolutionToPixelArea`**, **`pickTrackUnderPixelArea`**, and **`pickMaxResolutionVideoTrack` → `pickHighestResolutionVideoTrack`** (refactored to share the pixel-area helper). The background engine default picker uses the renamed helper; the adapter installs its own picker that applies the cap.
> 
> The **sandbox** drops manual per-rendition switching and Phase 1/roadmap chrome in favor of a **`maxResolution`** dropdown, a **Load** button (reassigns `src` to re-run picking), and a read-only rendition list that highlights the picker’s choice.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ad7b5a6f89e37e5b0a6f3a741e146e709d4d1e82. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/spf/src/media/primitives/select-tracks.ts
```diff
@@ -179,25 +179,77 @@ export function pickVideoTrack(presentation: Presentation, config: VideoSelectio
...
+ */
+export function pickMaxResolutionVideoTrack(
+  config?: MaxResolutionPickerConfig
+): (presentation: MaybeResolvedPresentation) => string | undefined {
+  return (presentation: MaybeResolvedPresentation) => {
+    const videoSet = presentation.selectionSets?.find((set) => set.type === 'video') as VideoSelectionSet | undefined;
+    const tracks = videoSet?.switchingSets[0]?.tracks;
+    if (!tracks?.length) return undefined;
+
+    const maxHeight = parseResolutionHeight(config?.maxResolution);
```
cjpillsbury: issue(blocking): we can't rely on config here for applying `maxResolution`, since that would require recreating the engine from scratch. I think we can simplify all of this by assuming the adapter will pass in its own function that relies on the closure or context binding (via `.bind()`) to get the _**current**_ value of `maxResolution` whenever the function is invoked. Remember that we don't want to write code that forces us to re-instantiate the engine between setting sources, and we want a newly set source to use whatever `maxResolution` happens to be on loading/track selection.

I suspect a lot of this can be greatly simplified as a result. We shouldn't need a higher order function for this use case, though we may want to write a function that can be unit tested in isolation.

cjpillsbury (changes requested):
I think we need to take a step back here. This isn't quite doing what we want it to (it's close), and is probably a bit more complex than is necessary. You should be able to start by writing a function that simply takes something like a `maxPixelArea` and an array of presumed (partially resolved or fully resolved) video tracks that finds the closest track whose pixel area is below that (or, if none are below that, the track that is closest to the pixel area). Feel free to use e.g. a sort and a default `maxPixelArea` of `Number.POSITIVE_INFIINITY` for the unset cases, since the total set of video tracks should never be so large that you need to optimize this significantly. This will allow you to have a unit-testable function. Then, in the adapter, rely on a closure-defined function or bound method to pass into the config as your picker that will invoke this function, but use whatever `this.maxResolution` (translated into pixel area) happens to be at the moment of invokation.

packages/spf/src/media/primitives/select-tracks.ts
```diff
@@ -179,31 +179,56 @@ export function pickVideoTrack(presentation: Presentation, config: VideoSelectio
...
+ */
+export function maxResolutionToPixelArea(value: string | number | undefined): number {
+  if (value === undefined || value === null) return Number.POSITIVE_INFINITY;
+  if (typeof value === 'number') return Number.isFinite(value) && value > 0 ? value : Number.POSITIVE_INFINITY;
+  const match = value.trim().match(/^(\d+)p?$/i);
+  if (!match) return Number.POSITIVE_INFINITY;
+  const height = Number(match[1]);
+  if (!(Number.isFinite(height) && height > 0)) return Number.POSITIVE_INFINITY;
+  return (height * height * 16) / 9;
+}
```
spuppo-mux, line 203: thought: to include the table in [this doc](https://www.mux.com/docs/pricing/overview#resolution-based-pricing), we could rework this so it receives a `{tier?, height?, width?, aspectRatio?, pixelArea?}`

packages/spf/src/playback/engines/background-looping-video/adapter.ts
```diff
@@ -233,8 +238,11 @@ export function BackgroundLoopingVideoMediaMixin<Base extends Constructor<any>>(
     // -------------------------------------------------------------------------
 
     #createEngine(): Composition<BackgroundLoopingVideoEngineState, BackgroundLoopingVideoEngineContext> {
-      const adapterPicker: TrackPicker = (presentation) =>
-        pickMaxResolutionVideoTrack({ maxResolution: this.#maxResolution })(presentation);
+      const adapterPicker: TrackPicker = (presentation) => {
```
cjpillsbury, line 247: praise: 😎 👌

packages/spf/src/media/primitives/select-tracks.ts
cjpillsbury: I think for now this should be good. Maybe update the function description comment for a succinct and clear description: Something like:
> translates a "max resolution" into a total total pixel area for comparisons with video track resolutions with an assumed 16:9 ratio.

Maybe with an example of: `720p = 720 * 1280 = 921600`

cjpillsbury (approved):
Looks great! One minor non-blocking suggestion in response to your callout.

PR #1658, opened 2026-06-04
fix(spf): refactor track switching to rules

cjpillsbury:
## TL;DR

Refactors `setupTrackSwitching` into a **rule-chain model**: track selection runs an ordered chain of rules (soft filters + one ranker) through an `applyRules` composer, instead of a hardcoded filter-then-rank. Along the way it unwinds the type coupling between the behavior and its rules — the behavior now declares only the signals/config it owns, and each rule re-declares what it reads as optional on its own view — and lifts the candidate-track derivation into a reactive seam ready for a future hard-constraints pre-pass. Video ABR (bandwidth-driven adaptive bitrate) is at parity; the one deliberate behavior change is that audio default-language selection is dropped (see *Breaking changes*). Internal SPF only (alpha) — no external consumers.

## For reviewers — how to read this PR

Effectively a single-file change: `track-switching.ts` is the PR; the other seven files are wiring, tests, and one sandbox preset.

**Read fully:** `track-switching.ts` (608) — four things to evaluate as shapes: (1) the rule contract + `applyRules` composer — fall-through (a rule that empties the set is skipped) + early-bail (stop at one survivor, so the ranker never reads bandwidth while a single-track selection holds); (2) the behavior↔rules decoupling — the behavior's state map + config carry only `presentation` / `selected*TrackId` / lifecycle config, while each rule's signals and config live on its own optional view (`UserSelection*` / `BandwidthRanker*`); (3) `rankByBandwidth` — one threshold-sort shared by video and audio, hysteresis preserved by boosting the current track's sort weight (no temporal state); (4) `candidateSet` — the playable-set computed read reactively, the seam a constraints pre-pass will occupy.

**Targeted careful read:** `share-signals.ts` (+16/−15) — `makeShareSignals` now materializes consumer-input slots (`user*TrackSelection`) via new `inputStateKeys`/`inputContextKeys`; verify it declares those as its `stateKeys`/`contextKeys` and that the consumer-boundary rationale holds (these slots are written by the embedder, read only by the filter rule).

**Skim file:** `track-switching.test.ts` (+87/−68) — assertion shape across the rule chain; the new resolved→resolved candidate-set reactivity test; removed picker / preferred-language tests.

**Skim structure only:** `engine.ts` (materializes both `user*TrackSelection` keys), `engine-audio-only.ts` (materializes `userAudioTrackSelection`), `select-tracks.ts` (new `matchesPartialTrack` helper, single caller), `all.ts` (`TrackSwitchingConfig` → `SwitchVideoTrackConfig` re-export rename), `sources.ts` (sandbox: one audio-only preset for the smoke test).

## Smoke test

Load locally (`pnpm dev:sandbox` → paste the path into the running Vite server) or via the PR's deploy preview. Use the **`spf-segment-loading`** harness — the default player skins don't surface rendition/audio-track controls, but this harness exposes rendition buttons, audio-track buttons, live throughput, and now-playing quality (exactly the track-switching surface this PR changes).

**1. Renditions + multi-language audio + ABR** — `/spf-segment-loading/?src=https://stream.mux.com/s41JYeqIpBMBzE4OzxDyGR2yrp2hD1CQ6gJN9SlVGDQ.m3u8`

- Press **Play** (the harness video is `preload="none"`).
- **Observe:**
  - The presentation resolves: the rendition list fills (640×286 → 3840×1714) and the audio-track list shows all languages (und/fr/es/de/it/pt/hi).
  - ABR (bandwidth-driven) selects a rendition for the measured throughput (marked `⟳`); throttle in Chrome DevTools → Network ("Slow 4G") and the selection steps **down**; clear the throttle and it steps back **up** (watch now-playing-quality + throughput).
  - Click a rendition → it locks (`🔒`, ABR disabled); "Re-enable ABR" returns to `⟳`.
  - Click an audio language → it switches **mid-stream with no stall or audio gap** (the track shows `🔒`; "Clear filter" returns to the default).
  - On first load the audio sits on `und · Default` — the bandwidth-ranked default, **not** an auto-picked preferred language (the intended change; see *Breaking changes*).

**2. Audio-only engine** — `/html-simple-hls-audio-only/?source=hls-audio-only-cmaf`

- Press play.
- **Observe:** the CMAF/fmp4 audio-only asset plays through cleanly (the audio-only composition omits the video behaviors).

## What changed — by surface

**Rule-chain model.** `setupTrackSwitching` no longer hardcodes a filter-then-rank pair. Selection is an ordered chain of rules run by `applyRules` (fall-through + early-bail); each variant supplies its chain via config and the helper just runs it and takes the head as the pick. Today both variants run `[filterByUserSelection, rankByBandwidth]`.

**Behavior↔rules decoupling (state + config).** The behavior previously declared every signal and config field any rule might read (`bandwidthState`, `user*TrackSelection`, ABR tuning). It now declares only what it owns — `presentation`, `selected*TrackId`, and its lifecycle config. Each rule re-declares the signal/config it reads as optional on its own view and reads it defensively. The consumer-input `user*TrackSelection` slots are materialized at the consumer boundary by `shareSignals`; `bandwidthState` comes from the buffer-actor sampler — `createComposition` hands full state to every behavior, so a composition without a sampler (audio-only) simply ranks on `initialBandwidth`.

**Shared bandwidth ranker.** One `rankByBandwidth` rule replaces the video ABR wrapper and the audio pin-to-current: fitting tracks (within the throughput threshold) highest-bitrate-first, over-throughput tracks least-over-first; the head is the pick. Hysteresis is kept without temporal state by boosting the current track's effective bitrate by `upgradeMargin` in the fitting sort.

**Constraints-readiness seam.** The candidate-track derivation moves out of the selection effect into a `candidateSet` computed, read reactively. A future hard-constraints pre-pass (capability probing, multi-CDN failover) narrows the set there; the effect re-picks when the playable set actually changes. No constraint code yet — just the reactive shape.

## Notable design decisions

- **Dropped the empty-slot audio picker now rather than carrying it alongside the chain.** The old `pickAudioTrack` set a preferred-language / default audio track on load; it's removed and `preferredAudioLanguage` is inert until it returns as a standing soft-filter rule. Alternative considered: keep the picker until its replacement rule lands so nothing regresses. Rejected because the picker and the chain both write `selectedAudioTrackId` — running them together reintroduces the multi-writer coupling the refactor removes. The regression is internal/alpha (flagged in callouts).
- **Tightly-coupled reads as the rule↔signal contract.** A rule names the state/context/config it needs on its own deps view and reads it at apply time. Alternative considered: a rule declares its dependencies as metadata the behavior aggregates up (keeping `stateKeys` exhaustive), or the behavior hands rules a prebuilt context. Rejected for now because metadata-aggregation collides with early-bail (you'd subscribe to every rule's signals up front, defeating the lazy chain), and this is the design's still-open contract question — tightly-coupled reads are the documented starting point.
- **`candidateSet` is read reactively (`.get()`), not peeked.** Alternative considered: keep the prior `peek(state.presentation)` (no subscription to the track set). Rejected because a future dynamic constraint (CDN failover) must re-pick when the playable set changes while a presentation stays resolved — `peek` can't see that. A track-id **set** equality on the computed keeps it from re-firing on presentation churn that leaves the set intact (a live playlist refresh swaps the presentation object but not the tracks).

- **[Author] `preferredAudioLanguage` is inert.** Audio default-language / default-track selection on load is gone until it returns as a standing soft-filter rule — the most user-visible change on the branch. Good enough for alpha; follow-up will reintroduce it as a rule.
- **[Author] `defineBehavior` context is threaded via a rest-spread stopgap.** A typed `context` param on the variant setup would widen `ContextMap` to its `AnySlotMap` bound (contravariant inference) and force the slot required, breaking direct `.setup({ state })` calls. The variants forward context untyped through `{ config, ...otherProps }`; it rides through at runtime. Proper fix = a typed pass-through-context affordance on `defineBehavior` (deferred).
- **[Author] The constraints seam is shape-only.** No `applyConstraints` / `constraints` config yet, and an empty playable set is a silent early-return, not a modeled "nothing playable" not-ready state — left to the constraints work.
- **[Deferred] Feature docs not yet updated.** `multi-language-audio.md` (dropped default selection), `audio-abr.md` (now the shared ranker), `video-abr.md` (impl surface), `audio-playback.md` (default-selection note re-opens) are owed a combined pass at merge — deferred because they track `main` and the recommendation doc is still draft.

## Breaking changes

Internal SPF only (alpha; no external consumers). Audio track selection drops its default-language / default-track pick on load (`preferredAudioLanguage` now unconsumed) and moves from pin-to-current to bandwidth ranking — audio with differing per-rendition bitrates now selects by bandwidth. The `TrackSwitchingConfig` → `SwitchVideoTrackConfig` export rename is internal SPF as well.

## Test plan

- [x] Full SPF suite green (`pnpm -F @videojs/spf test` — 925 passed / 13 skipped), including a new resolved→resolved candidate-set reactivity test and the `applyRules` composer unit tests.
- [x] Smoke-checked in the `spf-segment-loading` harness via automated browser: 7 renditions + 7 audio languages enumerate, manual rendition lock disables ABR, audio language switch is mid-stream with no stall, and the audio-only CMAF asset plays. ABR-under-throttle left as a reviewer hands-on check (see *Smoke test*).
- [ ] Feature-doc cascade — deferred to merge (see callouts).

---

> [!NOTE]
> **Medium Risk**
> Changes core playback selection (video ABR + audio defaults) and removes preferred-language on load; scope is internal alpha but affects mid-stream track behavior and embedder signal wiring.
> 
> **Overview**
> Refactors SPF **track switching** from a hardcoded filter-then-`selectQuality` path into an ordered **rule chain** (`filterByUserSelection` → `rankByBandwidth`) composed by **`applyRules`** (soft-filter fall-through + early-bail when one candidate remains so bandwidth isn’t subscribed under a locked user pick).
> 
> The behavior now owns only **`presentation`** and **`selected*TrackId`**; rules read **`user*TrackSelection`** and **`bandwidthState`** via optional deps. **`makeShareSignals`** can **materialize** embedder-written `user*TrackSelection` slots; HLS engines wire those keys. Playable candidates move to a reactive **`candidateSet`** computed (id-set equality) so picks update when the track set changes while still resolved.
> 
> Video ABR stays at parity via **`rankByBandwidth`** (hysteresis + **`resolutionArea`** tie-break). **Audio** drops empty-slot **`pickAudioTrack`** / **`preferredAudioLanguage`** and shares the bandwidth ranker (manifest-order default when bitrates tie). Public rename: **`TrackSwitchingConfig` → `SwitchVideoTrackConfig`**. Sandbox adds an **audio-only CMAF** source.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17c111e316d356158f7c4237c235a45519cc6514. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/spf/src/media/primitives/select-tracks.ts
+++ b/packages/spf/src/media/primitives/select-tracks.ts
@@ -113,6 +113,24 @@ export type TrackPicker<Config = unknown> = (
 ) => string | undefined;
 
 /**
+ * Test whether a track matches a partial-track description: every present,
+ * defined field of `filter` equals the track's. Absent or `undefined` filter
+ * fields don't constrain. Used to narrow candidates by a user selection
+ * (`{ id }`, `{ language }`, `{ height }`, …).
+ *
+ * @param track - The track to test
+ * @param filter - Partial-track description; only present, defined fields constrain
+ * @returns `true` when the track matches every constraining field
+ */
+export function matchesPartialTrack<T>(track: T, filter: Partial<T>): boolean {
+  for (const key in filter) {
+    const filterValue = filter[key as keyof T];
+    if (filterValue !== undefined && track[key as keyof T] !== filterValue) return false;
+  }
+  return true;
+}
+
+/**
  * Pick the first track of the given type from a presentation.
  *
  * Returns the first track in the first switching set of the matching
--- a/packages/spf/src/playback/behaviors/track-switching.ts
+++ b/packages/spf/src/playback/behaviors/track-switching.ts
@@ -88,7 +88,7 @@ export interface TrackSwitchingState {
 
 /**
  * Config for `switchVideoTrack` — the ABR tuning read by its ranker rule
- * (`rankByQuality`). `quality.safetyMargin` is the bandwidth-headroom
+ * (`rankByBandwidth`). `quality.safetyMargin` is the bandwidth-headroom
  * multiplier; `quality.upgradeMargin` the hysteresis ratio gating upgrades;
  * `bandwidth` tunes the estimator; `initialBandwidth` is the pre-sample
  * fallback. Defaults: `DEFAULT_QUALITY_CONFIG` (0.85 / 1.15),
@@ -201,9 +201,14 @@ type UserSelectionKey = 'userVideoTrackSelection' | 'userAudioTrackSelection';
 // it in detaches the user-selection mapped value from `P` and TS collapses
 // the intersection. T flows through `TrackSwitchingSetupConfig` instead;
 // the user-filter access casts at the read site (see below).
+//
+// `bandwidthState` is intentionally absent: only the bandwidth ranker rule
+// reads it, not the behavior's lifecycle. A rule needing a signal the behavior
+// doesn't use declares it *optional* on its own deps and reads it defensively
+// (see `BandwidthRankerStateMap`), so the behavior never assumes a rule-only
+// signal exists.
 type TrackSwitchingStateMap<S extends SelectionKey, U extends UserSelectionKey> = {
   presentation: ReadonlySignal<TrackSwitchingState['presentation']>;
-  bandwidthState: ReadonlySignal<TrackSwitchingState['bandwidthState']>;
 } & { [P in S]: Signal<TrackSwitchingState[P]> } & { [P in U]: ReadonlySignal<TrackSwitchingState[P]> };
 
 /**
@@ -233,6 +238,18 @@ type TrackSwitchingRuleDeps<
   U extends UserSelectionKey,
   T extends SwitchableTrack,
 > = SelectionRuleDeps<TrackSwitchingStateMap<S, U>, AnySlotMap, TrackSwitchingSetupConfig<S, U, T>>;
+
+/**
+ * State a bandwidth ranker reads: the behavior's lifecycle map plus an
+ * *optional* `bandwidthState`. The signal exists only when the composition
+ * includes a bandwidth sampler; the ranker reads it defensively and falls back
+ * to `initialBandwidth` (with a debug note) when it's absent. A behavior state
+ * map without `bandwidthState` is assignable here (the slot is optional), so a
+ * ranker drops into a chain whose behavior never declared the signal.
+ */
+type BandwidthRankerStateMap<S extends SelectionKey, U extends UserSelectionKey> = TrackSwitchingStateMap<S, U> & {
+  bandwidthState?: ReadonlySignal<TrackSwitchingState['bandwidthState']>;
+};
 
 type VideoTrackCandidate = PartiallyResolvedVideoTrack | VideoTrack;
 type AudioTrackCandidate = PartiallyResolvedAudioTrack | AudioTrack;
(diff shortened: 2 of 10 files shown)


packages/spf/src/media/primitives/select-tracks.ts
```diff
@@ -112,6 +112,24 @@ export type TrackPicker<Config = unknown> = (
...
+ * (`{ id }`, `{ language }`, `{ height }`, …).
+ *
+ * @param track - The track to test
+ * @param filter - Partial-track description; only present, defined fields constrain
+ * @returns `true` when the track matches every constraining field
+ */
+export function matchesPartialTrack<T>(track: T, filter: Partial<T>): boolean {
+  for (const key in filter) {
+    const filterValue = filter[key as keyof T];
+    if (filterValue !== undefined && track[key as keyof T] !== filterValue) return false;
```
spuppo-mux, line 128: question(non-blocking): Is checking for `filterValue !== undefined` just defensive? Is there any case where one would want to filter by {language: undefined}?

cjpillsbury: Just defensive! I don't think we'd ever want to filter on `undefined`, though if some weird edge case scenario comes up, we can always refactor.

spuppo-mux (approved):

packages/spf/src/playback/behaviors/track-switching.ts
```diff
@@ -415,24 +493,17 @@ const selectAudioCurrent = (
...
+      ...otherProps,
       state,
       config: {
-        ...config,
         selectionKey: 'selectedAudioTrackId',
         userSelectionKey: 'userAudioTrackSelection',
         getTracks: (presentation) => getTracksByType(presentation, 'audio') as readonly AudioTrackCandidate[],
-        selectOptimal: selectAudioCurrent,
-        picker: config?.picker ?? pickAudioTrack,
+        rules: [filterByUserSelection, rankByBandwidth],
```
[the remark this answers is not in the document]
cjpillsbury: This will be resolved in a followup PR.

Issue #1111, opened 2026-03-24
bug(html): premade selectors incompatible with createPlayer().PlayerController types


decepulis:
## Problem

Premade selectors like `selectPlayback`, `selectTime`, `selectVolume` are typed as `Selector<object, ...>`. When used with `PlayerController` obtained from `createPlayer({ features: videoFeatures })`, TypeScript rejects them because `PlayerController.Constructor<VideoPlayerStore>` expects `Selector<InferStoreState<VideoPlayerStore>, Result>`, and `object` doesn't satisfy the full intersection type.

Discovered while fixing `astro check` errors in #1109 — specifically in the `html-create-player` and `player-controller` demo files:
- `site/src/components/docs/demos/html-create-player/html/css/BasicUsage.ts`
- `site/src/components/docs/demos/player-controller/html/css/BasicUsage.ts`

## Why it works internally

Internal elements (e.g., `play-button-element.ts`, `time-slider-element.ts`) use the module-level `playerContext` typed as `PlayerContext<PlayerStore>` (the base type). `InferStoreState<PlayerStore>` resolves broadly enough that `Selector<object, ...>` is compatible. The public `createPlayer()` API locks the context to `VideoPlayerStore`, making the same selectors incompatible.

## Reproduction

```ts
import { createPlayer, MediaElement, selectPlayback } from '@videojs/html';
import { videoFeatures } from '@videojs/html/video';

const { PlayerController, context } = createPlayer({ features: videoFeatures });

class MyElement extends MediaElement {
  // TS error: Selector<object, MediaPlaybackState | undefined> is not assignable to
  // Selector<InferStoreState<VideoPlayerStore>, Result>
  readonly #player = new PlayerController(this, context, selectPlayback);
}
```

## Expected behavior

Premade selectors should type-check with `createPlayer().PlayerController` since this is the intended public API pattern.

## Possible fixes

- Widen `PlayerController.Constructor`'s selector constraint to accept `Selector<object, ...>`
- Type premade selectors against the store state rather than `object`
- Adjust `createSelector()` return type

decepulis:
### Temporary workaround landed in #1403

To unblock the `astro check` errors called out in #1109, the two demos that hit this bug (`html-create-player` and `player-controller`) now import `PlayerController` directly from `@videojs/html` instead of destructuring it from `createPlayer()`. A `NOTE` comment in each file points back to this issue.

**Please undo that workaround when this is fixed properly** — both demo files should go back to:

```ts
[...]

PR #1663, opened 2026-06-05
docs(react): clarify usePlayer typed vs untyped store access

### LINKED ISSUE #1483: Docs: usePlayer() Return Type Is Unclear — Selector Required for Typed Access
`usePlayer() `returns `unknown` when called without arguments. The correct usage requires passing a selector to get typed state, but this is not explained prominently in the docs.

`// ✅ typed access requires a selector
const playback = usePlayer(selectPlayback);
`

Jerricho93:
Closes #1483

## Summary

The `use-player` reference page mentioned the typed-vs-untyped distinction only in a single easy-to-miss sentence, so users assumed the standalone `usePlayer()` is supposed to return `unknown`. This makes the distinction prominent on the page and corrects the standalone hook's generated reference.

## Changes

- Add a note callout to the `use-player` reference page: `Player.usePlayer` (from `createPlayer`) is typed to the configured features and needs no selector, while the standalone `import { usePlayer }` returns an untyped `UnknownStore` and needs a premade selector to recover typing.
- Enrich the standalone `usePlayer` JSDoc so the generated reference explains the `UnknownStore` return — and that inline selectors stay `unknown` while premade selectors (e.g. `selectPlayback`) carry their type.

## Testing

- `pnpm -F site api-docs` — regenerates the util reference; confirmed `use-player.json` now includes the new JSDoc paragraphs.
- `pnpm build:site` — site builds and the page renders.

> Related: #1111 (premade selectors incompatible with `createPlayer().PlayerController` types) is a separate types bug and is **not** addressed here.

---

> [!NOTE]
> **Low Risk**
> Documentation and JSDoc only; no runtime or type behavior changes.
> 
> **Overview**
> Improves discoverability of **typed vs. untyped** `usePlayer` behavior so users don’t treat standalone `unknown` returns as a bug.
> 
> The **`use-player` reference** adds a note callout and a short example: `Player.usePlayer` from `createPlayer()` is feature-typed without a selector; standalone `import { usePlayer } from '@videojs/react'` yields an **`UnknownStore`**, with typing recoverable via premade selectors like `selectPlayback`.
> 
> **JSDoc** on the standalone `usePlayer` overloads now states the same — no feature knowledge, `UnknownStore` / `UnknownState`, inline selectors stay `unknown`, premade selectors carry their return type — so generated API reference matches the docs page.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c447e2e5a6d4874cbc8a671b2ded1131ca6e03d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/site/src/content/docs/reference/use-player.mdx
+++ b/site/src/content/docs/reference/use-player.mdx
@@ -4,6 +4,7 @@ description: Hook to access the player store from within a Player.Provider
 ---
 
 import UtilReference from "@/components/docs/api-reference/UtilReference.astro";
+import Aside from "@/components/Aside.astro";
 import Demo from "@/components/docs/demos/Demo.astro";
 import DocsLink from "@/components/docs/DocsLink.astro";
 import FrameworkCase from "@/components/docs/FrameworkCase.astro";
@@ -16,6 +17,18 @@ import selectorReactTsx from "@/components/docs/demos/use-player/react/css/Selec
 import selectorReactCss from "@/components/docs/demos/use-player/react/css/Selector.css?raw";
 
 `Player.usePlayer` gives you access to the player store from any component within a `Player.Provider`. It has two overloads — without arguments for direct store access, or with a selector for reactive state subscriptions. `Player.usePlayer` is typed to the features you passed to <DocsLink slug="reference/create-player">`createPlayer`</DocsLink>. It's also available as a standalone import (`import { usePlayer } from '@videojs/react'`), but the standalone version returns an untyped store.
+
+<Aside type="note" title="Typed vs. untyped access">
+`Player.usePlayer` (from <DocsLink slug="reference/create-player">`createPlayer`</DocsLink>) is typed to your features and needs no selector. The standalone `import { usePlayer } from '@videojs/react'` returns an untyped `UnknownStore` — pass a premade selector to recover typing.
+</Aside>
+
+```tsx
+import { usePlayer } from '@videojs/react';
+import { selectPlayback } from '@videojs/core/dom';
+
+usePlayer();               // UnknownStore (state: unknown)
+usePlayer(selectPlayback); // PlaybackState | undefined
+```
 
 ## Examples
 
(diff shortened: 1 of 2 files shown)


site/src/content/docs/reference/use-player.mdx
```diff
@@ -17,6 +18,18 @@ import selectorReactCss from "@/components/docs/demos/use-player/react/css/Selec
...
+`Player.usePlayer` (from <DocsLink slug="reference/create-player">`createPlayer`</DocsLink>) is typed to your features and needs no selector. The standalone `import { usePlayer } from '@videojs/react'` returns an untyped `UnknownStore` — pass a premade selector to recover typing.
+</Aside>
+
+```tsx
+import { usePlayer } from '@videojs/react';
+import { selectPlayback } from '@videojs/core/dom';
+
+usePlayer();               // UnknownStore (state: unknown)
+usePlayer(selectPlayback); // PlaybackState | undefined
+```
```
decepulis, line 27: 1. blocking: we try not to expose @videojs/core. Luckily, @videojs/react re-exports the important bits.
2. suggestion: I wonder if we need the aside. The code block feels visually-prominent enough.
3. suggestion: The aside feels redundant with some of what comes above it. Maybe we can combine the two... something like this?

What do you think of a change like this?
````
`Player.usePlayer` (from <DocsLink slug="reference/create-player">`createPlayer`</DocsLink>) gives you access to the player store from any component within a `Player.Provider`. It's typed to the features you passed to `createPlayer`. It has two overloads — without arguments for direct store access, or with a selector for reactive state subscriptions.

`usePlayer` also available as a standalone import (`import { usePlayer } from '@videojs/react'`), but the standalone version returns an `UnknownStore`. Pass a premade selector to recover typing:

```tsx
import { usePlayer, selectPlayback } from '@videojs/react';

usePlayer();               // UnknownStore (state: unknown)
usePlayer(selectPlayback); // PlaybackState | undefined
```

````

decepulis (changes requested):
Close. One blocking comment (we don't want to tell folks about @videojs/core) and two suggestions to tighten up the copy a bit

site/src/content/docs/reference/use-player.mdx
Jerricho93: This looks great! I'll push the changes right now. Will change the blocking one, and apply both suggestions, since they help understand cleanly what we want to inform the users

decepulis (approved):
Excellent. Thank you!

PR #1669, opened 2026-06-09
feat(packages): add time display toggle

### LINKED ISSUE #1422: Feature: Time Display Inversion (Elapsed / Remaining)
Click the time display to toggle between elapsed and remaining time. Configurable default direction.

## Prior art

- **Plyr:** `invertTime` config (default true), `toggleInvert` option
- **Vidstack:** `invertTime` and `toggleTime` props
- **Mux Player:** `default-show-remaining-time` attribute

sampotts:
## Summary

Slightly different to the issue in that:

- `type="duration" toggle` switches duration ↔ remaining.
- `type="remaining" toggle` still switches remaining ↔ duration.
- `type="current" toggle` still switches current ↔ remaining.

Why? I noticed Apple Music toggles between duration and remaining and that pattern also worked well with the default skin. Minimal's time layout suited the standard toggle between current and remaining. Happy to run through it on a call if need be. 

- Add a toggle prop/attribute for time displays across core, React, and HTML.
- Update docs and add focused coverage for core, React, and HTML time behavior.

## Validation
- pnpm -F @videojs/core test src/core/ui/time
- pnpm -F @videojs/react test src/ui/time
- pnpm -F @videojs/html test src/ui/time
- pnpm -F @videojs/skins build
- git diff --check

Closes #1422

---

> [!NOTE]
> **Low Risk**
> Player UI and accessibility labeling changes with broad test coverage; no auth, data, or playback pipeline changes.
> 
> **Overview**
> Adds a **`toggle`** prop/attribute so time displays can switch between paired modes: **`current` ↔ `remaining`**, or **`duration` ↔ `remaining`** (initial view still comes from `type`).
> 
> **`TimeCore`** gains toggle-aware labels and attrs: default accessibility text is the human **`phrase`** (not static role names), **`aria-valuetext` is removed**, and when toggle is on the element gets **`role="button"`**, **`tabIndex`**, and an **`aria-label`** that includes the next action (e.g. “Show remaining time.”).
> 
> **`TimeElement`** and **`Time.Value`** track an active display type, handle click/Enter/Space, respect `preventDefault`, and reset when `type` or `toggle` changes or media state is missing.
> 
> Built-in **audio/video skins** wire this up: **minimal** skins toggle **current** time; **default** skins use **`toggle type="remaining"`** on the end time slot (duration ↔ remaining). **Skin CSS** adds focus-visible styling for toggleable times. **Docs** and **tests** cover core, HTML, and React behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee8b10fd38cc8dbeed1dbf3d5bcdf28441d49eba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/core/ui/time/time-core.ts
+++ b/packages/core/src/core/ui/time/time-core.ts
@@ -15,6 +15,8 @@ export interface TimeProps {
   negativeSign?: string | undefined;
   /** Custom label for accessibility. */
   label?: string | ((state: TimeState) => string) | undefined;
+  /** Whether the time display can be toggled. */
+  toggle?: boolean | undefined;
 }
 
 export interface TimeState {
@@ -32,10 +34,10 @@ export interface TimeState {
   datetime: string;
 }
 
-const DEFAULT_LABELS: Record<TimeType, string> = {
-  current: 'Current time',
-  duration: 'Duration',
-  remaining: 'Remaining',
+const TOGGLE_LABELS: Record<TimeType, string> = {
+  current: 'Show elapsed time',
+  duration: 'Show duration',
+  remaining: 'Show remaining time',
 };
 
 export class TimeCore {
@@ -43,6 +45,7 @@ export class TimeCore {
     type: 'current',
     negativeSign: '-',
     label: '',
+    toggle: false,
   };
 
   #props = { ...TimeCore.defaultProps };
@@ -98,7 +101,15 @@ export class TimeCore {
     return secondsToIsoDuration(Math.abs(seconds));
   }
 
-  getLabel(state: TimeState): string {
+  #getToggleType(type: TimeType, currentType: TimeType): TimeType {
+    if (type === 'current') {
+      return currentType === 'remaining' ? 'current' : 'remaining';
+    }
+
+    return currentType === 'duration' ? 'remaining' : 'duration';
+  }
+
+  getLabel(state: TimeState, type = this.#props.type): string {
     const { label } = this.#props;
 
     if (isFunction(label)) {
@@ -108,13 +119,20 @@ export class TimeCore {
       return label;
     }
 
-    return DEFAULT_LABELS[this.#props.type];
+    if (!this.#props.toggle) {
+      return state.phrase;
+    }
+
+    const toggleType = this.#getToggleType(type, state.type);
+
+    return `${state.phrase}. ${TOGGLE_LABELS[toggleType]}.`;
   }
 
-  getAttrs(state: TimeState) {
+  getAttrs(state: TimeState, type = this.#props.type) {
     return {
-      'aria-label': this.getLabel(state),
-      'aria-valuetext': state.phrase,
+      'aria-label': this.getLabel(state, type),
+      role: this.#props.toggle ? 'button' : undefined,
+      tabIndex: this.#props.toggle ? 0 : undefined,
     };
   }
 
--- a/packages/html/src/define/audio/minimal-skin.tailwind.ts
+++ b/packages/html/src/define/audio/minimal-skin.tailwind.ts
@@ -89,7 +89,7 @@ function getTemplateHTML() {
 
           <div class="${time.controls}">
             <media-time-group class="${time.group}">
-              <media-time type="current" class="${time.current}"></media-time>
+              <media-time toggle type="current" class="${time.current}"></media-time>
               <media-time-separator class="${time.separator}"></media-time-separator>
               <media-time type="duration" class="${time.duration}"></media-time>
             </media-time-group>
--- a/packages/html/src/ui/time/time-element.ts
+++ b/packages/html/src/ui/time/time-element.ts
@@ -1,7 +1,7 @@
 import { TimeCore, TimeDataAttrs, type TimeType } from '@videojs/core';
 import { applyElementProps, applyStateDataAttrs, logMissingFeature, selectTime } from '@videojs/core/dom';
 import type { PropertyDeclarationMap, PropertyValues } from '@videojs/element';
-
+import { isInteractiveActivation } from '@videojs/utils/dom';
 import { playerContext } from '../../player/context';
 import { PlayerController } from '../../player/player-controller';
 import { MediaElement } from '../media-element';
@@ -13,11 +13,13 @@ export class TimeElement extends MediaElement {
     type: { type: String },
     negativeSign: { type: String, attribute: 'negative-sign' },
     label: { type: String },
+    toggle: { type: Boolean },
   } satisfies PropertyDeclarationMap<keyof TimeCore.Props>;
 
   type: TimeType = TimeCore.defaultProps.type;
   negativeSign = TimeCore.defaultProps.negativeSign;
   label = TimeCore.defaultProps.label;
+  toggle = TimeCore.defaultProps.toggle;
 
   readonly #core = new TimeCore();
   readonly #state = new PlayerController(this, playerContext, selectTime);
@@ -25,8 +27,15 @@ export class TimeElement extends MediaElement {
   readonly #signSpan = document.createElement('span');
   readonly #textNode = document.createTextNode('');
 
+  #disconnect: AbortController | null = null;
+  #listening = false;
+  #activeType: TimeType = TimeCore.defaultProps.type;
+
   override connectedCallback(): void {
     super.connectedCallback();
+
+    this.#disconnect = new AbortController();
+    this.#syncListeners();
 
     if (!this.#signSpan.parentNode) {
       this.#signSpan.setAttribute('aria-hidden', 'true');
@@ -40,18 +49,39 @@ export class TimeElement extends MediaElement {
     }
   }
 
+  override disconnectedCallback(): void {
+    super.disconnectedCallback();
+    this.#disconnect?.abort();
+    this.#disconnect = null;
+    this.#listening = false;
+  }
+
   protected override willUpdate(changed: PropertyValues): void {
     super.willUpdate(changed);
-    this.#core.setProps(this);
+    if (changed.has('type') || changed.has('toggle')) {
+      this.#activeType = this.type;
+    }
   }
 
   protected override update(changed: PropertyValues): void {
     super.update(changed);
 
+    if (changed.has('toggle')) {
+      this.#syncListeners();
+    }
+
     const media = this.#state.value;
+    if (!media) {
+      this.#clearAttrs();
+      return;
+    }
 
-    if (!media) return;
-
+    this.#core.setProps({
+      type: this.toggle ? this.#activeType : this.type,
+      negativeSign: this.negativeSign,
+      label: this.label,
+      toggle: this.toggle,
+    });
     this.#core.setMedia(media);
     const state = this.#core.getState();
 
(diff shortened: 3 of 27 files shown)


packages/core/src/core/ui/time/time-core.ts
```diff
@@ -38,11 +40,17 @@ const DEFAULT_LABELS: Record<TimeType, string> = {
   remaining: 'Remaining',
 };
 
+const TOGGLE_LABELS: Partial<Record<TimeType, string>> = {
+  current: 'Current time, press to show remaining time',
+  remaining: 'Remaining time, press to show current time',
+};
```
sampotts: Will need to update i18n PRs when this is merged.

packages/core/src/core/ui/time/time-core.ts
```diff
@@ -15,6 +15,8 @@ export interface TimeProps {
   negativeSign?: string | undefined;
   /** Custom label for accessibility. */
   label?: string | ((state: TimeState) => string) | undefined;
+  /** Whether current and remaining time can be toggled. */
+  toggle?: boolean | undefined;
```
sampotts, line 19: Happy to rename this prop but ideally I'd like to keep it a single word if possible since longer feels weird to me. Other possible options I thought of:

- `interactive` although it's a bit broad
- `toggleable` not a real word
- ?

packages/html/src/define/audio/minimal-skin.tailwind.ts
```diff
@@ -74,7 +74,7 @@ function getTemplateHTML() {
 
           <div class="${time.controls}">
             <media-time-group class="${time.group}">
-              <media-time type="current" class="${time.current}"></media-time>
+              <media-time toggle type="current" class="${time.current}"></media-time>
```
sampotts, line 92: I only added it to the minimal skin as the default skin felt a bit better with the remaining in place of the duration. We'll revisit this before GA though I guess as we may want to make minimal a bit more minimal in terms of UI. I'm just waiting for the menus stuff to merge so we're reasonably feature stable for the skins.

packages/core/src/core/ui/time/time-core.ts
```diff
@@ -38,11 +40,18 @@ const DEFAULT_LABELS: Record<TimeType, string> = {
   remaining: 'Remaining',
 };
 
+const TOGGLE_LABELS: Record<TimeType, string> = {
+  current: 'Current time, press to toggle display',
+  duration: 'Duration, press to toggle display',
+  remaining: 'Remaining time, press to toggle display',
+};
```
mihar-22, line 41: I assume this would complicate the i18n work. I think the ARIA attributes are already indicating that this can be pressed/toggled. YT seems to only require the addition of role and tabIndex. Ideally we can remove.

packages/html/src/ui/time/time-element.ts
```diff
@@ -13,45 +13,79 @@ export class TimeElement extends MediaElement {
...
+    if (!this.#disconnect) {
+      this.#disconnect = new AbortController();
+      applyElementProps(
+        this,
+        {
+          onClick: this.#handleClick,
+          onKeyDown: this.#handleKeyDown,
+        },
+        { signal: this.#disconnect.signal }
+      );
```
mihar-22: For perf reasons it'd be great if these event listeners are attached when `toggle` is true.

packages/html/src/ui/time/time-element.ts
```diff
@@ -62,4 +96,58 @@ export class TimeElement extends MediaElement {
...
+    return this.toggle;
+  }
+
+  #getMedia() {
+    try {
+      return this.#state.value;
+    } catch {
+      return undefined;
+    }
+  }
```
mihar-22: I think we can remove all of this?

packages/html/src/ui/time/time-element.ts
```diff
@@ -62,4 +96,58 @@ export class TimeElement extends MediaElement {
...
+
+  #handleClick = (event: MouseEvent): void => {
+    if (event.defaultPrevented || !this.#canToggle || !this.#getMedia()) return;
+    this.#toggleType();
+  };
+
+  #handleKeyDown = (event: KeyboardEvent): void => {
+    if (event.target !== event.currentTarget || event.defaultPrevented) return;
```
mihar-22: Why do we need this check?

packages/html/src/ui/time/time-element.ts
```diff
@@ -62,4 +96,58 @@ export class TimeElement extends MediaElement {
...
+  #getMedia() {
+    try {
+      return this.#state.value;
+    } catch {
+      return undefined;
+    }
+  }
+
+  #handleClick = (event: MouseEvent): void => {
+    if (event.defaultPrevented || !this.#canToggle || !this.#getMedia()) return;
```
mihar-22: Why do we need this check?

packages/html/src/ui/time/time-element.ts
```diff
@@ -62,4 +96,58 @@ export class TimeElement extends MediaElement {
...
+    this.#toggleType();
+  };
+
+  #handleKeyDown = (event: KeyboardEvent): void => {
+    if (event.target !== event.currentTarget || event.defaultPrevented) return;
+    if (event.key !== 'Enter' && event.key !== ' ') return;
```
mihar-22: I believe we have a utility for checking if a event key is a press/click.

packages/html/src/ui/time/time-element.ts
```diff
@@ -62,4 +96,58 @@ export class TimeElement extends MediaElement {
...
+    if (!this.#canToggle || !this.#getMedia()) return;
```
mihar-22: Don't see why we need this check.

packages/html/src/ui/time/time-element.ts
```diff
@@ -62,4 +96,58 @@ export class TimeElement extends MediaElement {
...
+
+    event.preventDefault();
```
mihar-22, line 105: What are we preventing?

packages/html/src/ui/time/time-element.ts
mihar-22, line 1: Most of the code here can be simplified into:

1. Attach event handlers only when `toggle` is true
2. When those event handlers are called, simply update the `displayType` and request an update
3. Handle updating props in the `update` hook and use `toggle` ternaries + `displayType` to set everything correctly.

packages/react/src/ui/time/time-value.tsx
```diff
@@ -25,12 +25,22 @@ export const Value = forwardRef(function Value(
   componentProps: ValueProps,
   forwardedRef: ForwardedRef<HTMLTimeElement>
 ) {
-  const { render, className, style, type, negativeSign, label, ...elementProps } = componentProps;
+  const { render, className, style, type, negativeSign, label, toggle, ...elementProps } = componentProps;
 
   const time = usePlayer(selectTime);
+  const defaultType = type ?? TimeCore.defaultProps.type;
+  const canToggle = !!toggle;
```
mihar-22: We can remove `canToggle` and default `toggle` above to `toggle = false`

packages/react/src/ui/time/time-value.tsx
```diff
@@ -25,12 +25,22 @@ export const Value = forwardRef(function Value(
...
 
   const [core] = useState(() => new TimeCore());
-  core.setProps({ type, negativeSign, label });
+  const [displayType, setDisplayType] = useState(defaultType);
+
+  useEffect(() => setDisplayType(defaultType), [defaultType]);
+  useEffect(() => {
+    if (!canToggle) setDisplayType(defaultType);
+  }, [canToggle, defaultType]);
```
mihar-22: This seems unnecessary as we rely on `activeType` below.

packages/react/src/ui/time/time-value.tsx
```diff
@@ -49,6 +59,25 @@ export const Value = forwardRef(function Value(
...
+    setDisplayType((value) => getNextType(value, defaultType));
+  };
+
+  const handleClick = (event: MouseEvent<HTMLTimeElement>) => {
+    if (event.defaultPrevented) return;
+    toggleType();
+  };
+
+  const handleKeyDown = (event: KeyboardEvent<HTMLTimeElement>) => {
+    if (event.defaultPrevented || (event.key !== 'Enter' && event.key !== ' ')) return;
```
mihar-22: Same comment as HTML, we might have a utility for key press/click, if not great to create one!

packages/react/src/ui/time/time-value.tsx
```diff
@@ -49,6 +59,25 @@ export const Value = forwardRef(function Value(
...
+    event.preventDefault();
```
mihar-22, line 76: What are we preventing?

packages/react/src/ui/time/time-value.tsx
```diff
@@ -49,6 +59,25 @@ export const Value = forwardRef(function Value(
     state.text
   );
 
+  const toggleType = () => {
+    if (!canToggle) return;
```
mihar-22: Not guarding anything meaningful. In addition, won't be called because listeners below won't be attached unless it can toggle.

packages/react/src/ui/time/time-value.tsx
```diff
@@ -61,13 +90,22 @@ export const Value = forwardRef(function Value(
...
 });
 
+function getNextType(type: TimeType, defaultType: TimeType): TimeType {
+  if (defaultType === 'current') {
+    return type === 'remaining' ? 'current' : 'remaining';
+  }
+
+  return type === 'duration' ? 'remaining' : 'duration';
+}
+
```
mihar-22: Only called in one place, fine to inline for now imo.

packages/react/src/ui/time/time-value.tsx
```diff
@@ -49,6 +59,25 @@ export const Value = forwardRef(function Value(
...
+  };
+
+  const handleKeyDown = (event: KeyboardEvent<HTMLTimeElement>) => {
+    if (event.defaultPrevented || (event.key !== 'Enter' && event.key !== ' ')) return;
+    event.preventDefault();
+    if (event.repeat) return;
+    toggleType();
+  };
+
+  const toggleProps = canToggle ? { onClick: handleClick, onKeyDown: handleKeyDown } : undefined;
```
mihar-22: I think it'd be fine to inline this but leave it up to you.

mihar-22 (approved):
Thanks for doing this! A second pass and simplifications on additions to Time components would be great, plus considering removing the toggle labels from core.

packages/core/src/core/ui/time/time-core.ts
sampotts: I've asked in the shared a11y about the best pattern here. I don't think YouTube has this quite right as there's no indication of what the button actually does, the label only informs of the current state. I could be wrong but waiting on clarification. I think we want the current state and next state to be obvious to screen readers.

packages/html/src/ui/time/time-element.ts
sampotts: The element can have children that could also fire events but we can remove it. Checking that default is prevented also allows folks to prevent the toggle but perhaps we don't need that either.

packages/html/src/ui/time/time-element.ts
sampotts: Checking that default is prevented allows folks to prevent the toggle but perhaps we don't need that. The rest can go if we're only binding when toggle is enabled.

packages/html/src/ui/time/time-element.ts
sampotts: Scrolling the page when pressing space.

packages/react/src/ui/time/time-value.tsx
sampotts: As above - space will scroll the page.

PR #1670, opened 2026-06-09
fix(packages): escape HTML special chars in serializeAttributes to prevent XSS

### LINKED ISSUE #1562: Bug: XSS via Unescaped Attribute Values in CustomMediaElement Shadow Root
> [!NOTE]

`serializeAttributes` in `packages/utils/src/dom/attributes.ts` interpolates attribute values into an HTML string without escaping quotes or `<`/`>`. The result is assigned to `this.shadowRoot!.innerHTML` in `CustomMediaElement`'s constructor (`packages/core/src/dom/media/custom-media-element/index.ts:226`), so any attribute on a custom media element that contains a `"` can break out of the attribute and inject arbitrary markup — including event handler attributes or `<script>` siblings — into the shadow root.

Attribute names come from `getAttrsFromProps(ctor.properties)`, but the values come from `this.attributes` unfiltered. Pages that render user-controlled data into a media element attribute (e.g. `<hls-video src={userInput}>`) inherit this sink. The duplicate `serializeAttributes` in `packages/html/src/media/background-video/index.ts:93` has the same bug.

Reported by Codex during review of the security guide in #1559.

## Tasks

- [ ] Escape `"`, `&`, `<`, `>` in attribute values inside `serializeAttributes` (and the background-video duplicate, or delete it in favor of the shared util).
- [ ] Add unit tests covering quote-injection, angle-bracket injection, and ampersand handling.
- [ ] Add an e2e/regression test that constructs a `CustomMediaElement` with a malicious attribute value and asserts no injected nodes end up in the shadow root.
- [ ] Audit other `innerHTML` sinks for the same pattern.

Jerricho93:
## Summary

`serializeAttributes` interpolated attribute values directly into HTML strings without encoding. Because the result is assigned to `shadowRoot.innerHTML` at construction time, and Shadow DOM does not sandbox script execution, any attribute value containing `"` (e.g. a user-controlled `poster`, `src`, or `crossorigin`) could break out of the attribute and inject event handlers or `<script>` siblings into the shadow root.

Changes per sink:

- **`utils`**: add private `escapeAttributeValue` that encodes `&`, `<`, `>`, `"` (ampersand first to prevent double-encoding); apply in `serializeAttributes`
- **`html/BackgroundVideo`**: remove the local duplicate `serializeAttributes`; import the shared util and `pick` against the existing attribute allowlist
- **`html/define/background`**: remove dead `_attrs` parameter (was passed to a template that never interpolated it)
- **`icons/scripts/build`**: inline `esc()` in the generated `renderIcon` function body to close the codegen sink
- **`site/build-ejected-skins`**: extend the partial local `escapeAttributeValue` to also cover `<` and `>`

## Testing

- New `packages/utils/src/dom/tests/attributes.test.ts`: 9 cases covering all four chars, boolean branch, ampersand-first ordering regression, and `namedNodeMapToObject` raw-value preservation
- New `packages/html/src/media/background-video/tests/background-video.test.ts`: quote breakout, angle-bracket injection, non-allowlisted attribute filtering, boolean attrs, safe value round-trip
- Extended `packages/core/.../custom-media-element.test.ts`: `describe('XSS prevention')` with 4 cases
- New `apps/e2e/tests/xss-prevention.spec.ts`: 3 Playwright tests exercising the construction-time path via `container.innerHTML`

## Validation

```bash
pnpm -F @videojs/utils test src/dom/tests/attributes.test.ts
pnpm -F @videojs/core test src/dom/media/custom-media-element
pnpm -F @videojs/html test src/media/background-video
pnpm test:e2e:vite
pnpm typecheck
pnpm lint
```

## Known issue (follow-up)

`CustomMediaElement` has a pre-existing attribute routing inconsistency: `getAttrsFromProps` uses `.toLowerCase()` to build `observedAttributes` (e.g. `crossorigin`, `controlslist`, `disableremoteplayback`), while `#define` uses `kebabCase` to key `mediaHostAttrToProp` (e.g. `cross-origin`, `controls-list`, `disable-remote-playback`). Because these never match, multi-word attributes are never added to the disallowed set and always flow through `serializeAttributes` rather than `attributeChangedCallback`. This means post-construction `setAttribute` calls for those attributes do not sync to the media host setter. This fix makes `serializeAttributes` safe, closing the injection vector — but the routing inconsistency itself should be tracked and corrected separately.

Closes #1562

---

> [!NOTE]
> **High Risk**
> Security fix on HTML string assembly used at custom-element construction; behavior change for attribute values containing `&<>"` but intended to block script/event injection in shadow templates.
> 
> **Overview**
> Closes an XSS path where **unescaped attribute values** were interpolated into HTML assigned to `shadowRoot.innerHTML` (and similar codegen sinks), allowing quote/angle-bracket breakout into event handlers or extra nodes inside shadow roots.
> 
> **`serializeAttributes`** in `@videojs/utils/dom` now runs values through **`escapeHtml`** (`&`, `<`, `>`, `"`, with ampersand first). **`BackgroundVideo`** drops its local serializer and uses the shared helper with an existing **video-attribute allowlist** via `pick`. **`background-video-skin`** stops passing unused attrs into a static template.
> 
> The same escaping pattern is applied to **generated `renderIcon` output** (icons build) and **ejected skin HTML** (`build-ejected-skins`), including `<`/`>` on icon names and attrs.
> 
> Coverage adds **unit tests** for `serializeAttributes` / `escapeHtml`, **XSS cases** on `CustomMediaElement` and `BackgroundVideo`, and **Playwright e2e** on `hls-video` construction via `innerHTML`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 101bd4aa80128f5724a732dd0af9cc8ed1c0c10b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/utils/src/dom/attributes.ts
```diff
@@ -9,15 +9,20 @@ export function namedNodeMapToObject(namedNodeMap: NamedNodeMap) {
...
  * Helper function to serialize attributes into a string.
  */
 export function serializeAttributes(attrs: Record<string, string>) {
   let html = '';
   for (const key in attrs) {
-    const value = attrs[key];
+    const value = attrs[key]!;
     if (value === '') html += ` ${key}`;
-    else html += ` ${key}="${value}"`;
+    else html += ` ${key}="${escapeAttributeValue(value)}"`;
```
luwes: 👍

packages/utils/src/dom/attributes.ts
```diff
@@ -9,15 +9,20 @@ export function namedNodeMapToObject(namedNodeMap: NamedNodeMap) {
   return obj;
 }
 
+// Ampersand must be escaped first to avoid double-encoding the entities below.
+function escapeAttributeValue(value: string): string {
+  return value.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;');
+}
```
luwes: can we move this to its own file and name it more generic `escapeHtml()`. it's pretty common as utility.
https://github.com/videojs/v10/blob/440145e9de980fda317b11ae2b7ee3abdc7e3ff7/packages/utils/src/string/escape-html.ts#L10

packages/html/src/define/background/skin.ts
```diff
@@ -27,7 +27,7 @@ export class BackgroundVideoSkinElement extends ReactiveElement {
 
     if (!this.shadowRoot) {
       this.attachShadow((this.constructor as typeof BackgroundVideoSkinElement).shadowRootOptions);
-      this.shadowRoot!.innerHTML = getTemplateHTML(namedNodeMapToObject(this.attributes));
```
luwes, line 30: why is this removed?

luwes (changes requested):
would like some small changes but looks good overall!

packages/html/src/define/background/skin.ts
Jerricho93: The `_attrs` parameter is not being used here. If we were to accept attrs in `getTemplateHTML` without escaping, it would introduce the same issues we're fixing with this PR. If there is a plan in the future to add this (by removing `_` to attrs), we can reintroduce it and add the escape logic now. Does it makes sense? Otherwise, at this point in time in the code, this is effectively dead code

packages/utils/src/dom/attributes.ts
Jerricho93: Updated!

luwes (approved):
LGTM

PR #1676, opened 2026-06-10
feat(spf): capability probing

cjpillsbury:
Capability probing: before `track-switching` picks a rendition, prune the ones
this environment can't decode. Two prongs — a synchronous codec probe
(`MediaSource.isTypeSupported`) wired as a new `excludeUnplayableTracks` hard
constraint, and non-fMP4 container detection (MPEG-TS, raw ADTS AAC) in the HLS
parser that relabels those renditions and asserts them unplayable. Riding along
are three parser-correctness fixes that make probing behave on real Mux streams
(non-AAC codec recognition, `CHANNELS` parse, `EXT-X-STREAM-INF` cross-product
dedup). Internal SPF only — no public API; the engine defaults the probe so
consumers get filtering for free.

## For reviewers — how to read this PR

> Base targets `feat/spf-multi-cdn-failover`, not main — repoints to main once
> that lands. The Files changed tab is already scoped to this PR.

### Bucket 1 — Runtime (focus here)

**Read fully:**
- `media/dom/capabilities.ts` (new, +66) — the `canPlayTrack` primitive: an
  `isTypeSupported` codec probe memoized by MIME, plus the non-fMP4-asserted-
  unplayable branch. Precedent for future cluster-D probes (key-system,
  `changeType`). Unprobeable tracks (no `mimeType` / no `codecs`) pass through as
  playable — the late `createSourceBuffer` check stays their backstop.
- `media/utils/tracks.ts` (+31, `applyContainerMimeType`) — immutable per-type
  MIME propagation; scoped to one track type (never cross audio↔video),
  idempotent.

**Targeted careful read:**
- `playback/behaviors/track-switching.ts` (+82/−16) — new `excludeUnplayableTracks`
  hard constraint pooled beside `excludeFailedCdns`, reading the injected
  `canPlayTrack`. No probe → passes everything through; empty-after-prune → no
  new pick. Verify the dropped speculative `noPlayable*` state left nothing
  write-only behind (slot, generic, effect write all gone).
- `media/hls/parse-media-playlist.ts` (+43) — container detection: no `EXT-X-MAP`
  + recognized extension (`.ts`/`.aac`). High-precision — fMP4 mandates the map,
  so never trips on fMP4; query string stripped; text tracks exempt.
- `media/hls/parse-multivariant.ts` (+33/−6) — `EXT-X-STREAM-INF` cross-product
  dedup by playlist URI + `CHANNELS` parse. Redundant-stream renditions (distinct
  per-CDN URIs) stay separate; lowest combined `BANDWIDTH` kept as the video-only
  proxy ABR ranks on; `audioGroupId` → `audioGroupIds` (old field had no readers).
- `media/hls/parse-attributes.ts` (+10/−1) — recognize non-AAC audio codecs,
  case-insensitive. An unrecognized audio codec parses empty → unprobeable →
  never pruned, so this set gates whether surround renditions can be filtered.
- `playback/behaviors/resolve-track.ts` (+17/−5) — propagate a detected container
  to every same-type rendition from one resolved playlist; same-type scope keeps
  the write disjoint from concurrent per-type resolution (no race).
- `media/types/index.ts` (+23/−1) — DOM-free `CanPlayTrack` predicate type +
  `audioGroupIds`.

**Skim structure only:** `engine.ts` (+11), `engine-audio-only.ts` (+11) —
default `canPlayTrack` into `finalConfig`; audio-only mirrors the default engine,
adapters already forward `...config`.

**Skim file:** Tests (8 files, +494) — assertion shape across codec filtering,
container detection, cross-product dedup, `CHANNELS`, no-pick-on-empty.

### Design doc

`internal/design/spf/features/capability-probing.md` (+132/−44) **[Skim file]** —
status advanced to the shipped shape; the removed `noPlayable*` surfacing recorded
as deferred.

## Smoke test

Drives `createSimpleHlsEngine` directly and takes the source as a path-only
`?src=` param (no SOURCES edit needed). The page's track panels show the
candidate list + selected track, so the prune/pick is directly visible.

**Non-fMP4 source** — Apple `bipbop_4x3` (muxed-TS video + raw-`.aac` audio):
`/spf-segment-loading/?src=https://devstreaming-cdn.apple.com/videos/streaming/examples/bipbop_4x3/bipbop_4x3_variant.m3u8`
- Observe: no rendition picked for either type (TS + raw-AAC asserted
  unplayable) — a loud no-pick, not a stall deep in the pipeline.

**AC-3 5.1 source** (`?enable_51_surround=true` surfaces an `ac-3` / `CHANNELS="6"`
rendition):
`/spf-segment-loading/?src=https://stream.mux.com/BKShkotuMGjubl4mSJK5y6hP7MizeNgpMU2Jw2Ul6Gc.m3u8?enable_51_surround=true`
- Observe: AAC stereo is selected and audio plays (the undecodable AC-3 5.1
  rendition is pruned); the video track panel shows 7 tracks, not 14.

## What changed — by surface

**Codec capability probing.** A new `canPlayTrack` primitive builds a track's MIME
codec string and asks `MediaSource.isTypeSupported`, memoized per MIME. It's
injected into `track-switching` as the `excludeUnplayableTracks` hard constraint,
so an undecodable variant (e.g. HEVC on a browser without it) is pruned before
selection instead of failing late at `createSourceBuffer`. The DOM-free
`CanPlayTrack` predicate type keeps the behavior from importing a DOM API; the
engine defaults the implementation. An emptied candidate set makes no pick — the
write-only `noPlayable*` flags that used to track this are dropped.

**Non-fMP4 container detection.** HLS TS and raw-AAC renditions carry the same
`CODECS` as fMP4 but no `EXT-X-MAP`, so the parser was mislabeling them
`video/mp4` / `audio/mp4`. The media-playlist parser now detects them by
init-segment-absence + extension and relabels to the container MIME;
`resolve-track` propagates that MIME across the same type from one resolved
playlist. `canPlayTrack` asserts both unplayable, so a non-fMP4 source makes no
pick (loud, observable) instead of stalling deep in the pipeline.

**HLS parser correctness on real streams.** Three fixes surfaced validating
probing against live Mux streams: `parseCodecs` recognizes non-AAC audio codecs
(an AC-3 5.1 track parsed with no audio codec, slipping past the probe);
`CHANNELS` is read off `EXT-X-MEDIA` (a 5.1 track modeled as stereo); and the
`EXT-X-STREAM-INF` cross-product is de-duplicated by URI (a 7-rendition source
surfaced 14 video tracks, doubling ABR's candidate set).

## Notable design decisions

- **Codec probe is a config predicate, not a state-slot writer.** Injected as
  `canPlayTrack`, read inline by the constraint. Alternative: a probing behavior
  writing per-track playability to state. Rejected because codec support is
  synchronous and static per environment — a predicate needs no slot or effect.
  The async cluster-D probes (key-system, `changeType`) *will* land as
  slot-writers, since their verdict resolves asynchronously.
- **Non-fMP4 containers are asserted unplayable, not run through
  `isTypeSupported`.** Alternative: trust the probe. Rejected because Chromium
  false-positives both `video/mp2t` and `audio/aac` (reports `true`, then appends
  produce no buffered range), and this engine has no TS transmux. Raw AAC is
  genuinely decodable but the append pipeline assumes an init segment — temporary,
  noted below.

- **[Author]** Raw ADTS AAC is asserted unplayable as a *temporary* limitation —
  the browser decodes it, but the segment/append pipeline assumes an `EXT-X-MAP`
  init segment, so it would fetch but never buffer (a silent stall). A loud
  no-pick beats that. Follow-up: drop the init-required assumption + switch to a
  bare-MIME probe.
- **[Deferred]** Surfacing "nothing playable" as observable state
  (fatal-vs-recoverable error) is deferred until a consumer (error mapping) needs
  it; the design questions are recorded in the doc.

## Test plan

- [x] Unit coverage added across 8 files: codec filtering, container detection +
  relabel, cross-product dedup, `CHANNELS` parse, no-pick-on-empty.
- [x] Manual validation in the SPF sandbox against live Mux streams — AC-3 5.1
  (AC-3 pruned, AAC stereo selected, audio plays; 14→7 video tracks) and Apple
  `bipbop_4x3` (muxed-TS video + raw-`.aac` audio → no pick on each type).

---

> [!NOTE]
> **Medium Risk**
> Changes core track-selection and engine defaults on the playback path; behavior is well covered by tests but wrong pruning could block playback or leave types unselected.
> 
> **Overview**
> Adds **upstream capability filtering** so undecodable renditions are dropped before `track-switching` picks a track, instead of failing late at `createSourceBuffer`.
> 
> **Codec probing:** New memoized `canPlayTrack` (`MediaSource.isTypeSupported`) is injected as config and wired into a new **`excludeUnplayableTracks`** hard constraint alongside `excludeFailedCdns` for video and audio. HLS engines default the DOM implementation; overrides are the Tier‑2 seam.
> 
> **Non-fMP4 containers:** Media-playlist parsing relabels MPEG‑TS (`.ts`) and raw ADTS AAC (`.aac`) when there is no `EXT-X-MAP`; `resolve-track` propagates that MIME across same-type renditions via `applyContainerMimeType`. Those containers are **asserted unplayable** in `canPlayTrack` (TS false-positives / no transmux; raw AAC blocked until init-segment pipeline work).
> 
> **Selection when nothing is playable:** If constraints empty a type that still has tracks, the behavior **clears** the selection (including stale picks after relabel) and logs a placeholder `console.error`; CDN-all-cooled-down now clears and re-picks on recovery instead of keeping the old pick.
> 
> **Parser fixes for real manifests:** Broader `parseCodecs` audio recognition (e.g. AC‑3), `CHANNELS` on `EXT-X-MEDIA`, and **dedup of the `EXT-X-STREAM-INF` cross-product** by video URI (`audioGroupId` → `audioGroupIds`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 728d210f8addc0ad64734f7796a5110829f0d0cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/spf/src/media/dom/capabilities.ts
+++ b/packages/spf/src/media/dom/capabilities.ts
@@ -19,6 +19,7 @@
  * predicate, since their verdict resolves asynchronously.
  */
 
+import { NON_FMP4_CONTAINER_MIMES } from '../hls/parse-media-playlist';
 import type { CanPlayTrack } from '../types';
 import { buildMimeCodec, isCodecSupported } from './mse/mediasource-setup';
 
@@ -31,8 +32,30 @@ const codecSupportCache = new Map<string, boolean>();
  * (CODECS is optional per the HLS spec) — is unprobeable and passes through as
  * playable (`true`) rather than being dropped; the late `createSourceBuffer`
  * check stays as the backstop for those.
+ *
+ * Detected non-fMP4 containers (`video/mp2t`, `audio/aac`) are asserted
+ * unsupported regardless of the probe, so they surface a loud, observable
+ * `noPlayable` instead of failing/stalling deep in the pipeline. Two different
+ * reasons, neither UA-based:
+ *
+ * - **MPEG-TS** can't be played at all here: `isTypeSupported('video/mp2t…')` is
+ *   a genuine false positive on Chromium (reports `true` but appends produce no
+ *   buffered range), and this engine has no TS transmux pipeline.
+ * - **Raw ADTS AAC** is a *temporary* limitation. The browser genuinely
+ *   supports it (Chrome/Safari decode `audio/aac`; Firefox doesn't), so it could
+ *   be made playable — but our segment actors / loading behaviors / append
+ *   pipeline assume every rendition has an `EXT-X-MAP` init segment (e.g. an
+ *   `append-init` task with an empty URL, fMP4-shaped append handling). Until
+ *   that init-segment assumption is removed, ADTS would fetch but never buffer
+ *   (a silent stall), so we assert it unplayable for now. FOLLOW-UP: drop the
+ *   init-required assumption in the pipeline and switch this to a bare-MIME
+ *   probe (`buildMimeCodec` would project `audio/aac` with no codecs) so it
+ *   plays where the browser supports it.
+ *
+ * Override via the engine's `canPlayTrack` config when those pipelines land.
  */
 export const canPlayTrack: CanPlayTrack = (track) => {
+  if (track.mimeType && NON_FMP4_CONTAINER_MIMES.has(track.mimeType)) return false;
   if (!track.mimeType || !track.codecs?.length) return true;
   const mimeCodec = buildMimeCodec({ mimeType: track.mimeType, codecs: track.codecs });
   const cached = codecSupportCache.get(mimeCodec);
--- a/packages/spf/src/media/dom/capabilities.ts
+++ b/packages/spf/src/media/dom/capabilities.ts
@@ -34,9 +34,9 @@ const codecSupportCache = new Map<string, boolean>();
  * check stays as the backstop for those.
  *
  * Detected non-fMP4 containers (`video/mp2t`, `audio/aac`) are asserted
- * unsupported regardless of the probe, so they surface a loud, observable
- * `noPlayable` instead of failing/stalling deep in the pipeline. Two different
- * reasons, neither UA-based:
+ * unsupported regardless of the probe, so they're pruned before selection
+ * (the type makes no pick) instead of failing/stalling deep in the pipeline.
+ * Two different reasons, neither UA-based:
  *
  * - **MPEG-TS** can't be played at all here: `isTypeSupported('video/mp2t…')` is
  *   a genuine false positive on Chromium (reports `true` but appends produce no
--- a/packages/spf/src/media/dom/capabilities.ts
+++ b/packages/spf/src/media/dom/capabilities.ts
@@ -1,0 +1,43 @@
+/**
+ * Capability probing — the engine's foundation for asking the browser what it
+ * can actually decode before committing a rendition to the pipeline.
+ *
+ * Today this is the synchronous codec half: `canPlayTrack` answers "can this
+ * environment play this track?" by building the track's MIME codec string and
+ * passing it to `MediaSource.isTypeSupported` (via `isCodecSupported`). It's the
+ * DOM implementation of the DOM-free `CanPlayTrack` predicate the
+ * track-switching hard-constraint pre-pass consumes — injected through engine
+ * config so the (DOM-free) behavior never imports a DOM API directly.
+ *
+ * Results are memoized by built MIME string: codec support is a pure function
+ * of (codec, environment) and never changes after load, so probing is lazy
+ * (per candidate, at constraint-apply time) but each unique MIME is asked once.
+ *
+ * Future cluster-D phases (async `requestMediaKeySystemAccess` key-system
+ * probing, `SourceBuffer.changeType()` availability) extend this surface; the
+ * async ones land as a state-slot writer behavior rather than a config
+ * predicate, since their verdict resolves asynchronously.
+ */
+
+import type { CanPlayTrack } from '../types';
+import { buildMimeCodec, isCodecSupported } from './mse/mediasource-setup';
+
+const codecSupportCache = new Map<string, boolean>();
+
+/**
+ * Whether the environment can decode `track`, by codec. Builds the track's
+ * MIME codec string and checks `MediaSource.isTypeSupported`, memoized by MIME.
+ * A track without enough to probe — no `mimeType`, or no declared `codecs`
+ * (CODECS is optional per the HLS spec) — is unprobeable and passes through as
+ * playable (`true`) rather than being dropped; the late `createSourceBuffer`
+ * check stays as the backstop for those.
+ */
+export const canPlayTrack: CanPlayTrack = (track) => {
+  if (!track.mimeType || !track.codecs?.length) return true;
+  const mimeCodec = buildMimeCodec({ mimeType: track.mimeType, codecs: track.codecs });
+  const cached = codecSupportCache.get(mimeCodec);
+  if (cached !== undefined) return cached;
+  const supported = isCodecSupported(mimeCodec);
+  codecSupportCache.set(mimeCodec, supported);
+  return supported;
+};
--- a/packages/spf/src/media/hls/tests/parse-multivariant.test.ts
+++ b/packages/spf/src/media/hls/tests/parse-multivariant.test.ts
@@ -288,6 +288,25 @@ video-lo.m3u8`;
       ]);
     });
 
+    it('parses the CHANNELS attribute on audio renditions (6 for 5.1, default 2 when absent)', () => {
+      const text = `#EXTM3U
+#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="surround",NAME="5.1",CHANNELS="6",URI="https://example.com/a51.m3u8"
+#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="stereo",NAME="Stereo",CHANNELS="2",URI="https://example.com/a2.m3u8"
+#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="nochan",NAME="Unspecified",URI="https://example.com/anc.m3u8"
+#EXT-X-STREAM-INF:BANDWIDTH=3000000,RESOLUTION=1920x1080,CODECS="avc1.640028,ac-3",AUDIO="surround"
+https://example.com/v.m3u8`;
+
+      const result = parseMultivariantPlaylist(text, { url: baseUrl });
+      const audioTracks = (result.selectionSets.find((s) => s.type === 'audio')?.switchingSets[0]?.tracks ??
+        []) as PartiallyResolvedAudioTrack[];
+      const byGroup = (groupId: string) => audioTracks.find((t) => t.groupId === groupId);
+
+      expect(byGroup('surround')?.channels).toBe(6);
+      expect(byGroup('stereo')?.channels).toBe(2);
+      // No CHANNELS attribute → stereo default.
+      expect(byGroup('nochan')?.channels).toBe(2);
+    });
+
     it('extracts audio codecs from referencing streams', () => {
       const result = parseMultivariantPlaylist(muxPlaylist, { url: baseUrl });
       const audioSet = result.selectionSets.find((s) => s.type === 'audio');
--- a/packages/spf/src/media/hls/tests/parse-multivariant.test.ts
+++ b/packages/spf/src/media/hls/tests/parse-multivariant.test.ts
@@ -74,7 +74,43 @@ video-1080p.m3u8`;
     expect(typeof track1080p.id).toBe('string');
     // Optional fields not present
     expect(track1080p.frameRate).toBeUndefined();
-    expect(track1080p.audioGroupId).toBeUndefined();
+    expect(track1080p.audioGroupIds).toBeUndefined();
+  });
+
+  it('de-duplicates the EXT-X-STREAM-INF cross-product: one track per video URI, accumulating audio groups', () => {
+    // The same two video renditions are each listed twice — once paired with a
+    // 5.1 group (higher BANDWIDTH, ac-3) and once with a stereo group (lower,
+    // mp4a) — sharing the same media-playlist URI. Mirrors the Mux
+    // enable_51_surround shape.
+    const text = `#EXTM3U
+#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="audio-51",NAME="Default",CHANNELS="6",URI="https://example.com/a51.m3u8"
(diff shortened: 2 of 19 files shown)


packages/spf/src/media/dom/capabilities.ts
```diff
@@ -0,0 +1,66 @@
...
+ * check stays as the backstop for those.
+ *
+ * Detected non-fMP4 containers (`video/mp2t`, `audio/aac`) are asserted
+ * unsupported regardless of the probe, so they're pruned before selection
+ * (the type makes no pick) instead of failing/stalling deep in the pipeline.
+ * Two different reasons, neither UA-based:
+ *
+ * - **MPEG-TS** can't be played at all here: `isTypeSupported('video/mp2t…')` is
+ *   a genuine false positive on Chromium (reports `true` but appends produce no
+ *   buffered range), and this engine has no TS transmux pipeline.
```
spuppo-mux, line 43: question(non-blocking): Is this still true? I notice I can play an HLS TS stream directly by url from Chrome, but not sure if there is something else going on under the hood. Or is this specifically a problem when working with buffers directly?

Image

`https://stream.mux.com/ihZa7qP1zY8oyLSQW9TS602VgwQvNdyIvlk9LInEGU2s.m3u8?redundant_streams=true`

cjpillsbury: Yessir, it is still true! What you're seeing is Chrome's new native HLS support, not for MSE/MSS-based playback

packages/spf/src/playback/behaviors/resolve-track.ts
```diff
@@ -132,9 +132,21 @@ function setupTrackResolution<K extends SelectedTrackKey>({
...
+                    // raw AAC) implies every rendition of *this* type matches —
+                    // relabel them all from one resolved playlist instead of
+                    // fetching each. Scoped to this track's own type: never cross
+                    // audio↔video (mixed-container sources exist, e.g. muxed-TS
+                    // video + raw-.aac audio), which also keeps per-type
+                    // resolutions' writes disjoint (no race).
+                    return NON_FMP4_CONTAINER_MIMES.has(mediaTrack.mimeType)
+                      ? applyContainerMimeType(patched, mediaTrack.type, mediaTrack.mimeType)
+                      : patched;
+                  });
```
spuppo-mux, line 149: thought(non-blocking): This is a bit confusing, marking it to re-read later

cjpillsbury: good catch. I think this may be vestigial/no longer necessary. Confirming now.

cjpillsbury: oh right it's not! (sorry been doing a lot of task switching). The TL;DR - the parser will update a specific track's `mimeType` based on the file extensions of the segment URLs. This updates all of the other tracks of the same track type to have the now-detected `mimeType`. E.g.
1. We (currently, at least) optimistically assume `video/mp4` (for `type='video'` tracks) / `audio/mp4` (for `type='audio'` tracks)
2. After actually resolving our first track of a type, we use the file extension (plus presence/absence of an init segment) to determine if we need to update that to e.g. `video/mp2t`
3. If we did, we update _**all**_ tracks of the track type as well. So e.g. all `type='video'` tracks will have their `mimeType` updated from `video/mp4` to `video/mp2t`

Make sense?

packages/spf/src/playback/behaviors/track-switching.ts
```diff
@@ -532,13 +595,16 @@ function setupTrackSwitching<
...
-            // exist) everything pruned. Nothing to pick. Surfacing "nothing
-            // playable" as a distinct not-ready state is left to the constraints
-            // work; for now it's a silent no-op, leaving any prior pick in place.
+
+            // No playable tracks: no tracks of this type, or the hard-constraints
+            // pre-pass pruned everything (every rendition undecodable, or every
+            // CDN cooled down). Nothing to pick — leave any prior pick in place.
+            // The late createSourceBuffer check is the backstop if an unplayable
+            // rendition reaches the pipeline; surfacing "nothing playable" as
+            // observable state is deferred until a consumer needs it.
```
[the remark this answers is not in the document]
cjpillsbury: This is a real bug (though an edge case for expected use cases). Still, going to add the relevant basic code here, with the anticipation of doing more once fatal errors are added to SPF.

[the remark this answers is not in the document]
cjpillsbury: A (wrong but not egregious) assumption made it through CDN failover - there was an assumption that if every CDN option failed, it should still try to play _**something**_ (by sticking with the last round of selections, which this early bail allowed for). Instead, this should actually unset the selected track for the track type, since constraints (unlike rules) are intended to signal things that _**should not**_ be played (temporarily or ever), and if that means there's "nothing left to play", then the code should reflect that. At a later phase, we'll add (plausibly abstracted) error signaling logic here.

packages/spf/src/media/types/index.ts
```diff
@@ -144,7 +144,14 @@ export type VideoTrack = Track &
...
     frameRate?: FrameRate;
-    audioGroupId?: string;
+    /**
+     * Audio groups (`EXT-X-STREAM-INF:AUDIO`) this video rendition can pair
+     * with. A list because one rendition is typically listed across multiple
+     * `EXT-X-STREAM-INF` entries — one per audio group (the HLS cross-product) —
+     * which the parser collapses into a single track carrying every group it
+     * advertised.
+     */
+    audioGroupIds?: string[];
```
spuppo-mux, line 154: comment: This change+comment looks odd, but I may not know the whole shape of the parser. afaik one EXT-X-STREAM-INF rendition (which I'm assuming translates to one Video Track) should only pair with one audio group. That audio group then consists of many audio tracks, and the group can pair many rendition

packages/spf/src/playback/behaviors/resolve-track.ts
spuppo-mux: Makes sense now!

packages/spf/src/playback/engines/hls/tests/engine.test.ts
```diff
@@ -285,6 +285,85 @@ describe('createSimpleHlsEngine', () => {
...
+    } as Presentation);
+    await flush();
+
+    // HEVC pruned upstream by the capability constraint; AVC selected.
+    expect(engine.state.selectedVideoTrackId.get()).toBe('1080p-avc');
+
+    engine.destroy();
+  });
+
+  it('makes no video pick when no rendition is decodable', async () => {
```
spuppo-mux, line 329: suggestion(non-blocking): test how the filter behaves if mimeType is not available. (although I'm not sure if that would be a valid state

packages/spf/src/media/types/index.ts
cjpillsbury: Yeah this is a result of "normalizing"/"collapsing" `#EXT-X-STREAM-INF`s to a single "track". The implementation details here may change over time, but the general concept has a decent amount of prior art (see, e.g. - https://github.com/video-dev/hls.js/blob/master/src/types/level.ts#L217-L219)

packages/spf/src/media/hls/tests/parse-multivariant.test.ts
```diff
@@ -74,7 +74,43 @@ video-1080p.m3u8`;
     expect(typeof track1080p.id).toBe('string');
     // Optional fields not present
     expect(track1080p.frameRate).toBeUndefined();
-    expect(track1080p.audioGroupId).toBeUndefined();
+    expect(track1080p.audioGroupIds).toBeUndefined();
+  });
+
+  it('de-duplicates the EXT-X-STREAM-INF cross-product: one track per video URI, accumulating audio groups', () => {
```
cjpillsbury, line 80: @spuppo-mux here's a test that relates to your audioGroups question FYI

spuppo-mux (approved):

packages/spf/src/playback/engines/hls/tests/engine.test.ts
cjpillsbury: Currently not a valid state, but I like where your head's at! For what it's worth, Casey and I talked about stuff related to this earlier in the week. Currently, we generally optimistically assume certain values in the CMAF-HAM model and then correct them as new information comes through. Instead, we may want to mark them as "unknown" in some manner (or support both possibilities) and then account for those cases in the architecture. Not currently planned work, but it's definitely on my mind.

packages/spf/src/playback/engines/hls/tests/engine.test.ts
```diff
@@ -8,15 +8,27 @@ vi.mock('../../../../media/dom/mse/append-segment', () => ({
   appendSegment: vi.fn().mockResolvedValue(undefined),
 }));
 
+// Fallback for URLs a test's mock doesn't handle explicitly. Segment/init
+// requests resolve with an empty body — the appendSegment mock makes the bytes
+// inert — so the failover monitor isn't tripped by unmocked segment fetches (a
+// single failed fetch trips that CDN into cooldown, which empties the candidate
+// set). Genuinely unknown URLs still reject loudly.
+function unmockedFetchFallback(url: string): Promise<Response> {
```
cjpillsbury, line 16: **NOTE:** This and several other test updates were the result of fixing the Cursor feedback. Here's the breakdown:
1. unmocked segment fetches were causing CDN failover to kick in (correct behavior)
2. before, we were not properly handling when constraints filtered tracks to an empty list (now fixed)
3. after the fix, this resulted in tests now failing because they were originally able to ignore the failed CDN scenarios (these errors would just happen silently in the background before), so the "correct" fix was to make the test environment look more like a (mocked) real environment, with actually succeeding segment fetches
4. this, in turn, meant the playback engine's behaviors would "get further along" in the test, exposing assumptions in how the tests were written that were relying on these silent background failures (see, e.g., all of the `readyState` changes to the tests). The tests weren't "wrong" in the strong sense, but **_were_** reliant on unrelated environment details to assert the core things they were testing/validating.
5. this, in turn, prompted updating the other now-failing tests to better reflect the environment/playback behaviors as expected.

We could take this further with real segments and/or more robust mocking of things like `MediaSource`/`SourceBuffer` and even the `HTMLMediaElement`, but these changes at least more closely model our actual expectations (again, just to reemphasize, the other tests were not "wrong", they were just relying on unrelated details to assert what was "right", including incorrect behavior definitions in multi-cdn-failover, which was an actual bug, albeit an edge case bug in the real world)

PR #1680, opened 2026-06-11
docs(site): document self-hosted and offline player builds

### LINKED ISSUE #1637: Docs: Self-Hosted Builds
> [!NOTE]

## Context

Discord thread: https://discord.com/channels/507627062434070529/1428837235003756658/1510945202179805237

The installation docs currently generate selected HTML installs as native ESM CDN bundles, for example the base player bundle plus an optional media bundle. Browsers can handle these module scripts directly, but users with offline/self-hosted deployment needs still want a clearer path for turning their selected player configuration into one or more files they can serve themselves.

In the Discord thread, the ask was effectively: after choosing a use case, skin, and media renderer in the installation docs, make it easy to get an offline/prebundled build for that exact selection.

## Problem

For common online usage, the CDN ESM bundles are enough. For offline or restricted-network environments, users often want one of these outcomes:

- mirror the selected ESM files locally and serve them without a CDN dependency
- bundle the selected imports into a single self-hosted module file
- generate a small project/entrypoint from the installation selections and run a documented bundler command

Right now the docs do not make that distinction explicit, so people may assume they need to manually stitch the CDN ESM bundles together.

R-Delfino95:
Closes #1637

## Summary

Add a dedicated **"Self-host the player"** guide (HTML-only) for serving the player without a public CDN.
- Two options: bundle a single file with your own bundler, or mirror the prebuilt CDN files.
- It's its own how-to (not a section on the install page), linked from the installation guide's "See also".

## Validation

- Reproduced both recipes (bundle + mirror) in standalone projects; each renders the player in a real browser with **no runtime requests to a public CDN**. The mirror serves `video.js` plus all shared chunks from localhost (no 404s).
- The guide is registered HTML-only (`frameworks: ['html']`), so it shows for HTML readers only — not React.

## Out of scope (separate follow-ups)

Surfaced while investigating but intentionally not part of this docs change:

- Maintainer-side packaging options that would simplify these instructions — pre-bundling the CDN entry files (no shared chunks) or adding `bundledDependencies` so the npm tarball is self-sufficient.
- A CLI command to generate a project/entrypoint (outcome 3 in the issue) — tracked in #1688.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime or API impact.
> 
> **Overview**
> Adds an HTML-only **Self-host the player** how-to that explains serving `@videojs/html` without a public CDN, for offline, air-gapped, or restricted networks.
> 
> The guide covers **bundling** a single ESM file (recommended entry imports + esbuild example) and **mirroring** the full `node_modules/@videojs/html/cdn` tree, with a caution that CDN entry scripts depend on hashed shared chunks. It also documents loading extra media modules (e.g. HLS) for both approaches.
> 
> The page is registered in `docs.config.ts` under How to with `frameworks: ['html']`, and the installation guide’s **See also** section links to it for the HTML framework view only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3e9a8d27bf09adbfe5dde54bba1234729c80bda9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


site/src/content/docs/how-to/installation.mdx
```diff
@@ -182,6 +182,75 @@ Add it to your components folder in a new file.
...
+Without the media module the player UI renders but the video never loads, with no console error.
+
+</FrameworkCase>
+
+<FrameworkCase frameworks={["react"]}>
+
+Your bundler includes `@videojs/react` in your app like any other dependency, so your production build already serves the player from your own origin with no CDN or runtime fetches. Offline and self-hosted deployments need no extra steps.
+
+</FrameworkCase>
+
```
decepulis: This isn't really providing any information to react devs. I don't think we should even show this information to react readers. 

Let's make this an HTML-only section.
Or, if we spin this out to its own guide, let's make it an HTML-only guide.

decepulis (commented):
imo this should be its own guide. I don't know if we want every user thinking about self-hosting.

Either that or a tab alongside cdn/npm/etc.

site/src/content/docs/how-to/installation.mdx
R-Delfino95: You are right, I reviewed it and it looks much better if we create a separate section in the HTML how-to.

I have already made the changes!

decepulis (approved):
Great, thanks!

PR #1692, opened 2026-06-15
feat(core): Support AirPlay on MSE

### LINKED ISSUE #1260: Feature: Support AirPlay for HLS Playback with MSE and Native Fallback Source
Safari can AirPlay HLS to Apple TV and other receivers when the `<video>` element exposes a **native-playable** URL (typically the same `.m3u8` used for the stream). Today, HLS in Video.js v10 is driven through **MSE** (e.g. hls.js), which does not satisfy that handoff by itself. We should define how v10 supports **AirPlay-capable HLS** alongside MSE playback.

**Intended behavior**

- Allow a **fallback `<source>`** (or equivalent) whose `src` is the original HLS manifest URL and whose type is one Safari treats as native HLS, while the element remains controlled via MSE for in-browser playback—following the dual-source pattern WebKit describes for MSE + AirPlay.

**Lifecycle**

- When the user starts **AirPlay / remote playback**, consider **stopping or suspending** MSE loading and buffer appends on the local `<video>` so the browser is not double-fetching or decoding in parallel with the remote pipeline.

- When playback **returns to the device**, **resume** MSE loading and normal playback so local UI and metrics stay correct.

Exact hooks (e.g. `webkitcurrentplaybacktargetiswirelesschanged`, Remote Playback API, or internal events) and how they map to our store/delegates should be spelled out during design.

## Context

- [How to use Media Source Extensions with AirPlay (WebKit)](https://webkit.org/blog/15036/how-to-use-media-source-extensions-with-airplay/) — pattern for combining MSE with a second source for AirPlay
- [hls.js — AirPlay streaming issue on iOS #6482](https://github.com/video-dev/hls.js/issues/6482) — prior discussion for MSE-based HLS and AirPlay
- [Delivering Video Content for Safari (Apple)](https://developer.apple.com/documentation/webkit/delivering-video-content-for-safari) — Safari / HLS playback expectations

spuppo-mux:
Closes #1260 

tl;dr: 
- adds a hls js media mixin to set up airplay for mse. Still not sure how/when to flip disableRemotePlayback (idea at the bottom of description).
- mild change in utils

## Problem

For AirPlay to work for MSE on Safari we need to follow the pattern described in ["How to use Media Source Extensions with AirPlay"](https://webkit.org/blog/15036/how-to-use-media-source-extensions-with-airplay/).

To do that we need to:

- inject a child `<source type="application/x-mpegURL" src="<manifest>">` as the native fallback.
- flip `disableRemotePlayback` off

Additionally, we want to suspend MSE loading while AirPlay is active so we don't double fetch segments, and resume once AirPlay is inactive.

One wrinkle on `disableRemotePlayback`: ManagedMediaSource requires it set to `true`, so we set it programmatically. The bridge has to distinguish *our* programmatic value from an author who set it to disable remote playback on purpose -> author intent wins.

## Approach: per-engine bridge in the media host

The bridge lives in the media host, close to the engine; not in the remote-playback / cast feature. The host already owns the `<video>` and the engine, which is everything the bridge needs access to.

The idea is to implement this in a way were each engine host opts in by composing without duplicating the wiring:
- For hls.js media element this is implemented as a Mixin.
- SPF design TBD

## Open questions

- **disableRemotePlayback user intent**: Currently we force set it to `false`. 
  - We probably would have to modify the Mixin chain to track read the user-set property somewhere in the `connectedCallback`.
  - Alternatively,  we could assume that if the developer opts-in on using this new mixin, then opts-in to `disableRemotePlayback` to be set to false
  - A similar idea could be to separate this logic into two mixins, one optional that only sets the property and another one in charge of setting the source and load control but doesn't modify the attribute.

---

> [!NOTE]
> **Medium Risk**
> Touches playback loading and `disableRemotePlayback` on the video element during cast handoffs; behavior is WebKit-scoped and tested, but conflicts with author intent to disable remote playback remain an open product question.
> 
> **Overview**
> Adds **AirPlay support for hls.js (MSE) playback on WebKit** by introducing `HlsJsMediaAirPlayMixin`, wired into `HlsJsOnlyMedia`.
> 
> On media attach, the mixin follows WebKit’s MSE+AirPlay pattern: it enables remote playback (`disableRemotePlayback = false`), appends a native HLS `<source>` pointing at the manifest (kept in sync on `MANIFEST_LOADING`), and **only suspends** hls.js via `stopLoad()` while `webkitCurrentPlaybackTargetIsWireless` is true—avoiding duplicate segment fetches and deliberately not calling `startLoad` when AirPlay ends (Safari re-sources locally). Cleanup runs on detach/destroy. Non-WebKit targets no-op.
> 
> `HTMLVideoElementHost` now exposes `webkitCurrentPlaybackTargetIsWireless`. Utils consolidate AirPlay typing onto `WebKitVideoElement` and update `isWebKitAirPlayCapable`. Vitest coverage documents connect-burst behavior and load-control edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 37e13f53d907e8b0dd7c05642bed1e506d0a593d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


luwes (approved):
👍

PR #1704, opened 2026-06-18
fix(core): keep controls visible while tapping controls on touch

### LINKED ISSUE #1682: Bug: Controls Auto-Hide While Tapping Control Buttons on iOS Safari
> [!NOTE]

> Checked #1446 (closed — Android *first-tap flash*, fixed by #1556) and #1549 (open — *volume popover* stays visible after auto-hide). This is a different problem: the controls bar itself hides while the user is actively tapping the buttons.

Core (@videojs/core), HTML (@videojs/html)

```
Browser: Safari
OS: iOS 26.3
Not reproducible on: Android Chrome
```

### Description

On iOS Safari, the controls auto-hide after ~2 seconds even while you are actively tapping the control buttons. For example, repeatedly tapping the mute button or the skip ±10s buttons doesn't keep the controls on screen — they disappear mid-interaction, which is annoying. The same gestures keep the controls visible on Android Chrome.

The short version of why: the controls hide after a couple of seconds of "inactivity", and that inactivity timer is supposed to reset whenever the user interacts. Tapping a control button isn't being counted as activity, so the timer keeps running and hides the controls while you're still using them. (Tapping the video area to show/hide the controls still works fine — it's specifically taps on the buttons that don't reset the timer.)

It surfaces on iOS Safari and not on Android because Android happens to fire extra pointer events around each tap that reset the timer as a side effect, while iOS Safari does not. The activity tracking lives in [`packages/core/src/dom/store/features/controls.ts`](https://github.com/videojs/v10/blob/main/packages/core/src/dom/store/features/controls.ts).

### Reduced test case

https://videojs.org/ (open on a real iOS device in Safari)

### Steps to reproduce

1. Open the Video.js demo on an iPhone in Safari.
2. Start playback so the controls are showing.
3. Repeatedly tap a control button — e.g. mute, or skip forward / back 10s — without tapping the video area.
4. After ~2 seconds the controls auto-hide even though you're actively tapping.

### Current Behavior

Tapping the control buttons does not reset the auto-hide timer on iOS Safari, so the controls disappear while the user is still interacting with them.

### Expected Behavior

Tapping any control should count as user activity and keep the controls visible, resetting the auto-hide timer — the controls should only hide after the user stops interacting.

### Errors

No console errors.

v10.0.0-beta.24

R-Delfino95:
### What's this?

On iOS Safari, the player controls would auto-hide after a couple of seconds **even while you were actively tapping the controls** — for example, repeatedly tapping mute or the skip ±10s buttons. The controls would vanish mid-tap, which felt broken.

The reason: the controls hide after a short period of "inactivity", and that timer is meant to reset whenever you interact. A tap on a control button wasn't being counted as activity, so the timer kept running and hid the controls while you were still using them. Android wasn't affected — it reports touch interactions slightly differently, which happened to keep the timer alive.

### The fix

Count a tap on any control as user activity, so the controls stay visible while you're interacting and only hide once you actually stop. Tapping the video area to show/hide the controls works exactly as before.

### Testing

- Added unit tests: a tap on a control now keeps the controls visible; a tap on the video area still defers to the show/hide gesture.
- All `@videojs/core` unit tests pass (1545).
- Worth a real-device check on iOS Safari to confirm the end-to-end behavior.

Closes #1682

---

> [!NOTE]
> **Low Risk**
> Localized touch/controls activity logic with targeted tests; no auth, data, or API surface changes.
> 
> **Overview**
> Fixes controls auto-hiding on touch (especially iOS Safari) while users tap **control buttons** when a `toggleControls` tap gesture is registered.
> 
> **`GestureCoordinator.claimsTap`** decides whether the gesture layer owns a tap for a given action—matching tap bindings and pointer type, excluding interactive targets, and still claiming when a binding is disabled so fallbacks do not steal the tap.
> 
> **`controlsFeature`** replaces the old “any `toggleControls` binding exists → bail out” check with `claimsTap`. **Claimed** taps on the video/container still defer to gestures and do not reset activity. **Unclaimed** taps (e.g. mute/skip buttons) no longer exit early; they reach **`setActive()`** and reset the idle timer so controls stay visible during repeated control use.
> 
> Unit tests cover `claimsTap` behavior and the controls idle-timer cases with and without a registered gesture.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bad291958b9ab08fdcf45b73cdb607e71b8d848f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/core/src/dom/store/features/controls.ts
```diff
@@ -106,6 +106,14 @@ export const controlsFeature = definePlayerFeature({
       }
 
       if (event.pointerType === 'touch' && Date.now() - pointerDownTime < TAP_THRESHOLD) {
+        // The gesture coordinator ignores interactive targets, so a tap on a control
+        // never reaches toggleControls — reset the idle timer here, or repeated
+        // control taps (mute, seek 10s) wouldn't keep controls visible on iOS Safari.
+        if (isInteractiveTarget(event)) {
+          setActive();
+          return;
+        }
```
mihar-22: Could we explore patching this in the gesture coordinator itself? Worried we'll have two places we're tracking this logic and they might conflict or run into bugs down the road.

mihar-22 (commented):
Thank you for catching this bug and proposing a fix!

packages/core/src/dom/store/features/controls.ts
R-Delfino95: Done!
Moved the decision into the coordinator via `claimsTap(event, action)` helper that owns both the interactive target filtering and the binding match. That puts the logic in one place.

Controls now just asks "is this tap yours?":
- If the coordinator claims it, controls stays out of the way.
- if not (e.g. a tap on a control button), it treats it as activity and keeps the controls visible.

mihar-22 (approved):
thank you!

PR #1722, opened 2026-06-22
fix(site): surface native media properties on element reference pages

### LINKED ISSUE #1717: Docs: Media Reference Pages Omit the Native Media Properties
Our media element reference pages omit the common native media properties — `currentTime`, `duration`, `volume`, `muted`, `paused`, and the rest — even though every element supports them. The other three sections (attributes, events, methods) each list what the element supports and link MDN for detail; properties is the one section that doesn't, so a reader cannot tell these properties exist. [Luwes flagged this on #1342](https://github.com/videojs/v10/pull/1342#issuecomment-4753951455).

### What's missing

Verified on the `hls-video` page at `main.videojs.org`:

- **Present:** `streamType`, `liveEdgeStart`, `targetLiveWindow`, `config`, `error`, `src`, `preload`, `engine`, renditions, tracks.
- **Absent:** `currentTime`, `duration`, `volume`, `muted`, `paused`, `playbackRate`, `buffered`, `textTracks`, `videoWidth`.

### Why it happens

The builder stops property extraction at the base host classes (`HTMLMediaElementHost`, `HTMLVideoElementHost`). Properties defined only there, the native passthroughs, get dropped. The Video.js-specific properties survive because the engine subclass re-declares them; methods survive because they are extracted separately. The omission is not intentional.

### The fix: treat properties like methods

Three of the four sections already work the same way. Each reads the element's supported surface from source, then links MDN for detail:

- **Attributes** come from `CustomMediaElement.properties` — the attributes the element actually registers and forwards.
- **Events** come from the typed `VideoEvents`/`AudioEvents` contract.
- **Methods** come from the base host classes.

Properties should do the same, from those same base host classes. The fix is to extend the methods extraction to cover properties.

We list rather than defer to MDN because the surface is a subset, not all of `HTMLMediaElement`. The element registers a fixed set of attributes, and the host implements a fixed set of properties — `networkState`, `srcObject`, `width`, and `height` are absent, for example. "These are real `<video>` elements, so every native property works" would be false. So list the supported subset from source, show the native ones with an MDN link for detail, and describe the Video.js-specific ones (`streamType`, `liveEdgeStart`, `targetLiveWindow`) inline.

### Align the sections

Once properties is a first-class section, a few cleanups make the page consistent:

- **Rename "Host Properties" to "Properties."** "Host" is an implementation term; the other headings are plain nouns. Dropping it restores parallelism.
- **Expect overlap between attributes and properties, and allow it.** An attribute is markup (`<video muted>`); a property is a JS member (`el.muted`). Properties are the larger set — many, like `currentTime`, have no attribute. Names that exist in both forms (`src`, `preload`) appear in both sections, the same way MDN splits content attributes from IDL properties.
- **Surface `stream-type` in Attributes.** It currently hides in the native attribute list under "the standard ones behave as MDN." Describe it the way Events describes its custom events.
- **Reconsider section order (optional, low stakes).** Markup before JS reads more naturally: Attributes, Properties, Methods, Events, CSS.

### Keep it honest for non-video elements

We will later support elements that are not backed by a native `<video>`, such as video-in-canvas. Drive any "real `<video>`" framing from element metadata, not a blanket sentence, so it stays true per element. One open question decides how far to trust an MDN link: will those elements wrap a native element, where MDN still applies, or synthesize the media API, where we lean on the contract descriptions instead?

decepulis:
Fixes #1717

## Summary

Make media element reference pages describe the complete public API exposed by each framework. The generator now derives HTML and React reference surfaces from normal source conventions, keeping references accurate without per-element documentation configuration.

## Changes

- Surface inherited native media properties such as `currentTime`, `duration`, `volume`, `buffered`, `textTracks`, and video dimensions.
- Generate standard and Video.js-specific attributes from runtime property mappings, host accessors, JSDoc, and defaults; `stream-type` is now documented from its source definition.
- Give HTML and React distinct reference models: HTML documents custom-element attributes, properties, methods, events, and CSS; React documents component props, its native ref target, and event handling.
- Respect the actual rendered target (`video`, `audio`, or `iframe`), so iframe-backed adapters do not inherit fictional native video APIs and list only events they implement.
- Add black-box end-to-end fixtures for native inheritance, React conventions, and iframe-backed media.

<details>

The generator treats existing runtime conventions as its input: `CustomMediaElement(...)` declares the HTML target and property mappings, `forwardRef<...>` declares the React ref target, native React attribute inheritance opts into standard DOM props, and the defaults object passed to `useSyncProps` declares Video.js-specific React props. The shared frontend renders a normalized, platform-specific model.

</details>

## Testing

- `pnpm build:packages`
- `pnpm -F site api-docs`
- `pnpm test`
- `pnpm -F site test` (490 tests)
- `pnpm typecheck`
- `pnpm exec astro check` (0 errors)
- `pnpm lint`
- `pnpm check:workspace`
- Manual browser verification of the HTML and React reference pages

---

> [!NOTE]
> **Medium Risk**
> Large schema and generator changes affect all generated media reference JSON and docs pages; behavior is heavily tested but any missed consumer of the old shape would break at build time.
> 
> **Overview**
> **Media element API reference pages** now describe the full public surface per framework instead of a single flat host-property list.
> 
> The **api-docs builder** reshapes output into `platforms.html` and optional `platforms.react`: HTML gets target tag (`video` / `audio` / `iframe`), standard vs Video.js-specific attributes, property definitions plus a compact **native passthrough** list (from shared `media-host` / `video-host` / `audio-host`), methods, events, and CSS. React is inferred from `forwardRef`, `VideoHTMLAttributes` / `AudioHTMLAttributes`, and `useSyncProps` defaults. **`CustomMediaElement`** discovery now records `iframe` targets and can find compositions on thin subclasses (e.g. Vimeo-style). Iframe-backed elements skip fictional native video APIs; only host-implemented attributes and dispatched/`emit` events are documented.
> 
> **Site rendering** splits into `HtmlMediaReference.astro` and `ReactMediaReference.astro`, with TOC/section models updated for platform-specific headings. **`streamType`** on `HTMLMediaElementHost` gains JSDoc used in generated docs.
> 
> E2E fixtures add **EmbedVideo** (iframe), React convention tests, and native-property inheritance coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 160c46be39f0316cdec5561a2ab8b0879aaff151. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/site/scripts/api-docs-builder/src/tests/fixtures/monorepo/packages/core/src/dom/media/custom-media-element/index.ts
+++ b/site/scripts/api-docs-builder/src/tests/fixtures/monorepo/packages/core/src/dom/media/custom-media-element/index.ts
@@ -81,6 +81,7 @@ export function CustomMediaElement(tag: string, Host: any) {
       poster: { type: String },
       preload: { type: String },
       src: { type: String },
+      streamType: { type: String, attribute: 'stream-type' },
     };
   }
 
--- a/packages/core/src/dom/media/media-host.ts
+++ b/packages/core/src/dom/media/media-host.ts
@@ -125,6 +125,11 @@ export class HTMLMediaElementHost<Target extends HTMLMediaTargetLike, Events ext
     this.dispatchEvent(new (event.constructor as typeof Event)(event.type, event));
   };
 
+  /**
+   * Current stream type (`'on-demand'`, `'live'`, or `'unknown'`). Defaults to
+   * `'unknown'`; detecting hosts update it automatically, and consumers can set
+   * it to override detection.
+   */
   get streamType() {
     return getProp(this, 'streamType') ?? this.#streamType;
   }
--- a/site/scripts/api-docs-builder/src/media-element-handler.ts
+++ b/site/scripts/api-docs-builder/src/media-element-handler.ts
@@ -24,7 +24,14 @@ import * as path from 'node:path';
 import * as ts from 'typescript';
 import * as tae from 'typescript-api-extractor';
 import { extractCSSVars } from './css-vars-handler.js';
-import type { HostPropertyDef, MediaElementReference, MediaElementResult, MediaEventDef } from './pipeline.js';
+import type {
+  HostPropertyDef,
+  MediaElementReference,
+  MediaElementResult,
+  MediaEventDef,
+  MediaTargetTag,
+  ReactMediaReference,
+} from './pipeline.js';
 
 // ─── Constants ──────────────────────────────────────────────────────
 
@@ -46,6 +53,12 @@ interface MediaElementSource {
   hostFilePath: string;
   hostClassName: string;
   mediaType: 'video' | 'audio';
+  targetTag: MediaTargetTag;
+}
+
+interface StaticMediaProperty {
+  property: string;
+  attribute: string;
 }
 
 // ─── Module Resolution ───────────────────────────────────────────────
@@ -206,6 +219,7 @@ function parseDefineFile(
     hostFilePath,
     hostClassName: hostInfo.hostClassName,
     mediaType: hostInfo.mediaType,
+    targetTag: hostInfo.targetTag,
   };
 }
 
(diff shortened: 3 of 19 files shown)


site/scripts/api-docs-builder/src/tests/fixtures/monorepo/packages/core/src/dom/media/custom-media-element/index.ts
```diff
@@ -81,6 +81,7 @@ export function CustomMediaElement(tag: string, Host: any) {
       poster: { type: String },
       preload: { type: String },
       src: { type: String },
+      streamType: { type: String, attribute: 'stream-type' },
```
luwes, line 84: 👍

luwes (approved):
if the output looks good 👍

PR #1726, opened 2026-06-23
fix(skin): aspect ratio related fixes

### LINKED ISSUE #1716: Bug: Aspect ratios set in the video skin are ignored

HTML (@videojs/html)

This affected all OS/Browser combinations that I have tried using the HTML version of Video.JS 10 and was initially reported over on Discord. I'm using a downloaded version of the Video JS libraries and not building with Node.

### Description

I'm working with a library of HLS videos where the aspect ratios of the videos themselves can vary, but will always have a 16:9 thumbnail for the poster image and the embeds need to always be 16:9 to accommodate fixed iframes on other websites. 

So I wanted to fix the aspect ratio of the Video.JS player to 16:9 with Video.JS simply padding the video with black bars if the video had an aspect ratio other than 16:9.

I tried setting the aspect in the video skin, according to advice received on Discord:

<video-skin style="aspect-ratio:16/9;">
    <hls-video id="vjs" src="4by3video.m3u8" poster="16by9image.jpg" playsinline>/hls-video>
</video-skin>

and tested using a 16:9 thumbnail and 4:3 underlying video. 

To illustrate what is happening, I've done a test page with a fixed red 16:9 border around VideoJS, when the page loads in Chrome, everything displays as expected.

Image

However when I click play, the video expands outside of the red box to accommodate the actual 4:3 aspect of the video file. My expectation is that the player will obey the 16:9 ratio set for the display area and will play the video with black bars down each. Once the player has loaded I would not expect it to change size, unless the user resizes the page.

Image

If I run the same test in Firefox, the correctly sized video appears for ~1 second during page load, but it the re-sizes and flows out of the box. I assume that FF is just pre-loading the meta data that triggers the resize vs Chrome not doing this until play is clicked.

Image

If I reverse the test, using a 16:9 video in a 4:3 box, then behaviour is as expected, the video plays with black bars on the top and bottom to pad out the space and the video player does not shrink to the height of the video.

Image

So the problem seems to be confined to videos that have an aspect ratio that is taller than that initially configured.

I have come up with a work-around for this problem which involves listening for the "canplay" event and then using some Javascript to find the video tag inside the VideoJS shadow root and manually setting the aspect-ratio style to 16/9. I think the fix that needs to be applied in Video.JS is for the aspect-ratio style that is set in the video-skin to be copied into the underlying video tag when the latter is created. I did try setting an aspect ratio style on the hls-video tags, but this did not work either.

### Reduced test case

_No response_

### Steps to reproduce

1.Create a Video.JS player with a 16:9 poster image and 4:3 video
2.Load page. In Firefox the player will initially display 16:9 but will change to 4:3 shortly after load. Chrome will change to 4:3 when play is clicked.
3.

### Current Behavior

The configured aspect ratio is ignored either during player load or when play is clicked, depending on browser.

### Expected Behavior

The configured aspect ratio for player display is obeyed and black bars are used to pad the video to match.

### Errors

_No response_

v10.0.0-beta.24

sampotts:
- Set `min-width: 0` and `min-height: 0` on `media-container` in the HTML skins so that slotted portrait videos can’t force the shadow container past the skin host’s authored aspect ratio. 
- Set a `max-height` on the slider thumbnails so that portrait thumbs don't overflow. I could have gone fancy here and used `cqb` units but I think this simple fix is enough for now. 
- Added a portrait video example (Dar's Tailwind video) as a source option for future testing. 

Closes #1716

---

> [!NOTE]
> **Low Risk**
> CSS and layout-only changes with new visual tests; no playback, auth, or data-path changes.
> 
> **Overview**
> Fixes layout when **portrait** video is slotted into a skin with an authored **16:9** (or other) aspect ratio.
> 
> **`media-container`** in shared HTML skin CSS now uses **`min-width: 0`** and **`min-height: 0`** so slotted portrait media cannot expand the shadow container past the host’s aspect ratio.
> 
> **Slider storyboard thumbnails** in default and minimal skins (CSS and Tailwind) gain **`--media-slider-thumbnail-max-height: 8rem`** on the preview image, replacing the old loading-only container cap so tall portrait thumbs stay bounded.
> 
> Adds Playwright checks for authored aspect ratio and thumbnail height, plus sandbox source **`hls-6`** (portrait Tailwind HLS) for manual testing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a64a67f60762de87c1501f4adec5a603f722ad50. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/e2e/tests/visual/video-skin.spec.ts
+++ b/apps/e2e/tests/visual/video-skin.spec.ts
@@ -62,7 +62,7 @@ test.describe('Visual — HTML Portrait Layout', () => {
 
       root.innerHTML = `
         <video-player>
-          <video-skin style="max-width: 800px; aspect-ratio: 16/9">
+          <video-skin style="width: 320px; aspect-ratio: 16/9">
             <video width="270" height="480" playsinline></video>
           </video-skin>
         </video-player>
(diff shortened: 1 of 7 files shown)


luwes (approved):
👍

mihar-22 (changes requested):
This is a skins related change so I'm just temporarily blocking here due to compiler work.

mihar-22 (approved):

PR #1727, opened 2026-06-23
feat(packages): remove native controls after mounting our controls

### LINKED ISSUE #1160: Feature: Remove Native Controls When Custom Control Is Loaded
Migrated from [this media-chrome issue](https://github.com/muxinc/media-chrome/issues/1281)

sampotts:
## Summary

Remove the native controls after  mounting our custom controls component. This allows for progressive enhancement and has been noted in the migration docs. 

Given it affects skins, I won't merge until @mihar-22 gives it the ok. 

Closes #1160 

---

> [!NOTE]
> **Medium Risk**
> Touches core player attach lifecycle and mutates native `controls` on media elements across HTML and React; behavior is covered by new tests but regressions could affect fallback playback without custom UI.
> 
> **Overview**
> Enables **progressive enhancement**: media can ship with native `controls` until a custom controls UI mounts, then native controls are turned off and restored if all custom controls unmount.
> 
> A new **controls registration** path tracks mounted control roots (HTML `media-controls` via player context; React `Controls.Root` via `registerControls` / `controlsMounted`). The provider mixin and React provider use `isMediaControlsCapable` to set `media.controls = false` only while at least one controls instance is registered, with refcounting when multiple roots exist.
> 
> Skins and presets wrap control chrome in **`media-controls` / `Controls.Root`** instead of plain divs, and `defineControls()` is wired into audio/live-audio UI registration. React **`Video` / `Audio`** omit the `controls` attribute when `controlsMounted` is true.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 14150711a3f16a8228385fddde5a2f8b0bff4cb6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/html/src/store/provider-mixin.ts
+++ b/packages/html/src/store/provider-mixin.ts
@@ -5,15 +5,17 @@ import {
   type PlayerStore,
   type PlayerTarget,
 } from '@videojs/core/dom';
+import { isMediaControlsCapable } from '@videojs/core/media/predicate';
 import { ContextProvider } from '@videojs/element/context';
 import { isNull } from '@videojs/utils/predicate';
 import type { MediaElementConstructor } from '@/ui/media-element';
-import type { ContainerContext, MediaContext, PlayerContext } from '../player/context';
+import type { ContainerContext, ControlsContext, MediaContext, PlayerContext } from '../player/context';
 import type { PlayerProvider, PlayerProviderConstructor } from './types';
 
 export interface ProviderMixinConfig<Store extends PlayerStore> {
   playerContext: PlayerContext<Store>;
   mediaContext: MediaContext;
+  controlsContext: ControlsContext;
   containerContext: ContainerContext;
   factory: () => Store;
 }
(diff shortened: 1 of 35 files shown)


packages/html/src/store/provider-mixin.ts
```diff
@@ -137,6 +138,7 @@ export function createProviderMixin<Store extends PlayerStore>(
         if (hasMediaChanged || hasContainerChanged) {
           this.#detachStore();
           this.#detach = store.attach(target);
+          if (isMediaControlsCapable(target.media) && target.media.controls) target.media.controls = false;
```
luwes: I read the issue but I find it slightly presumptuous. what if the user does want to use the native controls for some reason.

mihar-22 (commented):
I don't believe we have a progressive enhancement story? Is there an example where this would come up?

Edit: Ah, if someone passes in `<video>` on the HTML side, makes sense to me now 👍

packages/html/src/store/provider-mixin.ts
mihar-22: I can see that but at the same time I think it'd be unlikely someone has both provided custom controls and wants native controls?

sampotts: I'd have preferred to add the native controls removal logic to our controls component  but we don't actually have one (it's just a `<div>`) so we'd have to add a `<media-controls>` / `<Controls>` component, which would obviously add a tiny bit more to the bundle.

mihar-22: We do have controls components?

sampotts: "You're absolutely right!" 

I don't know why we're not using them currently in the skins (probably an oversight on my part way back) but I'll update and put the logic there instead.

sampotts:
I've changed it to be based on mounting of our custom controls instead of the container. This would allow consumers to use native controls if they wanted to.

mihar-22:
Going to close this out as `Controls` have been migrated over separately and I'd love for a simpler solution here for toggling native controls, preferably without a separate controls context.

PR #1732, opened 2026-06-23
feat: add DASH, Mux, and Vimeo as installation source types (UI + CLI)

decepulis:
Closes #1255.

Adds DASH, Mux, and Vimeo to the installation guide and installation CLI.

### Where?
- [React](https://deploy-preview-1732--vjs10-site.netlify.app/docs/framework/react/how-to/installation)
- [HTML](https://deploy-preview-1732--vjs10-site.netlify.app/docs/framework/html/how-to/installation)

### What?
Try the picker or drop in a Mux / Vimeo / DASH URL

CleanShot 2026-06-25 at 15 50 16@2x

## Out of scope (follow-ups)
#1749, #1750, #1751 

---

https://claude.ai/code/session_01Kd5A9UsSz5Ka7nYhEik3Fd

---

> [!NOTE]
> **Low Risk**
> Changes are confined to installation docs/codegen, CLI flags, and UI picker behavior with broad test coverage; no auth or core playback runtime changes in this diff.
> 
> **Overview**
> Adds **DASH**, **Mux** (video and audio), and **Vimeo** as selectable media source types on the installation guide and in `@videojs/cli docs how-to/installation`, including generated HTML/React snippets, imports, and CDN script tags where builds exist.
> 
> Renderer labels and picker ordering are centralized in shared `renderer-options` (site + CLI) so the UI and CLI stay aligned. CDN install snippets and the CDN tab are gated by the **cdn-media manifest**: DASH, HLS, and Mux get extra media scripts when listed; **Vimeo** is npm-only—the CLI errors on `--install-method cdn`, and the install tabs reset `installMethod` when CDN becomes unavailable so usage code doesn’t drop required JS imports.
> 
> URL auto-detection now prefers **Mux** over generic HLS for `stream.mux.com`, recognizes Vimeo hosts, maps **`.mpd`** to DASH, and falls through Mux video→audio rules by use case. Demo sample URLs for DASH and Vimeo were added for empty `--source-url`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f44fdfd8881f30b43abedf741e895e78f414589b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Issue #1255, opened 2026-04-07
Docs: Add Missing Media Elements to Installation Page


decepulis:
## Summary

The installation page's renderer picker (`site/src/components/installation/RendererSelect.tsx`) is missing several media elements that are already implemented in `packages/html/`.

## Currently active

- `background-video`
- `hls`
- `html5-audio`
- `html5-video`

## Missing (commented out or absent)

These media elements exist in `packages/html/src/define/media/` but aren't available in the installation picker:

- `dash` — `<dash-video>` element exists
- `native-hls-video` — `<native-hls-video>` element exists (absent from list entirely)
- `simple-hls-video` — `<simple-hls-video>` element exists (absent from list entirely)
- `mux-video` — `<mux-video>` element exists (commented out)
- `mux-audio` — `<mux-audio>` element exists (added in #1259, commented out in picker)

Additionally, several platform renderers are commented out (`cloudflare`, `jwplayer`, `mux-background-video`, `spotify`, `vimeo`, `wistia`, `youtube`). These may not have corresponding elements yet — only uncomment ones that are actually implemented and ready to use.

## What to do

1. Uncomment / add entries in `RENDERER_LABELS` for media elements that have working implementations
2. Add them to the appropriate `VALID_RENDERERS` use case arrays (in `site/src/stores/installation.ts`)
3. Add detection patterns in `site/src/utils/installation/detect-renderer.ts` if applicable
4. Verify the installation flow works end-to-end for each newly added renderer

## PR plan

Part of **PR3** alongside #1254 (installation code sample sizing fix).

Parent issue: #727

PR #1737, opened 2026-06-24
feat: add a CDN media-availability manifest and read it across install surfaces

decepulis:
So I'm working towards #1255 -- putting media elements on the installation page -- and it's come up that not all media elements have a cdn distro. So the installation page needs to know which ones do and don't.

This PR addresses that gap. 
1. It reads the source code and generates a manifest of CDN-ified media elements. 
2. It reads that manifest in the installation doc (for humans) and the CLI (for the robots)

This PR is a prerequisite to #1732. 

## How do I review this?
There's kinda three slices of code to know about

**Manifest generation**
1. `scripts/build-cdn-manifest.ts` is an ai-driven black box that does the reading. 
2. `turbo.json` makes sure it's called at the right time. 
3. `content.config.ts` tells astro what to expect from the manifest. 
4. `rendererSupportsCdn()` reads the manifest and does the checking.

**UI for humans**
- **`HTMLInstallTabs.astro` / `HTMLInstallTabsClient.tsx`** calls `rendererSupportsCdn()` and does some UI stuff in response

**CLI for Robots**
- **`@videojs/cli`** reads the same manifest (bundled by tsdown; a committed fixture backs the hermetic tests), exposes `supportsCdnInstall`, offers the CDN install method only when the renderer ships a CDN build, and rejects `--install-method cdn` for one that doesn't.

---

https://claude.ai/code/session_01Kd5A9UsSz5Ka7nYhEik3Fd

---

> [!NOTE]
> **Low Risk**
> Changes are limited to install UX, generated docs content, and build tooling; no runtime player or auth paths are modified.
> 
> **Overview**
> Adds a **build-time CDN media manifest** derived from `@videojs/html` CDN output so install flows know which renderers actually ship CDN bundles.
> 
> A new `cdn-manifest` script writes gitignored `cdn-media.json`, wired into site `dev`/`build` via Turbo (after `build:cdn`) and exposed as the `cdnMedia` Astro collection. Shared logic **`rendererSupportsCdn`** treats preset renderers as always CDN-capable and requires a media renderer’s subpath (e.g. `hls-video`) to appear in that manifest.
> 
> On the **installation doc**, HTML install tabs now load the manifest server-side and **hide the CDN tab** (with a short npm-only message) when the selected renderer has no CDN build, including a keyed remount so tab state stays in sync.
> 
> The **CLI** bundles the same manifest, offers CDN only in prompts when supported, and **errors on `--install-method cdn`** for unsupported renderers so generated snippets can’t reference missing CDN scripts. Hermetic CLI tests use a committed fixture mirroring the manifest shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea47cb6f5512f2435c522e77ee041d0f444cebee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


site/src/components/installation/HTMLInstallSection.astro
```diff
@@ -0,0 +1,10 @@
+---
+import { getCollection } from 'astro:content';
+import HTMLInstallTabs from './HTMLInstallTabs';
+
+// Read the generated CDN media manifest server-side and hand the list to the
+// client island, which decides per-renderer whether to offer the CDN option.
+const cdnMedia = (await getCollection('cdnMedia')).map((entry) => entry.id);
+---
+
+<HTMLInstallTabs client:idle cdnMedia={cdnMedia} />
```
decepulis: Elsewhere in the codebase, we have this pattern where we have a server half and a client half of the component. e.g., 
```
HTMLInstallTabs.astro
HTMLInstallTabsClient.tsx
```
I'd prefer you use that here to clearly communicate that these are two halves of the same whole

site/src/components/installation/HTMLInstallTabs.tsx
```diff
@@ -1,14 +1,37 @@
...
+      installMethod.set('npm');
+    }
+  }, [supportsCdn]);
+
+  // Observe tab changes and sync to installMethod store. The effect body
+  // doesn't read supportsCdn, but it must re-run when it changes: adding or
+  // removing the CDN tab changes the rendered tab set, and the observer has to
+  // re-attach to the current tabs (otherwise switching to/from a no-CDN
+  // renderer leaves installMethod tracking stale tabs).
+  // biome-ignore lint/correctness/useExhaustiveDependencies: re-attach observers when the tab set changes
```
decepulis: What implications does this awkward pattern have for react compiler?

site/src/components/installation/HTMLInstallTabs.tsx
```diff
@@ -59,6 +88,11 @@ export default function HTMLInstallTabs() {
           <ClientCode code="bun add @videojs/html" lang="bash" />
         </TabsPanel>
       </TabsRoot>
+      {!supportsCdn && (
+        <p className="my-4 mx-auto max-w-3xl">
+          This source type isn't available via CDN — install it with a package manager (npm, pnpm, yarn, or bun).
+        </p>
```
decepulis: I'd prefer you use the typography/P component. Or, I guess since this is a react context and you can't use astro components, at least use `~/v10/site/src/components/typography/styles.ts` to DRY the styles

decepulis (commented):
Claude, for when you're done with your current task, here's a few cleanup items

site/src/components/installation/HTMLInstallTabs.tsx
decepulis: Good question. The short answer: no real React Compiler implications, but the pattern was awkward for other reasons, so I removed the awkwardness.

On the compiler specifically: `babel-plugin-react-compiler` memoizes render output but it does **not** touch `useEffect` dependency arrays — it leaves them exactly as authored and never tries to infer effect deps. So the deliberately-narrowed `[supportsCdn]` + `biome-ignore useExhaustiveDependencies` was safe (no auto-"fix", no bailout); the imperative DOM reads happened in an effect, which the compiler permits. The awkwardness was really readability: depending on `supportsCdn` purely to re-attach observers when the tab set changes, while the effect body never reads it.

That's now gone. Instead of attaching one observer per `[role="tab"]` and re-running when the set changes, there's a single `MutationObserver` on the tabs root with `subtree: true` + `attributeFilter: ['data-tab-active']`. The subtree watch automatically covers the CDN tab when it's added/removed, so the effect attaches once on mount — deps are `[]` and honest, and the lint suppression is gone. Pushed in 060fae2.

---

packages/cli/src/utils/prompts.ts
```diff
@@ -1,10 +1,22 @@
...
+import { rendererSupportsCdn } from '@/utils/installation/cdn-code';
 import type { InstallationOptions } from '@/utils/installation/codegen';
 import { detectRenderer } from '@/utils/installation/detect-renderer';
 import type { InstallMethod, Renderer, Skin, UseCase } from '@/utils/installation/types';
 import { VALID_RENDERERS } from '@/utils/installation/types';
 import type { Framework } from './config.js';
 
+// Media subpaths that ship a CDN build, from the generated manifest (bundled at
+// build time). Mirrors how the installation page reads the `cdnMedia` content
+// collection — same source of truth, so the CLI gates CDN exactly like the UI.
```
decepulis: Unnecessary comment imo

packages/cli/src/utils/prompts.ts
```diff
@@ -1,10 +1,22 @@
...
+const CDN_MEDIA_SUBPATHS = cdnMedia.map((entry) => entry.id);
+
+/** Whether a renderer can be installed via CDN (parity with the install page). */
```
decepulis: Unnecessary comment; feels redundant with the name of the function to me

site/src/components/installation/HTMLInstallTabs.tsx
decepulis, line 1: Bummer that this is marked as a deletion, not a rename, in git.

site/src/components/installation/HTMLInstallTabsClient.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    const observer = new MutationObserver(() => {
+      const value = root.querySelector('[role="tab"][data-tab-active="true"]')?.getAttribute('data-value');
+      if (value) installMethod.set(value as InstallMethod);
+    });
+
+    observer.observe(root, { subtree: true, attributes: true, attributeFilter: ['data-tab-active'] });
+
+    return () => observer.disconnect();
+  }, []);
```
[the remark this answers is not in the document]
decepulis: Good catch — this is real (and only reachable once a no-CDN renderer like Vimeo exists, so it's latent here but live on #1732). When `supportsCdn` flips, the npm tab's `initial={!supportsCdn}` re-fires `Tab`'s `[initial]` effect and the CDN tab mounts/unmounts, each setting `data-tab-active` without clearing the previously-active tab — so two tabs end up active and the observer's "first active" can disagree with the visible tab + usage snippet.

Fixed in 1ceea75 by remounting the tab set on CDN-availability change (`key={supportsCdn ? 'with-cdn' : 'without-cdn'}`), which resets to a single active initial tab. The observer now watches from the stable wrapper instead of the tabs root so it survives the remount, and the separate cdn→npm reconcile effect is gone (redundant — the remount + observer handle it).

---

site/src/components/installation/HTMLInstallTabs.tsx
decepulis: Yeah — that's GitHub's rename detection, which is similarity-based on the *aggregate* base↔head content of the file, not on commit structure. The rename landed in the same change as the observer rewrite + styling refactor (and now the desync fix touches it again), so the new `HTMLInstallTabsClient.tsx` is similar enough to *less* than GitHub's threshold against the old `HTMLInstallTabs.tsx`, and it falls back to delete + add.

I can't force it back to a "renamed" badge without shrinking that net diff (which would mean undoing the refactors). Splitting the rename into its own pure `git mv` commit wouldn't help the squashed/Files-changed view either, since that compares final content regardless of commit boundaries — though it would read as a clean rename if you review commit-by-commit. Happy to restructure the commits that way if that's how you review; otherwise it's purely cosmetic and the file's history is intact via `git log --follow`.

---

site/src/components/typography/styles.ts
```diff
@@ -13,6 +13,8 @@ export const shared = {
   em: 'italic',
   li: '',
   ol: 'list-decimal list-outside pl-4 space-y-1',
+  p: 'my-4',
+  pMaxWidth: 'mx-auto max-w-3xl',
```
sampotts: question: Is this like a `prose` container? Just wondering if `pMaxWidth` is the best name but it's a nit pick for sure.

sampotts (approved):

site/src/components/typography/styles.ts
decepulis: Not a Tailwind `prose` container (no typography-plugin styles here) — it's literally just `mx-auto max-w-3xl`, i.e. the centered max reading width. I named it `pMaxWidth` to mirror `P.astro`'s existing `maxWidth` prop (which toggles exactly these classes), and to fit this file's element-keyed convention (`a`, `code`, `ol`, `p`, …) — `p` + its max-width variant.

Fair that the name under-sells the `mx-auto` half, though. If you'd prefer the typographic term I'm happy to rename it to `pMeasure` (or `pReadingWidth`) across the two consumers — your call. Leaving it as-is for now since it tracks the prop name a reader already knows.

---

sampotts: Yeah, I didn't mean a Tailwind-specific `prose` classname, more that it's _used for prose_ which `pReadingWidth` suggests it is? If that's the case I'd just go with `prose` for the property name but as I say, I don't feel super strongly about it.

decepulis: Makes sense — went with `prose`. Renamed across `styles.ts`, `P.astro`, and the install-tabs note in ea47cb6.

---

decepulis: Good job claude-dar. I didn't love the `pMaxWidth` name either. I'm not thrilled about `prose` either, but it's better, so, let's keep it!

PR #1755, opened 2026-06-26
fix(site): resolve api reference slug mismatch on airplay and hlsjs pages

luwes:
## Summary

The `airplay-button` and `hlsjs-video` reference pages were silently rendering without their API reference tables. Both `ComponentReference` and `MediaReference` look up generated JSON via `kebabCase(name)`, but those two entries don't follow that derivation, so the lookups returned nothing and the component bailed early.

## Changes

- Add an optional `slug` override to `ComponentReference` and `MediaReference`, mirroring the existing escape hatch on `UtilReference` (`slug ?? kebabCase(name)`).
- Wire the `slug` override through the `remarkConditionalHeadings` TOC injection so the headings still appear in the table of contents.
- Pass explicit slugs on the two affected pages, keeping the public-facing names `AirPlayButton` / `HlsJsVideo` intact.

<details>
<summary>Why these two mismatched</summary>

- `AirPlayButton` kebab-cases to `air-play-button`, but the generated file is `airplay-button.json` (the core component is named `AirplayButton`).
- `HlsJsVideo` kebab-cases to `hls-js-video`, but media files are named by element tag name, `hlsjs-video.json`.

All other reference pages happen to satisfy `kebabCase(name) === entry id`, so the override is opt-in and changes nothing for them.

</details>

## Testing

`pnpm turbo run build --filter=site` builds clean (the previous "Entry ... was not found" warnings are gone) and both pages now render their API Reference sections. `pnpm -F site test` passes (441 tests).

---

> [!NOTE]
> **Low Risk**
> Docs-site-only slug lookup and shared naming map; no runtime product or auth changes.
> 
> **Overview**
> **AirPlayButton** and **HlsJsVideo** API reference pages were empty because content lookups used `kebabCase(name)`, which does not match generated JSON slugs (`airplay-button`, `hlsjs-video`).
> 
> This PR introduces **`site/src/utils/api-reference-overrides.ts`** as the single **`NAME_OVERRIDES`** map (including **`airplay-button`** and **`hlsjs-video`**), with **`resolveReferenceSlug`** inverting it for PascalCase → slug and falling back to **`kebabCase`**. The api-docs builder imports that map from the site utils (re-exported from **`pipeline.ts`**), and **`ComponentReference`**, **`MediaReference`**, and **`remarkConditionalHeadings`** all use **`resolveReferenceSlug`** for **`getEntry`** and TOC JSON loading. Builder docs now point maintainers at the shared file; Vitest covers override and default kebab behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2819ebd9f0b563af24fda4e0cffec6c9cea8423. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude/skills/api-reference/references/builder-conventions.md
+++ b/.claude/skills/api-reference/references/builder-conventions.md
@@ -28,15 +28,18 @@ The builder derives PascalCase from kebab-case using `kebabCase` from es-toolkit
 
 ## NAME_OVERRIDES
 
-When kebab-to-pascal conversion doesn't produce the correct name, add an override in `site/scripts/api-docs-builder/src/index.ts`:
+When kebab-to-pascal conversion doesn't produce the correct name, add an override in `site/src/utils/api-reference-overrides.ts` (the shared map the builder imports and the reference pages invert for slug lookup):
 
 ```ts
-const NAME_OVERRIDES: Record<string, string> = {
+export const NAME_OVERRIDES: Record<string, string> = {
   'pip-button': 'PiPButton',
+  'airplay-button': 'AirPlayButton',
 };
 ```
 
 Use overrides only when the standard conversion fails (e.g., acronyms like PiP). Prefer aligning component naming with the standard conversion when possible.
+
+Media elements use a separate `MEDIA_SLUG_OVERRIDES` map in the same file for the rare case where a component's PascalCase name doesn't kebab-case to its element tag name (e.g. `HlsJsVideo` → `hlsjs-video`).
 
 ## Multi-Part Components
 
--- a/.claude/skills/api-reference/references/builder-conventions.md
+++ b/.claude/skills/api-reference/references/builder-conventions.md
@@ -39,7 +39,7 @@ export const NAME_OVERRIDES: Record<string, string> = {
 
 Use overrides only when the standard conversion fails (e.g., acronyms like PiP). Prefer aligning component naming with the standard conversion when possible.
 
-Media elements use a separate `MEDIA_SLUG_OVERRIDES` map in the same file for the rare case where a component's PascalCase name doesn't kebab-case to its element tag name (e.g. `HlsJsVideo` → `hlsjs-video`).
+The same map covers media elements whose PascalCase name doesn't kebab-case to their element tag name (e.g. `'hlsjs-video': 'HlsJsVideo'`). It is keyed by the generated-reference file slug regardless of component vs. media.
 
 ## Multi-Part Components
 
--- a/site/scripts/api-docs-builder/src/pipeline.ts
+++ b/site/scripts/api-docs-builder/src/pipeline.ts
@@ -34,6 +34,7 @@ import { kebabToPascal, partKebabFromSource, sortProps } from './utils.js';
 // Components whose PascalCase name doesn't match simple kebab-to-pascal conversion.
 export const NAME_OVERRIDES: Record<string, string> = {
   'pip-button': 'PiPButton',
+  'airplay-button': 'AirPlayButton',
 };
 
 // Parts whose HTML element file doesn't follow the `{component}-{part}-element.ts` convention.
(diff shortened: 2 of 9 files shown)


decepulis (commented):
Thanks for this! The slug override is the right call for the lookup mismatch, and I confirmed it's correctly wired through both `ComponentReference.astro`/`MediaReference.astro` **and** the `remarkConditionalHeadings` TOC injection.

One thing to flag before merge though: I ran the generator locally (fresh `pnpm build:packages` first) and the two pages don't behave the same way.

- **`hlsjs-video` is fully fixed.** `hlsjs-video.json` generates fine and the slug resolves it. ✅
- **`airplay-button` is still empty after this PR.** The slug points the lookup at `airplay-button.json`, but that file is **never generated**, so the page renders nothing regardless.

### Why airplay's JSON doesn't generate

The builder locates a component's exports by name: `extractCore` looks for `${componentName}Props` / `${componentName}State` (`core-handler.ts:18-24`), where `componentName = NAME_OVERRIDES[dir.name] ?? kebabToPascal(dir.name)` (`pipeline.ts:124`).

For `airplay-button` there's no override, so `componentName` = `kebabToPascal("airplay-button")` = `AirplayButton`, and it searches for `AirplayButtonProps`. But the source exports are brand-cased — `AirPlayButtonProps` / `class AirPlayButtonCore` (`airplay-button-core.ts:24`). Miss → `extractCore` returns null → the component is silently dropped (the run reports 25 components, not 26; no warning). So `NAME_OVERRIDES` is **not** just a cosmetic display-name map — it gates whether the JSON gets written at all.

This is the exact case `NAME_OVERRIDES` already exists for — it's why `'pip-button': 'PiPButton'` is in the map (the class is `PiPButtonCore`, not `PipButtonCore`). `hlsjs-video` avoids the issue because media references are discovered via define files and named by `tagName` (`index.ts:153`), not by `kebabToPascal`-ing a directory — which is why it's fully fixed and airplay isn't.

So airplay needs **both** halves: a `NAME_OVERRIDES` entry (so the JSON generates) *and* the slug (so the page finds it). I verified locally — adding the override below makes the builder find 26 components and emit `airplay-button.json` (name `AirPlayButton`, 2 props, 3 state).

### Minimal fix (recommended for this PR)

Add to `NAME_OVERRIDES` in `site/scripts/api-docs-builder/src/pipeline.ts`:

```ts
export const NAME_OVERRIDES: Record<string, string> = {
  'pip-button': 'PiPButton',
  'airplay-button': 'AirPlayButton',
};
```

Keep the `slug="airplay-button"` and the branded `component="AirPlayButton"` on the MDX page exactly as they are. Smallest change, mirrors the existing PiP pattern.

One note in case it comes up: an alternative would be to drop the slug and just change the prop to plain `component="AirplayButton"` (which `kebabCase`s to `airplay-button` on its own). That renders, but it would break the `api-reference-sync` workflow — it does `NAME=$(jq -r .name …)` then `grep "<ComponentReference component=\"$NAME\""` (`.github/workflows/api-reference-sync.yml:180,184`) to decide whether a reference page already exists. With `name` = `AirPlayButton`, a plain-cased prop wouldn't match the grep and the bot would open a duplicate "missing reference page" issue. The branded-prop-plus-slug combo in this PR matches that grep, so the slug route is the better of the two.

### Root-cause fix (optional, separate PR)

The deeper issue is that consumers *recompute* the lookup id via `kebabCase(displayName)` instead of reading an id the builder already knows (`result.kebab`). Emitting a `slug`/`id` field into the component + media JSON and looking up by that — which feature references already do (`feature-reference.ts:16`) — would kill this whole class of silent miss for every future irregular name, no per-page annotation.

Worth noting it does **not** replace `NAME_OVERRIDES`: that's about matching brand-cased TS exports at extraction time (and feeding the sync workflow's grep), which is distinct from the lookup slug. So the root-cause fix removes the need for per-page `slug` props, but the `airplay-button` override is still required either way.

Happy to leave the call on which route to you — just wanted to make sure the airplay page doesn't ship still-empty.

decepulis (approved):
Sorry for all the Claude noise. This looks good to me, now! Thank you!

PR #1541, opened 2026-05-13
fix(react): memoize Provider context value

R-Delfino95:
Related to #1296
 
## What changed
 
**Diff:** 2 files changed, 58 insertions(+), 7 deletions(-).

- **`packages/react/src/player/create-player.tsx`** — Provider context value wrapped in `useMemo` keyed on `[store, media, container, popupGroup]`. `setMedia` and `setContainer` are omitted (identity-stable `useState` setters). `popupGroup` (added on `main` since the original PR was opened) is included in both the value and the dep array.

- **`packages/react/src/player/tests/create-player.test.tsx`** — one regression test added: mounts a `usePlayerContext()` consumer, forces parent re-renders three times, asserts context value identity is preserved (`toBe`).

**Test results in `packages/react`:**

| Step | Result |
|---|---|
| Baseline (untouched `main`) | 17 files, **237 passing** |
| With fix applied | 17 files, **238 passing** (+1 new) |
| Source reverted, new test kept | 1 new test **fails** (fail-before/pass-after contract) |
| Source re-applied | **238/238 passing** |
| `pnpm typecheck` (root) | Clean |
| `pnpm lint` on the 2 changed files | Clean |

The `attachMediaElement` `WeakMap` fix has been fully discarded. Worth noting that #1376 has since landed on `main`, which introduced `useAttachMedia` (a `useCallback`-based hook keyed on `[media]`), so the ref-identity stability that fix #2 was guarding against is now handled at the hook level — fix #2 isn't just unwanted, it's obsolete against current `main`.

---

> [!NOTE]
> **Low Risk**
> Low risk: scoped to React context value memoization and adds a regression test; behavior should be unchanged except fewer unnecessary re-renders and attach/detach churn.
> 
> **Overview**
> Fixes `createPlayer().Provider` to **memoize the `PlayerContext` value** via `useMemo`, so consumers don’t re-render (and trigger media detach/reattach) on unrelated parent renders (regression #1296).
> 
> Adds a test asserting the context value identity remains stable across repeated parent re-renders.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e2915791e2c4684ca0c5299de58bb903be8f678. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Issue #1296, opened 2026-04-09
Bug: Provider context value not memoized causes ref churn in HlsVideo / DashVideo / etc., rewinding video at end of stream


akantic:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which package(s) or projects are affected?

React (@videojs/react)

### Browser/OS/Node environment

@videojs/react@10.0.0-beta.15
@videojs/core@10.0.0-beta.15
hls.js@1.6.15
react@19, Next.js 16, Chrome 142

## Summary

`createPlayer()`'s `Provider` builds a fresh context value object on every render, and `attachMediaElement()` returns a fresh callback ref on every call. Together these cause `<HlsVideo>`'s `<video>` element ref to be detached and re-attached on every parent re-render, which calls `hls.detachMedia()` / `hls.attachMedia()` and resets the MSE source. **At end of stream this rewinds the video to `t=0` and plays it again.**

### 1. Provider context value is rebuilt on every render

[`packages/react/src/player/create-player.tsx`](https://github.com/videojs/v10/blob/main/packages/react/src/player/create-player.tsx)

```tsx
return (
  <PlayerContextProvider value={{ store, media, setMedia, container, setContainer }}>
    {children}
  </PlayerContextProvider>
);
```

Every render of `Provider` produces a new `value` object identity, so every consumer of `useContext(PlayerContext)` re-renders — including `useMediaAttach()`, `useMedia()`, `useContainer()`, `useContainerAttach()`. This was made strictly worse in [#975](https://github.com/videojs/v10/pull/975) when `setMedia` and `setContainer` were added to the context.

### 2. `attachMediaElement()` returns a new callback ref every call

[`packages/react/src/utils/attach-media-element.ts`](https://github.com/videojs/v10/blob/main/packages/react/src/utils/attach-media-element.ts)

```ts
export function attachMediaElement(media) {
  return (element) => {
    if (element) media.attach(element);
    else media.detach();
    return () => media.detach();
  };
}
```

[`HlsVideo`](https://github.com/videojs/v10/blob/main/packages/react/src/media/hls-video/index.tsx) (and `DashVideo`, `MuxVideo`, `MuxAudio`, `SimpleHlsVideo`, `NativeHlsVideo`) call this directly inside the render body:

```tsx
const composedRef = useComposedRefs(attachMediaElement(mediaApi), ref);
```

[...]

PR #1787, opened 2026-07-01
fix(react): memoize Provider context value

mihar-22:
Replaces #1541 because the source fork has maintainer edits disabled, so the review cleanup could not be pushed there.

Related to #1296.

## What changed

- Memoizes the `createPlayer().Provider` context value so parent re-renders do not rebuild the value object unnecessarily.
- Adds a regression test that verifies the context value identity stays stable across parent re-renders.
- Applies the requested comment cleanup from #1541.

## Verification

- `pnpm --filter @videojs/react... build`
- `pnpm -F @videojs/react test src/player/tests/create-player.test.tsx`
- `pnpm lint:fix:file packages/react/src/player/create-player.tsx packages/react/src/player/tests/create-player.test.tsx`

---

> [!NOTE]
> **Low Risk**
> Small React Provider change with a focused regression test; no auth, data, or API surface changes.
> 
> **Overview**
> Fixes unnecessary context updates when the parent re-renders but player state is unchanged (**#1296**).
> 
> The `createPlayer().Provider` now builds its context value with **`useMemo`**, keyed on `store`, `media`, `container`, and `popupGroup`, instead of allocating a new object every render. Consumers that depend on context identity or shallow stability should no longer re-render solely because an ancestor updated.
> 
> A regression test forces several parent re-renders and asserts the context value reference stays the same.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf40bb2b51ddd4bc77f2a2d1c9796c1f7ecf83ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/react/src/player/create-player.tsx
+++ b/packages/react/src/player/create-player.tsx
@@ -91,15 +91,6 @@ export function createPlayer(config: CreatePlayerConfig<AnyPlayerFeature[]>): Cr
       return store.attach({ media, container });
     }, [media, container, store]);
 
-    // Memoize the context value so its identity only changes when one of its
-    // members actually changes. Without this, every render of a component
-    // above the Provider rebuilds the value object, which forces every
-    // `useContext(PlayerContext)` consumer to re-render even when nothing
-    // it depends on changed. `setMedia` and `setContainer` are `useState`
-    // setters and are identity-stable; `store` and `popupGroup` are also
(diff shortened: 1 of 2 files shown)


decepulis (approved):
looks lgtm to me

PR #1823, opened 2026-07-13
perf(core): batch menu viewport measurements

### LINKED ISSUE #1821: Settings menu open forces a synchronous layout ~12x per mounted panel (layout thrashing in menu measurement)
### Version

`@videojs/html` / `@videojs/core` `10.0.0-beta.25` (light-DOM usage: default video-skin markup ejected into the page, `skin.css` imported).

### Summary

Opening a `media-menu` settings menu (`.media-menu--settings` + `.media-menu__panel` pattern) force-measures **every mounted panel**, interleaving style writes with `getBoundingClientRect()` reads, so each read triggers a full synchronous style-recalc + layout. The cost scales linearly with panel count and with the host page's stylesheet size. On our production app (~8.5k CSS rules) a menu with 13 panels blocks the main thread for **~400 ms on every open**, and each submenu slide adds **~300 ms** more — visible lag and a janked open animation.

In a shadow-DOM skin this is partly masked (style recalc doesn't match the host document's rules), but the number of forced reflows is the same.

### Profiling data (Chrome stable, macOS, no CPU throttling)

Measured by wrapping `Element.prototype.getBoundingClientRect` and via DevTools performance traces:

| Scenario | forced `getBoundingClientRect` calls | forced-layout time |
|---|---|---|
| Menu open, 4 panels | 76 | ~160 ms |
| Menu open, 13 panels | 184 | ~400 ms |
| One submenu slide transition | ~150 | ~300 ms (in a 600 ms window) |
| Menu idle (open or closed) | ~0 | 0 |

- ~**12 forced reflows per mounted panel** on open; scaling is linear in panel count.
- A DevTools trace over ~7 s of menu interaction attributes **976 ms of forced reflow** to the internal flush helper (minified): `function In(i){i?.getBoundingClientRect()}`.
- Sampled stacks during open: `In ← Ra ← M1 ← <panel>.update` (panel size measurement) dominates; the from/to size transition helpers account for the rest.

### Cause

The panel measurement routine does, per panel: write `height:auto` / `min-width` / `max-width:none` → flush → `getBoundingClientRect` → conditionally write width → flush → read again → restore styles → flush. Because writes and reads interleave, no read is ever served from a clean layout — classic layout thrashing. During slide transitions the from/to size swap (custom-property writes + attribute toggles with flushes between) repeats the pattern on every animation frame.

Each forced flush costs a recalc of the dirtied scope against the *host document's* stylesheets — ~2 ms per flush on a real app page vs ~0.2 ms on a bare demo page, which is why the problem hides in minimal reproductions.

### Suggested directions

1. **Batch measurement**: apply the measurement styles to *all* panels first, flush once, read all rects, restore once — turns ~12×N reflows into ~2–3 total.
2. **Measure panels lazily** on first navigation to them (and cache until the panel's content mutates) instead of measuring the whole registry on open.
3. During transitions, reuse the already-measured from/to sizes instead of re-flushing per frame.

### Reproduction

Any page with a non-trivial stylesheet + a settings menu with several nested panels; wrap `getBoundingClientRect` as above and click the gear. Happy to provide the instrumentation snippet or a trace export if useful.

sampotts:
Closes #1821

## Summary

Opening an ejected light-DOM settings menu previously repeated root sizing for every initially hidden submenu and interleaved per-view style writes with layout reads. That work grew with the number of mounted panels and became expensive against a stylesheet-heavy host page.

This change skips the initial-hidden sync, batches the root and entering or exiting view measurements, and restores temporary styles without a final forced layout. The CSS transition barriers remain so submenu animation behavior is unchanged.

The meaningful changes are concentrated in three areas:

[`measureViews()` (line 167)](https://github.com/videojs/v10/blob/fix/forced-layout/packages/core/src/dom/ui/menu/menu-viewport-transition.ts#L167) — the main performance change. It applies measurement styles to all involved views before reading, does at most one constrained-width pass, and restores styles without a final layout flush.

[Entering/exiting transition setup (line 257)](https://github.com/videojs/v10/blob/fix/forced-layout/packages/core/src/dom/ui/menu/menu-viewport-transition.ts#L257) — entering now measures root + submenu together; exiting does the same at [line 311 (line 311)](https://github.com/videojs/v10/blob/fix/forced-layout/packages/core/src/dom/ui/menu/menu-viewport-transition.ts#L311). The forceLayout() calls here remain intentional CSS transition barriers.

[Hidden lifecycle guard (line 415)](https://github.com/videojs/v10/blob/fix/forced-layout/packages/core/src/dom/ui/menu/menu-viewport-transition.ts#L415) — lines 426–430 skip root measurement for initially hidden panels, while retaining the resync when a previously visible panel becomes hidden.

Most changes above line 167 are private-name cleanup, not behavior.

### Reported beta.25 production baseline

| Settings panels | Geometry reads | Forced-layout time |
| --- | ---: | ---: |
| 4 | 76 | ~160 ms |
| 13 | 184 | ~400 ms |

### Current light-DOM Tailwind v4 fixture

| Settings panels | Geometry reads after change | Forced-layout time after change* |
| --- | ---: | ---: |
| 4 | 8 | 1.8 ms |
| 13 | 8 | 2.5 ms |

*Measured as elapsed time inside the instrumented `getBoundingClientRect()` calls, with Tailwind v4 browser CDN CSS and 24 styled host-page cards. This is a controlled local approximation, not a DevTools Layout-event total from the reporter production stylesheet. It confirms the important property: opening cost is constant as mounted panel count grows.

## Changes

- Avoid root sizing when child panels first enter the hidden lifecycle state.
- Measure transition view pairs in shared natural-size and constrained-size passes.
- Keep active-to-hidden restoration and CSS transition state barriers intact.

## Verification

| Lifecycle case | Regression assertion |
| --- | --- |
| 13 initially hidden child panels | 0 root-view geometry reads |
| Natural-size measurement | 1 geometry read |
| Constrained-size measurement | 2 geometry reads |
| Entering submenu | Root and submenu measurements share the natural-size pass |
| Exiting submenu | Root and exiting view measurements share the natural-size pass while retaining the two CSS transition barriers |

A production reprofile against the reporter host stylesheet remains the final wall-clock validation.

## Testing

- `pnpm -F @videojs/core test src/dom/ui/menu/tests/menu-viewport-transition.test.ts`
- `pnpm -F @videojs/core build`
- `pnpm exec biome check packages/core/src/dom/ui/menu/menu-viewport-transition.ts packages/core/src/dom/ui/menu/tests/menu-viewport-transition.test.ts`


stow1x (approved):

decepulis (approved):
Seems sensible!

mihar-22 (approved):

PR #1824, opened 2026-07-13
fix(sandbox): Fix broken scripts on sandbox build

spuppo-mux:
When running `pnpm run build` the following was failing

1) On Sandbox: `Unknown file extension ".ts"`
    - `vite.config.ts` was doing a runtime `await import(pathToFileURL('scripts/shared.ts'))`, which bypassed esbuild and was used Node's native loader. Switched to a static `import { mirrorTemplatesToSrc } from './scripts/shared'` (bundled/transpiled with the config).

2) On Site: `DurationFormat is not a constructor`
`Intl.DurationFormat` exists in Node 24 but not Node 22:
    - Added a fallback to format.ts: `getDurationFormatter` now falls back when `Intl.DurationFormat` is absent. Gated the two Intl-delegation tests with `it.runIf` and added a fallback-path test.
    - Per `site/src/content/docs/concepts/browser-support.mdx:25`:
   > When we use a platform feature that hasn't landed in all supported browsers yet, we provide a fallback. The player works either way; the experience may differ slightly in older browsers. For example, tooltips and popovers use CSS anchor positioning where available and fall back to absolute positioning where it isn't.

This PR also includes a node version upgrade in `.node-version` and `.npmrc`, as well as switching the node-version to use node-version-file in two github workflows (as done in the rest of the workflows)

## Open question (resolved):
Node version mismatch:
    - 22.19.0: `.node-version` (read by fnm/asdf/nodenv/volta), `.npmrc`, `.node-version`, and `package.json` `engines >=22.19.0`.
    - 24.14.0: `.nvmrc` (nvm only), plus most CI jobs via `node-version-file`.

**Resolved**: We resolved to upgrade node version to 24.14.0.

 

---

> [!NOTE]
> **Medium Risk**
> Time-display fallback changes user-visible strings on older runtimes; Node 24 bump affects all contributors and CI. Sandbox build path change is localized but on the critical build pipeline.
> 
> **Overview**
> Fixes **`pnpm run build`** failures from sandbox Vite config loading and site builds on Node without **`Intl.DurationFormat`**, and aligns local/CI Node to **24.14.0**.
> 
> **Sandbox:** `vite.config.ts` no longer runtime-imports `scripts/shared.ts` via `pathToFileURL` (which triggered Node’s native `.ts` loader). It now **statically imports** `mirrorTemplatesToSrc` from `./scripts/shared` so the config is bundled/transpiled with the rest of the Vite setup.
> 
> **Time formatting:** `getDurationFormatter` in `packages/utils/src/time/format.ts` treats **`Intl.DurationFormat` as optional**. When it’s missing (e.g. Node &lt; 23 during SSR/prerender), it uses **`createFallbackFormatter`**: exact digital `H:MM:SS` output and English comma-separated phrases for localized styles. Tests gate Intl-specific cases with **`it.runIf` / `it.skipIf`** and add **`loadWithFallback()`** coverage for the fallback path.
> 
> **Node version:** **`.node-version`** and **`.npmrc`** move from **22.19.0** to **24.14.0** (matching **`.nvmrc`**). **`bundle-size.yml`** and **`website-tests.yml`** switch **`setup-node`** from hardcoded `node-version: 22` to **`node-version-file: '.nvmrc'`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97ee7df68cb3fdd70698cb835828c0965cb1e20a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/utils/src/time/tests/format.test.ts
```diff
@@ -2,6 +2,8 @@ import { describe, expect, it } from 'vitest';
 
 import { formatTime, formatTimeAsPhrase, secondsToIsoDuration } from '../format';
 
+const hasDurationFormat = typeof (Intl as { DurationFormat?: unknown }).DurationFormat === 'function';
+
 describe('formatTime', () => {
```
sampotts, line 27: Is it worth adding a test for the fallback?

sampotts:
As per Slack, let's normalize to `24.18.0` (latest 24 which is LTS) and perhaps add a test for the fallback.

packages/utils/src/time/tests/format.test.ts
spuppo-mux: Added tests forcing the fallback!

.github/workflows/bundle-size.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v5
         with:
-          node-version: 22
+          node-version-file: '.nvmrc'
```
spuppo-mux, line 23: @mihar-22 double checking with you if these were hardcoded to 22 on purpose or if these changes are ok

spuppo-mux:
note: only places I didn't update node was `package.json` > `engines.node` assuming we still want to support lower node versions and the corresponding mention in CONTRIBUTING.md. 

question: Related to the previous note, should I also update @types/node version in devDependencies to ^22 | ^24?

sampotts (approved):

PR #1826, opened 2026-07-13
feat(spf): expose media tracks on the SPF media adapter

### LINKED ISSUE #1795: Feature: Expose Media API Extensions on the SPF Adapter
The SPF adapter currently only surfaces `src`, `preload`, `engine`, `attach`, `detach`, `destroy`, and `play` so an SPF-backed element isn't API-compatible with HlsJsMedia, and consumers must reach through the engine escape hatch to build a quality menu, audio track selector, or live UI.

The goal of this ticket is to also expose other Media API extensions:

- `videoTracks` 
- `audioTracks`
- `videoRenditions`
- `audioRenditions`
- `streamType` (depends on https://github.com/videojs/v10/pull/1715)
- `liveEdgeStart` (depends on https://github.com/videojs/v10/pull/1715)
- `targetLiveWindow` (depends on https://github.com/videojs/v10/pull/1715)

Each extension needs both sides wired at the adapter:

- **Read** — project engine state onto the media-element property (e.g. the `VideoRenditionList`, `audioTracks`, `streamType`).
- **Write** — feed user intent back into the engine (e.g. setting the selected rendition/audio track drives `userAudioTrackSelection` / manual level switching).
- **Signal -> event** — bridge engine signal changes to the DOM events the contract dispatches (`change`, `addrendition`/`removerendition`, `streamtypechange`, etc.) so reactive consumers update without polling.

## Tasks

**Media tracks & renditions** (read / write / events):

- [ ] `videoTracks` — project content tracks onto a `VideoTrackList`; `selected` write drives track switching; emit `change` / `addtrack` / `removetrack`.
- [ ] `audioTracks` — read/write/events over `userAudioTrackSelection` /  `selectedAudioTrackId` (mirrors #1261).
- [ ] `videoRenditions` — project renditions onto a `VideoRenditionList`, map `selectedIndex` to manual switching and `-1` to auto, emit `addrendition` / `removerendition` / `change` (mirrors #1563).
- [ ] `audioRenditions` — read/write/events for the selected audio track's rendition list.

**Live & stream type** (depends on hls-live):

- [ ] `streamType` — surface detected state + `streamtypechange` (mirrors #1389).
- [ ] `liveEdgeStart` — read from the engine's live window (mirrors #1391).
- [ ] `targetLiveWindow` — read from the engine's live window (mirrors #1391).

**Parity & follow-ups:**

- [ ] Confirm the projected surface matches `core/media/types.ts` so the existing `streamTypeFeature`, `live` feature, and menu components bind with no adapter-specific code.
- [ ] **Lower priority:** fold in the autoplay fix. Autoplay currently only works once `preload` is set on the adapter; land it after the new surfaces to minimize merge conflicts.

## Context
-  `HlsJsMedia`
- Core contract this mirrors: #1563 (video renditions), #1261 (audio tracks),
  #1389 (stream type), #1262 (media components). The SPF adapter is the last
  strategy that hasn't implemented it.
- Consumers that build on this: #1785, #1782, #1779 (rendition caps for SPF) all
  presuppose a rendition surface on the adapter.

spuppo-mux:
Part 1 of https://github.com/videojs/v10/issues/1795 (video+audio → streamType → live → parity/autoplay).

**What**: `SimpleHlsMedia` now projects the SPF engine's video renditions and audio tracks onto the standard videoTracks / videoRenditions / audioTracks DOM surface (read, write, events) so a quality menu binds to it with no adapter-specific code, matching `HlsJsMedia`.

**How:** 
- SPF - New `@videojs/spf/media-tracks` module  Provides mapping from SPF Media Tracks (AudioTrack and VideoTrack lists) into MediaTrack API model. 
    - `dedupedAudioTracks` and `dedupedVideoTracks`- the distinct tracks of a presentation, deduped by properties (video: width+height+bandwidth; audio: language+name).
    - `toUserAudioTrackSelection` and `toUserVideoTrackSelection` - build the user*TrackSelection partials the engine's track-switching matches on.
    - `isSameVideoTrack` and `isSameAudioTrack` - compare two tracks by their dedupe key.
    - `frameRateToNumber`, plus the `VideoDedupeKey` / `AudioDedupeKey` shapes.
Mainly deduped lists of tracks grouped by certain keys. These same keys are used to create userAudioTrackSelection and userVideoTrackSelection using the existing SPF logic. Also provides a utility function to translate from the FrameRate model to a number (as used in API). This module uses SPF model and is purposely DOM-free.
 
- Core - A projection mixin (`SimpleHlsMediaMediaTracksMixin`): 
    - subscribes to composed signals (which map the presentation to it's deduped track lists), 
    - builds one selected main video track + a VideoRendition per item obtained from `dedupedVideoTracks` and one audioTrack per item in `dedupedAudioTracks`, reflecting engine selection.
    - and writes user selection back to engine by setting `userVideoTrackSelection` and `audioTrackSelection` using toUser*TrackSelection functions.

## Design decisions:
- Stuck to the "no SPF -> core dependency" rule.
   - Ideally this wouldn't belong to `@videojs/core`, but to do that we would need an independant `@videojs/media-track` package. Therefore, the mixin is added to core's `dom/media/simple-hls` which depends on `spf/media-tracks` and `spf/hls`. Tried to keep dom related vocabulary to the mixin and SPF model vocabulary in the new module.
- Order = manifest declaration order (what SPF already handled), deterministic but unsorted. Not sorted here; a bandwidth-ascending sort in the selector is a one-liner if we later want hls.js-style ordering.
- Video vs audio asymmetry is intentional: VideoRenditionList has an exclusive selectedIndex; AudioTrackList uses non-exclusive enabled (driven exclusively by the audio-track store feature), and audio's enabled doubles as both user input and engine reflection, hence the extra guard in the audio handler.

## Test plan:
- SPF media-tracks unit tests (transforms, property dedup incl. same-language commentary, selection round-tripped through the engine's real matchesPartialTrack).
- core projection tests (jsdom, signal-backed fake engine): rendition/audio projection, active/enabled reflection, selection -> user*TrackSelection, live-reload set-stability, source-unset teardown, destroy cleanup.

---

> [!NOTE]
> **Medium Risk**
> Touches playback track switching and bidirectional sync between DOM and engine signals; mistakes could mis-report quality/audio or cause selection loops, though behavior is heavily tested.
> 
> **Overview**
> **`SimpleHlsMedia`** now mirrors **`HlsJsMedia`**: the SPF engine’s presentation is projected onto **`videoTracks` / `videoRenditions` / `audioTracks`**, and UI-driven changes write back to **`userVideoTrackSelection`** and **`userAudioTrackSelection`**.
> 
> A new **`@videojs/spf/media-tracks`** export provides DOM-free helpers—**`dedupedVideoTracks`** / **`dedupedAudioTracks`** (multi-CDN collapse by resolution+bandwidth or language+name), **`toUser*TrackSelection`**, property-based **`isSame*`** matching, and **`find*TrackById`** for engine-resolved ids that may not appear in the deduped list.
> 
> **`SimpleHlsMediaMediaTracksMixin`** (stacked with **`MediaTracksMixin`** on **`SimpleHlsMedia`**) uses SPF effects to rebuild lists when track ids change, reflect **`active`** / **`enabled`** from engine selection, pin manual quality (**`selectedIndex`**, Auto clears the pin), handle non-exclusive audio **`enabled`** with a guard against feedback loops, clear pins when the manifest **URL** changes (but keep pins on same-URL live reload), and tear down on **`destroy`**.
> 
> Unit tests cover the SPF transforms and the DOM projection (dedup, failover reflection, source change, destroy).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 681312385543af9873d7e7663cf1d3e7c4500e72. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/media/src/dom/simple-hls/media-tracks.ts
+++ b/packages/media/src/dom/simple-hls/media-tracks.ts
@@ -1,0 +1,178 @@
+import { type Composition, computed, effect, untrack } from '@videojs/spf';
+import type { SimpleHlsEngineContext, SimpleHlsEngineState } from '@videojs/spf/hls';
+import {
+  type AudioTrack,
+  dedupedAudioTracks,
+  dedupedVideoTracks,
+  frameRateToNumber,
+  toUserAudioTrackSelection,
+  toUserVideoTrackSelection,
+  type VideoTrack,
+} from '@videojs/spf/media-tracks';
+import type { Constructor } from '@videojs/utils/types';
+import type {
+  MediaAudioTrackCapability,
+  MediaVideoRenditionCapability,
+  MediaVideoTrackCapability,
+} from '../../core/types';
+
+type SimpleHlsEngineHost = {
+  readonly engine: Composition<SimpleHlsEngineState, SimpleHlsEngineContext>;
+  destroy?(): void;
+};
+
+type MediaTracksHost = SimpleHlsEngineHost &
+  MediaVideoTrackCapability &
+  MediaAudioTrackCapability &
+  MediaVideoRenditionCapability;
+
+/** Two track lists carry the same set when their id sequences match. */
+const sameIds = (a: { id: string }[], b: { id: string }[]): boolean =>
+  a.length === b.length && a.every((item, i) => item.id === b[i]!.id);
+
+/**
+ * Projects the SPF engine's presentation onto the media element's
+ * `videoRenditions` / `audioTracks` lists, and wires user selection back to the
+ * engine's `userVideoTrackSelection` / `userAudioTrackSelection` signals.
+ *
+ * Requires the media-tracks mixin (track-list infrastructure) earlier in the
+ * chain so the host exposes `addVideoTrack`, `videoRenditions`, and friends.
+ */
+export function SimpleHlsMediaMediaTracksMixin<Base extends Constructor<MediaTracksHost>>(BaseClass: Base) {
+  class SimpleHlsMediaMediaTracks extends (BaseClass as Constructor<MediaTracksHost>) {
+    #abort = new AbortController();
+    // Memoized SPF model — the last `computed` result per type — so a DOM
+    // selection maps back to the engine's match criteria by id via the SPF
+    // selection helpers (the DOM lists carry only DOM-shape props).
+    #renditions: VideoTrack[] = [];
+    #audioTracks: AudioTrack[] = [];
+
+    constructor(...args: any[]) {
+      super(...args);
+
+      const { state } = this.engine;
+      const { signal } = this.#abort;
+
+      const renditionsSignal = computed(() => dedupedVideoTracks(state.presentation.get()), { equals: sameIds });
+      const audioTracksSignal = computed(() => dedupedAudioTracks(state.presentation.get()), { equals: sameIds });
+
+      // Rebuild video renditions when the set changes; seed `active` from the
+      // current resolved selection (read untracked — `active` deltas are the
+      // reflection effect's job).
+      const reflectRenditions = () => {
+        const renditions = renditionsSignal.get();
+        this.#renditions = renditions;
+        this.#removeVideoTracks();
+        if (!renditions.length) return;
+
+        const videoTrack = this.addVideoTrack('main');
+        videoTrack.selected = true;
+
+        const selectedId = untrack(() => state.selectedVideoTrackId.get());
+        for (const rendition of renditions) {
+          const domRendition = videoTrack.addRendition(
+            '',
+            rendition.width,
+            rendition.height,
+            rendition.codecs.join(','),
+            rendition.bandwidth,
+            rendition.frameRate ? frameRateToNumber(rendition.frameRate) : undefined
+          );
+          domRendition.id = rendition.id;
+          domRendition.active = rendition.id === selectedId;
+        }
+      };
+
+      const reflectSelectedVideo = () => {
+        const selectedId = state.selectedVideoTrackId.get();
+        for (const rendition of this.videoRenditions) {
+          rendition.active = rendition.id === selectedId;
+        }
+      };
+
+      // Rebuild audio tracks when the set changes; seed `enabled` from the
+      // current resolved selection.
+      const reflectAudioTracks = () => {
+        const tracks = audioTracksSignal.get();
+        this.#audioTracks = tracks;
+        this.#removeAudioTracks();
+        if (!tracks.length) return;
+
+        const selectedId = untrack(() => state.selectedAudioTrackId.get());
+        for (const track of tracks) {
+          const domTrack = this.addAudioTrack(track.default ? 'main' : 'alternative', track.name, track.language ?? '');
+          domTrack.id = track.id;
+          domTrack.enabled = track.id === selectedId;
+        }
+      };
+
+      const reflectSelectedAudio = () => {
+        const selectedId = state.selectedAudioTrackId.get();
+        for (const track of this.audioTracks) {
+          track.enabled = track.id === selectedId;
+        }
+      };
+
+      const effectCleanups = [
+        effect(reflectRenditions),
+        effect(reflectSelectedVideo),
+        effect(reflectAudioTracks),
+        effect(reflectSelectedAudio),
+      ];
+
+      this.videoRenditions.addEventListener('change', this.#selectRendition, { signal });
+      this.audioTracks.addEventListener('change', this.#selectAudio, { signal });
+      signal.addEventListener('abort', () => effectCleanups.forEach((cleanup) => cleanup()), { once: true });
+    }
+
+    destroy(): void {
+      this.#abort.abort();
+      this.#removeVideoTracks();
+      this.#removeAudioTracks();
+      super.destroy?.();
+    }
+
+    // Manual quality pin → the chosen rendition's width/height/bandwidth as the
+    // match criteria (the properties renditions were deduped on).
+    // `selectedIndex === -1` (Auto) clears the pin, resuming ABR.
+    #selectRendition = () => {
+      const { userVideoTrackSelection } = this.engine.state;
+      const index = this.videoRenditions.selectedIndex;
+      const domRendition = index < 0 ? undefined : this.videoRenditions[index];
+      const rendition = this.#renditions.find((candidate) => candidate.id === domRendition?.id);
+      userVideoTrackSelection.set(toUserVideoTrackSelection(rendition));
+    };
+
+    #selectAudio = () => {
+      const { selectedAudioTrackId, userAudioTrackSelection } = this.engine.state;
+      const current = [...this.audioTracks].find((track) => track.id === selectedAudioTrackId.get());
+
+      // `enabled` is not exclusive like video `selected`, so prefer a newly
+      // enabled track over the one that is already playing.
+      const enabled = [...this.audioTracks].filter((track) => track.enabled);
+      const target = enabled.find((track) => track !== current) ?? enabled[0];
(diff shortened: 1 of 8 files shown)


packages/media/src/dom/simple-hls/media-tracks.ts
```diff
@@ -0,0 +1,194 @@
...
+        effect(reflectRenditions),
+        effect(reflectSelectedVideo),
+        effect(reflectAudioTracks),
+        effect(reflectSelectedAudio),
+      ];
+
+      this.videoRenditions.addEventListener('change', this.#selectRendition, { signal });
+      this.audioTracks.addEventListener('change', this.#selectAudio, { signal });
+      signal.addEventListener('abort', () => effectCleanups.forEach((cleanup) => cleanup()), { once: true });
+    }
```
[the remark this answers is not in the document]
spuppo-mux: @cjpillsbury deferring to you since hls js mixin does the same. Not too sure how detach attach works

packages/spf/src/media/media-tracks/media-tracks.ts
```diff
@@ -0,0 +1,139 @@
...
+export function dedupedVideoTracks(presentation: MaybeResolvedPresentation | undefined): VideoTrack[] {
+  if (!presentation) return [];
+
+  return dedupe({
+    tracks: getTracksByType(presentation, 'video') as readonly VideoTrack[],
+    keyFn: videoRenditionKey,
+  });
+}
+
+const audioTrackKey = (track: AudioDedupeKey): string => `${track.language ?? ''}-${track.name}`;
```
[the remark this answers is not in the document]
spuppo-mux: Switching to objects as keys

packages/media/src/dom/simple-hls/tests/media-tracks.test.ts
```diff
@@ -0,0 +1,302 @@
+import { signal } from '@videojs/spf';
+import { describe, expect, it } from 'vitest';
+import { MediaTracksMixin } from '../../../../core/media/media-tracks';
+import { HTMLVideoElementHost } from '../../video-host';
+import { SimpleHlsMediaMediaTracksMixin } from '../media-tracks';
+
+// The projection only touches these five engine-state signals; a fake engine
+// backed by real signals is enough to drive the effects.
+function createEngine() {
```
cjpillsbury, line 9: praise: I like this approach for an integration test! 😎

packages/core/src/dom/media/simple-hls/tests/media-tracks.test.ts
```diff
@@ -0,0 +1,302 @@
...
+  type: 'audio',
+  url: 'https://cdn-a/a.m3u8',
+  bandwidth: 128_000,
+  codecs: ['mp4a.40.2'],
+  name: 'Audio',
+  ...over,
+});
+
+// Drain microtasks (effects) and the queued DOM track/rendition events.
+const flush = () => new Promise((resolve) => setTimeout(resolve, 0));
```
cjpillsbury: suggestion(non-blocking): consider tying this more directly to the `effect()` timing so it's more resilient, something like:

```ts
const flush = async () => {
  const sig = signal(0);
  return new Promise((resolve) => {
    effect(() => {
      resolve(sig.get());
    });
    sig.set(sig.get() + 1);
  });
}
```

packages/core/src/dom/media/simple-hls/tests/media-tracks.test.ts
```diff
@@ -0,0 +1,302 @@
...
+
+describe('SimpleHlsMediaMediaTracksMixin', () => {
+  it('projects video renditions into a selected main track, deduped by w/h/bandwidth', async () => {
+    const engine = createEngine();
```
cjpillsbury: nitpick(blocking): for this pattern, you should use `beforeEach()` and `afterEach()`

packages/media/src/dom/simple-hls/media-tracks.ts
```diff
@@ -0,0 +1,194 @@
...
+    #renditions: VideoTrack[] = [];
+    #audioTracks: AudioTrack[] = [];
+
+    constructor(...args: any[]) {
+      super(...args);
+
+      const { state } = this.engine;
+      const { signal } = this.#abort;
+
+      const renditionsSignal = computed(() => dedupedVideoTracks(state.presentation.get()), { equals: sameIds });
```
cjpillsbury, line 72: praise: Nice 👌

packages/media/src/dom/simple-hls/media-tracks.ts
```diff
@@ -0,0 +1,194 @@
...
+      // Rebuild video renditions when the set changes; seed `active` from the
+      // current resolved selection (read untracked — `active` deltas are the
+      // reflection effect's job).
+      const reflectRenditions = () => {
+        const renditions = renditionsSignal.get();
+        this.#renditions = renditions;
+        this.#removeVideoTracks();
+        if (!renditions.length) return;
+
+        const videoTrack = this.addVideoTrack('main');
```
cjpillsbury, line 94: suggestion(non-blocking): Might be good to add a code comment here explaining what a "video track" is and why you're hard-coding one and only one here, including something like "in the future, support of e.g. sources with multiple camera angles would require multiple, source-derived video tracks". You could also reference audio tracks for comparison.

packages/media/src/dom/simple-hls/media-tracks.ts
```diff
@@ -0,0 +1,194 @@
...
+            '',
+            rendition.width,
+            rendition.height,
+            rendition.codecs.join(','),
+            rendition.bandwidth,
+            rendition.frameRate ? frameRateToNumber(rendition.frameRate) : undefined
+          );
+          domRendition.id = rendition.id;
+          domRendition.active = isSameVideoTrack(toVideoKey(domRendition), resolved);
+        }
```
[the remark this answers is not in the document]
cjpillsbury: @spuppo-mux can you test this by selecting a rendition on one loaded source and then switching sources? I believe this is saying we may end up in a weird state under those conditions. Happy to talk this through if you want.

packages/media/src/dom/simple-hls/media-tracks.ts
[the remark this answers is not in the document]
cjpillsbury: Yeah I think this is a bug also in the hls-js mixin. If I'm following everything right, it's an edge case in our current expected use cases where the underlying `HTMLMediaElement` is "detached" and "re-attached". In those cases, we'll end up in a weird state. I think you can probably hold off on fixing this for now, but would be a good one to talk through/confirm with @luwes

packages/spf/src/media/media-tracks/media-tracks.ts
```diff
@@ -0,0 +1,153 @@
...
+    if (a[attr] !== b[attr]) return false;
+  }
+  return true;
+}
+
+/**
+ * Dedupe tracks by a key function, keeping the first occurrence of each key.
+ * Keys are compared field-by-field ({@link sameKey}).
+ */
+function dedupe<T, K extends object>({
```
cjpillsbury, line 107: thought(non-blocking): This could probably be refactored to more directly use `Object.entries()` in a `reduce` or similiar instead of `seen` + `kept` but this is probably also fine.

cjpillsbury (changes requested):
A few callouts, mostly minor or non-blocking with one testing ask. Otherwise and overall, looks 😎.

cjpillsbury (approved):
Ship it!

PR #1848, opened 2026-07-16
feat(i18n): convert to opaque keys

sampotts:
## Summary

As discussed in our meeting earlier, this moves over from using English as the key to opaque keys for the i18n locale files and usages. If we're happy with the DX of this change and the direction, then we can merge. If not, we can keep the existing English as a key. My preference would be to move to this but I realize there are aspects that weren't fully discussed earlier:

### 1. Drift 

We now have the default text inline _and_ the English locale catalogue and there's potential for them to drift - i.e. `<media-text token="buttons.play">Pause</media-text>` or `t('buttons.play', { default: 'Pause' })` which could cause SSR issues along with confusion. 

We can solve this by adding a build-time consistency check that validates known keys against `en.ts` - so `Pause` !== `Play` would throw. This can come in a later release if we decide to take this route.

### 2. State-derived text

At the end of the meeting we discussed relying on CSS to hide/show multiple `<media-text>` elements based on state like we do for icons, e.g.:

```html
<media-play-button commandfor="play-tooltip" class="media-button media-button--subtle media-button--icon media-button--play">
  ${renderIcon('restart', { class: 'media-icon media-icon--restart' })}
  ${renderIcon('play', { class: 'media-icon media-icon--play' })}
  ${renderIcon('pause', { class: 'media-icon media-icon--pause' })}
</media-play-button>
<media-tooltip id="play-tooltip" side="top" class="media-surface media-tooltip">
  <media-tooltip-label></media-tooltip-label>
  <media-tooltip-shortcut class="media-tooltip__kbd"></media-tooltip-shortcut>
</media-tooltip>
```

becomes:

```html
<media-play-button commandfor="play-tooltip" class="media-button media-button--subtle media-button--icon media-button--play">
  ${renderIcon('restart', { class: 'media-icon media-icon--restart' })}
  ${renderIcon('play', { class: 'media-icon media-icon--play' })}
  ${renderIcon('pause', { class: 'media-icon media-icon--pause' })}
  <media-text token="buttons.play" class="media-text media-text--play">Play</media-text>
  <media-text token="buttons.pause" class="media-text media-text--pause">Pause</media-text>
  <media-text token="buttons.replay" class="media-text media-text--restart">Replay</media-text>
</media-play-button>
<media-tooltip id="play-tooltip" side="top" class="media-surface media-tooltip">
  <media-tooltip-label>
    <media-text token="buttons.play" class="media-text media-text--play">Play</media-text>
    <media-text token="buttons.pause" class="media-text media-text--pause">Pause</media-text>
    <media-text token="buttons.replay" class="media-text media-text--restart">Replay</media-text>
  </media-tooltip-label>
  <media-tooltip-shortcut class="media-tooltip__kbd"></media-tooltip-shortcut>
</media-tooltip>
```

Notice we'd likely need to handle it in `<media-tooltip-label>` as well since that currently calls `getResolvedLabel()` on the trigger. If we're moving all of the state-derived text to the UI layer, we'd have to handle that too. 

## Size changes

The increase is primarily from the locale migration rather than runtime logic:

- Locale files now contain nested semantic translations for all shipped locales.
- Generated tree-shakeable text descriptors were added for built-in UI copy.
- The i18n generator now creates descriptors and validates locale completeness.
- Shared translation types, flattening utilities, and descriptor guards were added.
- UI components and platform adapters now resolve descriptors before rendering.
- Additional tests cover descriptor resolution, locale loading, browser translation, and platform output.

The runtime additions are relatively small; most of the size comes from locale data, generated files, and the associated test coverage.

---

> [!NOTE]
> **Medium Risk**
> Wide i18n API and key-shape change across core and all locales; wrong or missing keys would break UI strings, though build-time locale completeness validation reduces that risk.
> 
> **Overview**
> **Replaces English-as-key locale strings with nested opaque keys** (e.g. `buttons.play`, `errors.network`) across all shipped locale packs and moves typing from `types` to `params` with generated `TranslationParams` for placeholder contracts.
> 
> **Build pipeline:** `generate-i18n-locales` now flattens packs for `loadLocale`, validates every locale against English keys, and emits tree-shakeable `i18n/text/*` descriptors; a new `generate-i18n-types` step runs on prebuild. Package exports add `./i18n/text/*`.
> 
> **Runtime:** Registry and browser translation work on `FlatTranslations`; new `Text` descriptors plus `resolveText` / `translateText` support typed copy. Sandbox CDN checks use `['buttons.play']` instead of `.Play`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 782671717ba9850ed95251f079f8b9936d2fa79d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/core/src/core/i18n/text/airplay.ts
```diff
@@ -0,0 +1,14 @@
+/** Generated by packages/core/scripts/generate-i18n-locales.ts — do not edit. */
+import type { Text } from '../text';
+
+const prefix = 'airplay.';
+
+export const start = {
```
mihar-22: i think let's rename all of the text objects to have "Text" postfix (start -> `startText`)

packages/core/src/core/i18n/locales/en.ts
mihar-22, line 1: this will drift, shouldn't this just import and use the values defined in `i18n/text`?

packages/core/src/core/i18n/flatten.ts
mihar-22, line 1: lets probably move some of these utilities in `i18n/utils` directory, stuff like `interpolate.ts`, resolvers, and w.e else makes sense.

packages/core/src/core/i18n/types.ts
mihar-22, line 1: maybe rename this `params.ts` and move `Translator` related types to `translator.ts`

packages/core/src/core/ui/input-feedback/labels.ts
```diff
@@ -4,16 +4,16 @@ import type { InputIndicatorLabels } from './status';
...
+    volume: translator('volume.label', { default: 'Volume' }),
+    volumeWithValue: (value) => translator('volume.value', { value, default: 'Volume {value}' }),
+    captionsOn: translator('status.captionsOn', { default: 'Captions on' }),
+    captionsOff: translator('status.captionsOff', { default: 'Captions off' }),
+    paused: translator('status.paused', { default: 'Paused' }),
+    playing: translator('status.playing', { default: 'Playing' }),
+    fullscreen: translator('status.fullscreen', { default: 'Fullscreen' }),
+    exitFullscreen: translator('fullscreen.exit', { default: 'Exit fullscreen' }),
+    pictureInPicture: translator('status.pip', { default: 'Picture in picture' }),
+    exitPictureInPicture: translator('status.exitPip', { default: 'Exit picture in picture' }),
```
mihar-22: this should be using text objects I assume and not inlining strings? let's catch all these cases

packages/core/src/core/i18n/text.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function isText(value: unknown): value is Text {
+  return typeof value === 'object' && value !== null && 'key' in value && 'text' in value;
+}
+
+export type TextParams = Record<string, string | number>;
+
+export function textValue(value: Text | string): string {
+  return typeof value === 'string' ? value : value.text;
+}
```
mihar-22: rename suggestion: `resolveText`

packages/core/src/core/i18n/text.ts
```diff
@@ -0,0 +1,14 @@
+export interface Text {
+  readonly key: string;
+  readonly text: string;
+}
+
+export function isText(value: unknown): value is Text {
+  return typeof value === 'object' && value !== null && 'key' in value && 'text' in value;
```
mihar-22: we should have `isObject` type from utils package for `typeof value === 'object' && value !== null`

packages/core/src/core/i18n/resolve-text.ts
mihar-22, line 1: i think rename to `translateText`

packages/core/src/core/ui/error-dialog/error-dialog-i18n.ts
```diff
@@ -1,25 +1,30 @@
...
-  | 'You stopped media playback before it finished.'
-  | 'This media could not be loaded due to a network or server issue.'
-  | 'This media could not be played. It may be corrupted, or your browser may not support its format.'
-  | 'This media could not be loaded. It may be unavailable, or your browser may not support its format.'
-  | 'This media could not be played because it could not be decrypted.'
-  | ''
+  'errors.aborted' | 'errors.network' | 'errors.decode' | 'errors.source' | 'errors.encrypted' | 'common.empty'
 >;
 
 const MEDIA_ERROR_CODE_TO_KEY: Record<number, MediaErrorTranslationKey | undefined> = {
```
mihar-22: rename suggestion: `MEDIA_ERROR_CODE_TO_TRANSLATION_KEY`

packages/core/src/core/ui/error-dialog/error-dialog-i18n.ts
```diff
@@ -1,25 +1,30 @@
...
-  [MediaError.MEDIA_ERR_SRC_NOT_SUPPORTED]:
-    'This media could not be loaded. It may be unavailable, or your browser may not support its format.',
-  [MediaError.MEDIA_ERR_ENCRYPTED]: 'This media could not be played because it could not be decrypted.',
-  [MediaError.MEDIA_ERR_CUSTOM]: '',
+  [MediaError.MEDIA_ERR_ABORTED]: 'errors.aborted',
+  [MediaError.MEDIA_ERR_NETWORK]: 'errors.network',
+  [MediaError.MEDIA_ERR_DECODE]: 'errors.decode',
+  [MediaError.MEDIA_ERR_SRC_NOT_SUPPORTED]: 'errors.source',
+  [MediaError.MEDIA_ERR_ENCRYPTED]: 'errors.encrypted',
+  [MediaError.MEDIA_ERR_CUSTOM]: 'common.empty',
```
mihar-22: are these keys not defined in `core/i18n`? if so, we should use those rather than inline strings

packages/core/src/core/ui/error-dialog/error-dialog-i18n.ts
```diff
@@ -1,25 +1,30 @@
...
+};
+
+const MEDIA_ERROR_TEXT: Record<MediaErrorTranslationKey, Text> = {
+  'errors.aborted': aborted,
+  'errors.network': network,
+  'errors.decode': decode,
+  'errors.source': source,
+  'errors.encrypted': encrypted,
+  'common.empty': empty,
 };
```
mihar-22: does this object make sense to create? doesn't this already exist in `core/i18n/errors`? maybe we can use the module `import *`, not sure

packages/core/src/core/ui/error-dialog/error-dialog-i18n.ts
```diff
@@ -34,12 +39,16 @@ export function getMediaErrorTranslationKey(code: number): MediaErrorTranslation
...
+
+export function getErrorDialogDismissLabel(): Text {
+  return ok;
 }
 
-export function getErrorDialogDismissLabel(): string {
-  return 'OK';
+export function getErrorDialogUnexpectedLabel(): Text {
+  return unexpected;
 }
```
mihar-22: suggestion (not sure): get*Label -> get*Text

packages/core/src/core/ui/input-feedback/labels.ts
mihar-22: ideally keys and text objects come from root definitions in `core/i18n`

packages/html/src/define/video/minimal-skin.tailwind.ts
mihar-22, line 1: shouldn't we replace in all skins inlined text content with the text definitions in core?

e.g.

from:

```html
 <media-text token="menu.settings" id="settings-label" class="media-sr-only">Settings</media-text>
```

to:

```html
 <media-text token="${settingsText.key}" id="settings-label" class="media-sr-only">${settingsText.text}</media-text>
```

then we likely make a helper, something like:

```ts
${renderText(settingsText, { id: '', class: '' })}
```

packages/html/src/ui/menu/menu-back-element.ts
```diff
@@ -15,7 +15,7 @@ export class MenuBackElement extends MediaElement {
     label: { type: String },
   } satisfies PropertyDeclarationMap<'label'>;
 
-  label = 'Back';
+  label = 'menu.back';
```
mihar-22: usually i've seen `string | Text` making it clear there's a distinction in how string content is set, this is the first case where I'm seeing that line blurred and string is meaning key and implicitly acting as `Text`

packages/react/src/presets/audio/minimal-skin.tailwind.tsx
```diff
@@ -134,7 +134,12 @@ function PlaybackRateRadioGroup(): ReactNode {
   const { options, setValue, value } = state;
 
   return (
-    <Menu.RadioGroup className={menu.group} value={value} onValueChange={setValue} aria-label={t('Playback rate')}>
+    <Menu.RadioGroup
+      className={menu.group}
+      value={value}
+      onValueChange={setValue}
+      aria-label={t('menu.playbackRate', { default: 'Playback rate' })}
```
mihar-22: i imagine we could just pass a `Text` object into this `t` translator? otherwise we're inlining and running risks of drift all of the place

mihar-22: Ideally the `Text` objects in `core/i18n` are the source of truth for tokens and text values.

packages/core/src/core/i18n/locales/ar.ts
mihar-22, line 1: i think `generate-i18n-locales` already does, but just confirming it validates all key/values are present in each locale translation map?

packages/react/src/ui/menu/menu-back.tsx
```diff
@@ -16,7 +16,7 @@ export interface MenuBackProps extends UIComponentProps<'button', MenuState> {
 
 /** Button that navigates back to the parent menu view. Place at the top of a submenu Content. */
 export const MenuBack = forwardRef<HTMLButtonElement, MenuBackProps>(function MenuBack(
-  { render, className, style, label = 'Back', onClick, ...elementProps },
+  { render, className, style, label = 'menu.back', onClick, ...elementProps },
```
mihar-22: similar comment as in HTML where the line between `string` and `Text` is becoming too blurry

packages/utils/src/object/flatten.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @param prefix - A prefix to prepend to each flattened key.
+ * @returns A new object containing the flattened values.
+ *
+ * @example
+ * ```ts
+ * flatten({ buttons: { play: 'Play' } });
+ * // { 'buttons.play': 'Play' }
+ * ```
+ */
+export function flatten(object: Record<string, unknown>, prefix = ''): Record<string, unknown> {
```
mihar-22: maybe make prefix a config object incase we need other config options in the future.

packages/core/src/core/i18n/text/airplay.ts
sampotts: Can do. I kinda like the short naming though.

packages/core/src/core/i18n/locales/en.ts
sampotts: The text modules are generated from this file.

packages/core/src/core/i18n/flatten.ts
sampotts: Done.

packages/core/src/core/ui/error-dialog/error-dialog-i18n.ts
sampotts: Can do but we don't lose or gain here anything either way though, since we have type safety for those strings.

packages/core/src/core/i18n/locales/ar.ts
sampotts: Yep, it does.

packages/react/src/presets/audio/minimal-skin.tailwind.tsx
sampotts: We'd have to support both ways - passing a Text object or via key + config object. Reason being is if folks add custom keys they won't exist as text objects - they are driven purely by our translations. 

> Ideally the Text objects in core/i18n are the source of truth for tokens and text values.

Pretty much, en.ts is the source and the text objects are generated from it so in a way, they are.

packages/core/src/core/ui/error-dialog/error-dialog-i18n.ts
sampotts: Makes sense. Any method that returns a text should use the `Text` suffix as opposed to `Label` 👍🏼

packages/utils/src/object/flatten.ts
sampotts: Done!

packages/react/src/ui/menu/menu-back.tsx
sampotts: Yeah, not super clear. I've made the type `string | Text` and if a string is passed  it's used as-is otherwise we attempt a translation.

packages/html/src/ui/menu/menu-back-element.ts
sampotts: Yeah, not super clear. I've made the type `string | Text` and if a string is passed  it's used as-is otherwise we attempt a translation.

packages/core/src/core/ui/input-feedback/labels.ts
sampotts: Done.

packages/core/src/core/i18n/types.ts
sampotts: Done. I put generated types into their own file with a `.generated` suffix to make it a little more obvious.

packages/core/src/core/i18n/text/airplay.ts
sampotts: Done.

packages/core/src/core/i18n/text.ts
sampotts: Done.

packages/core/src/core/i18n/text.ts
sampotts: Done.

packages/core/src/core/i18n/resolve-text.ts
sampotts: Done.

packages/core/src/core/ui/error-dialog/error-dialog-i18n.ts
sampotts: Done.

packages/core/src/core/ui/error-dialog/error-dialog-i18n.ts
sampotts: Done. Also simplified and removed the MEDIA_ERROR_TEXT dictionary.

packages/core/src/core/ui/error-dialog/error-dialog-i18n.ts
sampotts: Removed.

packages/core/src/core/ui/error-dialog/error-dialog-i18n.ts
sampotts: Done.

packages/html/src/define/video/minimal-skin.tailwind.ts
sampotts: Done.

packages/core/src/core/i18n/locales/en.ts
mihar-22: I see, I missed the comment at the top. Sounds good. Maybe we can make that shout more that the file is auto-generated?

sampotts: Sure, do we have a standard header we can add perhaps? It's only a small line currently. I'll make it more obvious.

mihar-22 (approved):
Thank you for resolving all of my comments, this looks great!

PR #1880, opened 2026-07-28
feat(spf): Added autoplay support

spuppo-mux:
## Summary

`autoplay` had no effect on the SPF HLS adapter unless `preload` was also set.

Root cause: the adapter never bridged `autoplay` to the engine's load activation. With `preload` unset or `none`, the engine loads nothing until `loadActivated` flips true — but `loadActivated` only flips on a DOM `play`/`seeking` event, and nothing plays until data has loaded. Setting `preload` sidestepped it by loading independently, which is why autoplay only "worked" alongside a preload value.

Part of #1795 (media API extensions on the SPF adapter).

## Fix

`trackLoadTriggers` now treats `el.autoplay` as an upfront commitment to load, alongside the existing "already playing / mid-seek on entry" checks. When the attached media element has autoplay on entry to monitoring, `loadActivated` is set to true immediately (no play / seeking listeners attached) rather than waiting for a play event that preload="none" would never let fire.

```
if (el.autoplay || !el.paused || el.seeking) {
  setLoadActivated();
  return;
}
```

The browser's autoplay policy (muted-gating, unmuted blocking) is untouched. SPF just gets far enough for the browser to make its own call.

Autoplay is read straight off `context.mediaElement`, with no state slot and no adapter/engine changes. The element is already the source of truth: `<simple-hls-video autoplay>` reflects the attribute onto the inner `<video>` (via CustomMediaElement), and the low-level engine harness sets `video.autoplay` directly before attach.

## Repro
Go to Sandbox on `/?platform=html&styling=css&preset=simple-hls-video&skin=default&source=hls-multi-audio&autoplay=1&muted=1&loop=0&preload=none&locale=en` (simple-hls muted autoplay, no preload) and validate that before the changes this does not autoplay but after this changes it does.

Example: On another branch, this does not autoplay even though autoplay is on https://v10-sandbox-git-fork-spuppo-mux-feat-spf-adapter-med-3b75be-mux.vercel.app/?platform=html&styling=css&preset=simple-hls-video&skin=default&source=hls-multi-audio&autoplay=1&muted=1&loop=0&preload=none&locale=en

Go to the same path on this deployment and it should work.
https://v10-sandbox-kvd3wy019-mux.vercel.app/?platform=html&styling=css&preset=simple-hls-video&skin=default&source=hls-multi-audio&autoplay=1&muted=1&loop=0&preload=none&locale=en

Same for https://v10-sandbox-kvd3wy019-mux.vercel.app/spf-segment-loading/?muted=true&autoplay=true&preload=none

Some other cases:
- autoplay + muted -> autoplay starts normally
- autoplay + not muted -> autoplay doesn't start (unless hot reloading the page and you had already interacted with video); however will fetch 5 segments for both cases
- no autoplay -> wont start, respects preload (**except for the found bug in adapter**, see Notes)

## Testing

- trackLoadTriggers: new case asserting autoplay on entry sets loadActivated=true and attaches no play/seeking listeners; autoplay added to the fake-video helper.
- pnpm typecheck and pnpm -F @videojs/spf test (1044 passing) — green; Biome clean on changed files.

## Notes

- While testing noticed there is a bug with the sync-preload behavior and our Sandbox example. Essentially even if preload is set to none, it will always start as metadata. This is because sync-preload expects to read the preload attribute from the video element which never is never reached by our example. We should change either the behavior or the adapter.
- Autoplay read off the element with no backing signal → reactivity scoped to element identity (toggling autoplay on the same attached element post-monitoring isn't picked up; fine for autoplay semantics).
- No adapter/engine/sync-preload changes — earlier engine-slot and preload-folding iterations were reverted.

---

> [!NOTE]
> **Medium Risk**
> Changes when segment loading starts via `loadActivated`, which gates HLS load behavior, but the scope is narrow and covered by new unit tests.
> 
> **Overview**
> **`trackLoadTriggers` now treats `el.autoplay` like an upfront commitment to load**, not only `play` / `seeking` or an already-playing element. When the attached media element has **autoplay** on entry, **`loadActivated` is set to `true` immediately** and **`play` / `seeking` listeners are not attached**, so **`preload="none"`** no longer blocks loading while waiting for a `play` event that never comes.
> 
> Tests add **`autoplay` to the fake video helper** and assert the new on-entry activation path and that listeners stay detached.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 87df1aaf4645bdeb32cc1fe3ac9137b2b9ef9b38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/spf/src/playback/behaviors/dom/track-load-triggers.ts
+++ b/packages/spf/src/playback/behaviors/dom/track-load-triggers.ts
@@ -66,9 +66,9 @@ function deriveState(
  *
  * Writes `state.loadActivated = true` the first time a `play` or `seeking`
  * event fires on the attached media element for the current source — or
- * immediately on entry if the element is already in such a state
- * (`!el.paused` or `el.seeking`), mirroring autoplay / native-controls /
- * direct-DOM-`play()` scenarios.
+ * immediately on entry if the element is already committed to loading
+ * (`el.autoplay`, `!el.paused`, or `el.seeking`), covering autoplay,
+ * native-controls, and direct-DOM-`play()` scenarios.
  *
  * Sticky-true *within a source identity*: subsequent play/pause/seek
  * cycles don't flip back. Source identity = (mediaElement, presentation
@@ -125,7 +125,7 @@ function trackLoadTriggersSetup({
           const el = context.mediaElement.get()!;
 
           const setLoadActivated = () => state.loadActivated.set(true);
-          if (!el.paused || el.seeking) {
+          if (el.autoplay || !el.paused || el.seeking) {
             setLoadActivated();
             return;
           }
--- a/packages/spf/src/playback/behaviors/dom/track-load-triggers.ts
+++ b/packages/spf/src/playback/behaviors/dom/track-load-triggers.ts
@@ -2,7 +2,7 @@ import { listen } from '@videojs/utils/dom';
 import { defineBehavior } from '../../../core/composition/create-composition';
 import type { Reactor } from '../../../core/reactors/create-machine-reactor';
 import { createMachineReactor } from '../../../core/reactors/create-machine-reactor';
-import { computed, peek, type ReadonlySignal, type Signal } from '../../../core/signals/primitives';
+import { computed, type ReadonlySignal, type Signal } from '../../../core/signals/primitives';
 import type { MaybeResolvedPresentation } from '../../../media/types';
 
 /**
@@ -21,10 +21,6 @@ export interface LoadTriggersState {
   loadActivated?: boolean;
   /** Current presentation — URL is used to detect source changes. */
   presentation?: MaybeResolvedPresentation;
-  /**
-   * Standing autoplay intent. On source change the slot resets to *this* value.
-   */
-  autoplay?: boolean;
 }
 
 /**
@@ -77,8 +73,7 @@ function deriveState(
  * Sticky-true *within a source identity*: subsequent play/pause/seek
  * cycles don't flip back. Source identity = (mediaElement, presentation
  * URL). Either changing — including direct in-place swap with no
- * `undefined` intermediate — resets the slot. The reset target is
- * `state.autoplay`.
+ * `undefined` intermediate — resets the slot to `false`.
  *
  * Multi-writer with `hls/adapter.ts:play()` (which writes `true` directly
  * on programmatic play) is intentional — different domains. The adapter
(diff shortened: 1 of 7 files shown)


packages/spf/src/playback/engines/hls/adapter.ts
```diff
@@ -109,6 +111,25 @@ export function SimpleHlsMediaMixin<Base extends Constructor<any>>(BaseClass: Ba
...
+
+    set autoplay(value: boolean) {
+      // Reflect onto the inner <video> via media-host.
+      super.autoplay = value;
+      this.#signals.state.autoplay.set(value);
+      // Autoplay implies load intent; activate loading.
+      if (value) {
+        this.#signals.state.loadActivated.set(true);
+      }
+    }
```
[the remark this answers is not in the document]
cjpillsbury: Maybe fine for now. We can revisit if needed.

cjpillsbury (approved):
LGTM! Ship it.

packages/spf/src/playback/behaviors/dom/track-load-triggers.ts
```diff
@@ -148,15 +154,15 @@ function trackLoadTriggersSetup({
         effects: () => {
           context.mediaElement.get();
           urlSignal.get();
-          return () => state.loadActivated.set(false);
+          return () => state.loadActivated.set(peek(state.autoplay) ?? false);
```
cjpillsbury: question(blocking): is this needed if we're setting `loadActivated` based on `autoplay` in the adapter?

cjpillsbury (changes requested):

packages/spf/src/playback/behaviors/dom/track-load-triggers.ts
cjpillsbury: oh I see this is specifically handling source changes... hmmmmmm. Ok I think this broadly makes sense, but is happening in the wrong place maybe? take a look at the `monitoring` state.

packages/spf/src/playback/behaviors/sync-preload.ts
```diff
@@ -33,7 +37,10 @@ export interface SyncPreloadConfig {
   defaultPreload?: StandardPreload;
 }
 
-type State = Pick<PresentationState, 'preload' | 'presentation'>;
+type State = Pick<PresentationState, 'preload' | 'presentation'> & {
+  /** Standing autoplay intent - forces the effective preload to `'auto'`. */
+  autoplay?: boolean;
+};
```
spuppo-mux: thought: maybe add autoplay to PresentationState

cjpillsbury (approved):
I'm down. We may want to revisit this, but this also may be good enough for now (and maybe even the "right" solution overall).

Issue #1860, opened 2026-07-23
Architecture: Support Media Composition Classes as Markup (`<mux-data />` and `<google-cast />`)


decepulis:
```html
<video-player>
    <mux-video />
    <mux-data />
    <google-cast />
</video-player>
```

PR #1883, opened 2026-07-29
feat(packages)!: support media components as markup

luwes:
Refs #1860

## Summary

Mux Data and Google Cast were composed into media elements at construction. They are their own components now — `<mux-data>` / `<google-cast>` in HTML, `MuxData` / `GoogleCast` in React — registered against whichever media the surrounding player provides. Media elements compose nothing by default.

## API changes

**New components.** Both render nothing and sit next to the media element inside the player. HTML options are attributes, React options are props.

```html
<video-player>
  <media-container>
    <mux-video src="https://stream.mux.com/abc123.m3u8" playsinline></mux-video>
    <mux-data player-software-name="mux-video"></mux-data>
    <google-cast receiver="YOUR_APP_ID"></google-cast>
  </media-container>
</video-player>
```

```tsx
<Player.Provider>
  <Player.Container>
    <MuxVideo source={{ playbackId: 'abc123' }} playsInline />
    <MuxData playerSoftwareName="mux-video" />
    <GoogleCast receiver="YOUR_APP_ID" />
  </Player.Container>
</Player.Provider>
```

Entry points: `@videojs/html/media/{mux-data,google-cast}` (also `@videojs/html/cdn/media/*`) and `@videojs/react/media/{mux-data,google-cast}`.

**Removed: implicit composition.** `<mux-video>` / `<mux-audio>` no longer monitor playback, and the HLS and DASH elements no longer register a Cast sender.

```diff
- <hlsjs-video src="…"></hlsjs-video>
+ <hlsjs-video src="…"></hlsjs-video>
+ <google-cast></google-cast>
```

**Removed: `config` namespace routing.** Options go on the component directly.

```diff
- video.config = { googleCast: { receiver: 'YOUR_APP_ID' } };
+ document.querySelector('google-cast').receiver = 'YOUR_APP_ID';
```

```diff
- <MuxVideo config={{ muxData: { envKey: 'KEY' } }} />
+ <MuxVideo />
+ <MuxData envKey="KEY" />
```

`MediaComponentConfig` and `MediaComponentConstructor.configKey` are gone; `MediaConfig` is now `Record<string, unknown>`. `config` still carries host/engine settings (`preferPlayback`, `contentType`, `hlsJs`).

**New building blocks** for composing your own media component: `MediaComponentElement` (HTML) and `useMediaComponent` (React), plus `googleCastDefaultProps` / `muxDataDefaultProps` from `@videojs/media`.

<details>

`MediaComponentElement` consumes `mediaContext`, resolves the media host from the context value (a media custom element's `.host`, or the host itself), and registers on change, disconnect, and destroy. `useMediaComponent` does the same through the player context.

The component instance comes from an abstract `createComponent()` behind a lazy getter rather than a field. Upgrading an element that is already in the document runs its constructor while connected, so `addController` requests media context from within that constructor — before subclass field initializers run. A field there resolves to `undefined` and throws in `addMediaComponent`.

Mux Data needs no `envKey` for Mux-hosted playback: the view already reports the Mux playback ID as its `video_id`, which Mux attributes to the owning environment.

</details>

## Testing

- `pnpm -F @videojs/media test` (401), `pnpm -F @videojs/html test` (271), `pnpm -F @videojs/react test` (366), `pnpm -F @videojs/core test` (1325), `pnpm -F site test` (512)
- `pnpm typecheck`, `pnpm build:site`
- Verified in Chromium against the CDN bundles with markup parsed before the definitions load, so the elements upgrade in place while connected: both register, `player-software-name` reaches the component, no errors. happy-dom can't cover this — it reports `isConnected` as `false` during an upgrade.
- Manual: `pnpm dev` → `html-mux-video`, `react-mux-video`, `html-mux-audio`, `react-mux-audio`

---

> [!NOTE]
> **High Risk**
> Breaking change: apps that relied on automatic Mux monitoring or Cast on media elements must add new components and migrate off `config.muxData`/`config.googleCast`.
> 
> **Overview**
> **Mux Data** and **Google Cast** are no longer wired into `mux-video`, `mux-audio`, or HLS/DASH media elements at construction. They are **opt-in** siblings inside the player: `<mux-data>` / `<google-cast>` (HTML) and `MuxData` / `GoogleCast` (React), which register with whatever media the player context provides.
> 
> The **`config` namespace routing is removed** (`config.googleCast`, `config.muxData`, `MediaComponentConfig`, `configKey`). Cast and Mux options are set on the new components directly; `config` remains for host/engine settings only.
> 
> New composition primitives: **`MediaComponentElement`** (HTML) and **`useMediaComponent`** (React), plus CDN entry points for the new elements. **Host `destroy()`** now detaches and clears registrations without calling `destroy()` on components—the markup/React owner owns lifecycle. Sandboxes and docs are updated to show explicit composition and a new Mux Data concept page.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b534fdaa8623afd8c1c83c0de63c502e6bea1541. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/html/src/media/google-cast/google-cast-element.ts
+++ b/packages/html/src/media/google-cast/google-cast-element.ts
@@ -14,12 +14,12 @@ import { MediaComponentElement } from '../media-component-element';
  * ```html
  * <video-player>
  *   <hlsjs-video src="https://example.com/stream.m3u8"></hlsjs-video>
- *   <media-google-cast receiver="YOUR_APP_ID"></media-google-cast>
+ *   <google-cast receiver="YOUR_APP_ID"></google-cast>
  * </video-player>
  * ```
  */
 export class GoogleCastElement extends MediaComponentElement<GoogleCast> {
-  static readonly tagName = 'media-google-cast';
+  static readonly tagName = 'google-cast';
 
   static override properties = {
     src: { type: String },
--- a/packages/html/src/media/google-cast/google-cast-element.ts
+++ b/packages/html/src/media/google-cast/google-cast-element.ts
@@ -26,7 +26,8 @@ export class GoogleCastElement extends MediaComponentElement<GoogleCast> {
     contentType: { type: String, attribute: 'content-type' },
     streamType: { type: String, attribute: 'stream-type' },
     receiver: { type: String },
-  } satisfies PropertyDeclarationMap<'src' | 'contentType' | 'streamType' | 'receiver'>;
+    // `customData` takes an object, so it's a property-only prop.
+  } satisfies PropertyDeclarationMap<Exclude<keyof GoogleCastProps, 'customData'>>;
 
   protected createComponent(): GoogleCast {
     return new GoogleCast();
--- a/packages/html/src/media/google-cast/google-cast-element.ts
+++ b/packages/html/src/media/google-cast/google-cast-element.ts
@@ -1,0 +1,79 @@
+import type { PropertyDeclarationMap } from '@videojs/element';
+import { GoogleCast, type GoogleCastProps } from '@videojs/media/dom/google-cast';
+
+import { MediaComponentElement } from '../media-component-element';
+
+/**
+ * Adds Google Cast support to the surrounding player's media.
+ *
+ * Renders nothing — place it inside the player as a sibling of the media
+ * element and it registers a {@link GoogleCast} media component with the
+ * active media host.
+ *
+ * @example
+ * ```html
+ * <video-player>
+ *   <hlsjs-video src="https://example.com/stream.m3u8"></hlsjs-video>
+ *   <media-google-cast receiver="YOUR_APP_ID"></media-google-cast>
+ * </video-player>
+ * ```
+ */
+export class GoogleCastElement extends MediaComponentElement<GoogleCast> {
+  static readonly tagName = 'media-google-cast';
+
+  static override properties = {
+    src: { type: String },
+    contentType: { type: String, attribute: 'content-type' },
+    streamType: { type: String, attribute: 'stream-type' },
+    receiver: { type: String },
+  } satisfies PropertyDeclarationMap<'src' | 'contentType' | 'streamType' | 'receiver'>;
+
+  protected createComponent(): GoogleCast {
+    return new GoogleCast();
+  }
+
+  /** Source URL loaded on the Cast receiver. Falls back to the media's `src` / `currentSrc`. */
+  get src(): string {
+    return this.component.src ?? '';
+  }
+
+  set src(value: string | null | undefined) {
+    this.component.src = value ?? undefined;
+  }
+
+  /** MIME type of the Cast source. When unset, the receiver infers it from the URL. */
+  get contentType(): string | undefined {
+    return this.component.contentType;
+  }
+
+  set contentType(value: string | null | undefined) {
+    this.component.contentType = value ?? undefined;
+  }
+
+  /** Stream type used on the Cast receiver. Falls back to the media's `streamType`. */
+  get streamType(): GoogleCastProps['streamType'] {
+    return this.component.streamType;
+  }
+
+  set streamType(value: GoogleCastProps['streamType'] | null) {
+    this.component.streamType = value ?? undefined;
+  }
+
+  /** Cast receiver application ID. Defaults to Google's default media receiver. */
+  get receiver(): string | undefined {
+    return this.component.receiver;
+  }
+
+  set receiver(value: string | null | undefined) {
+    this.component.receiver = value ?? undefined;
+  }
+
+  /** Custom data sent to the Cast receiver with the load request. */
+  get customData(): GoogleCastProps['customData'] {
+    return this.component.customData;
+  }
+
+  set customData(value: GoogleCastProps['customData']) {
+    this.component.customData = value;
+  }
+}
--- a/packages/html/src/media/media-component-element.ts
+++ b/packages/html/src/media/media-component-element.ts
@@ -1,0 +1,83 @@
+import { ContextConsumer } from '@videojs/element/context';
+import type { Media } from '@videojs/media/dom';
+import {
+  addMediaComponent,
+  HTMLMediaElementHost,
+  type HTMLMediaTargetLike,
+  type MediaComponent,
+} from '@videojs/media/dom/media-host';
+
+import { mediaContext } from '../player/context';
+import { MediaElement } from '../ui/media-element';
+
+type MediaHost = HTMLMediaElementHost<HTMLMediaTargetLike, any>;
+
+/** Resolve the media host from a context media value (a media custom element or the host itself). */
+function resolveMediaHost(media: Media | null): MediaHost | null {
+  if (media instanceof HTMLMediaElementHost) return media;
+  const host = (media as { host?: unknown } | null)?.host;
+  return host instanceof HTMLMediaElementHost ? host : null;
+}
+
+/**
+ * Abstract base for elements that register a media component (e.g. Mux Data,
+ * Google Cast) with the media provided by the surrounding player.
+ *
+ * Place inside a player, as a sibling of the media element. The component is
+ * registered when a media host becomes available, follows the media when it
+ * changes, is removed when this element disconnects, and is destroyed with
+ * this element.
+ */
+export abstract class MediaComponentElement<Component extends MediaComponent> extends MediaElement {
+  #component: Component | null = null;
+  #host: MediaHost | null = null;
+  #removeComponent: (() => void) | null = null;
+
+  /**
+   * Create the media component this element registers. Called once, lazily.
+   *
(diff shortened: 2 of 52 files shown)


packages/html/src/media/media-component-element.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  #setHost(host: MediaHost | null): void {
+    if (this.#host === host) return;
+
+    this.#removeComponent?.();
+    this.#removeComponent = null;
+    this.#host = host;
+
+    if (host) this.#removeComponent = addMediaComponent(host, this.component);
+  }
```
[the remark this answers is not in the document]
luwes: Fixed in 6486b8af3. `addMediaComponent` now calls `detach()` when the active component is unregistered or replaced. `MuxData.setMedia()` transfers its `loadstart` listener between hosts, and `destroy()` now tears down active monitoring through `detach()`. Added regression coverage for unregister, replacement, media swaps, and destroying an active `video.mux` monitor.

packages/media/src/dom/google-cast/media.ts
```diff
@@ -18,15 +18,15 @@ export interface GoogleCastProps {
...
+  src: undefined,
+  contentType: undefined,
+  streamType: undefined,
+  receiver: undefined,
+  customData: undefined,
+};
 
 export class GoogleCast implements GoogleCastProps, MediaComponent {
-  static readonly configKey = 'googleCast';
-
```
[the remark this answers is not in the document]
luwes: Fixed in 690ead69a. Native `EventTarget` already deduplicates identical listener registrations, but `GoogleCast.setMedia()` now makes the intended lifecycle explicit: the provider and its `connect`/`disconnect` listeners are initialized together exactly once. Added a regression test that re-registers against another host and asserts one registration per event.

packages/react/src/media/mux-data/mux-data.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * <Player.Provider>
+ *   <MuxVideo source={{ playbackId: 'abc123' }} />
+ *   <MuxData playerSoftwareName="mux-video" />
+ * </Player.Provider>
+ * ```
+ */
+export function MuxData(props: MuxDataProps): ReactNode {
+  const component = useMediaComponent(MuxDataComponent);
+
+  useSyncProps(component, props, muxDataDefaultProps);
```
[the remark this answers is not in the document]
luwes: Fixed in 690ead69a. `MuxData` now handles `MuxDataSdk` separately from generic `useSyncProps`: an own prop with the value `undefined` disables monitoring, while removing the prop restores `muxDataDefaultProps.MuxDataSdk`. Added coverage for configured SDK → explicit undefined → omitted/default.

packages/media/src/dom/google-cast/media.ts
```diff
@@ -45,10 +45,13 @@ export class GoogleCast implements GoogleCastProps, MediaComponent {
 
     this.#media = host;
 
-    this.#provider ??= new GoogleCastProvider(this);
+    if (!this.#provider) {
+      this.#provider = new GoogleCastProvider(this);
+      this.#provider.remote.addEventListener('connect', this.#onStateChange);
+      this.#provider.remote.addEventListener('disconnect', this.#onStateChange);
+    }
+
```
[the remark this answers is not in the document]
luwes: Fixed in 3b06dab9b. The disconnected remote-only override is now created only with the provider. Re-registering against another media host leaves the current override unchanged, so a connected session continues routing through the provider. Added a regression test for connect → media change → provider override preserved.

packages/html/src/media/google-cast/google-cast-element.ts
```diff
@@ -0,0 +1,79 @@
...
+ */
+export class GoogleCastElement extends MediaComponentElement<GoogleCast> {
+  static readonly tagName = 'google-cast';
+
+  static override properties = {
+    src: { type: String },
+    contentType: { type: String, attribute: 'content-type' },
+    streamType: { type: String, attribute: 'stream-type' },
+    receiver: { type: String },
+  } satisfies PropertyDeclarationMap<'src' | 'contentType' | 'streamType' | 'receiver'>;
```
mihar-22: can these prop keys come from the existing `GoogleCastProps`?

packages/html/src/media/mux-data/mux-data-element.ts
```diff
@@ -0,0 +1,131 @@
...
+    playerInitTime: { type: Number, attribute: 'player-init-time' },
+  } satisfies PropertyDeclarationMap<
+    | 'envKey'
+    | 'beaconCollectionDomain'
+    | 'debug'
+    | 'disableCookies'
+    | 'playerSoftwareName'
+    | 'playerSoftwareVersion'
+    | 'playerInitTime'
+  >;
```
mihar-22: Curious if these types can come from an existing interface instead of inlined

packages/html/src/media/media-component-element.ts
mihar-22, line 1: nice 👌

packages/react/src/media/mux-data/mux-data.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * <Player.Provider>
+ *   <MuxVideo source={{ playbackId: 'abc123' }} />
+ *   <MuxData playerSoftwareName="mux-video" />
+ * </Player.Provider>
+ * ```
+ */
+export function MuxData(props: MuxDataProps): ReactNode {
+  const component = useMediaComponent(MuxDataComponent);
+  const { MuxDataSdk, ...rest } = props;
+  const sdk = 'MuxDataSdk' in props ? MuxDataSdk : muxDataDefaultProps.MuxDataSdk;
```
mihar-22: (question): not entirely sure what this line is doing

mihar-22 (approved):

packages/html/src/media/google-cast/google-cast-element.ts
luwes: Yes — done in f77f20d02. It's now `PropertyDeclarationMap<Exclude<keyof GoogleCastProps, 'customData'>>`, so the keys are derived from the props interface and adding or renaming one fails typecheck. `customData` is object-valued with no attribute form, so it stays property-only and is excluded explicitly.

packages/html/src/media/mux-data/mux-data-element.ts
luwes: Done in f77f20d02 — same treatment as `GoogleCastElement`: `PropertyDeclarationMap<Exclude<keyof MuxDataProps, 'metadata' | 'MuxDataSdk'>>` replaces the seven-member union. `metadata` and `MuxDataSdk` take objects, so they're property-only and excluded explicitly.

packages/react/src/media/mux-data/mux-data.tsx
luwes: It picks the SDK to sync while distinguishing "prop omitted" from "prop passed as `undefined`". `useSyncProps` skips `undefined` values and resets removed props to their defaults, and the default `MuxDataSdk` is the real Mux SDK — so routing it through `useSyncProps` would make `MuxDataSdk={undefined}` (the documented way to disable monitoring) silently re-enable it. The `in` check means an own prop wins even when its value is `undefined`, and only omitting the prop falls back to `muxDataDefaultProps.MuxDataSdk`.

Added a clarifying comment in 0702c5900. Behavior is covered by `disables monitoring when MuxDataSdk is explicitly undefined` and `resets a removed prop to its default` in `packages/react/src/media/tests/mux-data.test.tsx`.

packages/react/src/utils/use-media-component.ts
```diff
@@ -0,0 +1,33 @@
...
+): Component {
+  const media = useMedia();
+  const [component] = useState(() => new ComponentClass());
+
+  useDestroy(component);
+
+  useEffect(() => {
+    if (!(media instanceof HTMLMediaElementHost)) return;
+    return addMediaComponent(media as HTMLMediaElementHost<HTMLMediaTargetLike, any>, component);
+  }, [media, component]);
```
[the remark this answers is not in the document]
luwes: Valid — fixed in b534fdaa8. Media components are now always externally owned (the only production registrations are `MediaComponentElement` and `useMediaComponent`), so `HTMLMediaElementHost.destroy()` destroying them was an ownership violation. It now detaches and unregisters instead, and destruction stays with the owner: `destroyCallback` in HTML, `useDestroy` in React. A still-mounted `MuxData` / `GoogleCast` survives the media being destroyed and re-registers with the next host; the stale cleanup from the old registration is already a no-op via the identity check in `addMediaComponent`.

Regression coverage in `packages/media/src/dom/tests/media-host.test.ts` (host destroy detaches but does not destroy) and `packages/react/src/media/tests/mux-data.test.tsx` (component survives `media.destroy()` while mounted and follows the next media). Both fail against the previous implementation.

Issue #1838, opened 2026-07-14
SPF-Backed Live HLS Support


decepulis:
Add live HLS playback to the SPF playback engine, covering both `<simple-hls-video>` and the SPF-backed `<mux-video>` element. Scope is parity with Mux Player's live features **minus DVR and LL-HLS**: live stream-type detection, live-edge tracking, a live indicator, and seek-to-live — no DVR seek window (live rewind) and no low-latency HLS.

DVR / live rewind is tracked in #1427; low-latency HLS in #1419.

## Prior art
- [Mux Player live streaming](https://www.mux.com/docs/guides/video/start-live-streaming)

> [!NOTE]
> This is an AI-generated spec. Requirements should be independently confirmed.

cjpillsbury:
Done and merged.

PR #1884, opened 2026-07-29
feat(spf): live hls playback on the presentation-timeline model

cjpillsbury:
## Summary

**Scope: sliding-window ("simple") live HLS, consumed through the live Video.js 10 skin.** DVR / `EXT-X-PLAYLIST-TYPE:EVENT` is deliberately **not** in scope — it partly works and has known rough edges, tracked in Deferred below. LL-HLS is out of scope entirely (an LL-HLS source plays, at regular-live latency).

**Try it — no setup, autoplay + muted:**
https://v10-sandbox-git-feat-spf-hls-live-v2-mux.vercel.app/?platform=html&styling=css&preset=simple-hls-video&skin=default&source=hls-live&autoplay=1&muted=1&loop=0&preload=metadata&locale=en

That's the "HLS - Live Stream Big Buck Bunny" Mux live source playing through `<simple-hls-video>` inside `live-video-player` + `live-video-skin` — the path this PR is aimed at: the SPF engine driving the real live player surface, not the engine in isolation.

Under that, this re-lands live HLS on the `startMediaTime`/`timestampOffset` presentation-timeline model: a frozen per-source anchor derived from `EXT-X-PROGRAM-DATE-TIME`, sliding-window playback with reload-driven segment placement, live seekable range and live-edge start, and the media surface (`streamType` / `targetLiveWindow` / `liveEdgeStart`) the live skins consume.

Supersedes the buffer-sourced anchor stack from the original `feat/spf-hls-live` branch (kept as reference, not merged). Coordinate contract: `internal/design/spf/live-presentation-timeline-model.md`.

## Changes

- **Anchor establishment.** `establishStartMediaTime` owns the whole coordinate triple: the reference track's first parse freezes the wall-clock anchor and stamps it onto every track; every other track's first parse gates on that stamp, so segment placement is settled at parse time — the old model's "buffer pin shifts the window later" problem is gone by construction.
- **PDT-primary placement.** Live windows re-place from PDT against the frozen anchor on every reload — self-correcting across reloads, history-free across full window turnovers; `EXTINF`/media-sequence carry-forward remains only as the PDT-less fallback.
- **Reload loop.** Track resolution re-runs on the target-duration cadence (half on an unchanged window) until a playlist completes, driven by an injected `reschedule` seam rather than a separate reloading behavior.
- **Live window + edge.** `seekable` is the intersection of the selected A/V windows, mirrored into `MediaSource.setLiveSeekableRange`. Playback starts at the live edge minus HOLD-BACK via `state.startPosition` — the same one-shot command AirPlay session-restore uses — so the startup playhead has a single writer and the segment loaders aim at the edge from their *first* fetch instead of loading from the window start and being redirected. A window-exit guard rescues a playhead stranded behind the sliding window while leaving in-window pauses and scrub-backs untouched.
- **Live latency from the server.** `EXT-X-SERVER-CONTROL` `HOLD-BACK` is honored when declared, falling back to the spec default of 3× target duration. Kept in the playlist metadata rather than on `Track`, since it's unclear a wall-clock holdback generalizes across delivery formats. `PART-HOLD-BACK` is deliberately unread — see Deferred.
- **Stream end.** Ongoing live never ends off its rolling edge; a playlist that turns complete (`ENDLIST`) opts in to `endOfStream`. Live duration is written as `Infinity`.
- **Media surface.** The simple-hls adapter exposes `streamType` (detected, user-overridable), `targetLiveWindow`, and a read-time `liveEdgeStart` — the same contract `HlsJsMedia` implements, so the live presets / skins / store features work unchanged over SPF.
- **Records.** `live-stream-support` advances to `partial`/`sketched` with Implementation surface and Verification populated; `ll-hls-support` and `dvr-event-stream-support` no longer claim live is an unimplemented prerequisite.
- **Test tooling.** An ffmpeg-based local live source (`apps/sandbox/scripts/live/`, demuxed CMAF + PDT, no Mux account) and live status/controls in the `spf-segment-loading` harness.

<details>
<summary>Deferred — not in this PR</summary>

- **DVR / `EVENT` rough edges.** Decided that DVR starts at the live edge like sliding-window live; start-in-place stays an unimplemented seam. Separately, the one real `EVENT` source tested has an undiagnosed anchor failure (model lands behind native, negative seekable start, back-seek stalls) — that RCA is the first task for DVR, ahead of any windowing work. See `dvr-event-stream-support.md`.
- **Unguarded `setLiveSeekableRange`.** `liveWindowFromState` doesn't floor the window start at 0 when the *reference* window is negative, and the writer has no try/catch. Unreachable for sliding-window live (the reference track defines presentation-0, so its start grows from 0); reachable for a mis-anchored `EVENT` source. Tests pin the actual contract.
- **Reload jitter / backoff** and the **unchanged-playlist miss-counter** termination fallback. Termination is `ENDLIST`-only.
- **`PART-HOLD-BACK`** — deliberately unread until LL-HLS partial-segment playback exists. It's the holdback for clients playing *partial* segments; using it while fetching whole segments would seat the playhead ahead of the last complete segment. Not stored at all, rather than stored write-only.
- **Format neutrality** — `liveWindowFor` still reads `getMediaPlaylistMetadata` / `targetDuration` below the `resolveLiveLatency` seam that exists to keep behaviors format-agnostic. Needs unwinding before DASH. (The latency policy itself now sits above the seam, where format-specific vocabulary belongs.)
- Fallback policy for live that loses PDT mid-stream; presentation-level anchor hoist.
- Audio Only live support. This will be treated as a followup effort.

</details>

## Testing

**Primary path** — the deployed sandbox preview, no local setup:

https://v10-sandbox-git-feat-spf-hls-live-v2-mux.vercel.app/?platform=html&styling=css&preset=simple-hls-video&skin=default&source=hls-live&autoplay=1&muted=1&loop=0&preload=metadata&locale=en

Expect: starts near the live edge (not the window start), the LIVE button reports at-edge and un-latches on a back-seek, the seekable window advances, no console errors. Swap `source=hls-4` for the VOD control on the same page.

Verified on that deployment: `streamType: 'live'`, `targetLiveWindow: 0`, `duration: Infinity`, playing at the edge (`currentTime` 32.6 against `liveEdgeStart` 32), window `18–38`, LIVE button at-edge.

Also measured locally on the same path: startup at `seekableEnd − 3 × targetDuration`, back-seek held ~11.5s behind across ~12 reloads with no spurious repositioning, and the box-parsed origin segment is the live-edge segment (`segmentStartTime: 14`, not `0` — evidence the loaders aim at the edge from their first fetch). VOD regression clean, with no `startPosition` written. Local equivalent: `pnpm dev:sandbox`, same query string on the port it prints.

**Offline alternative** (no Mux dependency, needs `ffmpeg`): `apps/sandbox/scripts/live/local-live.sh` serves a sliding-window demuxed CMAF stream with PDT on `http://localhost:5399`, then the `spf-segment-loading` page with `?src=http://localhost:5399/master.m3u8&muted=true&autoplay=true&preload=auto` — that harness also shows a live status strip and a seek-to-live-edge control.

**Automated:** `pnpm -F @videojs/spf test` — 1339 tests. Full workspace suite green (16/16), `pnpm typecheck` and `pnpm build` clean.

Worth noting for the holdback change: it has **no observable effect on the sources available for smoke testing** — the Mux live stream advertises only `PART-HOLD-BACK`, so latency stays at 3× target duration (verified: still 6s, unchanged). It's a spec-conformance improvement covered by fixtures, not something the sandbox will show.

<details>
<summary>Pre-existing issues you'll hit while reviewing (not from this PR)</summary>

- **The live button's status dot renders gray, not red.** `#1474` gives `[aria-disabled="true"]` `filter: grayscale(1)`, and the live button is aria-disabled precisely when at the live edge. Skin-layer, both default and minimal, out of scope here.
- **The live skin renders no time UI.** `live-video/skin.ts` has an empty `media-time-controls` div.
- **E2E is red**: 66 failures here vs 71 on `main` (`video-controls`, `visual`, `xss-prevention`, `audio-controls`) — none SPF, load-sensitive, verified against a `main` baseline.
- **`?source=hls-1` renders a dead player** — track-switching clears the video selection. Reproduces on `main`. Use `hls-4` for VOD smoke.

</details>

---

> [!NOTE]
> **High Risk**
> Touches core HLS parsing, presentation anchoring, MSE live seekable range, and startup seek behavior—any regression affects live playback and timing. Scope is bounded to sliding-window live with broad unit coverage; known DVR/EVENT and unguarded negative seekable-range edge cases are documented but not fixed here.
> 
> **Overview**
> Adds **end-to-end sliding-window live HLS** to `createSimpleHlsEngine`: media playlists reload on target-duration pacing, growing playlists get `Infinity` duration, the browser seekable range tracks the live window, and playback starts near the live edge with a guard when the playhead falls behind the sliding window. DVR/`EVENT` and LL-HLS mechanisms stay out of scope.
> 
> **Timeline & parsing.** Live segment placement is **PDT-primary** against a frozen per-source wall-clock anchor (reference track first, others gated); reloads re-place from PDT instead of relying only on media-sequence carry-forward. `parseMediaPlaylist` now surfaces PDT, playlist metadata (`targetDuration`, `HOLD-BACK`, `playlistType`, etc.), and finite vs infinite duration; muxed renditions without an `AUDIO` group keep both video and audio codecs for correct MSE mime types.
> 
> **Playback plumbing.** `resolve-track` can run on a **`RecurringRunner`** via injected `reschedule` (`delayedReschedule` + `mediaPlaylistReloadDelay`). New DOM behaviors **`syncLiveSeekableRange`** and **`seekToLiveEdge`** share `liveWindowFromState` / server `HOLD-BACK` (not `PART-HOLD-BACK`). `endOfStream` only fires when the playlist is complete. Core tasks gain memoized `run()`, `clone()`, and **`when()`** for async signal waits.
> 
> **Product surface & tooling.** The HLS adapter exposes live media capabilities (`streamType`, `targetLiveWindow`, `liveEdgeStart`). Sandbox adds ffmpeg/Mux **live test scripts** and live status/seek controls on `spf-segment-loading`. Design docs record live as **partial** implemented and document the live presentation timeline contract.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ad8f56dcd44c37a39f28b468cc537109cd44ae76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/spf/src/playback/behaviors/resolve-track.ts
```diff
@@ -57,19 +74,29 @@ interface TrackResolutionConfig<K extends SelectedTrackKey> {
...
   config: TrackResolutionConfig<K>;
 }) {
-  // NOTE: This can/maybe will be pulled into a per-use case factory (e.g. something like createTaskRunner() with args TBD),
-  // likely eventually passed down via config or a new "definitions" argument. This will allow us to decide if we want our task runner/scheduler
-  // to e.g. run concurrently (like we currently are), serially with a queue, or abort the previous task and replace it with the newly scheduled one. (CJP).
-  const runner = new ConcurrentRunner();
+  // Recurrence lives in the runner: with a `reschedule` (live) it re-runs the
+  // task until the policy stops; `runOnce` (VOD) runs it exactly once. Single-
+  // slot — a selection change re-schedules (abort-and-replace).
+  const runner = new RecurringRunner<ResolvedTrack>(reschedule ?? runOnce);
```
cjpillsbury, line 99: **NOTE**: eventually we may want to abstract this away so that we can use different runners for different compositions. For now, since RecurringRunner can be used for both live and on demand (in terms of expected functionality), this is acceptable.

packages/spf/src/playback/behaviors/resolve-track.ts
```diff
@@ -104,26 +131,73 @@ function setupTrackResolution<K extends SelectedTrackKey>({
...
+                  // fetch).
+                  const text = await fetchResolvableText(current, { signal });
+
+                  // Hold placement — never the fetch — until this track's
+                  // first-parse placement inputs settle (for live, the
+                  // wall-clock anchor; see `primitives/gate-first-parse`).
+                  // First parses only — a reload's window is already on the
+                  // established timeline. Aborting rejects the wait, so a
+                  // source change can't strand a gated task.
+                  if (gateFirstParse && !isResolvedTrack(current)) {
```
cjpillsbury, line 174: **NOTE**: This is another one that would merit some cleaner abstraction. TL;DR - For live, we end up needing to make sure both audio and video are available (assuming they're used/defined - see optional chaining of e.g. `selected{Type}TrackId?.get()`) so that we can appropriately figure out the time alignment of segment/track timelines using PDT/wallclock time. This isn't needed for on demand or non-sliding-window DVR, so figuring out an architecture that allows this (or any replacement abstraction) to be opt in via composition would be ideal. However, since this is fully backwards compatible, we can keep it as is for now.

packages/spf/src/playback/behaviors/resolve-track.ts
```diff
@@ -104,26 +131,73 @@ function setupTrackResolution<K extends SelectedTrackKey>({
...
+                  // the anchor `startDate` onto this track's shell — must feed
+                  // the parse, not be clobbered (anchoring is establish-once;
+                  // parsing the pre-fetch snapshot would strand the track off
+                  // the anchor for good). Correctness rests on a
+                  // run-to-completion invariant: NOTHING may yield (await)
+                  // between this re-read and the write below, so no writer can
+                  // interleave. `parseMediaPlaylist` is synchronous — keep it
+                  // that way, or move the read into the updater.
+                  const live = peek(state.presentation);
+                  const previous = live ? findTrackToResolve(live, trackId) : undefined;
```
cjpillsbury, line 201: **NOTE**: Another example of live-specific changes that we should try to revisit in a followup architecture revisit/cleanup if possible.

packages/spf/src/playback/behaviors/resolve-track.ts
```diff
@@ -143,14 +217,18 @@ function setupTrackResolution<K extends SelectedTrackKey>({
                     // audio↔video (mixed-container sources exist, e.g. muxed-TS
                     // video + raw-.aac audio), which also keeps per-type
                     // resolutions' writes disjoint (no race).
-                    return NON_FMP4_CONTAINER_MIMES.has(mediaTrack.mimeType)
+                    const relabeled = NON_FMP4_CONTAINER_MIMES.has(mediaTrack.mimeType)
                       ? applyContainerMimeType(patched, mediaTrack.type, mediaTrack.mimeType)
                       : patched;
+                    // Stream nature (live vs on-demand) — stable once parsed.
+                    return { ...relabeled, streamType: deriveStreamType(getMediaPlaylistMetadata(mediaTrack)) };
```
cjpillsbury, line 233: **NOTE**: Another example of live-specific changes that we should try to revisit in a followup architecture revisit/cleanup if possible. This one feels less urgent to _**fully**_ abstract, since `streamType` is a useful concept to model, even though it only became relevant/necessary as a result of introducing a stream type other than "on demand". Previously, we could implicitly assume everything was on demand, and we still effectively do that in a lot of places where the distinction doesn't matter. In most places, "on demand" is the status quo behavior/functionality, and live sometimes needs to be treated as an exception type/behavior/functionality.

packages/spf/src/playback/behaviors/resolve-track.ts
```diff
@@ -143,14 +220,18 @@ function setupTrackResolution<K extends SelectedTrackKey>({
...
```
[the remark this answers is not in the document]
cjpillsbury: This shouldn't be an issue in real world usage (at least for now based on what's in scope), since we shouldn't expect different playlists to have different playlist types per spec.

R-Delfino95:
question(non-blocking): Attaching a screen recording from the deployed preview.
If I pause near the live edge and hit play again a bit later, the LIVE badge keeps flickering between the at-edge and behind-edge states instead of settling.

I'm not sure yet whether this is coming from the engine or the skin, so I'm asking about the engine side: after a pause, is the playhead supposed to end up back at the live edge, or is it expected to resume where it left off and stay
behind? It seems to depend on how long I waited.

https://github.com/user-attachments/assets/29c52acf-dc91-4516-99fd-3e6d245ec1f9

R-Delfino95:
note (non-blocking): Smoke-tested this on both the deployed sandbox preview and locally against the ffmpeg live source.
Everything I tried worked.

**Live**
- Cold load starts near the live edge, not at the beginning of the window.
- Jumped back to the live edge, let the playhead drift behind, and went back to live again. Both directions behaved.
- Left it paused for more than 2 minutes and came back. Tried it with the tab focused and with the tab in the background. Recovered cleanly both times.
- Ended the stream mid-playback with `#EXT-X-ENDLIST`. The player noticed the playlist was complete, stopped reloading it, and closed the stream out instead of sitting there polling for segments that were never coming.

**VOD**
- Ran the on-demand source as a regression check. Nothing looked different from before this branch.

All on Chrome.

R-Delfino95 (approved):

packages/spf/src/playback/behaviors/resolve-track.ts
```diff
@@ -104,26 +131,76 @@ function setupTrackResolution<K extends SelectedTrackKey>({
...
+
+                  // Re-read `previous` after the awaits: a concurrent write
+                  // during them — notably the establishment reactor stamping
+                  // the anchor `startDate` onto this track's shell — must feed
+                  // the parse, not be clobbered (anchoring is establish-once;
+                  // parsing the pre-fetch snapshot would strand the track off
+                  // the anchor for good). Correctness rests on a
+                  // run-to-completion invariant: NOTHING may yield (await)
+                  // between this re-read and the write below, so no writer can
+                  // interleave. `parseMediaPlaylist` is synchronous — keep it
```
R-Delfino95, line 198: question(non-blocking): Is there a test that would fail if someone made it async later? Right now the invariant looks like it's only enforced by the comment

packages/spf/src/playback/behaviors/resolve-track.ts
```diff
@@ -104,26 +131,76 @@ function setupTrackResolution<K extends SelectedTrackKey>({
...
 
             const track = findTrackToResolve(presentation, trackId);
-            if (!track || isResolvedTrack(track)) return;
+            // Skip a complete or missing track; a resolved-but-incomplete (live)
+            // window still reloads — its window may have slid past the playhead.
+            if (!track || (isResolvedTrack(track) && Number.isFinite(track.duration))) return;
 
-            runner.schedule(
+            // Abort (selection/source change) settles quietly; a genuine resolve
+            // failure rejects — swallowed for now (TODO: surface to state).
```
R-Delfino95, line 146: question(non-blocking): What does the viewer see if the media playlist starts failing mid-stream? A frozen picture with no error?

packages/spf/src/playback/engines/hls/adapter.ts
R-Delfino95, line 1: question(non-blocking): Am I right that the audio-only engine (`adapter-audio-only.ts`) doesn't get any of this? Since the repo ships a `live-audio` skin, is audio-only live deliberately out of scope here, or is it a follow-up worth naming in the description?

packages/spf/src/playback/behaviors/resolve-track.ts
cjpillsbury: Yeah our error story in general is going to be a WIP, especially for network/server unreliability. Literally the next (currently stacked/based on this branch) effort is getting some basic error signaling in place (https://github.com/videojs/v10/pull/1936), though, even there, this will not account for all of the relevant network errors just yet.

cjpillsbury: That is out of scope for this effort/PR, though.

packages/spf/src/playback/behaviors/resolve-track.ts
cjpillsbury: no, but I'm not sure there's a clear/clean way to write that test, since the TypeScript types currently require `parseMediaPlaylist` to be synchronous?

packages/spf/src/playback/engines/hls/adapter.ts
cjpillsbury: Yeah this is a really good line of questioning... I did intentionally leave off audio only + live based on priorities, but that isn't clear. I'll at least update the description. Good call!

cjpillsbury:
> I'm not sure yet whether this is coming from the engine or the skin, so I'm asking about the engine side: after a pause, is the playhead supposed to end up back at the live edge, or is it expected to resume where it left off and stay behind? It seems to depend on how long I waited.

@R-Delfino95 Yeah this is intentional and was based off of some conversations with @littlespex on how he built his playback engine's behavior. The short version is: pausing+unpausing will work "normally" as long as the playhead/`currentTime` hasn't fallen outside of the "live window" (determined by the playlist), but will jump to live if/when it does fall out of the "live window". For Mux Video (non-DVR/"EVENT") live streams, that live window isn't especially long, but it does mean if someone briefly pauses and unpauses the media, they can still playback normally. It was an "executive decision" I made, in part to allow roughly the same code/logic to work for, live, "sliding window" DVR, and standard DVR (and even a lot of shared logic with on demand).

cjpillsbury:
> If I pause near the live edge and hit play again a bit later, the LIVE badge keeps flickering between the at-edge and behind-edge states instead of settling.

Ok actually confirmed. This is (sadly) expected behavior. The TL;DR - Since HTTP Adaptive Streaming (both HLS and DASH) aren't "true streams", this means the live edge actually updates at discrete times, meaning there will always be edge cases with our current live edge definition where this would happen. I confirmed that we aren't doing it "in the wrong place", but it will still happen. It's not explicitly called out here, but this doc and this section relate to it (just imagine `seekable.end()` updating by small jumps based on refetches of the live media playlists) - https://github.com/video-dev/media-ui-extensions/blob/main/proposals/0007-live-edge.md#recommended-computation-for-rfc8216bis12-aka-hls

cjpillsbury:
Merging but left your playlist test callout open in case you wanted to discuss further.

cjpillsbury:
resolves #1838

packages/spf/src/playback/behaviors/resolve-track.ts
R-Delfino95: Makes sense. The types do catch the obvious version, and I agree a test here would be awkward.
Fine to leave it as the comment!

PR #1885, opened 2026-07-29
fix(packages): prevent controls click triggering interactions

sampotts:
Add the `data-interactive` escape hatch the controls container, preventing clicks from triggering any click interactions bound to the main container. 

---

> [!NOTE]
> **Low Risk**
> Small, targeted DOM attribute change with tests; affects gesture hit-testing only, not playback or auth.
> 
> **Overview**
> Clicks on the **controls bar** (not just individual buttons or sliders) were still bubbling into the player’s container tap gesture and could toggle play/pause. This PR marks the whole controls root as an interactive surface so those pointer events are ignored by container gestures.
> 
> **HTML** `ControlsElement` sets `data-interactive` in `connectedCallback`. **React** `ControlsRoot` passes the same attribute on its root `div`. Unit tests cover both implementations, core gesture behavior for descendants inside a `data-interactive` ancestor, and **e2e** tests for HTML and React verify pointer events on the controls container do not change playback state.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3be0a229f51c35700e38f096957d1ce8eaca146. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/html/src/ui/controls/controls-element.ts
+++ b/packages/html/src/ui/controls/controls-element.ts
@@ -20,6 +20,8 @@ export class ControlsElement extends MediaElement {
   override connectedCallback(): void {
     super.connectedCallback();
 
+    this.setAttribute('data-interactive', '');
+
     if (__DEV__ && !this.#mediaState.value && this.#mediaState.displayName) {
       logMissingFeature(this.localName, this.#mediaState.displayName);
     }
--- a/packages/react/src/ui/controls/controls-root.tsx
+++ b/packages/react/src/ui/controls/controls-root.tsx
@@ -42,7 +42,7 @@ export const ControlsRoot = forwardRef(function ControlsRoot(
           state,
           stateAttrMap: ControlsDataAttrs,
           ref: [forwardedRef],
-          props: [{ children }, elementProps],
+          props: [{ children }, elementProps, { 'data-interactive': '' }],
         }
       )}
     </ControlsContextProvider>
--- a/apps/e2e/tests/gestures.spec.ts
+++ b/apps/e2e/tests/gestures.spec.ts
@@ -53,6 +53,13 @@ test.describe('Mouse Gestures', () => {
     await expect(player.playButton).not.toHaveAttribute(DATA_ATTRS.paused, { timeout: 5_000 });
   });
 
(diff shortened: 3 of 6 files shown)


packages/html/src/ui/controls/controls-element.ts
```diff
@@ -20,6 +20,8 @@ export class ControlsElement extends MediaElement {
   override connectedCallback(): void {
     super.connectedCallback();
 
+    this.setAttribute('data-interactive', '');
```
cjpillsbury, line 23: thought(non-blocking): this should not block merging/releasing imo, but, as a followup, we may want to consider another attribute or a different name, since this name is kind of "lying." These components aren't interactive. If anything, we're trying to identify them to disable a gesture-based interaction.

packages/react/src/ui/controls/controls-root.tsx
```diff
@@ -42,7 +42,7 @@ export const ControlsRoot = forwardRef(function ControlsRoot(
           state,
           stateAttrMap: ControlsDataAttrs,
           ref: [forwardedRef],
-          props: [{ children }, elementProps],
+          props: [{ children }, elementProps, { 'data-interactive': '' }],
```
cjpillsbury, line 45: suggestion(non-blocking): We may want to switch the order of `elementProps` and `{ 'data-interactive': '' }` here so that folks could override the default (I think?)?

cjpillsbury (approved):
Approved with a couple of suggestions (one non-blocking/as a followup)

packages/html/src/ui/controls/controls-element.ts
sampotts: I guess it's more "contains interactive" since it tells the gesture logic to ignore interactions from this element but yeah, we can rename it.

packages/react/src/ui/controls/controls-root.tsx
sampotts: Would we want them to allow the interactions with the controls container to trigger gestures? I mean I guess we _could_ 🤔

PR #1889, opened 2026-07-30
fix: Track disableRemotePlayback Preference

### LINKED ISSUE #1800: Bug: Track Author Preference for disableRemotePlayback on AirPlay on MSE
Follow up for work done in https://github.com/videojs/v10/issues/1260

The AirPlay bridge currently sets `disableRemotePlayback=true` when setting up AirPlay, the goal of this ticket is to add some media api/renderer-level getter/setter for `disableRemotePlayback` to distinguish between a developer explicitly setting that value vs. the bridge programmatically setting it so we can respect the developer's decision.

spuppo-mux:
Closes #1800.

## What

When a page author sets `disableRemotePlayback`, the hls.js AirPlay bridge now respects it instead of unconditionally clearing it.

## How

`HlsJsMedia` snapshots the element's `disableRemotePlayback` in `attach()`, before any engine attaches and while the value still reflects author intent, and hands it to the inner `HlsJsOnlyMedia` delegate (which hosts the AirPlay bridge) in `load()`. 

The bridge only clears the flag when the author didn't set it:

```ts
if (!this.authorDisableRemotePlayback) {
  target.disableRemotePlayback = false;
}
```

This works uniformly for both platforms without a shared-plumbing change: in HTML the `<hlsjs-video disableRemotePlayback>` attribute is mirrored onto the inner `<video>` at construction; in React `disableRemotePlayback` is a plain prop React applies to the `<video>`. Either way the facade reads it before hls.js overwrites it.

## Testing

- `airplay-bridge.test.ts`: preserves `disableRemotePlayback` when the author set it; still enables AirPlay when it was set programmatically (no author intent).
- Manual: Safari + an AirPlay receiver, picker suppressed with the attribute, offered without it.

---

> [!NOTE]
> **Low Risk**
> Playback-flag plumbing for AirPlay only; no auth, data, or engine-core changes. Risk is limited to Safari remote-playback picker behavior.
> 
> **Overview**
> The hls.js AirPlay bridge no longer always clears `disableRemotePlayback` after attach. Author opt-out (markup, React props, or the media API) is preserved so AirPlay stays suppressed when requested.
> 
> The mixin samples the element flag on each `attach()` (before hls.js overwrites it for Managed Media Source) and records explicit setter calls. It only forces the flag off when neither source opted out. `HlsJsMedia` forwards the property to the MSE delegate so API writes reach the bridge.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3bd0327995333ef76fcc13551a6d12c36b04dc76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/media/src/dom/hls-js/airplay-bridge.ts
```diff
@@ -20,6 +20,9 @@ export function HlsJsMediaAirPlayMixin<Base extends Constructor<HlsEngineHost>>(
     #sourceEl: HTMLSourceElement | null = null;
     #disconnect: AbortController | null = null;
 
+    /** Author's `disableRemotePlayback` intent, set by `HlsJsMedia` before hls.js overwrites it. */
+    authorDisableRemotePlayback = false;
```
luwes: blocking: could `authorDisableRemotePlayback` be not part of the public Media API?

luwes:
wondering if the user intent can be captured in a disableRemotePlayback set override

spuppo-mux:
> wondering if the user intent can be captured in a disableRemotePlayback set override

@luwes, Which class would you say could hold the disableRemotePlayback override?

Initially I had implemented it in a similar way with a `RemotePlaybackPreference` MediaComponent (see first commit https://github.com/videojs/v10/pull/1889/changes/332a64fae8f450eff01f2793ba89f4f4a1f5a3ba), but I couldn't make it work.

I know you mean something simpler than this but sharing in case you can see where I went wrong

luwes:
yes, with the components is too much hassle, with the weak map works yes.
I thought it could be set on the class mixin itself?

spuppo-mux:
@luwes just updated. Now we grab the author intent on attach (but before hlsjs attach) and keep it as a private attribute.

Not sure why I didn't go for this initially. I think I tunnel-visioned on the media component idea and missed this after it didnt work. Feels much simpler now. Thanks!

packages/media/src/dom/hls-js/airplay-bridge.ts
```diff
@@ -31,16 +33,24 @@ export function HlsJsMediaAirPlayMixin<Base extends Constructor<HlsEngineHost>>(
       });
     }
 
+    override attach(target: HTMLVideoElement): void {
+      // Snapshot the author's intent before `super.attach` runs hls.js's
+      // `attachMedia`, which forces `disableRemotePlayback` on for MMS.
+      this.#authorDisableRemotePlayback = target.disableRemotePlayback ?? false;
```
luwes: hmm generally the user will call our Media API first and that will then (when a native video is attached) be forwarded to the native video element. 

I think it will have to go in a `override set disableRemotePlayback(...)` instead of here in attach.

spuppo-mux: It seems like I need it on attach because if set from markup (e.g `<hlsjs-video disableremoteplayback>`) it doesnt go through the setter. Correct me if I'm wrong though, in case I'm missing something from the new arch.
The overrides were necessary however if set before attaching. Just pushed the changes.

luwes: good you caught that, I would like to see this differently if it doesn't cause other issues, it will require changes to CustomMediaElement I think where the html attributes are converted to the media host props after a new instance is created from the media host. it will also require a lookup by using the static properties to go from the lower case attribute to the property naming. could you give this a go?

spuppo-mux:
@luwes, I updated this PR. Would appreciate if you can double check that the custom media elements changes make sense.

I also added `authorDisableRemotePlayback` to both the airplay-bridge mixin and the media host. On the mixin to actually guard against author intent and on the media host mainly to forward it to the mixin. I did have to keep a private value to track the changes before the delegate is created and the forward it on load.

One particular edge case I could not resolve and wanted to verify with you is that if `disableRemotePlayback` is set before the media element is attached, this is not reflected on the video element. As far as I understand this means that GoogleCast component wont see it since it reads directly from target.disableRemotePlayback. However, I am not entirely sure if this has any real effect so I wanted to run it through you first. lmk if that makes sense

luwes (approved):
LGTM 👍 thanks

Issue #1866, opened 2026-07-24
Design: A Common Shape for Configuring Media


decepulis:
Media elements today take a `config` object for everything that isn't the source — engine tuning, Mux Data metadata, Google Cast settings. Meanwhile #1797 has landed on a `src: string` + `source: {}` model, and `source` already carries source-adjacent configuration (`drm: { token }`, `thumbnail`, `storyboard`). So we now have two plausible homes for media configuration and no stated rule for which one a given option belongs in.

Here's what `config` looks like on `<mux-video>` today ([`HlsMediaConfig`](https://github.com/videojs/v10/blob/main/packages/core/src/dom/media/hls-js/index.ts#L37-L41) extending [`MediaConfig`](https://github.com/videojs/v10/blob/main/packages/core/src/dom/media/media-host.ts#L39-L43)):

```js
const el = document.querySelector('mux-video');

el.config = {
  // host-level: which delegate to use, how to interpret the src
  preferPlayback: 'mse',
  contentType: 'application/vnd.apple.mpegurl',
  // raw hls.js passthrough
  hlsJs: {
    maxBufferLength: 20,
    capLevelToPlayerSize: true,
  },
  // per-component namespaces, keyed by each component's static `configKey`
  muxData: {
    envKey: 'env-key',
    metadata: { video_title: 'Test VOD', viewer_user_id: 'user-id-007' },
  },
  googleCast: {
    receiver: 'ABC12345',
  },
};
```

React takes the same object as a prop: `<MuxVideo config={{ /* as above */ }} />`.

Three different things are already living in one bag: host settings at the top level, an untyped-by-us passthrough under `hlsJs`, and per-component namespaces that components opt into by declaring a `configKey`. Nothing in the shape says which tier a new option belongs to, none of it has an attribute surface, and `hlsJs` leaks hls.js naming straight into our public API. `<hlsjs-video>` has the same `config` type but only registers `GoogleCast`, so `muxData` silently does nothing there.

We should have a design conversation to decide: do we keep the `config` shape, move configuration onto a new key in `source`, or do something else? Worth settling what counts as "part of the source" versus "how the engine behaves", whether the answer differs per engine, and how this reads as attributes on the HTML element versus props in React.

Four issues are waiting on this to pick an API surface:

- #1777 — Max Auto Resolution Cap for `<hlsjs-video>` and hlsjs-powered `<mux-video>`
- #1772 — DRM for hls.js
- #1773 — DRM for DASH
[...]

PR #1903, opened 2026-07-31
refactor(packages)!: restructure media source, engine, and Mux image APIs

### LINKED ISSUE #1891: Architecture: Replace the Media config Property With source.engine
Media elements accept a single `config` bag for everything that isn't the source. It mixes unrelated tiers into one untyped namespace, and it has no attribute surface. This issue settles the question raised in #1866 by moving engine configuration onto the already-landed structured `source` object, and retiring `config`.

**Design settled.** The open questions below are answered; see "Settled design".

## Today

The core contract is deliberately untyped — `packages/media/src/core/types.ts`:

```ts
export interface MediaConfigCapability {
  config: Record<string, unknown>;
}
```

All typing comes from DOM-layer declaration merging (`MediaConfig` on `HTMLMediaElementHost`) plus per-host extension — `packages/media/src/dom/hls-js/media.ts`:

```ts
export interface HlsMediaConfig extends MediaConfig {
  preferPlayback?: PlaybackType | undefined;
  contentType?: SourceType | undefined;
  hlsJs?: Partial<HlsJsConfig>;
}
```

So one property carries engine construction options (`hlsJs`, `preferPlayback`, `contentType`) and, at the time of filing, per-component namespaces, with nothing in the shape indicating which tier a new option belongs to.

> [!NOTE]
> Two things changed after this issue was filed. #1860 and #1883 moved `muxData` / `googleCast` out of `config` into `<mux-data>` / `<google-cast>` elements and components with their own props, and deleted the `static configKey` routing. `MediaConfig` is now plain `Record<string, unknown>`. That answers open question 1 by removing the tier rather than relocating it.

## Problems

- **Tiers are conflated.** Engine options, source overrides, and (formerly) component settings share one namespace with no rule for where a new option goes.
- **Replace-not-merge, one-way.** The setter overwrites `#config` wholesale; a dropped key is never reset.
- **Bag replacement can silently recreate the engine.** `set config` triggers `#requestLoad()` when the shallow `#engineConfigKey()` changes, and that key spreads all of `hlsJs` plus `preferPlayback` / `contentType`. Assigning an unrelated key on a media that previously set `preferPlayback` destroys and recreates the delegate. Every test uses the defensive `{ ...media.config, … }` idiom as a result.
- **No attribute surface.** `config` is JS-only because `observedAttributes` derives solely from `static properties`, and the reflected-attribute branch would stringify an object.
- **Coverage is uneven.** `HlsMediaConfig` is inherited by `<dash-video>` and `<simple-hls-video>`, but `dom/dash/` consumes no config at all, and SPF engine config is constructor-only while `CustomMediaElement` constructs hosts with no arguments — so `<simple-hls-video>.config` is inert. `VimeoMedia` declares an unrelated flat `config: VimeoConfig`.
- **`hlsJs` leaks vendor naming** into the public API, as noted in #1866.

## Settled design

`config` is replaced by the structured `source` object. `engine` **is** the playback engine's own configuration object, typed per host and passed through untouched — there is no Video.js-owned wrapper inside it and no vendor-named key.

```ts
el.source = {
  src: 'https://example.com/stream.m3u8',  // identity
  type: 'application/vnd.apple.mpegurl',   // how to interpret it
  preferPlayback: 'native',                // an option Video.js normalizes
  engine: { maxBufferLength: 60 },         // hls.js's own config, passed through
};
```

| Tier | Home | Example |
| --- | --- | --- |
| Source identity | `source` fields (`src`, or a host's own identity fields) | `src`, `MuxSource.playbackId` |
| How to interpret the source | `source.type` | `'application/vnd.apple.mpegurl'` |
| Options Video.js normalizes | `source`, alongside `engine` | `source.preferPlayback` |
| The engine's own configuration | `source.engine` | hls.js `maxBufferLength`, dash.js `streaming.abr` |
| A side-car component's settings | that component's own props | `<mux-data env-key>`, `<google-cast receiver>` |

`engine` is typed per host: `Partial<HlsConfig>` for HLS, `dashjs.MediaPlayerSettingClass` for DASH, Vimeo embed parameters for Vimeo.

The mechanism is proven: `MuxMedia` already exposes an object `source` that derives `src`, compares structurally via `deepEqual`, and dispatches `sourcechange`. `MediaSourceObject<Engine>` promotes that shape into the shared contract, and `MuxSource` extends it.

## Settled questions

1. **Where do `muxData` / `googleCast` live?** Already resolved by #1860 and #1883 — each component owns its own props. Nothing to relocate; the decision record states the boundary.
2. **`preferPlayback` / `contentType` — source or engine?** `contentType` becomes `source.type`; it describes the source. `preferPlayback` sits on `source` as a Video.js-normalized option — it selects the delegate and is not part of any engine's config.
3. **Is `engine` a passthrough, a normalized shape, or both?** Passthrough, exclusively. `engine` is the engine's own config object, so it is exhaustive by construction and a new hls.js option needs no change from us. Options we normalize across engines live on `source` beside it. No `hlsJs` / `dashJs` nesting.
4. **Merge or replace?** Replace, compared structurally with `deepEqual`, so reassigning an equivalent object (an inline React prop) is a no-op. Assigning `src` replaces identity only and preserves every other option.
5. **Attribute story?** No attribute for `source`. `src` stays the attribute surface and mirrors `source.src` both ways; `sourcechange` fires on either direction.
6. **Required on `MediaSourceCapability` / `MediaTargetLike`?** No — `MediaSourceCapability` keeps `src` only, and `MediaConfigCapability` is deleted. `MediaSourceObject<Engine>` is the shared shape, but `source` is declared only by hosts that consume it: `HlsJsMedia` (and `MuxMedia`), `DashMedia`, `VimeoMedia`. Putting it on the base host would give `<native-hls-video>` and `<simple-hls-video>` an `engine` nothing reads — the same inert-property problem, renamed.

## Acceptance criteria

- [ ] `source` is the documented way to configure a media element, with `engine` carrying engine configuration, on both custom elements and React media components.
- [ ] `config` is removed from `MediaConfigCapability`, `MediaConfig`, `HlsMediaConfig`, `VimeoConfig`, and every host/element/React prop that exposes it.
- [ ] The tier boundary is recorded, including that `googleCast` / `muxData` already live on their own components.
- [ ] Assigning unrelated configuration no longer recreates the playback engine; re-creation is keyed only on `engine`, `preferPlayback`, and the resolved content type.
- [ ] `<dash-video>` honors `source.engine` (dash.js settings, applied live). `<native-hls-video>` and `<simple-hls-video>` drop the inert `config` and document that they expose no engine configuration.
- [ ] Tests updated: `dom/tests/media-host.test.ts`, `dom/custom-media-element/tests/custom-media-element.test.ts`, `dom/hls-js/tests/hls-media.test.ts`, `dom/mux/tests/mux-media.test.ts`, `dom/dash/tests/dash-media.test.ts`, `dom/vimeo/tests/media.test.ts`, `html/src/media/tests/mux-video.test.ts`, `react/src/media/tests/mux-video.test.tsx`.
- [ ] Docs: first-class `source` documentation, since it is currently undocumented on the site. `concepts/cast.mdx` needs no change — #1883 already migrated it off `config`.
- [ ] A decision record captures the tier boundary so future options have an unambiguous home.

## Notes

- Resolves the design question in #1866; unblocks #1777, #1772, #1773, and #1783. Those four now have a home for normalized options on `source` instead of landing as raw `hlsJs` passthrough.
- `EMPTY_CONFIG` in `packages/media/src/core/constants.ts` has no consumers and can be removed with this change.
- Wiring `source.engine` through to the SPF engine behind `<simple-hls-video>` is out of scope: SPF reads its config once from a constructor argument the custom-element layer never supplies, so it needs a change in SPF rather than a type-level rename here.
- The changelog entry noting the break from `10.0.0-beta.25` comes from a breaking Conventional Commit; release-please and git-cliff generate the files.

luwes:
Closes #1891
Refs #1866

Retires the untyped `config` bag: everything describing a source now lives on `source`. Mux's derived URLs are read from the media rather than built by hand.

### Engine configuration

```ts
// before
el.config = {
  preferPlayback: 'native',
  contentType: 'application/vnd.apple.mpegurl',
  hlsJs: { maxBufferLength: 60 },
};

// after
el.source = {
  src: 'https://example.com/stream.m3u8',
  type: 'application/vnd.apple.mpegurl',
  preferPlayback: 'native',
  engine: { maxBufferLength: 60 },
};
```

`engine` is the engine's own config object, passed through untouched and typed per host:

```ts
HlsSource['engine'];   // Partial<HlsConfig>
DashSource['engine'];  // dashjs.MediaPlayerSettingClass
VimeoSource['engine']; // Vimeo embed parameters
```

`<dash-video>` gains working engine configuration for the first time, applied live via `updateSettings`. `<native-hls-video>` and `<simple-hls-video>` lose the `config` they inherited but never read.

### Mux images

```ts
// before
el.source = { playbackId, thumbnail: { time: 12 } };
el.thumbnail;    // derived URL, or an override you set
el.storyboard = url;

// after
el.source = { playbackId, poster: { time: 12 } };
el.contentData;  // { poster, storyboard } — read-only, derived from source
```

```html
<!-- `source` has no attribute, so the poster frame gets one of its own -->
<mux-video src="https://stream.mux.com/PLAYBACK_ID.m3u8" poster-time="12"></mux-video>
```

`source.poster` and `source.storyboard` are data — nothing applies them to the media, apart from the thumbnail track `<mux-video>` adds for you. The URLs they describe are read together from `contentData`, keyed by what each one is and typed `Record<string, string>`, so the next derived URL doesn't need another property on the media. A key is absent, rather than empty, when its URL can't be built — no playback ID, or signed playback without a matching image token.

The URL builders are internal; `@videojs/media/dom/mux` exports `MuxMedia`, `MuxData`, and the types reachable from `MuxSource`.

### Vimeo metadata

`contentData` is not Mux-only. Because it's a keyed record, each media can report whatever it has; `VimeoVideo` fills `title`:

```ts
el.contentData; // { title: 'Sample Video' } — once the embed has loaded
```

Nothing here is derivable from `src` the way Mux builds URLs, so the embed reports it: `getVideoTitle()` resolves alongside the rest of the metadata in `loadedmetadata`, and a source change clears it. A top-level `title` property would collide with `HTMLElement.prototype.title` — the same clash that kept Mux's derived poster out of `poster`.

No poster for Vimeo: the player SDK has no thumbnail method, and the only route that does (Vimeo's oEmbed endpoint) costs a request to `vimeo.com` per source. Left out until that trade is worth making.

### Behavior

```ts
el.source = { src: 'a.m3u8', preferPlayback: 'native' };

el.src = 'b.m3u8';
el.source; // { src: 'b.m3u8', preferPlayback: 'native' }  ← options survive a URL change

const source = el.source;
el.source = source;                                      // same object → nothing happens
el.source = { src: 'b.m3u8', preferPlayback: 'native' }; // new but equal → announced, no reload
el.source = { src: 'c.m3u8', preferPlayback: 'native' }; // new url → announced, reloads
```

- Re-assigning the same object does nothing; any new object fires `sourcechange`.
- Reloads only on the resolved `src`, the engine config, or the content type — so a fresh object literal every React render never restarts playback.
- `source` has no attribute. `src` stays the attribute and mirrors `source.src` both ways.

### Breaking changes

- `config` is removed, with `MediaConfigCapability`, `MediaConfig`, `HlsMediaConfig`, `VimeoConfig`, `EMPTY_CONFIG`, and `MediaSourceObject`. Map `config.preferPlayback` → `source.preferPlayback`, `config.contentType` → `source.type`, `config.hlsJs` → `source.engine` (one level shallower — the `hlsJs` / `dashJs` nesting is gone). Vimeo's `config` → `source.engine`. `config.muxData` / `config.googleCast` already moved to `<mux-data>` / `<google-cast>`.
- `thumbnail` and `storyboard` properties and attributes are removed from `<mux-video>`, `<mux-audio>`, `MuxVideo`, and `MuxAudio` → read-only `contentData.poster` / `contentData.storyboard`. To point at a storyboard we cannot derive, author the `<track kind="metadata">` yourself.
- `MuxSource.thumbnail` → `MuxSource.poster`, taking a single params object; the array form is gone. `MuxThumbnailParams`, `MuxThumbnailFitMode`, `MuxThumbnailExt` → `MuxPosterParams`, `MuxPosterFitMode`, `MuxImageExt`.
- No longer exported from `@videojs/media/dom/mux`: `createMuxPosterURL`, `createMuxStoryboardURL`, `createMuxVideoURL`, `parseMuxVideoURL`, `createMuxQuery`, `MUX_VIDEO_DOMAIN`, and the `MuxJWT` type. `isSameMuxSource` is removed.
- `MuxSource.playbackId` is now optional, so `<mux-video>` can play a non-Mux URL through `source.src`.
- `<native-hls-video>`, `<simple-hls-video>`, and `<simple-hls-audio-only>` no longer accept `config`.

### Testing

```
pnpm -F @videojs/media test    # 449 passed
pnpm -F @videojs/html test     # 291 passed
pnpm -F @videojs/react test    # 385 passed
pnpm typecheck
pnpm -F site build
```

New coverage for engine re-creation keyed only on engine-relevant values, options surviving a `src` change, equivalent sources not reloading, dash settings resetting rather than merging, the derived `contentData` URLs, the Vimeo title, and the React inline-literal case. `pnpm typecheck` is clean apart from pre-existing errors in `packages/spf`.

Docs: new `concepts/media-sources` guide — structured `source` was previously undocumented despite `<mux-video>` already shipping it.

---

> [!NOTE]
> **High Risk**
> Large breaking change across HLS, DASH, Mux, and Vimeo playback configuration and load/reload semantics; consumers must migrate from `config` and Mux thumbnail/storyboard overrides.
> 
> **Overview**
> **Breaking API reshape:** the untyped **`config`** bag is removed from media hosts and core types. Playback is configured through a structured **`source`** object (`src` / identity, optional `type`, `preferPlayback`, and **`engine`** passed straight to hls.js, dash.js, or Vimeo). **`src` and `source` stay in sync**; changing only the URL keeps engine options. Assignments fire **`sourcechange`**; structurally equal sources avoid reloads (including inline React literals), while engine / URL / content-type changes still tear down or reattach as before.
> 
> **Mux:** **`thumbnail` → `source.poster`** (single params object); override **`thumbnail` / `storyboard` properties and attributes** are gone. Derived poster and storyboard URLs are **read-only via `contentData`**. `<mux-video>` adds **`poster-time`** → `source.poster.time` and drives the storyboard track from **`contentData.storyboard`**. Several Mux URL helpers are no longer public exports.
> 
> **Dash / Vimeo:** Dash gains first-class **`source.engine`** with reset-then-apply settings; Vimeo moves embed params to **`source.engine`** and exposes loaded **`title`** on **`contentData`**, with load-barrier fixes for source swaps.
> 
> Docs add **`concepts/media-sources`** and React/HTML wrappers are updated for the new types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 795fadfede174081e8b18b5b2f89b6fa81b79551. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/decisions/media/source-configuration-tiers.md
mihar-22, line 1: I think we could consider dropping this document as the decision largely will live in the code. If you'd like to retain some important context then maybe compacting this down would be nice. Trying to keep these decision docs more concise and directly related to discussions we have. If you feel stronlgy about anything feel free to keep it!

packages/media/src/core/types.ts
```diff
@@ -132,6 +132,27 @@ export interface MediaSourceCapability {
...
+ * `MuxSource`'s `playbackId`).
+ */
+export interface MediaSourceObject<Engine = unknown> {
+  /** Source URL. Mirrors the host's `src` property. */
+  src?: string | undefined;
+  /** MIME type of the source. Takes precedence over inference from `src`. */
+  type?: string | undefined;
+  /** The playback engine's own configuration. Changing it recreates the engine unless the engine accepts live updates. */
+  engine?: Engine | undefined;
+}
```
mihar-22: I am wondering whether we need this as a shared interface with `engine` in it specifically. Feels a little off to me but hard to pin down why. Maybe we can push it down to specific Media and leave this as src/type? Hard to see why would need `engine` in the base interface atm.

packages/media/src/dom/dash/media.ts
```diff
@@ -1,24 +1,38 @@
...
 import { MediaTracksMixin } from '../../core/media-tracks';
-import type { MediaEngineHost } from '../../core/types';
+import type { MediaEngineHost, MediaSourceObject } from '../../core/types';
 import { HTMLVideoElementHost } from '../video-host';
 
+/**
+ * Structured DASH source: the MPD URL in `src`, dash.js's own settings in
+ * `engine`. Replacing `engine` resets any previously applied settings.
+ */
+export interface DashSource extends MediaSourceObject<dashjs.MediaPlayerSettingClass> {}
```
mihar-22: I can see the `engine` configuration type living here

packages/media/src/core/media-source.ts
```diff
@@ -0,0 +1,21 @@
...
+ * themselves rather than calling this.
+ */
+export function resolveSourceObject<Source extends MediaSourceObject>(
+  src: string,
+  previous?: Source | null
+): Source | null {
+  const { src: _previousSrc, ...options } = previous ?? {};
+  const source = { ...options, ...(src ? { src } : null) } as Source;
+  return Object.keys(source).length > 0 ? source : null;
+}
```
mihar-22: ⚠️ This is merging two different source objects together? Concerning?

packages/media/src/dom/dash/media.ts
```diff
@@ -56,8 +70,45 @@ export class DashMedia
     return this.#src;
   }
 
-  set src(src) {
-    this.#src = src;
-    this.#engine.attachSource(src);
+  /** MPD URL. Setting it re-derives `source`, carrying over its options. */
+  set src(value) {
+    const source = resolveSourceObject<DashSource>(value, this.#source);
```
mihar-22: Is this the right behaviour when setting `src`? I expect it to be one or the other otherwise strange things might start happening.

packages/media/src/dom/dash/media.ts
```diff
@@ -56,8 +70,45 @@ export class DashMedia
...
+    const changed = !deepEqual(this.#source, source);
+    this.#src = value;
+    this.#source = source;
+    this.#engine.attachSource(value);
+    if (changed) this.dispatchEvent(new Event('sourcechange'));
```
mihar-22: Should we just assume any set here is a "change"? Little worried we're calling the engine `attachSource` which might have its own rules for what "change" means. Wondering if there's any engine events we could use here or make sure we're aligned with what source change means.

packages/media/src/dom/dash/media.ts
```diff
@@ -56,8 +70,45 @@ export class DashMedia
...
+    const engineChanged = !deepEqual(this.#source?.engine ?? null, value?.engine ?? null);
+    this.#source = value ?? null;
+    if (engineChanged) this.#applyEngineConfig(value?.engine);
+    const src = value?.src ?? '';
+    if (this.#src !== src) {
+      this.#src = src;
+      this.#engine.attachSource(src);
+    }
+    this.dispatchEvent(new Event('sourcechange'));
+  }
```
mihar-22, line 120: Equally here I'm wondering whether we just assume any set is config changed, attach the new src, and how we can be best aligned with the engine's definition of source change.

packages/media/src/dom/dash/media.ts
```diff
@@ -56,8 +70,45 @@ export class DashMedia
...
+
+  // `engine` is replaced, not merged, but dash.js merges every
+  // `updateSettings()` call into the current settings — reset first so dropping
+  // a key clears it instead of leaving the previous value behind.
+  #applyEngineConfig(engine?: dashjs.MediaPlayerSettingClass) {
+    this.#engine.resetSettings();
+    if (engine) this.#engine.updateSettings(engine);
   }
```
mihar-22, line 128: really good catch 👌

packages/media/src/dom/dash/media.ts
mihar-22: Might consider `engineChanged` => `configChanged`

packages/media/src/dom/hls-js/media.ts
```diff
@@ -29,40 +29,71 @@ export const StreamTypes = MediaStreamTypes;
...
+ * `preferPlayback` and `engine` are both read when the engine is constructed, so
+ * changing either recreates it.
+ */
+export interface HlsSource extends MediaSourceObject<Partial<HlsJsConfig>> {
+  type?: SourceType | undefined;
+  /**
+   * Preferred playback path: `'mse'` for hls.js, `'native'` for the browser's
+   * own HLS support. Ignored when the preferred path cannot play the source.
+   */
   preferPlayback?: PlaybackType | undefined;
```
mihar-22, line 52: Could this be clearer as maybe just a boolean like `preferNative` or `preferMse`? Just some suggestions, not sure.  Only because `preferPlayback` as a config option sounds odd to me in the context of the `source`

packages/media/src/dom/hls-js/media.ts
```diff
@@ -29,40 +29,71 @@ export const StreamTypes = MediaStreamTypes;
...
+  }
+
+  /**
+   * Resolve the structured source a media URL describes. Subclasses override
+   * this to parse identity out of the URL. `type` and `engine` carry over from
+   * `previous`, so assigning `src` never drops engine configuration.
+   */
+  static resolveSource(src: string, previous: HlsSource | null): HlsSource | null {
+    return resolveSourceObject(src, previous);
+  }
```
mihar-22: How come these are statics here? They weren't exposed on dash so consistency call out too.

packages/media/src/dom/hls-js/media.ts
```diff
@@ -135,15 +152,54 @@ export class HlsJsMedia extends HTMLVideoElementHost implements HlsMediaProps {
...
 
   set src(src: string) {
+    const source = this.#ctor.resolveSource(src, this.#source);
+    const changed = !deepEqual(this.#source, source);
+
     this.#src = src;
+    this.#source = source;
+
+    if (changed) this.dispatchEvent(new Event('sourcechange'));
+    this.#requestLoad();
```
mihar-22: Same comments as I made in dash media.

packages/media/src/dom/hls-js/media.ts
```diff
@@ -135,15 +152,54 @@ export class HlsJsMedia extends HTMLVideoElementHost implements HlsMediaProps {
...
+  set source(value: HlsSource | null) {
+    const source = value ?? null;
+    if (deepEqual(this.#source, source)) return;
+
+    this.#source = source;
+    this.#src = this.#ctor.resolveSrc(source);
+
+    this.dispatchEvent(new Event('sourcechange'));
     this.#requestLoad();
   }
```
mihar-22, line 196: Same comments as I made in dash media.

packages/media/src/dom/mux/media.ts
```diff
@@ -26,10 +25,36 @@ export const muxMediaDefaultProps: MuxMediaProps = {
...
+  static override resolveSource(src: string, previous: MuxSource | null): MuxSource | null {
+    const { type, preferPlayback, engine } = previous ?? {};
+    const source: MuxSource = {
+      ...(type && { type }),
+      ...(preferPlayback && { preferPlayback }),
+      ...(engine && { engine }),
+      ...(parseMuxVideoURL(src) ?? (src ? { src } : null)),
+    };
+    return Object.keys(source).length > 0 ? source : null;
+  }
```
mihar-22: 1. Should we consider dropping these statics for now? 
2. I'm a little concerned as previous comments on the idea of merging sequential sources.

packages/media/src/dom/mux/media.ts
```diff
@@ -41,34 +66,26 @@ export class MuxMedia extends HlsJsMedia implements MuxMediaProps {
...
-    const src = createMuxVideoURL(value) ?? '';
-    if (super.src !== src) super.src = src;
-    this.dispatchEvent(new Event('sourcechange'));
+    super.source = value;
   }
 
   /** Thumbnail image URL. Falls back to one derived from `source`. */
   get thumbnail(): string {
-    return this.#thumbnail || (createMuxThumbnailURL(this.#source) ?? '');
+    return this.#thumbnail || (createMuxThumbnailURL(this.source) ?? '');
```
mihar-22: Not sure here... instead of falling back should setting `source` also just set/reset the `thumbnail` property? Concerned you could set a `thumbnail`, then set a new `source`, and it would fall out of sync.

packages/media/src/dom/mux/media.ts
```diff
@@ -77,7 +94,7 @@ export class MuxMedia extends HlsJsMedia implements MuxMediaProps {
 
   /** Storyboard (thumbnail sprite) VTT URL. Falls back to one derived from `source`. */
   get storyboard(): string {
-    return this.#storyboard || (createMuxStoryboardURL(this.#source) ?? '');
+    return this.#storyboard || (createMuxStoryboardURL(this.source) ?? '');
   }
```
mihar-22: Same concern as `thumbnail`

packages/media/src/core/types.ts
```diff
@@ -132,6 +132,27 @@ export interface MediaSourceCapability {
...
+ * no-op. Hosts may also add identity fields of their own in place of `src` (see
+ * `MuxSource`'s `playbackId`).
+ */
+export interface MediaSourceObject<Engine = unknown> {
+  /** Source URL. Mirrors the host's `src` property. */
+  src?: string | undefined;
+  /** MIME type of the source. Takes precedence over inference from `src`. */
+  type?: string | undefined;
+  /** The playback engine's own configuration. Changing it recreates the engine unless the engine accepts live updates. */
+  engine?: Engine | undefined;
```
mihar-22: One thing I'm wondering is whether we should consider keeping `engine` reserved so in the future it can be our normalized/shared config property (if one ever arises). I think also reading the code that it wasn't clear what "engine" meant in the context of the source. 

Should we for now just lean on the engine names to avoid future potential conflicts/issues? E.g., `engine` => `hlsJs`, `dashJs`, `vimeo`, etc. Not exactly sure but something like that. 

```ts
source = {
  hlsJs: {
    ...
  }
}
```

I originally thought we don't need namespaces but what if for some odd reason you're switching between engines like SPF and hls.js based on some criteria. You would probably want to safely separate configurations.

mihar-22 (approved):
Thanks for doing this! I've left comments on things I think are worth discussing further — at least from what I've gathered reviewing in the diff :)

packages/media/src/dom/mux/media.ts
luwes: this prop was decided to be removed and replaced with a getter `contentPoster`

luwes:
Thanks for the thorough pass — most of it is in. Flagging what changed because the branch has moved a fair bit since you reviewed, and several of your comments now point at code that no longer exists. Worth re-reading the Mux poster area from scratch rather than from the old diff.

## Addressed

| Your comment | What changed |
| --- | --- |
| Drop the decision doc | Deleted (`033bae48b`). The tier boundary is in the `concepts/media-sources` guide instead. |
| `core/types.ts:154` — do we need a shared interface with `engine` in it? | `MediaSourceObject` is gone from core (`87bb2521c`). Each host declares its own source shape, listing only what it reads — which also dropped `type` from the DASH and Vimeo sources, since neither ever read it. |
| `dash/media.ts:12` — engine config type could live here | It does now, same commit. |
| `core/media-source.ts:21` — ⚠️ merging two different source objects? | Fair, and the rest-spread was the reason it read that way. The util is deleted (`5a4d6c40d`); each host derives its own source in `set src` and names the fields it carries. What survives a URL change is now visible at the point it happens. |
| `hls-js/media.ts:95`, `mux/media.ts:49` — why statics? | Dropped (`0d2f49f2a`). `MuxMedia` keeps the source it was assigned and hands it down with `src` resolved, so the base still owns storage, the reload decision, and `sourcechange` — one code path, no extension mechanism, and consistent with DASH and Vimeo. |
| `dash/media.ts:105` — `engineChanged` → `configChanged` | Renamed (`8eb5a57cd`). |
| `mux/media.ts:88`, `:98` — `thumbnail` / `storyboard` falling out of sync with `source` | Both props are removed entirely (`f028fc21a`), so there is nothing left to fall out of sync. `source.poster` and `source.storyboard` are data; read-only `contentPoster` / `contentStoryboard` getters build the URLs. All the Mux URL builders are internal now (`ef5f82aad`) — the entry exports `MuxMedia`, `MuxData`, and the types reachable from `MuxSource`, nothing else. |

packages/html/src/media/mux-video/media.ts
```diff
@@ -19,13 +22,49 @@ export class MuxVideo extends MuxVideoBase {
...
   }
 
   attributeChangedCallback(name: string, oldValue: string | null, newValue: string | null) {
+    if (name === 'poster-time') {
+      this.#applyPosterTime(this.#posterTimeAttr());
+      return;
+    }
     super.attributeChangedCallback(name, oldValue, newValue);
-    if (name === 'storyboard') this.#syncStoryboard();
+  }
```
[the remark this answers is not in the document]
luwes: Fixed in c47a8e90e. Removing the attribute still clears what it set, but a value that is present and not a number is now ignored rather than treated as removal — so it can no longer wipe a `source.poster.time` set through JS. That matches how `#syncPosterTime` already behaved, which is the inconsistency you spotted.

Covered by `keeps an existing poster time when poster-time is not a number`, which fails against the previous version.

packages/media/src/dom/dash/media.ts
```diff
@@ -56,8 +75,51 @@ export class DashMedia
...
+    const srcChanged = this.#src !== src;
+
+    this.#source = source;
     this.#src = src;
-    this.#engine.attachSource(src);
+
+    if (configChanged) this.#applyEngineConfig(source?.engine);
+    if (srcChanged) this.#engine.attachSource(src);
+
+    this.dispatchEvent(new Event('sourcechange'));
```
mihar-22, line 119: Should the event dispatch be inside the `srcChange` if block? See related comments on hls media for more context.

packages/media/src/dom/hls-js/media.ts
```diff
@@ -135,13 +137,58 @@ export class HlsJsMedia extends HTMLVideoElementHost implements HlsMediaProps {
...
     this.#src = src;
-    this.#requestLoad();
+
+    // Assigning is always a source change, so it is always announced.
+    this.dispatchEvent(new Event('sourcechange'));
+
+    // Only reload for something playback depends on. Subclasses add params that
+    // describe images rather than the stream — Mux's `poster` and `storyboard` —
+    // and changing one of those must not restart what is already playing.
+    if (srcChanged || this.#shouldEngineUpdate(this.#engineConfigKey())) this.#requestLoad();
```
mihar-22, line 195: Should we move `sourcechange` event dispatch inside this block? 

Few questions coming up here:
1. I assume ideally the event is only dispatched when the source actually changes on the engine? Any timing concerns? 
2. This logic was a little different on the dash side, intentional? I don't think a config change there indicated a load request. Should we be consistent or is there reasons not to be?
3. One of the things I took away from our team conversations around Media like HLS/DASH was that any source change meant a fresh instance to avoid unexpected issues that come up with sharing the instance. Is that what we would expect here?

mihar-22 (approved):
Thanks for making all those updates :) Looks great! Left a few comments around the specifics on what "source change" means, when the event is fired specifically, and how we react to it (do we build a new instance?).

packages/media/src/dom/hls-js/media.ts
luwes: I'm going to update this so the `sourcechange` event only is updated if the value of the source object changes. in other words if the same POJO is passed (shallow equal) it will not run. 

Yes, the new instance is something I want to test again for initially. I think we might have been a bit too dramatic with this argument. There have been some improvements around these things.

packages/media/src/dom/vimeo/media.ts
```diff
@@ -75,6 +91,7 @@ export class VimeoMedia extends VimeoMediaBase implements Partial<Video> {
   #videoWidth = Number.NaN;
   #videoHeight = Number.NaN;
   #readyState = READY_STATE_HAVE_NOTHING;
+  #title = '';
```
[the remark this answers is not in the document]
luwes: Fixed in 778d64925 — good catch. `load()` reset *after* the empty-src bail, so a cleared source never reached `#resetState()`.

The title was only the visible symptom: `duration` kept reporting `60` for a video the media no longer had, and so did `videoWidth`/`videoHeight`/`error`. So the reset now runs before the bail rather than being special-cased for `#title`:

```ts
if (!this.#player) return;
this.#resetState();
if (!this.#src) return;
this.#loadComplete = createPublicPromise<void>();
```

The `emptied` / `loadstart` dispatches and the `#loadComplete` barrier deliberately stay below the bail — replacing the barrier when there is nothing to load would leave a pending `play()` / `requestFullscreen()` await on a promise nothing resolves.

Covered by `clears state reported about the old video when the source is cleared`, which loads a source, clears it, and asserts both `contentData` and `duration`.

packages/media/src/dom/vimeo/media.ts
```diff
@@ -142,12 +163,16 @@ export class VimeoMedia extends VimeoMediaBase implements Partial<Video> {
 
   /** Reload the current source via Vimeo's `loadVideo`; no-op until `attach()`. */
   async load() {
-    if (!this.#player || !this.#src) return;
+    if (!this.#player) return;
+    // Reset before bailing on an empty src: a cleared source has nothing to load,
+    // but what we report about the old video still has to go. The events and the
+    // load barrier are left alone, so nothing already awaiting a load hangs.
     this.#resetState();
+    if (!this.#src) return;
```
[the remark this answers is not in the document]
luwes: Fixed in ff69f68ff. Both halves were real.

**The embed kept running.** Clearing reset local state but left the iframe playing, so `timeupdate` / `durationchange` wrote the cleared values straight back and `play()` still drove a video the media no longer had. `load()` now unloads on the empty-src path, which is what returns the iframe to its initial state:

```ts
if (!this.#src) {
  await this.#player.unload().catch(() => {});
  return;
}
```

**In-flight metadata could land late.** `#onLoaded` awaits a round-trip to the iframe, so a clear or source change mid-flight landed the old `duration` / `title` afterwards — and on a clear nothing later corrected it. There's now a load generation bumped per `load()`; a superseded read drops its writes and its events.

`this.#loadComplete.resolve()` deliberately stays outside that guard: a superseded load can still have `play()` / `requestFullscreen()` awaiting the barrier, and skipping it would hang them on a promise nothing resolves.

Two tests cover it: `clears state reported about the old video when the source is cleared` (asserts `unload` was called alongside the cleared state) and `ignores metadata that arrives after the source is cleared`, which holds `getVideoTitle` open across the clear. I verified the second one fails without the generation guard rather than passing vacuously.

packages/media/src/dom/vimeo/media.ts
```diff
@@ -142,12 +164,23 @@ export class VimeoMedia extends VimeoMediaBase implements Partial<Video> {
...
+    this.#loadId++;
+    // Reset before bailing on an empty src: a cleared source has nothing to load,
+    // but what we report about the old video still has to go. The events and the
+    // load barrier are left alone, so nothing already awaiting a load hangs.
     this.#resetState();
+    if (!this.#src) {
+      // The embed has to stop too. Left running it keeps playing and writes the
+      // state just cleared straight back through its own events.
+      await this.#player.unload().catch(() => {});
+      return;
```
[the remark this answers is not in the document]
luwes: I don't think this one holds, but checking it turned up a real hang next door, fixed in 7e1351287.

**On the reported race.** `unload()` and `loadVideo()` are both `callMethod`, which posts through the same `ready()` promise (`@vimeo/player/src/player.js:181`):

```js
return this.ready().then(() => {
  storeCallback(this, name, { resolve, reject });
  postMessage(this, name, args);
}).catch(reject);
```

`ready()` hands back the same stored promise every time, so the `.then` callbacks run in registration order — the clear's `postMessage('unload')` is dispatched before the new source's `postMessage('loadVideo')`, and the iframe handles messages in arrival order. The commands are serialized by the message channel in call order regardless of `await` vs `void`.

What resolves late is `unload()`'s *response callback*, and that only settles a promise; it can't retroactively undo a command the embed already processed. So a late-resolving `unload()` can't wipe the video that started loading after it. `await` there is for direct `await media.load()` callers, not for ordering.

**What was actually broken.** Every `load()` replaced `#loadComplete` with a fresh promise, so anything already awaiting the outgoing one — `play()`, `requestFullscreen()`, `requestPictureInPicture()` — was left holding a promise nothing would ever resolve. Reachable on any source replace or clear before the load finished:

```ts
this.#loadId++;
this.#loadComplete.resolve();  // ← added; unblock the outgoing load
this.#resetState();
```

No-op for a load that completed, and it unblocks the one that didn't. Covered by `unblocks a pending play() when the source is replaced mid-load`, which hangs to a timeout without the line — so the guarantee I claimed on the last two comments now holds end to end instead of only inside `#onLoaded`.

packages/media/src/dom/vimeo/media.ts
```diff
@@ -142,12 +164,25 @@ export class VimeoMedia extends VimeoMediaBase implements Partial<Video> {
...
   async load() {
-    if (!this.#player || !this.#src) return;
+    if (!this.#player) return;
+    // Supersedes any metadata read still in flight for the outgoing source.
+    this.#loadId++;
+    // Unblock whoever is awaiting the outgoing load before the barrier is replaced
+    // below. The replacement leaves them holding a promise nothing resolves.
+    this.#loadComplete.resolve();
+    // Reset before bailing on an empty src: a cleared source has nothing to load,
+    // but what we report about the old video still has to go.
```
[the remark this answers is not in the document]
luwes: Correct, and it's the interaction rather than either line on its own — resolving the outgoing barrier in `load()` turned this unconditional `resolve()` into one that settled the *incoming* load. Rather than add a fourth condition, I restructured the lifecycle in 795fadfed.

**The barrier is now the load's identity**, so there is no separate generation counter to keep in agreement with it:

```ts
#beginLoad(): PublicPromise<void> {
  this.#loadComplete.resolve();               // outgoing load stops stranding waiters
  this.#loadComplete = createPublicPromise<void>();
  return this.#loadComplete;
}
```

Everything that holds a barrier settles the one it was handed, never `this.#loadComplete` as it stands later:

```ts
const load = this.#loadComplete;
// ...
if (load !== this.#loadComplete) return;  // superseded; already settled by #beginLoad
// ...
load.resolve();
```

So the case you flagged is now structural: a superseded `#onLoaded` has no barrier of its own left to settle and cannot touch the one that replaced it.

Two more strandings fell out of the same rule while applying it:

- An unparsable `src` never reaches the player, so no `loaded` could ever settle that load — `load()` settles it directly.
- `detach()` leaves no player to finish anything, so it settles too.

Both were pre-existing hangs on `play()` / `requestFullscreen()` / `requestPictureInPicture()`.

New coverage: `does not settle the new load when a superseded one finishes` (holds the first load's reads open, supersedes it, asserts a pending `play()` stays pending until the *new* load reports `loaded`) and `settles the load for a src the player can never load`. I checked both fail against the previous behavior rather than passing vacuously — the first releases `play()` early, the second hangs to a timeout.

packages/media/src/dom/hls-js/media.ts
mihar-22: Sounds great to me!

PR #1921, opened 2026-08-03
fix(packages): keep live edge indicator colored

### LINKED ISSUE #1920: Keep The Live Edge Indicator Red When Disabled
## Context

The live button intentionally becomes non-interactive at the live edge and exposes both `aria-disabled="true"` and `data-live-edge`. The shipped skins use `data-live-edge` to make the status dot red.

After #1474 changed generic disabled-button styling from `[disabled]` to `[aria-disabled="true"]`, the parent button receives `filter: grayscale(1)`. That filter is applied to the rendered `::before` indicator as well, so the red live-edge dot appears gray. This affects the default and minimal skins in both their CSS and Tailwind variants.

## Reproduction

1. Load a live stream with a shipped live video or live audio skin.
2. Play within the live-edge window so the live button has `aria-disabled="true"` and `data-live-edge`.
3. Observe the live indicator dot.

The dot appears gray because the generic disabled filter desaturates the whole button.

## Expected behavior

The live button remains focusable and exposes `aria-disabled="true"` at the live edge, but its status indicator remains red. When playback is behind the live edge, the enabled seek-to-live button uses the neutral gray indicator.

## Acceptance criteria

- Preserve the live button’s existing `aria-disabled` behavior at the live edge.
- Render the live-edge indicator red without grayscale.
- Keep the behind-live indicator neutral and the button actionable.
- Keep opacity and cursor styling for disabled controls without desaturating their colors.
- Keep default/minimal and CSS/Tailwind skin variants in parity.

Related: #1390, #1473, #1474

mihar-22:
Closes #1920

## Summary

Keep the live-edge status indicator red while preserving the live button’s focusable `aria-disabled` state and inactive visual treatment. The availability-to-ARIA migration made the generic disabled grayscale filter apply to the entire live button, including its red status dot.

## Changes

- Remove grayscale from generic disabled-button styling while retaining opacity and cursor feedback
- Keep default/minimal and CSS/Tailwind skin variants in parity
- Add browser coverage proving live and ordinary disabled buttons preserve their colors and reduced opacity

## Testing

- `pnpm --filter @videojs/html... build`
- `pnpm build:sandbox`
- `pnpm --dir apps/e2e exec playwright test tests/visual/video-skin.spec.ts --project=vite-chromium --project=vite-webkit --grep "keeps the live-edge indicator colored"`
- `pnpm typecheck`

### DIFF
--- a/packages/skins/src/default/css/components/button.css
+++ b/packages/skins/src/default/css/components/button.css
@@ -37,7 +37,10 @@
   &[aria-disabled="true"] {
     cursor: not-allowed;
     opacity: 0.5;
-    filter: grayscale(1);
+
+    &:not([data-live-edge]) {
+      filter: grayscale(1);
+    }
   }
 }
 
--- a/packages/skins/src/default/css/components/button.css
+++ b/packages/skins/src/default/css/components/button.css
@@ -37,10 +37,6 @@
   &[aria-disabled="true"] {
     cursor: not-allowed;
     opacity: 0.5;
-
-    &:not([data-live-edge]) {
-      filter: grayscale(1);
-    }
   }
 }
 
--- a/packages/skins/src/minimal/css/components/button.css
+++ b/packages/skins/src/minimal/css/components/button.css
@@ -37,7 +37,10 @@
   &[aria-disabled="true"] {
     cursor: not-allowed;
     opacity: 0.5;
-    filter: grayscale(1);
+
+    &:not([data-live-edge]) {
+      filter: grayscale(1);
+    }
   }
 }
 
--- a/packages/skins/src/minimal/css/components/button.css
+++ b/packages/skins/src/minimal/css/components/button.css
@@ -37,10 +37,6 @@
   &[aria-disabled="true"] {
     cursor: not-allowed;
     opacity: 0.5;
-
-    &:not([data-live-edge]) {
-      filter: grayscale(1);
-    }
   }
 }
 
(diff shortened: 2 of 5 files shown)


packages/skins/src/default/css/components/button.css
```diff
@@ -37,7 +37,10 @@
   &[aria-disabled="true"] {
     cursor: not-allowed;
     opacity: 0.5;
-    filter: grayscale(1);
+
+    &:not([data-live-edge]) {
```
cjpillsbury: nitpick(non-blocking): This feels hacky to shove this into the generic `button.css`, but I don't think this should be a blocker for PR merging.

packages/skins/src/minimal/css/components/button.css
```diff
@@ -37,7 +37,10 @@
   &[aria-disabled="true"] {
     cursor: not-allowed;
     opacity: 0.5;
-    filter: grayscale(1);
+
+    &:not([data-live-edge]) {
```
cjpillsbury: nitpick(non-blocking): same as above.

cjpillsbury (approved):
gud enuff for gubment work.

packages/skins/src/default/css/components/button.css
sampotts: We could always just ditch the grayscale completely since we don't have any other colored buttons anyway.

mihar-22: Agreed — removed grayscale from the generic disabled-button styling in 0d2f4039b. Disabled controls retain the opacity and cursor feedback, and the browser test now verifies that both live and ordinary disabled buttons preserve their colors.

packages/skins/src/minimal/css/components/button.css
mihar-22: Updated here too; both default and minimal skin variants now omit grayscale while retaining the disabled opacity and cursor treatment.

Issue #1944, opened 2026-08-04
Bug: Reset Source State Before Direct Store Reattachment


decepulis:
## Context

A normal store detach resets every source-state key to the slice's initial state while preserving provider configuration. React and HTML providers call that cleanup before attaching a replacement target.

The lower-level store API also supports calling `store.attach(B)` while A is still attached. That path aborts A's signals and installs B, but it does not reset source state first. Each feature must overwrite every value it owns. If B does not support an optional capability and the feature returns early, values from A can remain published.

This behavior predates player metadata and affects generic store lifecycle semantics. Metadata should follow the same capability-guard pattern as existing features rather than carrying a one-off clear.

## Expected behavior

Direct reattachment should have the same source-state boundary as an explicit detach followed by attach:

- Abort and clean up the previous attachment.
- Reset source state to the slice-declared initial state.
- Preserve provider-lifetime config.
- Attach and synchronize the replacement target.
- Recompute derived state through the normal atomic publication path.

## Acceptance criteria

- Direct A-to-B reattachment cannot retain unsupported or unsynchronized state from A.
- Provider config survives reattachment.
- Attach hooks that read source state see the documented reset state.
- Existing listener/signal cleanup still runs exactly once.
- Subscriber behavior does not expose an unintended intermediate snapshot.
- Tests compare direct reattachment with explicit detach-then-attach, including an optional capability that B does not support.

Discovered while implementing #1875.

decepulis:
not now

Issue #1943, opened 2026-08-04
Align Existing Media with the Content Data API


decepulis:
## Context

The existing Mux and Vimeo media implementations already expose parts of the standardized `contentData` bag. The finalized contract in #1875 makes capability belong to the bag as a whole and uses one aggregate `contentdatachange` event whenever that bag changes.

These implementations should be aligned with that contract so player features can consume their metadata without provider-specific wiring.

## Scope

- Audit the content data that the existing Mux and Vimeo implementations can authoritatively provide.
- Keep each implementation's `contentData` bag current as its source, provider configuration, or asynchronously loaded metadata changes.
- Emit `contentdatachange` whenever the bag changes; do not add per-field change events.
- Preserve the capability while individual keys appear or disappear. A defined empty bag remains capable; `undefined` for the whole bag means the capability is unsupported.
- For Mux, cover the existing source-derived `poster` and `storyboard` fields and any additional authoritative fields supported by the shared API.
- For Vimeo, cover the embed-reported `title`, including its asynchronous arrival and clearing on source replacement.
- Avoid coupling player stores or UI components directly to provider internals.

## Acceptance criteria

- Mux and Vimeo expose the correct initial `contentData` bag.
- Relevant provider and source changes update the bag and emit `contentdatachange`.
- Vimeo emits when its title arrives and when that title is cleared.
- Mux emits when source-derived content data changes.
- Media-host and custom-media wrapper propagation work without provider-specific player code.
- Tests cover initial values, asynchronous updates, clearing, source replacement, and unchanged values where applicable.
- Once #1941 adds the resolved content-poster path, Mux poster data can flow through it without another provider-specific contract.

Follow-up to #1875 and #1941.

decepulis:
## Findings carried over from #1886 (being discarded)

#1886 wired the media side of content metadata differently than #1946 and turned up two things that matter here. Inlining the substance since that branch may be deleted.

### 1. Donors need `empty: null` on the `CustomMediaElement` declaration — and the neighbouring declarations are the wrong precedent

#1946 adds the `contentTitle` capability and the `HTMLMediaElementHost` getter, but does not touch `CustomMediaElement`. So today a donor can only provide `contentTitle` as a JS property — which is why #1946's sandbox has to `Object.defineProperty` a raw `<video>` to demo the media tier.

For attribute-driven donation, the declaration in `packages/media/src/dom/custom-media-element/custom-media-element.ts` needs:

```ts
[...]

Issue #1942, opened 2026-08-04
Move Orientation Lock to Provider Configuration


decepulis:
## Context

Orientation lock currently uses the legacy factory-configured feature form. The provider-lifetime feature configuration added by #1875 gives HTML and React consumers a consistent, responsive way to configure selected features after player creation.

#1875 intentionally generates `String` HTML attributes because its first configuration values are nullable text. This follow-up is the next provider-configuration migration and should establish how feature declarations describe HTML conversion when future configuration values are not strings.

## Scope

- Move orientation-lock configuration from its feature factory argument to declared provider configuration.
- Expose the configuration through the selected player's React props and HTML properties/attributes.
- Apply configuration changes during the provider lifetime without recreating the player store.
- Define feature-owned HTML conversion metadata that can support `String`, `Boolean`, `Number`, and other future configuration types instead of inferring conversion solely from a nullable default.
- Preserve the existing fullscreen lock/unlock behavior and browser-API fallbacks.
- Remove the legacy factory-configured form once its callers and tests are migrated.

## Acceptance criteria

- Orientation lock can be configured and updated through React and HTML providers.
- Custom players without the feature do not expose its provider configuration in their types.
- React props, HTML properties, and HTML attributes normalize values according to the feature's declared configuration types.
- Conversion behavior has focused coverage, including a non-string fixture if orientation lock itself remains string-valued.
- Existing default behavior remains unchanged.
- Tests cover initial configuration, updates, clearing/restoring the declared default, and detach/destroy cleanup.

Follow-up to #1875.

decepulis:
## Findings carried over from #1886 (being discarded)

#1886 migrated orientation lock onto provider config as part of its POC. That branch is not going to land, but it surfaced two things worth having here before anyone starts. Inlining the substance rather than just linking, since the branch may be deleted.

### 1. `createScreenOrientationLock` captures `type` at construction, not at lock time

`packages/core/src/dom/presentation/orientation.ts:27` destructures in the parameter list:

```ts
export function createScreenOrientationLock({
  type = 'landscape',
}: ScreenOrientationLockConfig = {}): ScreenOrientationLock {
```

`type` is bound once when the lock is created. This is harmless today, because `forConfig` in `feature.ts` freezes feature config at construction anyway — nothing can change it later. It stops being harmless the moment config becomes responsive: the store half of this migration will look finished while `lock()` keeps using whatever value was set at attach, with no error and no warning.

[...]

Issue #1941, opened 2026-08-04
Add Content Poster to Player Metadata


decepulis:
## Context

The metadata feature introduced by #1875 intentionally starts with content title. Player authors and media providers also need a content-poster value with the same ownership and precedence model.

This must remain independent from legacy `title` and `poster` concepts. The resolved content poster is intended to take precedence over the existing poster configuration flow, so the work must reconcile those paths rather than creating two competing displayed-poster values.

## Scope

- Add user-configured `contentPoster` and `defaultContentPoster` inputs to the metadata feature.
- Let capable media donate `contentPoster` and announce changes through the shared media metadata API.
- Resolve user override, media value, user default, and feature fallback in the established order.
- Reconcile the resolved content poster with the existing poster configuration and rendering flow.
- Preserve the established empty-string, `null`, and unsupported-value semantics.

## Acceptance criteria

- React and HTML expose the new provider inputs only when metadata is selected.
- Media hosts and wrappers propagate the content-poster capability and change event.
- One resolved poster path determines the displayed poster, with content poster taking the documented precedence.
- Source changes, media detach, provider updates, and imperative writers behave consistently with content title.
- Tests cover precedence and the existing poster integration.

Follow-up to #1875.

decepulis:
## Findings carried over from #1886 (being discarded)

#1886 built `contentPoster` and `contentPosterAlt` end to end, including the React `<Poster>` and HTML `<media-poster>` plumbing. That branch is not landing, but it hit three traps that are cheap to know about in advance and expensive to find by hand. Inlining the substance since the branch may be deleted; the working code is on `feat/content-metadata-poc` if it's still around when this starts.

### 1. An empty resolved poster must remove `src`, not set it to `""`

`<img src="">` resolves to the current document URL and re-requests the page. Since the resolve chain's final fallback is the empty string, "no poster" arrives as `''` and must not be piped into `src`.

Two consequences beyond the obvious one:

- HTML: `removeAttribute('src')` rather than `setAttribute('src', '')`.
- React: returning `null` instead of rendering the element at all — otherwise every player without a poster carries a meaningless `<img>`, and the blur-up placeholder (which keys off `img[data-visible]:not([data-loaded])`) stays on screen forever waiting for a load that can never happen. `srcSet` and `render` are the escape hatches for `<picture>` and framework image components, so their presence means there *is* something to render.

### 2. `alt` ownership is decided by presence, never emptiness

[...]

PR #1946, opened 2026-08-04
feat(packages): resolve feature state from user and media values

### LINKED ISSUE #1875: Architecture: User-Configured Feature State and Provider Reconciliation
Media metadata arrives from two directions: the media provider reports it, and the author configures it. For #1798 both paths have to coexist, which means the store needs a general way for users to configure state in a feature slice — not only at setup, but at any point during the session — and a defined way to reconcile that user-configured state with whatever the provider supplies.

Reconciliation is the hard part. Which side wins when both are set, what happens when the user clears an override, and whether a user-set value survives a source change or a provider re-emitting its own metadata. This is a core/store mechanism rather than a metadata-specific one; metadata is just the first consumer.

The design work in #1799 isn't locked in, but we've had enough conversations to start building. Nothing here should block on it.

## Context

- vidstack puts author-supplied `title`/`artist` directly on the player alongside provider-derived state — `MediaPlayerProps extends MediaState` in [player-props.ts](https://github.com/vidstack/player/blob/main/packages/vidstack/src/core/api/player-props.ts).
- Base UI's controlled/uncontrolled split — `value` plus `onValueChange` versus `defaultValue`, as in [Accordion](https://base-ui.com/react/components/accordion) — is the closest UI-library analogue for "the author owns this value" versus "the component owns it".

decepulis:
Refs #1798
Closes #1875

## Why

> [!TIP]
> [Try the metadata sandbox](https://v10-sandbox-git-codex-user-configured-feature-mux.vercel.app/metadata/). Toggle each source and edit its value to see which title the store chooses.

A player title can come from the app or the media. The app may know the title from its CMS; a media wrapper may discover one later. Player UI needs one resolved value, and those inputs have different lifetimes: user input must survive media detach, while media-owned input must reset.

This PR defines a reusable feature-state pattern and uses content metadata as its example:

```ts
userContentTitle ?? mediaContentTitle ?? userDefaultContentTitle ?? ""
```

User and media inputs live in private source state. A feature-owned `config` map tells providers how to write user inputs and tells the store which user-owned keys survive detach. Generic `derived` formulas publish the value consumers read.

### React

A player gets these props only when it includes `metadataFeature`.

Use `contentTitle` to override media:

```tsx
const Player = createPlayer({ features: [metadataFeature] });

<Player.Provider contentTitle="Title from my CMS">
  <Video />
  <Title />
</Player.Provider>
```

Or provide a fallback that media may override:

```tsx
<Player.Provider defaultContentTitle="Title if media provides none">
  <Video />
  <Title />
</Player.Provider>
```

Read or update the resolved title through the store:

```tsx
const contentTitle = Player.usePlayer((state) => state.contentTitle);
const setContentTitle = Player.usePlayer((state) => state.setContentTitle);
```

### HTML

HTML exposes the same inputs as kebab-case attributes:

```html
<video-player content-title="Title from my CMS">
  <video></video>
</video-player>
```

```html
<video-player default-content-title="Title if media provides none">
  <video></video>
</video-player>
```

An empty string is a real title. Removing an attribute clears that user tier. Properties are one-way inputs: store actions do not rewrite properties or attributes.

> [!NOTE]
> This PR generates string-valued HTML inputs because its configuration is nullable text. Runtime conversion metadata for future non-string inputs is deferred to [#1942](https://github.com/videojs/v10/issues/1942), where there is a concrete use case.

### Store features

A feature declares each provider-facing key as a private action plus its provider-owned state key:

```ts
const MEDIA_CONTENT_TITLE = Symbol('vjs.contentTitle.media');
const USER_CONTENT_TITLE = Symbol('vjs.contentTitle.user');
const USER_DEFAULT_CONTENT_TITLE = Symbol('vjs.defaultContentTitle.user');
const SET_USER_CONTENT_TITLE = Symbol('vjs.contentTitle.user.set');
const SET_USER_DEFAULT_CONTENT_TITLE = Symbol('vjs.defaultContentTitle.user.set');

export const metadataFeature = definePlayerFeature({
  name: 'metadata',
  config: {
    contentTitle: {
      action: SET_USER_CONTENT_TITLE,
      state: USER_CONTENT_TITLE,
    },
    defaultContentTitle: {
      action: SET_USER_DEFAULT_CONTENT_TITLE,
      state: USER_DEFAULT_CONTENT_TITLE,
    },
  } satisfies PlayerFeatureConfig<MetadataSourceState>,
  state: ({ set }): MetadataSourceState => {
    const setUserContentTitle = (value: MediaContentValue) => set({ [USER_CONTENT_TITLE]: value });
    const setUserDefaultContentTitle = (value: MediaContentValue) => set({ [USER_DEFAULT_CONTENT_TITLE]: value });

    return {
      [MEDIA_CONTENT_TITLE]: undefined,
      [USER_CONTENT_TITLE]: undefined,
      [USER_DEFAULT_CONTENT_TITLE]: undefined,
      [SET_USER_CONTENT_TITLE]: setUserContentTitle,
      [SET_USER_DEFAULT_CONTENT_TITLE]: setUserDefaultContentTitle,
      setContentTitle: setUserContentTitle,
      setDefaultContentTitle: setUserDefaultContentTitle,
    };
  },
  derived: {
    contentTitle: ({ get }) => get()[USER_CONTENT_TITLE] ?? get()[MEDIA_CONTENT_TITLE] ?? get()[USER_DEFAULT_CONTENT_TITLE] ?? DEFAULT_CONTENT_TITLE,
  },
  // attach...
});
```

Each map is checked with `satisfies PlayerFeatureConfig<SourceState>`. This rejects actions or configured state keys outside that feature's source state, and requires mapped actions to accept `string | null | undefined` so omitted and removed inputs clear their tier. The mapped action parameter supplies the React and HTML TypeScript type.

The explicit `satisfies` is deliberate: it keeps `definePlayerFeature({...})` inference unchanged and reports mapping mistakes beside the map. `definePlayerFeature` translates each config `state` entry into the slice's generic `preserve` metadata. The config map stores no values, and the generic store has no `$config` controller.

On detach, ordinary source state returns to its initial value. Preserved user keys keep their current values. `combine()` unions those keys across slices. Derived formulas run before publication, so source and resolved state commit together.

### Media

A media wrapper opts into the general content-data capability by returning any defined bag, including an empty one:

```ts
media.contentData = {};
```

It may add or remove keys during the same attachment and emits the shared event whenever the bag changes:

```ts
media.contentData = { ...media.contentData, title: 'Title from media' };
media.dispatchEvent(new Event('contentdatachange'));
```

The whole bag being `undefined` means the capability is unsupported. Missing, `undefined`, and `null` titles fall through; strings, including `""`, win. Existing `media.title` remains independent.

## How to review

1. **Start with the DX.** [Try the sandbox](https://v10-sandbox-git-codex-user-configured-feature-mux.vercel.app/metadata/) and read the compact [resolved feature-state design](https://github.com/videojs/v10/pull/1946/changes#diff-2fd505027d50208ee2d31a7ed57eea8c2960271d13e9187f68bdf73b2052d480). The [AI `/grill-me` log](https://app.notion.com/p/mux/Sol-s-metadata-scratchpad-3b297a7f89d080809e7ffaed96ec653b?source=copy_link) has the longer decision history.
2. **Review the feature contract.** [Player feature routing and inference](https://github.com/videojs/v10/pull/1946/changes#diff-003d2e15560fd1e2832cebc3c68b6f018bb47be2573ef03de3cedbacf8c593e1) connect provider keys to actions and detach persistence. [Player types](https://github.com/videojs/v10/pull/1946/changes#diff-4b2dab1021d846f0ba0ba397321e3e0e25245ec06d2e18da176457c17b22d117) carry the inferred provider surface.
3. **Review the example feature.** [`metadataFeature`](https://github.com/videojs/v10/pull/1946/changes#diff-2dac285d3d1813b63ee412fa6e839800bc60fc4216645f2e45491da9feb535ea) owns the precedence formula, private inputs, public writers, and media listener. [Its tests](https://github.com/videojs/v10/pull/1946/changes#diff-a827b6eae6c37d4787924c44fff8cd8fb5dfc1a00ce83023286bc0579fab9000) cover precedence, empty strings, asynchronous media data, detach, presets, and selectors.
4. **Review the store mechanics.** [Slice types](https://github.com/videojs/v10/pull/1946/changes#diff-66dba168b36d7aa2262a4dc0c13597698a7412f1643f693c5afd2876de85e9b7) define eager derivation and detach persistence. [The store](https://github.com/videojs/v10/pull/1946/changes#diff-39cdf67dfbad271e6c57d0c35cbf3f43097c7ebe024da869ccff712e93150d83) keeps symbols private, derives before commit, and resets only unpreserved state. [`combine()`](https://github.com/videojs/v10/pull/1946/changes#diff-876bf33a10f88fa57f3ccf9916edc16c8b04333525b428dd2073ec00acda1330) merges formulas and persistence keys.
5. **Review the adapters.** [React](https://github.com/videojs/v10/pull/1946/changes#diff-d1b17d05da5ca0386d73dbb4a41902f8454cd87e42e9131426f4fa2f5a58d631) seeds new stores for SSR and first render, then forwards only committed prop changes. [HTML](https://github.com/videojs/v10/pull/1946/changes#diff-d58957110a5723d10d7f0454cbd910a37b474f11df88f4af71c65ecf3cc5f6e5) follows existing reactive-property and attribute conventions.
6. **Review the media boundary.** [Content-data types](https://github.com/videojs/v10/pull/1946/changes#diff-9bf90842d4a45221f8dedd7ab05ed866888a8875a089803a20fba19e529583bc), the [capability predicate](https://github.com/videojs/v10/pull/1946/changes#diff-2f2d18743fc8e8ff10f63aa0ac5aa65df338067178d1acc0e014111132d75bfc), and [media host forwarding](https://github.com/videojs/v10/pull/1946/changes#diff-91b042c81f3c25875d7ff474e1b1352d8c5aa5e76bab605c38dc30778a224998) standardize media-owned catalog data.

## Trade-offs exposed by this PR

- Each configurable value needs a private action and a provider-owned state key. That is more feature code than a generic config snapshot, but it keeps values in the existing state/action model and makes lifetime ownership explicit beside the feature.
- Provider adapters invoke actions sequentially. Normal same-turn subscriber delivery is coalesced, but if a later action or formula throws, earlier actions remain applied. This PR does not add batching without a demonstrated need.
- `derived` remains generic store machinery. Removing it would force every mutation path—including media events and detach—to remember to recompute public values.

## Out of scope

- [Title UI](https://github.com/videojs/v10/issues/1123)
- [`contentPoster` and existing-poster reconciliation](https://github.com/videojs/v10/issues/1941)
- [Existing media content-data cleanup](https://github.com/videojs/v10/issues/1943)
- [Responsive orientation-lock configuration](https://github.com/videojs/v10/issues/1942)
- [Feature-configuration API references](https://github.com/videojs/v10/issues/1940)
- [Generic source-state reset on direct reattachment](https://github.com/videojs/v10/issues/1944)

Existing `title` remains intentionally independent from `contentTitle`. The sandbox overlay only demonstrates the resolved store value.

## Architecture notes

Yes, there's a fair bit of machinery in here. (And somehow, still less than the [first time](https://github.com/videojs/v10/pull/1886) I built this.) And I'm glad you two are pushing on it! If anything in here is unnecessary—if it's just AI psychosis—I'm really happy to eliminate it.

Let me tell you about some hard constraints that drove my decision-making. In that, hopefully you'll understand why we need `config` at the player-feature layer and `derived` in the store. (Though there is some wiggle room in how those concepts are expressed.)

### Why do we need separate buckets?

- Just `state: { contentTitle }` isn't enough. We need separate buckets for media-provided state and user-provided state so it's not "last write wins." That's the only way `state = user ?? media ?? default` is possible.
- User state and media state have different lifecycles. When media detaches, media-provided state should be reset. User-provided state should persist.

### Why do we need `config`?

- When media detaches, ordinary source state resets to its initial values. Because user-provided values also live in source state, we need a way to identify the keys that should survive detach.
- The provider also needs a way of identifying what in a feature is user-configurable.
- Those two facts lead me to say we need some sort of `config` key. It doesn't have to look how I implemented it, but, there's gotta be something to say what's what and where

### Why do we need `derived`?

- We also need some mechanism to combine the media and user buckets. Some sort of `derived` mechanism. That mechanism needs to run the `state = user ?? media ?? default` formula every time one of those dependencies changes and store the result in a way users can easily access.
  - I chose a top-level `derived: {}` object that stores formulas. The results of those formulas get put onto the frozen `usePlayer` state object users consume on every patch.
  - I also explored `state: { contentTitle: derived(...) }` functions that live inside the slice `state` object. I had trouble coming up with an elegant implementation of that given our existing store. And no matter what, this approach resulted in *extremely* messy types.
  - I also explored running the formula directly in the setter, e.g. `state: { setUserTitle: v => set({ userTitle: v, contentTitle: v ?? mediaTitle ?? defaultTitle }) }`. That got messy because every state update needs to go through the right setter, putting a burden on every mutation site. It felt fragile.
  - In this implementation, derived formulas run on every patch instead of only when a dependency changes, so I don't have to write dependency tracking. The performance cost is negligible here.

### Anything else?

- The mechanism we build here should be generic, not specific to `metadataFeature`. `orientationLockFeature` already demonstrates that other features might require user configuration.

---

> [!NOTE]
> **Medium Risk**
> Touches core store commit/detach semantics and player provider lifecycles across React/HTML; behavior is heavily tested but regressions could affect all feature state and attach/detach flows.
> 
> **Overview**
> Introduces a **resolved feature-state** pattern: private source inputs, feature-owned `config` maps (provider key → action + detach-persistent state), and **eager `derived`** formulas so only resolved values appear on the public store. The store gains **source vs public snapshots**, **`preserve` on detach**, symbol-backed config actions, and **identity-preserving updates** when derived output is unchanged.
> 
> **`metadataFeature`** is the reference implementation: `contentTitle` resolves `user ?? media contentData.title ?? user default ?? ""`, listens for `contentdatachange`, and ships in standard audio/video/live presets (not background). Media adds optional **`contentData`** / capability checks, separate from legacy `title`.
> 
> **React and HTML** `createPlayer` merges feature configs into typed provider props/attributes (`contentTitle`, `default-content-title`), one-way into store actions, with SSR/hydration and reattach seeding covered by tests. A **metadata sandbox** demonstrates precedence toggles. Design docs record the decision and updated reactive-store rationale.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 93ed860af5e0061d237eaf86c96daafd4eaea3ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


decepulis:
@cursor review

sampotts:
> ✅ Bugbot reviewed your changes and found no new issues!

Show off 😛

internal/design/store/resolved-feature-state.md
```diff
@@ -0,0 +1,95 @@
+---
+status: implemented
+date: 2026-08-04
+---
+
+# Resolved player feature state
+
+This record defines how a player feature resolves one public value from inputs with different owners and lifetimes. Content metadata illustrates the pattern but does not define its boundary. Source, exports, and tests define the current API.
```
decepulis:
```suggestion
This record defines how a player feature resolves one public value from inputs with different owners and lifetimes. Source, exports, and tests define the current API.
```

"Content metadata illustrates the pattern but does not define its boundary." is weirdly specific and imo unnecessary.

internal/design/store/resolved-feature-state.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Consumer experience
+
+A selected feature may expose user inputs through its provider. Those inputs appear in React and HTML only when the player includes that feature.
+
+For example, a content-title feature may accept a user override:
```
decepulis:
```suggestion
For example, a `metadataFeature` may accept user input for its `contentTitle` state:
```

internal/design/store/resolved-feature-state.md
```diff
@@ -0,0 +1,95 @@
+---
+status: implemented
+date: 2026-08-04
+---
+
+# Resolved player feature state
+
+This record defines how a player feature resolves one public value from inputs with different owners and lifetimes. Content metadata illustrates the pattern but does not define its boundary. Source, exports, and tests define the current API.
+
+## Consumer experience
```
decepulis, line 14: Before this section, I think we need a brief "Media interaction" section noting that features can be configured by media. Features listen to media events and set internal state. 

With that context, the fallback comment later makes more sense.

internal/design/store/resolved-feature-state.md
```diff
@@ -0,0 +1,95 @@
...
+- `state` holds attachment-owned source values and feature actions. It resets when media detaches.
+- `derived` eagerly resolves public state from `config` and `state` before one frozen snapshot is published.
+
+Features choose their own formula. A fallback-shaped value can use:
+
+```ts
+providerOverride ?? attachedMediaValue ?? providerFallback ?? featureFallback
+```
+
+Internal source values may use symbol keys. They follow the normal attachment update and reset path but do not appear in the public store. Lower-precedence inputs remain current beneath an override, so removing that override reveals the latest available value.
```
decepulis: goodness this whole guide is jargony. take a pass with `writing-style.md`

decepulis: In a similar "reconsider" vein, once again take a look at the guidance that drove https://github.com/videojs/v10/pull/1835 -- see how skills and design docs were cut back -- to make sure that this guide has the minimum necessary information and relies on source for truth.

packages/core/src/dom/feature.ts
```diff
@@ -7,24 +13,32 @@ export interface ConfigurablePlayerFeature<Config, State> extends PlayerFeature<
...
+  extends Omit<SliceConfig<PlayerTarget, State>, 'attach' | 'config' | 'derived' | 'state'> {
   state: (ctx: StateContext<PlayerTarget>, config: Config) => State;
   attach?: (ctx: AttachContext<PlayerTarget, State>, config: Config) => void;
 }
 
 const definePlayerSlice = defineSlice<PlayerTarget>();
 
+/** Define a static player feature. */
 export function definePlayerFeature<State>(config: SliceConfig<PlayerTarget, State>): PlayerFeature<State>;
+/** Define a responsive player feature with provider-lifetime config. */
```
decepulis:
```suggestion
/** Define a player feature with responsive user-facing config. */
```

packages/core/src/dom/feature.ts
```diff
@@ -7,24 +13,32 @@ export interface ConfigurablePlayerFeature<Config, State> extends PlayerFeature<
...
+export function definePlayerFeature<Config>(): SliceFactory<PlayerTarget, Config>;
+/** Define the legacy factory-configured form used by static feature variants. */
```
decepulis:
```suggestion
/** Define the legacy factory-configured form used by static feature variants. Currently used for orientation lock, scheduled for removal in #1942 */
```

packages/media/src/core/types.ts
```diff
@@ -501,6 +501,26 @@ export interface MediaPosterCapability {
...
+
+/** Standardized content metadata reported by a media implementation. */
+export type MediaContentData = Readonly<Record<string, MediaContentValue>>;
+
+/** Events emitted when a media implementation's content data changes. */
+export interface MediaContentDataEvents {
+  contentdatachange: EventLike;
+}
+
+/** Optional media-owned content metadata, independent from legacy media properties. */
```
decepulis:
```suggestion
/** Optional media-owned content metadata */
```

packages/react/src/player/create-player.tsx
```diff
@@ -70,22 +70,50 @@ export function createPlayer<const Features extends AnyPlayerFeature[]>(
...
 
 export function createPlayer(config: CreatePlayerConfig<AnyPlayerFeature[]>): CreatePlayerResult<AnyPlayerStore> {
-  function Provider({ children }: ProviderProps): ReactNode {
-    const [store, setStore] = useState(() => createStore<PlayerTarget>()(combine(...config.features)));
+  const slice = combine(...config.features);
+  const configKeys = Object.keys(slice.config ?? {});
+
+  function Provider(props: ProviderProps<any>): ReactNode {
+    const { children } = props;
+    const providerConfig = readProviderConfig(props, configKeys);
```
decepulis: I could use a comment explaining why this is necessary. Or a better-named function than `readProviderConfig`

packages/react/src/player/create-player.tsx
```diff
@@ -70,22 +70,50 @@ export function createPlayer<const Features extends AnyPlayerFeature[]>(
...
+    const [store, setStore] = useState(() => createStore<PlayerTarget>()(slice, { config: providerConfig }));
+    const syncedConfig = useRef({ store, values: providerConfig });
     const [popupGroup] = useState(() => createPopupGroup());
     const [media, setMedia] = useState<Media | null>(null);
     const [container, setContainer] = useState<HTMLElement | null>(null);
 
     useDestroy(store);
 
+    useLayoutEffect(() => {
```
decepulis, line 102: I could use a comment somewhere explaining that this layout effect is syncing the props to the store

packages/store/src/core/combine.ts
decepulis, line 1: for future human reviewer: 

tldr `combine` used to be `state + attach` and now it's `user config  +  media state  +  derived formulas  +  attach`

packages/store/src/core/slice.ts
```diff
@@ -26,9 +41,11 @@ export interface AttachContext<Target, State> {
 // State Context
 // ----------------------------------------
 
-export interface StateContext<Target> {
+export interface StateContext<Target, Config = object> {
   /** Returns the current target. Throws if not attached. */
   target: () => Target;
+  /** Provider-lifetime configuration for feature-authored actions. */
```
decepulis:
```suggestion
  /** User-defined, provider-lifetime configuration. */
```

packages/store/src/core/slice.ts
```diff
@@ -40,43 +57,110 @@ export interface StateContext<Target> {
    *   (e.g., loading a new source cancels pending seeks).
    */
   signals: AbortControllerRegistry;
-  /** Read current slice state. Safe to use inside action closures (not during `state()` init). */
-  get: () => Readonly<Record<string, unknown>>;
-  /** Patch the slice state. Safe to use inside action closures (not during `state()` init). */
-  set: (partial: Record<string, unknown>) => void;
+  /** Read current slice source state. Safe to use inside action closures (not during `state()` init). */
```
decepulis: Generally you use the term `source` and `media` interchangeably. Consider unifying to just `media`. While `source` is more technically correct, `media` might be more familiar to a maintainer. Apply this comment everywhere (unless I'm misunderstanding a distinction here).

packages/store/src/core/state.ts
```diff
@@ -69,6 +70,15 @@ class StateContainer<T> implements WritableState<T> {
     }
   }
 
+  replace(next: T): void {
+    // Internal source/config inputs can change without changing the public
+    // snapshot. Preserve its identity and do not notify subscribers in that case.
```
decepulis: Provide an example, since this is an important case to grok

internal/design/store/resolved-feature-state.md
decepulis: Applied in fb1c42a9. I removed the metadata-specific sentence so the introduction states only the durable purpose and points to source for the current API.

internal/design/store/resolved-feature-state.md
decepulis: Applied in fb1c42a9. The example now names `metadataFeature` and its `contentTitle` state directly.

internal/design/store/resolved-feature-state.md
decepulis: Added a short Media interaction section before the consumer example in fb1c42a9. It explains that features read attached media, listen for its events, write internal state, and clear media-owned state on detach without clearing user input.

internal/design/store/resolved-feature-state.md
decepulis: Reworked the whole record in fb1c42a9 using `writing-style.md` and the PR #1835 guidance. It is down to 494 words, uses simpler terms, removes implementation mechanics, and points to source for current behavior.

packages/core/src/dom/feature.ts
decepulis: Applied the suggested wording in fb1c42a9.

packages/core/src/dom/feature.ts
decepulis: Applied in fb1c42a9. The JSDoc now identifies the legacy static-variant form, its orientation-lock use, and planned removal in #1942.

packages/media/src/core/types.ts
decepulis: Applied the shorter comment in fb1c42a9.

packages/store/src/core/combine.ts
decepulis: Agreed. That is the right high-level summary of this file; no code change needed for this informational note.

packages/media/src/core/types.ts
decepulis: Applied the shorter comment in fb1c42a9.

packages/store/src/core/slice.ts
decepulis: There is a distinction here. In the player design, I now use `media` for media-owned values. Inside generic `@videojs/store`, source state means the raw slice state before derived values; it can also contain actions or non-media target state. I kept the internal type names, but changed the public JSDoc in fb1c42a9 to say `slice state before derived values` so maintainers do not need to infer that distinction.

packages/react/src/player/create-player.tsx
decepulis: Addressed in fb1c42a9. Renamed the helper to `pickFeatureConfig` and added a local comment explaining that only config declared by selected features belongs in the store.

packages/store/src/core/state.ts
decepulis: Added the requested example in fb1c42a9: media metadata can change beneath a user override while the resolved title stays the same, so the public snapshot should remain stable.

internal/design/store/resolved-feature-state.md
mihar-22, line 1: Most of these decisions will live in the code, are you comfortable removing this file entirely or compacting?

packages/core/src/dom/store/features/metadata.ts
```diff
@@ -0,0 +1,68 @@
...
+import { isMediaContentDataCapable } from '@videojs/media';
+import { listen } from '@videojs/utils/dom';
+import { definePlayerFeature } from '../../feature';
+import type { PlayerFeatureConfig } from '../../player';
+
+const MEDIA_CONTENT_TITLE = Symbol('vjs.contentTitle.media');
+const USER_CONTENT_TITLE = Symbol('vjs.contentTitle.user');
+const USER_DEFAULT_CONTENT_TITLE = Symbol('vjs.defaultContentTitle.user');
+const SET_USER_CONTENT_TITLE = Symbol('vjs.contentTitle.user.set');
+const SET_USER_DEFAULT_CONTENT_TITLE = Symbol('vjs.defaultContentTitle.user.set');
```
mihar-22, line 11: Symbol descriptions so far follow the pattern `@videojs/*` and kebab-case. 

Example: `vjs.contentTitle.media` -> `@videojs/media-content-title`

packages/core/src/dom/store/features/metadata.ts
```diff
@@ -0,0 +1,68 @@
...
+      state: USER_CONTENT_TITLE,
+    },
+    defaultContentTitle: {
+      action: SET_USER_DEFAULT_CONTENT_TITLE,
+      state: USER_DEFAULT_CONTENT_TITLE,
+    },
+  } satisfies PlayerFeatureConfig<MetadataSourceState>,
+  state: ({ set }): MetadataSourceState => {
+    const setUserContentTitle = (value: MediaContentValue) => set({ [USER_CONTENT_TITLE]: value });
+    const setUserDefaultContentTitle = (value: MediaContentValue) => set({ [USER_DEFAULT_CONTENT_TITLE]: value });
```
mihar-22, line 40: if we can inline these that'd be great, it should already be typed on return object.

packages/core/src/dom/store/features/metadata.ts
```diff
@@ -0,0 +1,68 @@
...
+      setContentTitle: setUserContentTitle,
+      setDefaultContentTitle: setUserDefaultContentTitle,
+    };
+  },
+  derived: {
+    contentTitle: ({ get }) =>
+      get()[USER_CONTENT_TITLE] ??
+      get()[MEDIA_CONTENT_TITLE] ??
+      get()[USER_DEFAULT_CONTENT_TITLE] ??
+      DEFAULT_CONTENT_TITLE,
```
mihar-22, line 57: easy to reason about 👌

packages/react/src/player/create-player.tsx
```diff
@@ -70,22 +76,54 @@ export function createPlayer<const Features extends AnyPlayerFeature[]>(
...
+        return;
+      }
+
+      for (const key of configKeys) {
+        if (Object.is(previous.values[key], configValues[key])) continue;
+        setPlayerConfigValue(store, featureConfig[key]!, configValues[key]);
+      }
+
+      syncedValues.current = { store, values: configValues };
+    });
```
mihar-22, line 117: Worth calling out that this won't run on the server and it won't set the store state correctly on initial pass. I think we pull this out in the render path

packages/react/src/player/create-player.tsx
```diff
@@ -116,3 +154,13 @@ export function createPlayer(config: CreatePlayerConfig<AnyPlayerFeature[]>): Cr
     useMedia,
   };
 }
+
+function pickConfigValues(props: Record<string, unknown>, keys: readonly string[]): Record<string, unknown> {
+  return Object.fromEntries(keys.map((key) => [key, props[key]]));
+}
```
mihar-22, line 160: Looks like a generic utility? Do we not have a pick keys from object?

packages/store/src/core/state.ts
```diff
@@ -101,6 +111,18 @@ class StateContainer<T> implements WritableState<T> {
...
+  const aKeys = Reflect.ownKeys(a as object);
+  const bKeys = Reflect.ownKeys(b as object);
+  if (aKeys.length !== bKeys.length) return false;
+
+  for (const key of aKeys) {
+    if (!hasOwnProp.call(b, key) || !Object.is(a[key as keyof T], b[key as keyof T])) return false;
+  }
+
+  return true;
+}
```
mihar-22, line 124: this comes up a lot, ideally it's a generic utility from `utils` package, either here on in a follow-up nice to catch all inlined versions (if any).

packages/store/src/core/store.ts
```diff
@@ -2,20 +2,39 @@ import { isNull, isObject } from '@videojs/utils/predicate';
...
-  options?: StoreOptions<Target, State>
-) => Store<Target, State> {
-  return <State>(slice: Slice<Target, State>, options: StoreOptions<Target, State> = {}): Store<Target, State> => {
-    type TargetStore = Store<Target, State>;
+export interface StoreFactory<Target> {
+  <S extends AnySlice<Target>>(
+    slice: S,
+    options?: StoreOptions<Target, InferSliceState<S>>
+  ): Store<Target, InferSliceState<S>>;
+  <State>(slice: import('./slice').Slice<Target, State>, options?: StoreOptions<Target, State>): Store<Target, State>;
```
mihar-22, line 26: why is it importing the `Slice` type like this lol?

packages/store/src/core/store.ts
```diff
@@ -2,20 +2,39 @@ import { isNull, isObject } from '@videojs/utils/predicate';
...
+  InferSliceDerivedState,
+  InferSliceSourceState,
+  InferSliceState,
+  StateContext,
+} from './slice';
 import type { StateChange, State as StateContainer, SubscribeOptions, UnknownState, WritableState } from './state';
 import { createState } from './state';
 
 const STORE_SYMBOL = Symbol.for('@videojs/store');
+const hasOwnProp = Object.prototype.hasOwnProperty;
```
mihar-22, line 17: This is a good perf trick

packages/store/src/core/store.ts
mihar-22, line 1: A lot of code added here which is unfortunate but maybe unavoidable. As we discussed together, we can revisit all of this store shenanigans.

mihar-22 (approved):
Great work, overall pretty damn happy with where we landed. We'll keep iterating and improving in the future 💪

packages/store/src/core/store.ts
decepulis: Can not _wait_ to

packages/core/src/dom/store/features/metadata.ts
decepulis: Agreed.

packages/core/src/dom/store/features/metadata.ts
decepulis: Oh yeah. huh. I wonder why it decided to do that. I didn't give it much thought.

internal/design/store/resolved-feature-state.md
decepulis: Oh yeah totally comfortable. I'm still not sure what we decide to do a doc on or not.

packages/core/src/dom/store/features/metadata.ts
decepulis: Will address!

packages/react/src/player/create-player.tsx
decepulis: There's similar logic in `useState` that seeds the store on the server!

decepulis: I'm _really_ wary of doing it in the render path. What if React discards a render mid-run 😨

packages/store/src/core/state.ts
decepulis: Will address

packages/react/src/player/create-player.tsx
decepulis: Will look

packages/store/src/core/store.ts
decepulis: ...huh

Issue #1951, opened 2026-08-05
Shadcn Registry


mihar-22:
## Summary

Publish the validated React and Tailwind VJSC source contract as a Shadcn-compatible registry.

## Product outcome

Users can discover and install supported VJSC components and complete skin blocks through standard Shadcn registry tooling.

## Delivery flow

```txt
VJSC Cutover (#2185)
        |
        v
Catalog (#1971) + GitHub Registry (#1970)
        |
        v
Validation (#2261)
        |
        v
Direct Registry Docs (#2186)
```

## Scope

- Publish supported React and Tailwind components and complete Skin blocks.
- Define stable item names, metadata, dependencies, and references.
- Keep checked-in registry source reviewable and reproducible.
- Support standard discovery and installation workflows.

Registry emission from VJSC is already covered by #2264 and #2265. This public follow-up does not gate the Eject CLI or VJSC cutover.

## Acceptance criteria

- The catalog and GitHub registry expose the agreed public items.
- Items install with complete dependencies and valid references.
- Validation covers discovery, installation, repetition, and overlapping items.
- Direct registry documentation is published after validation.

mihar-22:
#2018 completes the registry foundation and catalog boundary: five validated manifests, 28 generated items, individual components, a branch-only core controls block, drift validation, and branch/SHA clean-install evidence. It will close #1970 and #1971 on merge; CLI add/eject work remains in #1974 and #1975.

mihar-22:
## Post-#2021 registry status

#2021 generates and validates a contained seven-item React/Tailwind shadcn registry from the canonical Skin catalog. It proves component/block manifest generation and dependency partitioning, but it is not yet the public Video.js GitHub registry promised by this epic.

#1970/#1971 remain open: the repository has no public root/nested `videojs/v10` registry surface, versioned-ref release contract, list/search/MCP discovery proof, or current clean-project branch/SHA install evidence. HTML and vanilla-CSS registry publication are not implemented; current policy explicitly publishes React/Tailwind only. #1974 and #1975 should start only after the public/versioned install contract is proven so the CLI wraps the same artifacts rather than defining a second format.

mihar-22:
Planning cleanup: #1956 now belongs to this epic because publishing the icon package and proving install-safe icon closure is part of public source distribution, not the active Video component markup/style stack in #1949.

Issue #1950, opened 2026-08-05
Epic: VJSC Skin Styling, Tokens, and Themes


mihar-22:
Parent epic: #1055

## Summary

Define the shared styling system used by canonical/source-owned components, packaged Skins, Tailwind output, scoped vanilla CSS, Default and Minimal themes, and future host design-system integration.

Component-specific styles remain owned by each component. This epic owns the cross-cutting source architecture, token contract, theme boundary, and resilience expectations.

## Established decisions

- #1964 / #2021 established canonical component recipes, shared base styles, separate theme values, React/HTML projection, editable Tailwind output, scoped vanilla CSS, exact style closure, and generated-output drift checks.
- #570 is resolved: stable semantic `media-*` classes are contained by catalog-owned Skin scopes; no configurable class-prefix feature is required for the first delivery.
- #571 is resolved: Tailwind output integrates with the host design system and defines Video.js semantic theme aliases; scoped vanilla CSS is the self-contained alternative for projects that replace Tailwind defaults.
- #1741 / #2094 stabilized the packaged public/private CSS-variable convention and added HTML/React × Default/Minimal × CSS/Tailwind customization coverage.

## Remaining boundary

- #1965 owns the complete public/private semantic token inventory, Default/Minimal expression, and accessibility-preference contract.
- Component-specific markup and styles continue to land with the Video component slices in #1949.
- #1972 under #2183 owns final canonical React/HTML × Tailwind/vanilla CSS visual and style validation.
- Documentation for Tailwind and source-owned Skin styling belongs to #840/#2184.

## Success criteria

- [x] Component styles and Skin theme values have separate owners.
- [x] Tailwind and vanilla CSS derive from one semantic source.
- [x] Vanilla component CSS is scoped by catalog metadata.
- [x] Tailwind and vanilla outputs have explicit, different host-integration contracts.
- [ ] Every public Skin variable has a documented disposition.
- [ ] Default and Minimal express complete visual differences through the accepted component/theme boundaries.
- [ ] High contrast, reduced motion, and reduced transparency are represented where required.
- [ ] Final canonical output proves multiple-Skin, host-collision, responsive, and customization behavior.
- [ ] Installed components include only their required style closure.

## Required context

- #1965
- #1949
- #1972
- PR #2021
- PR #2094
- `packages/skins/canonical/styles/`
- `packages/skins/build/styles/`

mihar-22:
[...]

Issue #1949, opened 2026-08-05
Epic: VJSC Components and Configuration


mihar-22:
Parent epic: #1055

## Summary

Define and complete the canonical component model and complete Default and Minimal Skins used by the on-demand Video preset.

This epic owns component markup, component-owned styles, framework-neutral canonical composition, and generated React/HTML projections. Final executable parity, package cutover, and documentation are downstream in #1952.

## Current state

- #2021 established constrained canonical JSX, compiler projections, the Skin-owned catalog, style generation, and the first generated component slice.
- #2179 established the canonical Container, Poster, and Overlay boundaries and made `skin.tsx` the complete root composition.
- #2181 is aligning the visible control set while keeping SeekButton independently installable.
- The temporary component-parity master plan describes the remaining Default/Minimal video slices one component gap at a time.

## Active implementation boundary

Complete these reusable Video Skin gaps in stacked order:

- BufferingIndicator.
- ErrorDialog.
- CaptionsButton.
- Cast, AirPlay, and picture-in-picture controls.
- Input indicators and StatusAnnouncer composition.
- VolumePopover availability and Default/Minimal layout variants.
- TimeSlider chapter composition.
- Video settings menu and its submenu components.
- Standard video hotkey and gesture composition.
- Complete Default and Minimal video assembly.

Default and Minimal should share domain components and differ only where layout, meaningful props, or component-owned styles require it.

## Architectural constraints

- `skin.tsx` is the canonical root composition; there is no shell concept.
- Canonical layout components are Container, Poster, and Overlay.
- Canonical source remains framework-neutral.
- Components own their source and styles.
- Target lowering remains in `html.ts` and `react.ts`, supported by generic compiler APIs.
- Independently installable SeekButton remains cataloged but is absent from Default and Minimal video compositions.
- Raw platform elements require a deliberate semantic component, compound part, Template boundary, or target lowering.

## Success criteria

- [ ] Canonical Default and Minimal video source contains the complete accepted component graph.
- [ ] Every visible structure maps to a deliberate component, compound part, private Skin composition, Template boundary, or target lowering.
- [ ] Reusable components are complete, styled, and independently installable where intended.
- [ ] Default and Minimal reuse behavior components where behavior is identical.
- [ ] React and HTML lower the same canonical intent idiomatically.
- [ ] Tailwind and vanilla CSS preserve the accepted markup and style behavior.
- [ ] Complete generated Default and Minimal output is ready for the downstream parity gates in #2183.

[...]

Issue #1055, opened 2026-03-20
Source-Owned UI


decepulis:
## Summary

Developers should be able to use packaged Video.js Skins, add individual source-owned UI components, or eject a complete Skin into their project.

VJSC is the canonical source for packaged React and HTML Skins, editable source output, Tailwind and vanilla CSS, examples, fixtures, CLI workflows, and the later public Shadcn registry.

Media ejection remains out of scope until SPF is stable and after Video.js v10 GA.

## Product outcome

Video.js provides a clear ownership curve:

1. Use a packaged Preset or Skin.
2. Add an individual complete and styled component.
3. Eject a complete Skin into the project.
4. Later, own Preset feature composition while Media, Store, and feature implementations remain packaged.

Source-owned files belong to the developer. Video.js does not silently update or merge locally modified source.

## Plan

```text
VJSC Vite Plugin (#2470)
          |
          v
VJSC Video Skins (#2183)
          |
          +--> Audio Skins (#2112) -----------+
          +--> Live Skins (#2113) ------------+--> VJSC Presets complete (#2187)
          `--> Background Skin (#2114) -------+
                                                    |
                                                    v
                                          Eject CLI (#2262)
                                                    |
                                                    v
                                     Eject Documentation (#2184)
                                                    |
                                                    v
                                          VJSC Cutover (#2185)
                                                    |
                                                    v
                                        Shadcn Registry (#1951)
```

Live work is split into the shared Live Button (#2115), Live Video Skins (#2454), and Live Audio Skins (#2455).

## Workstreams

- **VJSC Vite Plugin:** reusable Vite integration plus validated development, HMR, diagnostics, and production workflows.
- **VJSC Presets:** canonical source, React/HTML output, Tailwind/vanilla CSS, and executable parity for every shipped Skin.
- **Eject CLI:** add, view, dry-run, diff, and safe ejection directly from the VJSC catalog; no public registry prerequisite.
- **Documentation:** verified package and CLI workflows plus VJSC-owned examples; direct Shadcn guidance is deferred.
- **VJSC Cutover:** packages/CDN, sandbox, site, E2E, legacy removal, and final clean-consumer validation.
- **Shadcn Registry:** public React/Tailwind source delivery after the internal cutover.

## User workflows

```sh
# Guided Video.js workflows
videojs add skin video
videojs add skin video --dry-run
videojs add skin video --diff
videojs view skin video
videojs eject skin video

# Direct Shadcn registry follow-up
[...]

PR #1976, opened 2026-08-05
chore: configure native Eject issue hierarchy

mihar-22:
## Summary

Adds a one-time GitHub Actions workflow that configures the native sub-issue relationships for #1055 and its five sub-epics.

The workflow:

- keeps only #1949, #245, #1950, #1951, and #1952 as direct sub-issues of #1055;
- moves active and reused implementation issues under their owning sub-epic using `replace_parent`;
- removes unrelated or stale direct sub-issues from each managed parent;
- verifies the resulting hierarchy through GitHub's sub-issues API.

After the workflow succeeds, this workflow file will be removed in a follow-up cleanup PR.

---

> [!NOTE]
> **Medium Risk**
> The workflow mutates live issue relationships via the Issues API on every qualifying PR until removed; mistakes in the hard-coded map or `replace_parent` could re-parent the wrong issues.
> 
> **Overview**
> Adds a **GitHub Actions workflow** (`.github/workflows/configure-eject-subissues.yml`) that runs on PRs touching that file and uses `actions/github-script` to sync GitHub’s **native sub-issue** links for epic **#1055** and five sub-epics.
> 
> For each parent in a hard-coded map, the job **removes** sub-issues not in the desired list, **adds** missing children via the sub-issues API with **`replace_parent: true`** (so implementation issues sit under the right sub-epic), then **fails the run** if the final hierarchy doesn’t match. API calls are throttled with short sleeps; permissions are `issues: write` only.
> 
> Intended as a **one-time** setup step; the workflow file is meant to be deleted after it succeeds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40c10c5edcd61797189a57dabaec959564f0900f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/configure-eject-subissues.yml
+++ b/.github/workflows/configure-eject-subissues.yml
@@ -1,7 +1,8 @@
 name: Configure Eject issue hierarchy
 
 on:
-  push:
+  pull_request:
+    types: [opened, synchronize, reopened]
     paths:
       - .github/workflows/configure-eject-subissues.yml
 
--- a/.github/workflows/configure-eject-subissues.yml
+++ b/.github/workflows/configure-eject-subissues.yml
@@ -1,0 +1,117 @@
+name: Configure Eject issue hierarchy
+
+on:
+  push:
+    paths:
+      - .github/workflows/configure-eject-subissues.yml
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  configure:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


mihar-22:
The one-time pull-request workflow completed successfully and verified the native hierarchy through GitHub's sub-issues API. The workflow file does not need to land on `main`, so this PR can be closed.

Issue #1798, opened 2026-07-02
Feature: Communicate Media Metadata to the UI


decepulis:
Oftentimes, videos have aspects about them that can be displayed in the UI. Take, for example, `title`. `mux-video` and other media can provide this information and show it in a `media-title` component. 

How should the media communicate these details? What details should be communicated? (title, description, storyboards, posters, chapters, idk just suggestions). Where does the user configure overrides? (on the media? store? eject and ui? all? none?)

PR #1997, opened 2026-08-06
feat(packages): add title component

decepulis:
Closes part of #1123. Refs #1798.

## Sandbox
- [Metadata](https://v10-sandbox-git-claude-title-state-store-yzkrvh-mux.vercel.app/metadata/)

## What is this?
1. `<media-title>` and `<Title>` display the content title the store resolves.
2. `TitleCore` derives two things from `metadata` and nothing else: the resolved `title`, and `hidden` when there is no title to show.
3. HTML sets the native `hidden` attribute plus `data-hidden`. React renders nothing.
4. A reference page with HTML and React demos.

This revives the UI half of the closed #1603. That PR shipped its own metadata state, which #1875 and #1946 have since replaced, so nothing else survived.

## The API

Set the title on the player:

```
<video-player content-title="Big Buck Bunny">

<Player title="Big Buck Bunny">
```

Then drop it in:

```
<media-title></media-title>

<Title />
```

Nothing else is required — an empty title takes itself out of the layout. Style it against a sibling when you want it to travel with the chrome:

```
media-title { opacity: 0; transition: opacity 150ms ease-out; }
media-controls[data-visible] ~ media-title { opacity: 1; }
```

## Rebased onto the config-only title

#2176 landed while this was open. It renamed the resolved state `contentTitle` to `title`,
dropped `setContentTitle` / `setDefaultContentTitle`, and removed the `defaultContentTitle`
tier, so a title now resolves from two places instead of three: what the author set, then
what the media reports. The title is player input, not writable store state.

What that changed here:

- `TitleCore` reads `media.title`. `TitleState` is untouched, since it already called its own field `title`.
- The element and component tests drove the title through the removed store action. They now set it through the feature's own config with `setPlayerConfigValue`, the way a provider does and the way `metadata.test.ts` does.
- The reference page documents two sources rather than three, and shows setting the input again instead of calling a setter.
- #2039 then made `poster` a required member of `MediaMetadataState`, so the title fixtures carry it. `TitleCore` never reads it.

## What this isn't
- Skins. #2036 stacks on this and renders the title in the video skins.
- Registration in the skin entries. `@videojs/html/ui/title` is the only way in until a skin renders one, so no skin pays for bytes it doesn't use.
- Naming the player region from the title. The container has its own don't-clobber-the-author logic in both platforms, so that's its own change.
- A `Title` entry in `packages/html/vjsc/entries.generated.ts`. Every build regenerates that file to add one, but per @mihar-22's note on #2063 this branch leaves all compiler-generated output alone. Nothing imports the entry yet, so the next regeneration can pick it up.

## Funny business
- _Visibility is a composition concern, not the title's._ This started out the other way: #1603 wrote `.media-controls[data-visible]:has([data-paused]) ~ .media-title`, and the first pass here moved that expression into `TitleCore`. Getting `paused` out of a selector was worth doing; bringing `controlsVisible` along was not, since controls are a sibling and the skins already peer off them (`peer-data-visible/controls:` in `overlay.ts`). Per @mihar-22's review the title now knows only whether it has a title, and the skin decides the rest.
- _`metadataFeature` is the only feature it touches._ Not "the only one it needs" with fallbacks for the others — it no longer reads playback or controls at all, so there are no fallbacks and no optional slices.
- _`hidden`, not `hasTitle`._ `hasX` appears nowhere else in `core/ui`; `hidden: boolean` + `data-hidden` is what `thumbnail`, `pip-button`, `cast-button`, and `captions-button` use. Native `hidden` also gets the browser default and the accessibility-tree removal for free, so demos no longer need a `display: none` rule.
- _No `setMedia`._ `TitleCore` holds no state, so `getState(media)` takes the slice as an argument the way `ThumbnailCore` does. 13 of the 33 cores in `core/ui` have no `setMedia`.
- _No ARIA._ #1603 put `aria-label` and a conditional `aria-hidden` on a plain `<span>`. `aria-label` on a role-less element is invalid, most AT ignores it, and it duplicated the visible text. The text node is already in the accessibility tree. The `label` prop existed only to feed that attribute and went with it.
- _The component owns its text._ Set it through the player's `title` prop or `content-title` attribute, not children. A DOM-content override would compete with the chain `internal/design/store/resolved-feature-state.md` defines, and primitives here take no slots.
- _Children stack with the text, and now hide with it._ `<media-title>Hello</media-title>` renders `Hello` followed by the resolved title, because the element appends its text node without clearing what it finds — and with no resolved title the whole element goes `hidden`, `Hello` included. `TimeElement` has the same append behavior, so straightening that out is a change for both. Flagging rather than fixing here since children aren't part of the contract; say the word if you'd rather the element only hide itself when it has nothing of its own to show.

https://claude.ai/code/session_0118ZxrT7hBHhZ57GdgCTvUy

---

> [!NOTE]
> **Low Risk**
> New opt-in UI surface with broad test coverage; no changes to playback, auth, or existing skin bundles unless consumers import the title module.
> 
> **Overview**
> Adds **`TitleCore`**, **`media-title`**, and **`<Title>`** so apps can show the player’s resolved content title from **`metadataFeature`** only—visibility is **`hidden`** / **`data-hidden`** (HTML) or no render (React) when the title is empty, with titles set via player **`contentTitle`** / **`content-title`**, not children.
> 
> Ships opt-in **`@videojs/html/ui/title`** registration, package exports, unit tests, a **Title** reference page with HTML/React demos, and updates the metadata sandbox preview to use **`<Title>`**. **`TimeElement`** gets a small internal tweak (`new Text()` + **`append()`** for its text node).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63e13913320254f9a443142b5e767172c5383e7f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/core/src/core/ui/title/title-core.ts
```diff
@@ -0,0 +1,46 @@
+import type { MediaControlsState, MediaMetadataState, MediaPlaybackState } from '@videojs/media';
+
+/**
+ * Media state the title reads. Controls and playback are optional so a player
```
decepulis: Why are these optional? Would this code be more simple they weren't? Is there precedent for this in the codebase?

packages/html/src/define/video/minimal-ui.ts
decepulis, line 1: all of these definitions should be moved to the PR where the title is actually added to the skins, so they wouldn't clutter this PR.

site/src/content/docs/reference/title.mdx
```diff
@@ -0,0 +1,160 @@
...
+<FrameworkCase frameworks={["html"]}>
+  ```html
+  <media-title></media-title>
+  ```
+</FrameworkCase>
+
+## Behavior
+
+The component displays the title resolved by the metadata feature. It owns its own
+text content — set the title on the player rather than by passing children.
```
decepulis: "resolved by the metadata feature" is hella jargony. Heavily consider the guidance in `writing-style.md`

site/src/content/docs/reference/title.mdx
```diff
@@ -0,0 +1,160 @@
...
+<FrameworkCase frameworks={["html"]}>
+  ```html
+  <video-player content-title="Big Buck Bunny">
+  ```
+</FrameworkCase>
+
+The metadata feature resolves one value from several sources, in order: the title you
+set, the title the media reports, the fallback you set with `defaultContentTitle`, and
+finally an empty string. You can also update it at any time by calling `setContentTitle`
+on the store.
```
decepulis: This paragraph is _so so so_ important. Make sure it communicates with the utmost clarity and minimal jargon

decepulis (commented):
Just some comments for claude

packages/core/src/core/ui/title/title-core.ts
decepulis: Good catch — one of the two was buying nothing, and the whole type was wider than it needed to be. Fixed in 1c48b8d.

**Precedent.** No core in the repo takes a `Partial<FeatureState>`; the only `Partial<>` uses are label overrides. The two shapes that do exist:

- `LiveButtonCore` narrows with `Pick`: `Pick<MediaTimeState, 'currentTime' | 'seek'> & Pick<MediaBufferState, 'seekable'> & MediaLiveState`.
- `TimeSliderCore` requires `MediaTimeState & MediaBufferState`, and `TimeSliderElement` supplies `{ buffered: [], seekable: [] }` when the buffer feature is absent — the fallback lives in the adapter, not the core.

**Playback: made required.** Every preset includes `playbackFeature`, so `Partial` there was dead weight. The adapters now bail and warn like `Poster` does.

**Controls: still optional, but for a concrete reason.** `audioFeatures` and `liveAudioFeatures` leave out `controlsFeature` — audio chrome never auto-hides, so there is no visibility to follow. `controlsVisible ?? true` is what "this player has no hiding controls" means, and the JSDoc now says exactly that instead of hand-waving about features in general.

Net effect on the type:

```ts
export type TitleMediaState = Pick<MediaMetadataState, 'contentTitle'> &
  Pick<MediaPlaybackState, 'paused'> & {
    controlsVisible?: MediaControlsState['controlsVisible'] | undefined;
  };
```

Three fields instead of three whole feature states, one fallback instead of two, and the tests no longer need to mock `setContentTitle`, `requestControlsLock`, and friends just to construct a state.

---

packages/html/src/define/video/minimal-ui.ts
decepulis: Done in 1c48b8d — removed from all six `ui.ts` / `minimal-ui.ts` entries. `define/ui/title.ts` stays, so `<media-title>` is still available as a granular import (that is what the docs demo uses).

Worth noting this also fixes a size regression the bundle bot flagged on this PR: every skin entry does `import './ui'`, so registering the element there grew all thirteen skin entries (`/video/skin` +257 B) for a component none of them render. The stacked skin PR adds the registrations back where they pay for themselves.

---

packages/core/src/core/ui/title/title-core.ts
decepulis: Correcting my reply above — I took "make it required" one step too far and broke a real consumer.

`createPlayer({ features: [metadataFeature] })` has no playback slice, and that is exactly what `apps/sandbox/templates/metadata` composes. With playback required the adapters bailed out, so that template rendered nothing. This PR updates that template, so I broke the thing it was meant to demonstrate.

Fixed in 560b7c3 by pushing your question one level further out. `TitleMediaState` now requires all three fields, and `TitleCore` has no `??` left in it at all:

```ts
export type TitleMediaState = Pick<MediaMetadataState, 'contentTitle'> &
  Pick<MediaPlaybackState, 'paused'> &
  Pick<MediaControlsState, 'controlsVisible'>;
```

The degradation moved out to the adapters, which is where `TimeSliderElement` already keeps it for its optional buffer slice (`...(buffer ?? { buffered: [], seekable: [] })`, `time-slider-element.ts:161`):

```ts
paused: this.#playbackState.value?.paused ?? true,
controlsVisible: this.#controlsState.value?.controlsVisible ?? true,
```

So the answer to your question is yes, and more than my first pass managed: no optionality in the core, no branch to reason about, and a player missing either slice still works — a player with nothing that hides the title simply doesn't hide it. Both reduced compositions now have tests: no `controlsFeature` (the audio presets) and `metadataFeature` alone (the sandbox).

---

site/src/content/docs/reference/title.mdx
decepulis: Rewritten. "Resolved by the metadata feature" is gone, along with "feature", "state", and "precedence" as reader-facing words. The mechanism is now stated as three places a title can come from, in the order the player tries them:

> A title can come from three places. The player uses the first one that has a value:
>
> 1. `contentTitle`, the title you set.
> 2. The title the media reports about itself.
> 3. `defaultContentTitle`, the fallback you set for media that does not name itself.

---

site/src/content/docs/reference/title.mdx
decepulis: Agreed — this is the paragraph people will actually need, so it now states the rule as a rule and then says what it buys the viewer:

> The title shows when three things are true: a title exists, the controls are on screen, and playback is paused. Press play and it hides, so it stays out of the way of the video.
>
> A player that leaves out a feature simply has less to hide behind. Audio players have no auto-hiding controls, so their title follows playback alone; a player with no playback at all shows its title for as long as one resolves.

The second paragraph is new in 560b7c3 and describes real behavior now, not a hypothetical — both compositions are covered by tests.

One nearby claim was also just wrong: the page said an empty title "renders nothing." It doesn't — the element stays in the page with no text. That is the whole reason `data-has-title` exists, so the corrected sentence points at it rather than leaving the reader to discover an invisible empty element sitting on their layout.

---

site/src/content/docs/reference/title.mdx
```diff
@@ -0,0 +1,168 @@
...
+  ```tsx
+  <Player.Provider contentTitle="Big Buck Bunny">
+  ```
+</FrameworkCase>
+
+<FrameworkCase frameworks={["html"]}>
+  ```html
+  <video-player content-title="Big Buck Bunny">
+  ```
+</FrameworkCase>
```
decepulis, line 61: show a full example with both the player and the title inside.

site/src/content/docs/reference/title.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+A title can come from three places. The player uses the first one that has a value:
+
+1. `contentTitle`, the title you set.
+2. The title the media reports about itself.
+3. `defaultContentTitle`, the fallback you set for media that does not name itself.
+
+If none of them has a value, the title is empty. The element stays in the page with no
+text, and `data-has-title` comes off so your CSS can hide it. To change the title while
+the player is running, call `setContentTitle`.
```
decepulis: I'd like a short snippet showing how to call setContentTitle

site/src/content/docs/reference/title.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+The title shows when three things are true: a title exists, the controls are on screen,
+and playback is paused. Press play and it hides, so it stays out of the way of the video.
+
+A player that leaves out a feature simply has less to hide behind. Audio players have no
+auto-hiding controls, so their title follows playback alone; a player with no playback at
+all shows its title for as long as one resolves.
```
decepulis: imo you can drop this paragraph

decepulis (commented):
this is looking pretty good!

site/src/content/docs/reference/title.mdx
decepulis: Done in a4e1f08 — the behavior example is now a whole player with the title inside it, in both frameworks, matching the demo at the bottom of the page.

---

site/src/content/docs/reference/title.mdx
decepulis: Added in a4e1f08, one snippet per framework. React gets the store from `Player.usePlayer()`, HTML reaches it through `player.store`.

---

site/src/content/docs/reference/title.mdx
decepulis: Dropped in a4e1f08.

One thing that goes with it: the page now states the visibility rule as three conditions, and an audio player has no auto-hiding controls, so a reader with one has no way to tell from this page that their title follows playback alone. The behavior is covered by tests either way. Say the word if you want a sentence back for that case.

---

decepulis:
## The title no longer hides on play (4dd03c3)

Raised in review: it's weird that the title disappears separately from the controls. It is, and nothing chose it deliberately.

**Where the rule came from.** #1603, as a skin selector:

```css
.media-default-skin .media-controls[data-visible]:has([data-paused]) ~ .media-title { … }
```

When this PR moved the visibility decision into `TitleCore`, so skins react to `[data-visible]` instead of reaching across the DOM, the `&& paused` came along unexamined.

**Nothing asked for it.** #1123 doesn't specify visibility, and the prior art it cites treats the title as chrome — Media Chrome slots `<media-title>` into a chrome region and it inherits that region's visibility.

**Where they diverged.** `controlsVisible` is already `lock || userActive || paused || casting` (`controls.ts:64`), so the title's rule was the controls' rule plus `&& paused`:

| | controls | title (before) |
|---|---|---|
| paused | shown | shown |
| playing, pointer moving | shown | **hidden** |
| playing, idle 2s | hidden | hidden |
| playing, menu open | shown | **hidden** |
| casting | shown | **hidden** |

**And it left the top buttons without a scrim.** Stacked #2036 moves the top gradient onto the title, per @sampotts's "collapse the two overlay gradients into one" — `.media-overlay` darkens `to top` only. Below `32rem` the default video skin pins `.media-controls--secondary` top-right, so while playing with the controls up those buttons sat on bare video. #1603 had the same gap; moving the gradient onto the title made it load-bearing.

### What changed

```ts
visible: hasTitle && media.controlsVisible,
```

The title doesn't read playback at all now, so `TitleMediaState` drops its `MediaPlaybackState` pick and both adapters drop `selectPlayback` and the `paused ?? true` fallback. `controlsVisible` stays optional for the audio presets.

Two skin comments that described a coupling — "Leaves with the secondary controls, which share the top of the player" and "Slides off the top edge, mirroring the control bar leaving the bottom" — are now accurate rather than aspirational, so #2036's CSS is unchanged.

Tests updated in all three packages. The HTML one asserts the store saw the play before checking the title is still up, so it can't pass on a stale frame.

### One thing to push back on

This puts a sentence about audio players back into the reference page, in the thread where you said to drop the paragraph. Different reason this time: controls are now the only thing that gates the title, and an audio player has no controls feature, so without a sentence a reader with one would expect hiding that never happens.

> The title is part of the player's chrome. It shows whenever a title exists and the controls are on screen, and it leaves when they do, so it never sits over the video on its own. An audio player has no controls that come and go, so its title stays on screen whenever there is one to show.

Cut the last sentence if you still don't want it.

---

site/src/content/docs/reference/title.mdx
decepulis: Answering my own question, and reversing part of this: 4dd03c3 drops the `paused` term from the visibility rule, so a sentence about audio players went back in — one sentence, not the paragraph you cut, and for a different reason.

Controls are now the only thing that gates the title. An audio player has no controls feature, so its title never hides, and the page would otherwise promise hiding that doesn't happen.

Full reasoning in [this comment](https://github.com/videojs/v10/pull/1997#issuecomment-5260072112). Cut the sentence if you disagree.

---

packages/html/src/ui/title/title-element.ts
```diff
@@ -0,0 +1,56 @@
...
+ * children.
+ */
+export class TitleElement extends MediaElement {
+  static readonly tagName = 'media-title';
+
+  readonly #core = new TitleCore();
+  readonly #metadataState = new PlayerController(this, playerContext, selectMetadata);
+  readonly #controlsState = new PlayerController(this, playerContext, selectControls);
+
+  readonly #textNode = document.createTextNode('');
```
luwes:
```suggestion
  readonly #textNode = new Text();
```

code 🏌️‍♂️

decepulis: Taken in d1f4393. ⛳

Swept `TimeElement` with it — it's the only other element that owns a text node, so changing one and not the other would leave two spellings for the same thing. Drop that hunk if you'd rather keep this PR to the title.

---

packages/html/src/ui/title/title-element.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  override connectedCallback(): void {
+    super.connectedCallback();
+
+    if (!this.#textNode.parentNode) {
+      this.appendChild(this.#textNode);
```
luwes:
```suggestion
      this.append(this.#textNode);
```

decepulis: Taken in 2fc2d54.

You're right that this was already the house style — the radio groups and `StatusAnnouncer` all use `this.append(...)`, and `appendChild` on `this` only survived in these two files. `TimeElement` came along again, where its two calls collapse into one variadic append:

```ts
this.append(this.#signSpan, this.#textNode);
```

`ThumbnailElement` still uses `shadow.appendChild` — different receiver, and outside this PR.

---

luwes (approved):
👍

packages/core/src/core/ui/title/title-core.ts
```diff
@@ -0,0 +1,47 @@
...
+  hasTitle: boolean;
+  /** Whether the title should be displayed. */
+  visible: boolean;
+}
+
+export class TitleCore {
+  #media: TitleMediaState | null = null;
+
+  setMedia(media: TitleMediaState): void {
+    this.#media = media;
```
heff: Is this a reference to the actual media, or is this the player store state? Want to make sure we're not using the term "media" too liberally when it has specific meaning in this ecosystem.

decepulis: Neither, strictly. It's the media *state* slices that `@videojs/media` itself defines — `MediaMetadataState`, `MediaControlsState` and friends from `packages/media/src/core/state.ts`. Not a `Media` (`packages/media/src/core/types.ts:543`), and not the store either: the adapters select those slices out of the player store and hand the core a plain object.

The naming isn't mine, though. It's the `core/ui` contract:

```ts
/** Constraint for core UI classes that derive component state from media state. */
export interface MediaUIComponent<Props = object, State extends object = object> extends UIComponent<Props, State> {
  setMedia(media: object): void;
}

/** Extracts the media state parameter type from a core's `setMedia` method. */
export type InferMediaState<Core extends MediaUIComponent> = Parameters<Core['setMedia']>[0];
```

20 cores implement it with the same `#media` field and `setMedia` setter. `ControlsCore` and `PosterCore` are the same shape as this one down to the line breaks. So `TitleCore` reads the way its 19 siblings do, and `TitleMediaState` follows `LiveButtonMediaState`, the one other core that names a composite of picks.

That said, your instinct isn't wrong — `setMedia(media)` taking a bag of state rather than a `Media` is genuinely a bit of a false friend, and I had to trace it myself when I started. But it's a `core/ui`-wide rename: 20 cores, the `MediaUIComponent` constraint, `InferMediaState`, and both adapter layers. Too big to smuggle into a title PR, and it'd conflict with anything else in flight.

Happy to open an issue for it if you think it's worth doing. Or if you'd rather this one component read differently now, say so and I'll change it — it'd just be the odd one out until the rest follows.

---

packages/core/src/core/ui/title/title-core.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
```
mihar-22: seems unnecessary and not a great pattern or name, if we do it elsewhere then we can address later in refactors (noted), i think for now just pass the TitleMediaState into `getState` which can then derive the `TitleState`

packages/core/src/core/ui/title/title-data-attrs.ts
```diff
@@ -0,0 +1,9 @@
+import type { StateAttrMap } from '../types';
+import type { TitleState } from './title-core';
+
+export const TitleDataAttrs = {
+  /** Present when a title is available to display. */
+  hasTitle: 'data-has-title',
+  /** Present when the title should be displayed. */
+  visible: 'data-visible',
```
mihar-22: We don’t typically model component availability as `hasX`, so I don’t think `hasTitle` belongs in state or the data attributes.

`visible` also isn’t the right state here. More importantly, `Title` shouldn’t derive its presentation from controls visibility - that belongs to the Controls or skin composition.

If we want to expose title availability, `hidden` follows our existing convention: HTML can apply native `hidden` and `data-hidden`, while React can return `null` when no title and children are available. 

Consumers could technically use `:empty` (text content would be empty), but explicit `hidden` behavior gives the browser the correct default and provides a familiar styling hook at very little cost.

packages/core/src/core/ui/title/title-core.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Media state the title reads, composed by the HTML and React `Title` adapters
+ * from the `metadata` and `controls` store slices.
+ *
+ * Every field is required. A player can leave out `controlsFeature` — the audio
+ * presets do — so the adapters substitute the neutral value that means "nothing
+ * here hides the title".
+ */
+export type TitleMediaState = Pick<MediaMetadataState, 'contentTitle'> & Pick<MediaControlsState, 'controlsVisible'>;
```
mihar-22: How is controls visibility relevant to the title itself? If the title should appear and disappear with the controls, that behavior should be owned by the `Controls` component or the surrounding UI composition.

packages/html/src/ui/title/title-element.ts
```diff
@@ -0,0 +1,56 @@
...
+ * The element owns its text content. Set the title through the player's
+ * `content-title` attribute or `setContentTitle` rather than by writing
+ * children.
+ */
+export class TitleElement extends MediaElement {
+  static readonly tagName = 'media-title';
+
+  readonly #core = new TitleCore();
+  readonly #metadataState = new PlayerController(this, playerContext, selectMetadata);
+  readonly #controlsState = new PlayerController(this, playerContext, selectControls);
```
mihar-22: as per core comments, wouldn't expect this here

packages/core/src/core/ui/title/title-core.ts
mihar-22: it can probably just accept `MediaMetadataState` to not create a new type alias

mihar-22 (changes requested):
Thanks for doing this!! Requesting some small changes 🙏

packages/core/src/core/ui/title/title-core.ts
mihar-22: I'd leave this as a composition-concern.

decepulis: Reasonable! I have no clue what I was thinking here.

packages/html/src/ui/title/title-element.ts
decepulis: This felt like such a simple component. I told AI "look at other components that read the store and display it and do it that way" and wow did it go off the rails. Nice reminder that AI is not a substitute for understanding things, myself

packages/core/src/core/ui/title/title-data-attrs.ts
decepulis: All of this taken in fa21107. `TitleState` is now just `title` and `hidden`, and the whole review nets out at −119 lines.

Checked both naming claims before changing them, and you're right on both counts. `hasX` appears zero times in `core/ui`. `hidden: boolean` + `data-hidden` appears in `thumbnail`, `pip-button`, `cast-button`, and `captions-button`, with the buttons also emitting the native attribute via `getAttrs`. So this was the odd one out, not a new convention.

- HTML sets `this.hidden = state.hidden` alongside `data-hidden`. The demo stylesheet dropped its `:not([data-has-title]) { display: none }` rule entirely — the UA sheet already does it, and nothing in that file sets `display`, so the gradient stops sitting over empty space for free.
- React returns `null` when no title resolves, matching `Poster` and the ~20 other components that early-return.

I left `Omit<UIComponentProps, 'children'>` alone rather than accepting children as a fallback, since `Time.Value` omits children the same way and the component owning its own text is the documented contract. Happy to revisit if you meant that as part of the ask rather than a description of the shape.

---

packages/core/src/core/ui/title/title-core.ts
decepulis: Both done in fa21107.

```ts
export class TitleCore {
  getState(media: MediaMetadataState): TitleState {
    const title = media.contentTitle;

    return { title, hidden: title.length === 0 };
  }
}
```

I was going to push back on dropping `setMedia`, on the grounds that it would leave this the only core without one. That turned out to be wrong, so recording the correction rather than the objection: `core/ui` has 33 cores and 13 of them have no `setMedia`. `ThumbnailCore` is the same shape as this now — stateless, inputs as `getState` arguments, `hidden` in the returned state. So this isn't a deviation waiting on the sweep you noted; it's the pattern that already fits a core with nothing to hold onto.

Dropping the alias also cleans up the test, which now builds a real `MediaMetadataState` with `vi.fn()` setters exactly like `PosterCore`'s does, instead of a hand-rolled composite.

---

packages/html/src/ui/title/title-element.ts
decepulis: Controller's gone in fa21107, along with the controls slice in the core and the `controlsVisible` fallback in both adapters. The title now reads `metadata` and nothing else.

Tracing where the coupling came from, since it's worth recording: it wasn't invented here. #1603 expressed it as a skin selector — `.media-controls[data-visible]:has([data-paused]) ~ .media-title` — and this PR moved that expression into TypeScript so skins could key off one attribute instead of reaching across the DOM for playback state. That argument holds for `paused`, which really was awkward in a selector. It does not hold for `controlsVisible`, which is a sibling relationship the skins already express directly (`peer-data-visible/controls:opacity-100` in `overlay.ts`). I ported both halves when only one had a reason.

**Heads-up for #2036, which now needs a matching change:** it styles the title off `data-visible`, and that attribute no longer exists. Four skin modules move to `peer-data-visible/controls:` in Tailwind and `media-controls[data-visible] ~ media-title` in CSS — the same peer/sibling pattern `overlay` already uses. cc @luwes since that PR is yours now. The docs page here shows the sibling form under Styling if it's useful as a reference.

---

packages/core/src/core/ui/title/title-core.ts
decepulis: Done in fa21107 — the controls slice is gone from the core and from both adapters, and the title reads `metadata` and nothing else.

On the two owners you offered, it has to be the surrounding composition rather than `Controls`, and for a more concrete reason than I'd appreciated: `Controls.Root` has no box to nest into. Both skins set it to `display: contents` (`video.css:188`, `video.tailwind.ts:126`) and let its two groups position themselves to opposite edges — `--primary` at `bottom: var(--inset)`, `--secondary` at `top`/`right`. Only past 32rem does the root become the positioned flex bar, and then the groups flip to `contents` instead. A nested title would lay out against the skin root either way, so the nesting would buy nothing.

It wouldn't be free, either. Inside `ControlsElement` a title also inherits `#closeOwnedOverlays()`, which closes every `POPUP_HOST_SELECTOR` descendant when the controls hide; `.media-error[data-open] ~ .media-controls` (`video.css:321`), which would take the title down with the bar; `@container/media-controls` for any container query it wants; and `data-interactive`.

The composition already has the shape for this. `.media-overlay` is a sibling of `Controls.Root` that fades with it through `peer-data-visible/controls:` (`overlay.ts:16`), and the title's relationship to controls is the same one — shares their rhythm, isn't one of them. Which is the version of your point I missed on the first pass: `controlsVisible` in `TitleCore` made the title a control in the type signature while the DOM never treated it as one.

One ordering note for wiring this up in the skins: `peer` compiles to `~`, so the title has to come after `Controls.Root` in the markup, the way `.media-overlay` does. If it needs to come earlier for reading order, that's a `:has()` on the skin root instead.

---

packages/core/src/core/ui/title/title-core.ts
decepulis: Updating this, since the code moved out from under my answer and one of my numbers was wrong.

Per @mihar-22's review, `TitleCore` no longer has `setMedia`, a `#media` field, or a `TitleMediaState` alias. It's now `getState(media: MediaMetadataState)` and nothing else. So the "it reads the way its siblings do" defense doesn't apply to this file anymore — and the count I gave was off: `core/ui` has 33 cores, 13 of which have no `setMedia`. `ThumbnailCore` is the shape this one follows now.

The substance of the answer holds: `media` here is the state slice `@videojs/media` defines (`MediaMetadataState`, `packages/media/src/core/state.ts`), not a `Media` and not the store. But your point lands harder than it did before. What's left in this file is a parameter named `media` whose type is a `Media*State`, which is the same false friend in miniature — and it's now the only thing carrying that name here, rather than one instance of a 20-file pattern.

Still inclined to leave the sweep out of this PR, but the naming question is a live one rather than something the convention answers. Happy to open an issue for it.

---

mihar-22 (approved):

decepulis:
Merged main and pushed one fix on top of the approvals — flagging it so the new commit doesn't need re-reviewing as a behavior change.

Main had moved 59 commits since CI last ran here, and one of them changed `createPlayer` to return `{ Player, usePlayer, useMedia }` instead of a namespace with `Provider` and `Container`. Nothing conflicted textually, and GitHub reported the PR mergeable, but the title reference page rendered `Player.Provider` as `undefined` and failed the site build. The approved commit would have gone red on merge.

`c2050d9` migrates the two files main couldn't migrate for me, because they only exist on this branch:

- the React demo — `<Player>` with a separate `<Container>`
- the `setContentTitle` snippet — bare `usePlayer()`

The metadata sandbox template merged cleanly on its own; main had already migrated it and git combined that with the `<Title>` changes here.

Verified locally on the merge: `build`, `typecheck`, `lint`, `check:workspace`, site build, and the full suites for the three touched packages — core 1437, html 352, react 458, including all 17 title tests.

One incidental note, since it's a small vindication of the review: main independently added `hidden: boolean` to `RadioOptionsState` in `core/ui/types.ts` while this sat, with nearly the same JSDoc as the one here. The convention @mihar-22 pointed at is spreading on its own.

---

PR #1998, opened 2026-08-06
feat(packages): make Mux and Vimeo content-data donors

### LINKED ISSUE #1943: Align Existing Media with the Content Data API

decepulis:
Closes #1943.

## What is this?
1. Ensure Mux and Vimeo elements consistently donate `contentData` and emit `contentdatachange` events
2. Give `MediaContentData` named keys for the vocabulary player features read

This is the media half of the contract only. The store half — resolving a `poster` and a `posterPlaceholder` out of the bag, and the setters behind them — moved out to the branches that build the consumers, so each resolved value lands with the component that reads it: poster in #2039, poster placeholder in #2063.

## What this isn't
- The store fields. See above.
- `posterPlaceholder` as a bag key — it left with the store half, since no media donates one and the key only means something once the store resolves it.
- I didn't add `storyboard` to the store since it feels like our timed metadata story is still in flight
- All the other work scoped in #1798

## Funny business
- *`mux/media` now requires `#refreshContentData` to be called when `contentData` is changed* so that we can check if the object changed before emitting an event. This is the same thing that `streamType` and `targetLiveWindow` do in order to keep noise down.
- `vimeo/media` does something similar, but it's less complicated because we're just looking at `#title` there.
- *The SPF-backed Mux Medias needed the same treatment.* #2045 landed them on main with a `contentData` that re-derived on every read and announced nothing, which would have left the store stale on a source change. `MuxMediaMixin` now caches and dispatches like its hls.js counterpart, and one mixin backs both the video and audio flavors.

https://claude.ai/code/session_01GsEG638e7EhTTtBLtMgKUi

---

> [!NOTE]
> **Medium Risk**
> Behavioral contract changes for metadata consumers (`contentdatachange` timing and reference stability); limited to Mux/Vimeo media paths with broad test coverage.
> 
> **Overview**
> **Media implementations now donate `contentData` and fire `contentdatachange` only when the bag actually changes**, so upstream store/features can stay in sync without polling.
> 
> `MediaContentData` is now an interface with documented shared keys (`title`, `poster`, `storyboard`) and guidance on omitting keys vs using `''` to block fallbacks.
> 
> **Mux (hls.js `MuxMedia` and SPF `MuxMediaMixin`)** cache derived poster/storyboard URLs, refresh before `sourcechange`, dedupe events when playback params change but image URLs do not, and clear the bag when `source` is dropped.
> 
> **Vimeo** keeps a cached `contentData` bag, sets `title` after embed load, clears on reset/source change with `contentdatachange` ordered after other reset fields, treats blank titles as absent keys, and dedupes repeat loads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26b0bb7b29835aeb42057e0d30be4bf08efd3f39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


decepulis:
## Handoff: one commit for this PR to absorb

`posterPlaceholder` was sitting in #2039, but it's this PR's kind of change. It's ready to cherry-pick:

```
git fetch origin claude/poster-picture-model
git cherry-pick 151a69f2974a1b0e1296a9d54ddedb69fbbc6d62
```

It's the first commit on that branch and applies cleanly to `0414637`, which is where it was written.

### What it is

`feat(packages): resolve a poster placeholder alongside the poster` — 4 files, +91/−2:

- `packages/core/src/dom/store/features/metadata.ts` — `posterPlaceholder` and `defaultPosterPlaceholder` on the same four tiers as `poster`, with `MediaContentData.posterPlaceholder` feeding the media tier
- `packages/core/src/dom/store/features/tests/metadata.test.ts` — two cases: resolution order, and that it resolves independently of the poster
- `packages/media/src/core/state.ts`, `types.ts` — the two type members

It touches only files this PR already touches, and it's additive: nothing reads the resolved value.

### Why it belongs here rather than in #2039

#2039, #2009 and #2010 each added this field, and the three `metadata.ts` results are the *same file* — git gives all three the content hash `e031fb5`. Across three competing designs for the poster component, the store layer was identical. So it isn't a design choice under review; it's plumbing that any of them needed, which makes it this PR's business.

Moving it also means #2039's diff is entirely about the component, which is what it's actually asking to be reviewed on.

### One thing deliberately left out

The JSDoc on those two type members is neutral about what the value holds. The decision that it's a **CSS image rather than a URL** stays in #2039 as a separate commit, because that follows from how that PR exposes the value. If a different design lands instead, the wording should follow it — so please don't tighten those two doc comments here.

### Checks run on the commit alone

- `pnpm -F @videojs/media build`, `pnpm -F @videojs/core build` — clean
- `pnpm -F @videojs/core test src/dom/store/features/tests/metadata.test.ts` — 13 passed
- `pnpm -F @videojs/media test` — 539 passed, 30 files

Once it's pushed here I'll rebase #2039 onto this branch; the copy there drops out on its own, since the diffs match.

Worth a line in this PR's "What is this?" — the body currently says storyboard was left out, and it's now poster plus placeholder that go in.

---

packages/core/src/dom/store/features/metadata.ts
```diff
@@ -7,56 +7,103 @@ import type { PlayerFeatureConfig } from '../../player';
...
 const USER_DEFAULT_CONTENT_TITLE = Symbol('@videojs/user-default-content-title');
-const SET_USER_CONTENT_TITLE = Symbol('@videojs/set-user-content-title');
-const SET_USER_DEFAULT_CONTENT_TITLE = Symbol('@videojs/set-user-default-content-title');
 const DEFAULT_CONTENT_TITLE = '';
 
-interface MetadataSourceState extends Omit<MediaMetadataState, 'contentTitle'> {
+const MEDIA_POSTER = Symbol('@videojs/media-poster');
+const USER_POSTER = Symbol('@videojs/user-poster');
+const USER_DEFAULT_POSTER = Symbol('@videojs/user-default-poster');
+const DEFAULT_POSTER = '';
```
luwes: non-blocking: in html `` does try to load something, argument for using `undefined` or `null`

packages/core/src/dom/store/features/metadata.ts
```diff
@@ -7,56 +7,103 @@ import type { PlayerFeatureConfig } from '../../player';
...
+
+const MEDIA_POSTER_PLACEHOLDER = Symbol('@videojs/media-poster-placeholder');
+const USER_POSTER_PLACEHOLDER = Symbol('@videojs/user-poster-placeholder');
+const USER_DEFAULT_POSTER_PLACEHOLDER = Symbol('@videojs/user-default-poster-placeholder');
+const DEFAULT_POSTER_PLACEHOLDER = '';
```
luwes: same here

packages/core/src/dom/store/features/metadata.ts
decepulis: Right about the mechanism. Covering both this and line 20, since they're the same question.

Where it lands today: nothing consumes either resolved value yet, so neither default reaches an `img`.

- `PosterCore.PosterState` carries no URL — it only wraps `MediaPlaybackState`
- React `Poster` reads `src` from author props (`poster.tsx:40`), not from the store
- HTML `media-poster` takes a `placeholdersrc` attribute and already guards on truthiness before setting `--media-poster-placeholder`, so `''` there means "no placeholder" rather than an empty `url()`

So `''` is inert at the moment, and the hazard starts the instant something writes ``'&lt;img src={poster}&gt;'``. That wiring is #2039's — it's building the component that consumes these, and the store side only just moved there from here.

Which is why I'd rather the call be made there than here: whoever writes the consumer knows whether they want to branch on falsiness or rely on the attribute simply being absent. Both shapes work — keep `''` and have the consumer omit the attribute when falsy, which is what `media-poster` already does, or make the default `undefined` so absence is structural and can't be got wrong downstream.

Leaving as-is for now, since it's non-blocking and `''` matches `DEFAULT_CONTENT_TITLE` two declarations up. Switching is cheap if you'd prefer the store foreclose it: widen `MediaMetadataState.poster` and `posterPlaceholder` to include `null` or `undefined`, plus a handful of `toBe('')` assertions. No structural change, so it's an easy follow-up in whichever PR ends up owning the consumer.

---

luwes (approved):
LGTM

packages/core/src/dom/store/features/metadata.ts
mihar-22: I don't think we should rush into the concept of a "default poster". Doesn't feel natural, not sure where that would come up atm. If you have a desire for it then feel free to share :)

packages/media/src/core/state.ts
```diff
@@ -141,10 +141,22 @@ export interface MediaStreamTypeState {
...
+  /** Set the fallback used when neither the user nor media supplies a title, or clear it. */
+  setDefaultContentTitle(value: string | null | undefined): void;
+  /** Set the user poster override, or clear it with `null` or `undefined`. */
+  setPoster(value: string | null | undefined): void;
+  /** Set the fallback used when neither the user nor media supplies a poster, or clear it. */
+  setDefaultPoster(value: string | null | undefined): void;
+  /** Set the user poster placeholder override, or clear it with `null` or `undefined`. */
+  setPosterPlaceholder(value: string | null | undefined): void;
+  /** Set the fallback used when neither the user nor media supplies a poster placeholder, or clear it. */
+  setDefaultPosterPlaceholder(value: string | null | undefined): void;
```
mihar-22: I get it but exposing these hooks scares me a little because it opens the door two way data binding - bugs

mihar-22: Wonder if we just leave this out for the time being, including title-related

packages/media/src/core/types.ts
```diff
@@ -508,15 +508,45 @@ export interface MediaPosterCapability {
...
+ * an empty string is a deliberate blank that stops a feature's fallback chain,
+ * so reporting `''` for "not loaded yet" suppresses the author's fallback.
+ */
+export interface MediaContentData {
+  /** Title of the content. */
+  readonly title?: MediaContentValue;
+  /** URL of a still image representing the content. */
+  readonly poster?: MediaContentValue;
+  /** A low-resolution stand-in shown while the poster loads. */
+  readonly posterPlaceholder?: MediaContentValue;
```
mihar-22: Probably the biggest question I have is whether poster placeholder is a the right concept inside our player? Wondering if this is just a application-level concern and we shouldn't do this. Thoughts?

mihar-22 (changes requested):
I'm going to temporarily block this just until we can talk over it tomorrow more. [Related to this comment](https://github.com/videojs/v10/pull/1998/changes#r3770057002). Sorry to be a pain, just want to run over this IRL again. Could very well be the right idea here, just once over together would be great 😄

packages/core/src/dom/store/features/metadata.ts
decepulis: Do you have the same hesitations with defaultTitle?

packages/media/src/core/state.ts
decepulis: I hear those concerns. But also, I feel like we covered this at the onsite.

At the onsite, we decided to include this feature, and if someone's controlling state from the outside _and_ setting state from the inside, it's last-write-wins and the author is at fault. Idk if that's the main class of bug you're worried about but it is for me.

packages/media/src/core/types.ts
decepulis: I thought we talked about this in person. I thought I convinced you that it should be a player-level concept since we need some way for media to pass placeholders through the store to skins

decepulis: But if that's not ringing a bell and you're still feeling unsure, let's re-open tomorrow at topics

mihar-22: Let's just go over it one more time tomorrow 🙏

packages/media/src/dom/mux/media.ts
```diff
@@ -108,18 +118,29 @@ export class MuxMedia extends HlsJsMedia implements MuxMediaProps {
...
+
+    const next: MuxContentData = {
       ...(poster && { poster }),
       ...(storyboard && { storyboard }),
     };
+
+    if (shallowEqual(this.#contentData, next)) return false;
+
+    this.#contentData = next;
+    return true;
```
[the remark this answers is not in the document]
decepulis: Correct, and fixed in 0aeb248a.

The divergence appeared when #2045 landed on main: it added the SPF-backed Mux Medias with their own `contentData`, re-derived on read, while this branch was upgrading the hls.js one to a cached bag plus `contentdatachange`. Merging main brought the two side by side, and `metadataFeature` syncs on that event alone — so SPF Mux would have donated a poster the store never picked up.

`MuxMediaMixin` now caches behind the same `shallowEqual` guard and dispatches the event when the derived URLs move, refreshed before `sourcechange` so listeners reading `contentData` from that event see the new values. One mixin backs both flavors, so `MuxVideoMedia` and `MuxAudioMedia` get it together — covered by four new tests on the video flavor and one on the audio flavor.

---

mihar-22 (approved):

PR #1999, opened 2026-08-06
feat(packages)!: configure orientation lock through providers

### LINKED ISSUE #1942: Move Orientation Lock to Provider Configuration

decepulis:
Closes #1942.

Last in the stack: #2039 → #2063 → #2000 → this. Merge those first; the base is set to #2000's branch so this diff shows only what's new here.

Moves the orientation lock type from a feature-factory argument to declared provider configuration, and deletes the factory-configured feature form now that its only caller is gone.

## Docs preview

- [Orientation lock](https://deploy-preview-1999--vjs10-site.netlify.app/docs/framework/html/reference/feature-orientation-lock)
- [Fullscreen](https://deploy-preview-1999--vjs10-site.netlify.app/docs/framework/html/reference/feature-fullscreen)

## What is this?

1. `orientationLockType` becomes provider configuration — an attribute in HTML, a prop in React — and can change while the player runs
2. `features.orientationLock({ type })` is gone, along with `ConfigurablePlayerFeature`, `ConfigurablePlayerFeatureConfig`, `OrientationLockFeatureConfig`, `forConfig`, `definePlayerSlice`, and `definePlayerFeature`'s two-argument overload
3. `selectOrientationLock` joins the other feature selectors, declared beside its feature so it stays out of bundles that never select it
4. `createScreenOrientationLock` takes the type per `lock()` call instead of capturing it at construction, and serializes platform requests so the screen cannot desync from the store
5. `ScreenOrientationLockType` is exported, having previously been only structurally reachable
6. A config input may be typed by a union narrower than plain text, so the reference page's Configuration table shows the real orientation types and their `landscape` default

## The API

Select the feature as a value, not a call:

```diff
- features: [...videoFeatures, features.orientationLock({ type: 'portrait' })]
+ features: [...videoFeatures, features.orientationLock]
```

Then set the type on the provider — `orientationLockType="portrait"` as a prop on `Player` in React, `orientation-lock-type="portrait"` as an attribute on the player element in HTML. Clearing the value restores `landscape`. Read it with `selectOrientationLock`, or set it imperatively with `setOrientationLockType`.

The PR title carries the `!` so the squash-merged commit marks the break for release-please.

## What this isn't

- **No feature-owned HTML conversion metadata.** The issue sketched `String` / `Boolean` / `Number` sentinels and a non-string fixture. Orientation lock is string-valued and needs none of it; `Boolean` is unsound under the current model, because a removed boolean attribute converts to `false` rather than absent (`reactive-element.ts:186`), so a `Boolean` key defaulting to `true` could never be cleared through markup; and once the sentinel is checked against the action's parameter type it only restates what the compiler already knows. Worth designing against a feature that actually has a non-string value.
- **No validation.** A typo'd `orientation-lock-type="banana"` reaches `screen.orientation.lock()`, which rejects, and the existing `catch` swallows it — so it silently does not lock rather than falling back to the default. Left out to keep the config API unchanged. The empty string is the exception, since a valueless attribute is ordinary markup rather than a typo; it restores the default.
- **No per-adapter attribute, property, and prop coverage.** `contentTitle` already proves that plumbing, so each adapter gets one wiring test.

## Funny business

- *`ConfigActionKey` required actions to be mutually assignable with `string | null | undefined`,* which rejects `ScreenOrientationLockType | null | undefined`. The reverse clause is dropped so a narrower union keeps its own type on the provider input. Actions must still accept `null | undefined` so providers can clear.

  The builder's mock of that constraint has to move with it. It exists to reject what production rejects, so left stricter it would reject what production now accepts — the fixture declares a union-valued input, which fails its own type check if the constraint drifts back. Two other steps still assumed a private symbol: an input's initial value was read only from a computed key in `state()`, so a published key produced no default. #2000's string-name matching covers the rest, and its checker-based lookup also resolves an inherited action, which is why the syntactic version this branch first wrote is gone.

- *`selectOrientationLock` is declared beside its feature instead of in `selectors.ts`.* `createSelector` evaluates its slice at module load and every selector in that module shares one evaluation, so an entry there retains this feature in every bundle importing any selector. The other features ship in presets, so their cost is already paid; this one is in no preset, and an entry there measured **+290 B per `@videojs/html` UI component and +298 B per React one**. Declared beside the feature, 9 of 291 entries move, all inside compressor noise except `@videojs/core/dom` — the aggregate entry that includes everything — at +34 B. A `/* @__PURE__ */` annotation on the `selectors.ts` form was measured too and does not fix it: 76 entries still grow, by up to 116 B.

  The consequence is that `api-docs-builder` scans `selectors.ts` as a leaf entry point, so a selector declared outside it silently gets no reference page. The feature module is added to `UTIL_ENTRY_POINTS`, which yields `select-orientation-lock.json` and no stray pages. The placement rationale is a plain comment rather than JSDoc, since the extractor publishes JSDoc verbatim to the reference page.

- *Making the type responsive exposed three ways the lock could desync from the screen,* all from one variable standing for both "last requested" and "type we hold" (fixed in d945a1f):
  - A rejected re-lock was recorded as held. After landscape succeeded and a portrait request rejected, the guard suppressed every later portrait request — the screen stayed landscape permanently while the store and the lock both reported portrait.
  - Overlapping requests hit the same problem from the other side. The platform does not guarantee settle order, so whichever request landed last won the screen while the bookkeeping described the other.
  - Where the platform rejects — desktop Chrome, outside mobile form factors — nothing ever converged, so every published store change re-issued a rejected `screen.orientation.lock()`. The store publishes at several hertz during playback.

  The primitive now tracks `desired` and `held` separately and reconciles between them one request at a time. A re-entrant call records the new type and returns; the running pass re-reads `desired` before it exits, so the last request wins and platform requests never overlap. A rejection leaves `held` alone, which keeps it describing the platform and lets a later request for the same type through.

- *`attach` compares a resolved value rather than detecting a fullscreen edge.* Edge detection would not have fired on a config change during steady-state fullscreen, so one `sync` now serves fullscreen events and config changes alike. Comparing the resolved `(fullscreen, type)` value is also what keeps the whole-store subscription from re-requesting a rejected lock.

- *A single-owner value does not need a private symbol.* Nothing in media donates an orientation preference, so there is nothing to resolve: `orientationLockType` is a published source key defaulting to `landscape`, with `setOrientationLockType` restoring that default for absent input — no symbol and no `derived` formula, unlike `contentTitle`, which needs both because three owners compete for it. Publishing the key is also what makes it observable, since `publish` copies string keys only (`store.ts:132`) and a symbol-keyed change notifies nobody — and this behavior reacts to its own configuration. `internal/design/store/resolved-feature-state.md` previously stated config inputs are always symbol-keyed; it now records this case.

- *The HTML docs example was inert.* It set `orientation-lock-type` on a player whose `videoFeatures` preset does not include orientation lock, contradicting the same page's note that the feature is not in the default presets. It now defines an element that selects the feature. The fullscreen page's orientation section keeps a code example, updated from the deleted factory call to the provider input, and framework-gated — the sample it replaced was React-only code shown to HTML readers too.

- *The examples were rewritten again for the `createPlayer` change on main.* `createPlayer` no longer returns `Container`, and the compound `Player.Provider` / `Player.Container` form is gone, so the React samples destructure `Player` and nest a separately imported `Container`. The HTML samples drop `slot="media"`, which the container reference now states is unnecessary.

## Testing

Verified on the merged stack. `pnpm typecheck`, `pnpm lint`, `pnpm check:workspace` (10/10), and `pnpm -F site astro check` (0 errors, 0 warnings) pass. Package suites pass individually — core 1439, html 352, react 422, store 124, and site 551, which includes the api-docs-builder suite.

Coverage added for initial config, updates, clearing to the default, detach preservation, post-detach quiescence, re-lock during fullscreen, selector identity and shape, and one wiring test per adapter. The builder's `poster` fixture covers the two steps a published config key still has to carry: its initial value, and a value type narrower than plain text.

The lock state machine is pinned by exact platform call counts rather than by observing a later `unlock()` — the earlier supersede test passed under a mutation of the branch it claimed to cover. Each new guard was mutation-checked: dropping serialization, recording a rejection as held, `??` for `||`, and dropping the `sync` comparison each fail at least one test.

Also fixes cross-test leakage in `orientation-lock.test.ts`, where stores were never destroyed, so their `document` fullscreen listeners stayed live and reacted to later tests' dispatches.

https://claude.ai/code/session_01DdtP5uKxwgyhrDLEg5T6in

---

> [!NOTE]
> **Medium Risk**
> Breaking public API for orientation lock plus a rewritten lock state machine that talks to `screen.orientation`. Wrong reconciliation could leave the device locked or spam rejected lock() calls during playback.
> 
> **Overview**
> Orientation lock type is now live provider configuration (`orientationLockType` / `orientation-lock-type`) instead of a one-shot `features.orientationLock({ type })` factory argument. Clearing the value restores `landscape`. The type can change while fullscreen is already active.
> 
> Removes the factory-configured feature form (`ConfigurablePlayerFeature` and `definePlayerFeature`'s two-argument overload). Config actions may now be a narrower string union plus `null | undefined`, so the provider input keeps the Screen Orientation enum.
> 
> `createScreenOrientationLock` takes the type on each `lock()` and serializes platform requests (`desired` vs `held`) so overlapping or rejected locks cannot desync the screen. `selectOrientationLock` is declared beside the feature to avoid pulling it into every selector bundle.
> 
> Docs and adapters (React prop, HTML attribute) are updated for the new API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0538d98875787a53e7ccf22da430023cc2afe0e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/dom/feature.ts
+++ b/packages/core/src/dom/feature.ts
@@ -1,24 +1,5 @@
-import {
-  type AttachContext,
-  type DerivedContext,
-  defineSlice,
-  type SliceConfig,
-  type StateContext,
-} from '@videojs/store';
-import { isUndefined } from '@videojs/utils/predicate';
+import type { DerivedContext, SliceConfig } from '@videojs/store';
 import type { AnyPlayerFeature, PlayerFeature, PlayerFeatureConfig, PlayerTarget } from './player';
-
-export interface ConfigurablePlayerFeature<Config, State> extends PlayerFeature<State> {
-  (config?: Config): PlayerFeature<State>;
-}
-
-export interface ConfigurablePlayerFeatureConfig<Config, State>
-  extends Omit<SliceConfig<PlayerTarget, State>, 'attach' | 'derived' | 'preserve' | 'state'> {
-  state: (ctx: StateContext<PlayerTarget>, config: Config) => State;
-  attach?: (ctx: AttachContext<PlayerTarget, State>, config: Config) => void;
-}
-
-const definePlayerSlice = defineSlice<PlayerTarget>();
 
 type DerivedFunctions<State> = Record<string, (ctx: DerivedContext<State>) => unknown>;
 
@@ -26,69 +7,36 @@ type DerivedValues<Definitions extends Record<string, (...args: any[]) => unknow
   [Key in keyof Definitions]: ReturnType<Definitions[Key]>;
 };
 
-type StaticPlayerFeatureConfig<State, Derived, Config extends PlayerFeatureConfig> = Omit<
+type PlayerFeatureDefinition<State, Derived, Config extends PlayerFeatureConfig> = Omit<
   SliceConfig<PlayerTarget, State, Derived>,
   'preserve'
 > & {
   config?: Config;
 };
 
-/** Define a static player feature with derived state and optional configuration inputs. */
+/** Define a player feature with derived state and optional configuration inputs. */
 export function definePlayerFeature<
   State,
   const Definitions extends DerivedFunctions<State>,
   const Config extends PlayerFeatureConfig = Record<never, never>,
 >(
-  definition: Omit<StaticPlayerFeatureConfig<State, DerivedValues<Definitions>, Config>, 'derived'> & {
+  definition: Omit<PlayerFeatureDefinition<State, DerivedValues<Definitions>, Config>, 'derived'> & {
     derived: Definitions;
   }
 ): PlayerFeature<State, DerivedValues<Definitions>, Config>;
-/** Define a static player feature with optional configuration inputs. */
+/** Define a player feature with optional configuration inputs. */
 export function definePlayerFeature<State, const Config extends PlayerFeatureConfig = Record<never, never>>(
-  definition: Omit<StaticPlayerFeatureConfig<State, object, Config>, 'derived'> & { derived?: never }
+  definition: Omit<PlayerFeatureDefinition<State, object, Config>, 'derived'> & { derived?: never }
 ): PlayerFeature<State, object, Config>;
-/**
- * Define the legacy factory-configured form used by static feature variants.
- * Currently used for orientation lock and scheduled for removal in #1942.
- */
-export function definePlayerFeature<Config, State>(
-  definition: ConfigurablePlayerFeatureConfig<Config, State>,
-  defaultConfig: Config
-): ConfigurablePlayerFeature<Config, State>;
-export function definePlayerFeature<Config, State>(
-  definition:
-    | StaticPlayerFeatureConfig<State, object, PlayerFeatureConfig>
-    | ConfigurablePlayerFeatureConfig<Config, State>,
-  defaultConfig?: Config
-): PlayerFeature<State> | ConfigurablePlayerFeature<Config, State> {
-  if (arguments.length === 1) {
-    const feature = definition as StaticPlayerFeatureConfig<State, object, PlayerFeatureConfig>;
-    const preserved = Object.values(feature.config ?? {}).map((entry) => entry.state);
+export function definePlayerFeature<State>(
+  definition: PlayerFeatureDefinition<State, object, PlayerFeatureConfig>
+): PlayerFeature<State> {
+  const preserved = Object.values(definition.config ?? {}).map((entry) => entry.state);
 
-    return {
-      ...feature,
-      ...(preserved.length > 0 ? { preserve: preserved } : {}),
-    } as PlayerFeature<State>;
-  }
-
-  const { name, state, attach } = definition as ConfigurablePlayerFeatureConfig<Config, State>;
-
-  const forConfig = (featureConfig: Config): PlayerFeature<State> =>
-    definePlayerSlice({
-      ...(isUndefined(name) ? {} : { name }),
-      state: (ctx) => state(ctx, featureConfig),
-      ...(attach ? { attach: (ctx) => attach(ctx, featureConfig) } : {}),
-    });
-
-  const defaultFeature = forConfig(defaultConfig as Config);
-  const feature = ((featureConfig?: Config) =>
-    isUndefined(featureConfig) ? defaultFeature : forConfig(featureConfig)) as ConfigurablePlayerFeature<Config, State>;
-
-  feature.state = defaultFeature.state;
-  if (defaultFeature.attach) feature.attach = defaultFeature.attach;
-  if (!isUndefined(name)) Object.defineProperty(feature, 'name', { value: name });
-
-  return feature;
+  return {
+    ...definition,
+    ...(preserved.length > 0 ? { preserve: preserved } : {}),
+  } as PlayerFeature<State>;
 }
 
 /** Merge the configuration declarations from the selected player features. */
--- a/packages/core/src/dom/store/features/orientation-lock.ts
+++ b/packages/core/src/dom/store/features/orientation-lock.ts
@@ -1,52 +1,112 @@
-import { listen } from '@videojs/utils/dom';
+import { createSelector } from '@videojs/store';
+import { listen, type WebKitVideoElement } from '@videojs/utils/dom';
+import { isNull } from '@videojs/utils/predicate';
 
 import { definePlayerFeature } from '../../feature';
+import type { PlayerFeatureConfig } from '../../player';
 import { isFullscreen } from '../../presentation/fullscreen';
 import { createScreenOrientationLock, type ScreenOrientationLockType } from '../../presentation/orientation';
 
-export interface OrientationLockFeatureConfig {
-  /** Screen orientation type to lock while fullscreen is active. */
-  type?: ScreenOrientationLockType | undefined;
+export type { ScreenOrientationLockType };
+
+const DEFAULT_ORIENTATION_LOCK_TYPE: ScreenOrientationLockType = 'landscape';
+
+/** Orientation lock configuration and its user-config writer. */
+export interface OrientationLockState {
+  /** Screen orientation type locked while fullscreen is active. */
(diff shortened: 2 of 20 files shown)


packages/core/src/dom/store/features/metadata.ts
```diff
@@ -45,6 +47,7 @@ export const metadataFeature = definePlayerFeature({
     setDefaultContentTitle: (value) => set({ [USER_DEFAULT_CONTENT_TITLE]: value }),
   }),
   derived: {
+    /** The resolved content title. */
     contentTitle: ({ get }) =>
```
mihar-22, line 91: Might be best to lift and use a interface with jsdoc here over inlining jsdoc comments

packages/core/src/dom/store/features/orientation-lock.ts
```diff
@@ -1,52 +1,115 @@
...
+  /**
+   * Sets the locked orientation type. Absent input — nullish, or the empty
+   * string a valueless HTML attribute produces — restores the default.
+   */
+  setOrientationLockType(value: ScreenOrientationLockType | null | undefined): void;
 }
 
 interface WebKitPresentationMedia extends HTMLMediaElement {
   webkitPresentationMode?: string;
 }
```
mihar-22: not related to this PR but this type is out of place

packages/core/src/dom/feature.ts
```diff
@@ -47,48 +28,15 @@ export function definePlayerFeature<
...
-  const feature = ((featureConfig?: Config) =>
-    isUndefined(featureConfig) ? defaultFeature : forConfig(featureConfig)) as ConfigurablePlayerFeature<Config, State>;
-
-  feature.state = defaultFeature.state;
-  if (defaultFeature.attach) feature.attach = defaultFeature.attach;
-  if (!isUndefined(name)) Object.defineProperty(feature, 'name', { value: name });
-
-  return feature;
+export function definePlayerFeature<State>(
+  definition: StaticPlayerFeatureConfig<State, object, PlayerFeatureConfig>
```
mihar-22: `StaticPlayerFeatureConfig` but param called `definition` -> likely need some naming revision here

packages/core/src/dom/feature.ts
mihar-22, line 1: so much red, love it, shiny 💍

mihar-22 (approved):

packages/core/src/dom/store/features/metadata.ts
decepulis: This line is #2000's — it's in this diff only because this branch stacks on that one.

Worth flagging that the inline JSDoc isn't only a style choice here. The api-docs-builder reads these descriptions straight off the object-literal property (`getJSDocDescription(prop)` in `feature-handler.ts`, for both `config` and `derived`), so lifting them to an interface would empty the Details column in the generated tables unless the extractor learns to follow the type. Where an interface already exists — the state shapes in `media/state.ts` — it does read JSDoc from the members, so that half already works the way you're describing. It's `config` and `derived` that have no interface to hang it on.

Happy to teach the extractor to follow a declared type and move these, if you'd rather have it that way — say where you want it, #2000 or a follow-up.

---

mihar-22: Yes Claude bot, but you won't need to walk object literals if you just walk the types. So there's likely a gap here in how you discover and walk config/derived types. I don't know what mess you've made over there and now you're justifying it :)

decepulis: Fair, and you've named the actual gap. `getJSDocDescription` is purely syntactic — it reads JSDoc off whichever AST node the walker is standing on, never through the checker — so the walker's shape is what dictates where the docs have to live. That's a property of the extractor, not a reason the docs belong inline, and I shouldn't have framed it as the latter.

Resolving each property to its symbol and asking the checker for the comment is the fix, and it subsumes both spellings: the declaration behind an inline property and behind an interface member are each reachable that way, so placement goes back to being a style call — and then I agree the interface wins. Worth checking what falls out for `config` specifically, since `PlayerFeatureConfig<State>` is a `Record`, so the named keys only exist on the `satisfies` expression's own literal type rather than on anything declared.

That's the whole handler rather than the published-key matching I added here, so I'd rather not grow this PR with it. Taking it as a follow-up unless you want it blocking.

---

PR #2009, opened 2026-08-07
feat(packages): give the poster placeholder its own component

decepulis:
Stacked on #1998. Groundwork for getting the poster and its placeholder out of the skin (#1798).

> [!NOTE]
> Repurposed. This PR used to converge the poster DOM by wrapping React's `Poster` in a container. That's been discarded — it broke the one-component-one-element rule and changed the React poster's public shape to solve a problem that wasn't about the poster. The poster placeholder is the thing that actually needed to move.

## The problem

The blur-up placeholder lived on the poster: painted into `media-poster::before`, fed a URL by a CSS custom property that the skin threaded down from a prop.

That works where the poster is a wrapper element. It cannot work in React, where `Poster` renders an `img` — a replaced element generates no `::before` box and cannot have children. So the React skins painted the placeholder on the skin **container** instead, then reached back across the tree with `:has()` to find out whether the poster had loaded.

One concept, two implementations, and the divergence was structural rather than incidental. Neither side could adopt the other's.

Configuration had the same split: the placeholder was a skin prop, so it was only reachable from a packaged skin. Compose components by hand, or eject a skin, and there was no way to set one.

## The change

A component of its own, the one shape both platforms can render:

```html
&lt;video-player poster="poster.jpg" poster-placeholder="tiny.jpg"&gt;
  &lt;media-container&gt;
    &lt;video src="video.mp4" playsinline&gt;&lt;/video&gt;
    &lt;media-poster-placeholder&gt;&lt;/media-poster-placeholder&gt;
    &lt;media-poster&gt;&lt;/media-poster&gt;
  &lt;/media-container&gt;
&lt;/video-player&gt;
```

```tsx
&lt;Player.Provider poster="poster.jpg" posterPlaceholder="tiny.jpg"&gt;
  &lt;Player.Container&gt;
    &lt;Video src="video.mp4" playsInline /&gt;
    &lt;PosterPlaceholder /&gt;
    &lt;Poster /&gt;
  &lt;/Player.Container&gt;
&lt;/Player.Provider&gt;
```

One empty element that sets `background-image` from the store and leaves sizing, position, and blur to CSS.

A plain box beside the poster is the only shape both platforms share — React can't put a layer inside an `img`, and a custom element can't *be* one (customized built-ins never shipped in WebKit). One consequence worth noting: the skin rules need no `isShadowDOM` branch, because there's no inner element to reach past.

## Painting over, not handing off

The placeholder sits behind the poster and stays put. The poster covers it as it loads, and where the poster doesn't reach — the bars left by `object-fit: contain` — it keeps showing through.

Nothing observes the poster's load state, so the cross-element `:has()` selector goes away entirely. This is also what the HTML path already did for author-supplied images.

Visibility mirrors the poster: both carry `data-visible`, both drop it once playback starts. `visible` deliberately doesn't depend on having a URL — with nothing configured the element paints nothing either way, and coupling them would only make the contract harder to state.

## Configuration

`posterPlaceholder` and `defaultPosterPlaceholder` join the metadata feature on the same four tiers as the poster: user, media, user default, `''`. A hand-authored layout and an ejected skin can now set one, and the media tier is open to a `MediaContentData` donor.

`BaseVideoSkinProps.placeholder` and the skin's `placeholdersrc` attribute are gone. So is `--media-poster-placeholder`, since the URL is no longer a custom property; `--media-poster-placeholder-blur` keeps its name and moves to the new element.

## Why the `poster` prefix

Every public name carries it: the components, the store keys, the `poster-placeholder` attribute, `MediaContentData.posterPlaceholder`, and the `.media-poster-placeholder` class.

"Placeholder" alone is a generic word in UI work — it's the `placeholder` attribute on an input, and a stand-in for content that hasn't arrived. This one is specifically the stand-in for the poster: it resolves next to `poster`, it's painted behind `poster`, and it means nothing without one. Naming it after what it stands in for makes that relationship readable in an import list, an attribute, and a CSS selector alike.

## `cssUrl`

New in `@videojs/utils/style`. An unquoted `url()` ends at the first paren, so a signed URL carrying one produced a declaration the browser dropped — silently, since `setProperty` with an invalid value is a no-op and the image just never painted.

## Docs

New reference page with HTML and React demos, plus a pointer to it from the poster page. `internal/design/ui/poster-placeholder.md` records the decision, including why the name carries the prefix. It supersedes the old record, which moves to `poster-placeholder-layer.md` — named after the shape it chose, so the new record can own the plain name.

## Testing

Suites green per package: core 1413, react 398, html 303, media 539, utils 419, site 530. `pnpm typecheck` reports only the 24 pre-existing `packages/store` test errors; `check:workspace` and `astro check` clean.

New coverage: tier resolution and its independence from the poster, `cssUrl` escaping, and both bindings painting the resolved URL, clearing it, and hiding on play.

Verified in Chromium that `poster-placeholder` on the player reaches the store through the provider's attribute plumbing, and that the HTML and React skins produce pixel-identical output — which is the whole point of the change.


decepulis:
Closing in favour of #2039, which drops the separate placeholder component. The poster offers the resolved value as `--media-poster-placeholder` and the packaged skins paint it as the poster image's own `background-image` — an `` has a background box like anything else, so one element covers both and the poster arrives on top of its own stand-in.

What carried over: `posterPlaceholder` and `defaultPosterPlaceholder` on the metadata feature resolved on the same four tiers as the poster, the `poster` prefix on every public name and the reasoning behind it, `placeholdersrc` and the React `placeholder` prop going away, and no `isShadowDOM` branch for the placeholder.

What #2039 does differently:

- No `<media-poster-placeholder>` element, and no second box in the tree.
- `--media-poster-placeholder` keeps its name and its job rather than being removed — it is set by the component that resolves the value instead of by `SkinElement` from an attribute.
- No `cssUrl`. The store takes a CSS image and passes it through untouched, so quoting is the caller's. That is a real cost and it is recorded as one; your point about `setProperty` failing silently on an invalid value is exactly why.
- `--media-poster-placeholder-blur` is gone rather than relocated. Blur is baked into the image, which is what image pipelines already return.

`internal/design/ui/poster-placeholder.md` is updated in place in #2039, so the naming rationale from this PR survives there rather than in a superseded record.

---

PR #2010, opened 2026-08-07
feat(packages): resolve the poster from the store, not from markup

### LINKED ISSUE #1798: Feature: Communicate Media Metadata to the UI

decepulis:
Stacked on #2009. Closes #1798.

## The problem

The poster was configured in markup: an `img` carrying `slot="poster"` in HTML, a `poster` prop on the React skin. That put content configuration in the skin's hands, made it unreachable from a hand-authored layout, and left the store's resolved `poster` — user, media, user default — with no reader.

## The change

Before, in HTML:

```html
&lt;video-skin&gt;
  &lt;video src="video.mp4" playsinline&gt;&lt;/video&gt;
  &lt;img slot="poster" src="poster.jpg" alt="Video poster" /&gt;
&lt;/video-skin&gt;
```

After:

```html
&lt;video-player poster="poster.jpg"&gt;
  &lt;video-skin&gt;
    &lt;video src="video.mp4" playsinline&gt;&lt;/video&gt;
  &lt;/video-skin&gt;
&lt;/video-player&gt;
```

Before, in React:

```tsx
&lt;Player.Provider&gt;
  &lt;VideoSkin poster="poster.jpg"&gt;
    &lt;Video src="video.mp4" playsInline /&gt;
  &lt;/VideoSkin&gt;
&lt;/Player.Provider&gt;
```

After:

```tsx
&lt;Player.Provider poster="poster.jpg"&gt;
  &lt;VideoSkin&gt;
    &lt;Video src="video.mp4" playsInline /&gt;
  &lt;/VideoSkin&gt;
&lt;/Player.Provider&gt;
```

`PosterCore` composes `playback` and `metadata`, so the chain is written and tested once and every consumer inherits it: a packaged skin, an ejected one, a layout you wrote yourself. `BaseVideoSkinProps` is gone — the video skins take the same three props the audio ones always did.

## Rendering an image without taking it away

[`internal/design/ui/poster.md`](https://github.com/videojs/v10/blob/claude/poster-store-config-c31gf2/internal/design/ui/poster.md) treats author control of the image as the point of the component, and that still holds. The store supplies a default, it doesn't take over.

**React's `Poster` stays the `img` it was.** `src` fills in from the store when none is passed; every other image attribute is untouched.

```tsx
&lt;Poster /&gt;
&lt;Poster srcSet="poster-480.jpg 480w, poster-1080.jpg 1080w" sizes="100vw" /&gt;
&lt;Poster render={(props) =&gt; &lt;Image {...props} alt="" fill /&gt;} /&gt;
```

It also gains `alt=""` — it had no `alt` at all, so screen readers announced the URL.

**HTML is the harder half**, because `media-poster` never rendered an image; it positioned one you supplied. It now owns an `img` carrying `part="img"` in a shadow root, beside a slot, the same shape `media-thumbnail` already uses. Fill the slot and the owned image steps aside:

```html
&lt;media-poster&gt;
  &lt;img src="poster.jpg" srcset="poster-480.jpg 480w" alt="Keynote speaker" /&gt;
&lt;/media-poster&gt;
```

Hiding it isn't enough on its own: an image that isn't rendered **still downloads**, which is not hypothetical now that Mux and Vimeo donate posters. The `src` is withheld too.

### Why `assignedElements({ flatten: true })`

A skin forwards its own `slot` carrying `name="poster"` into the element, and that empty slot is itself an assigned node — so counting assigned nodes finds one whether or not you supplied anything. Verified in Chromium:

```
author slots an image  →  assigned: 1 (the skin's slot), flattened: [IMG]
author slots nothing   →  assigned: 1 (the skin's slot), flattened: []
```

Same reason the owned image sits *beside* the slot rather than inside it as fallback content: fallback renders only when nothing is assigned, and inside a skin something always is.

### Why `data-loaded` moved to the host

A selector can't reach past `::part()` to read an attribute. `media-poster:not([data-loaded])::part(img)` is expressible; `::part(img):not([data-loaded])` is not. Skins reach the owned image as a part and leave yours alone, which is what held the blur-up sequence together before.

## Fixes that fall out

- **Ejected Tailwind HTML skins rendered an unpositioned poster.** The shadow branch styled only `::slotted(img)`, but `site/scripts/ejected-skins/html.ts` inlines a literal `img` where the slot used to be, which `::slotted` never matches. All three variants — part, slotted, descendant — are emitted now.
- **The React skins' poster was styled through `.media-default-skin > img`.** It's the explicit `.media-poster` now, matching the `.media-poster-placeholder` its sibling uses.

## Testing

Suites green per package: core 1411, react 406, html 311, media 539, utils 419, skins 4, site 530. `pnpm typecheck` reports only the 24 pre-existing `packages/store` test errors; `check:workspace` and `astro check` clean.

New coverage: `PosterCore` resolving the URL; the element rendering it, stepping aside for a slotted image without fetching its own, taking back over when that image is removed, and reporting load on the host; and the React side filling in `src`, preferring one passed to it, defaulting `alt`, and handing the resolved `src` to a `render` override.

Verified in Chromium across both skins and both stylings: the store-resolved poster paints over the poster placeholder, a slotted image replaces it with the owned one hidden and unfetched, and Tailwind compiles the `[&::part(img)]` variants that reach it.

### DIFF
--- a/apps/sandbox/templates/html-mux-video/main.ts
+++ b/apps/sandbox/templates/html-mux-video/main.ts
@@ -39,14 +39,13 @@ async function render() {
   const srcAttr = source ? '' : ` src="${url}"`;
 
   document.getElementById('root')!.innerHTML = wrapSandboxHtmlI18n(html`
-    <${playerTag}${posterPlaceholder ? ` poster-placeholder="${posterPlaceholder}"` : ''}>
+    <${playerTag}${poster ? ` poster="${poster}"` : ''}${posterPlaceholder ? ` poster-placeholder="${posterPlaceholder}"` : ''}>
       <${tag} class="aspect-video max-w-4xl mx-auto">
         <!-- The storyboard track is derived automatically from the Mux src. -->
         <mux-video${srcAttr} ${mediaAttrs} playsinline crossorigin="anonymous"></mux-video>
         <!-- Mux Data and Cast are opt-in media components; no env key is needed for Mux-hosted sources. -->
         <mux-data player-software-name="mux-video"></mux-data>
         <google-cast></google-cast>
-        ${poster ? html`<img slot="poster" src="${poster}" alt="Video poster" />` : ''}
       </${tag}>
     </${playerTag}>
   `);
--- a/apps/sandbox/templates/react-hlsjs-video/main.tsx
+++ b/apps/sandbox/templates/react-hlsjs-video/main.tsx
@@ -39,14 +39,8 @@ function App() {
 
   return (
     <SandboxI18nProvider>
-      <Provider>
-        <VideoSkinComponent
-          poster={poster}
-          skin={skin}
-          styling={styling}
-          live={live}
-          className="aspect-video max-w-4xl mx-auto"
-        >
+      <Provider poster={poster}>
+        <VideoSkinComponent skin={skin} styling={styling} live={live} className="aspect-video max-w-4xl mx-auto">
           <HlsJsVideo
             {...(hlsSource ? { source: hlsSource } : { src: url ?? '' })}
             autoPlay={autoplay}
--- a/apps/sandbox/templates/react-mux-video/main.tsx
+++ b/apps/sandbox/templates/react-mux-video/main.tsx
@@ -41,14 +41,8 @@ function App() {
 
   return (
     <SandboxI18nProvider>
-      <Provider posterPlaceholder={posterPlaceholder}>
-        <VideoSkinComponent
-          poster={poster}
-          skin={skin}
-          styling={styling}
-          live={live}
-          className="aspect-video max-w-4xl mx-auto"
-        >
+      <Provider poster={poster} posterPlaceholder={posterPlaceholder}>
+        <VideoSkinComponent skin={skin} styling={styling} live={live} className="aspect-video max-w-4xl mx-auto">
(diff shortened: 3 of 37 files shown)


decepulis:
Closing in favour of #2039, which takes the compound-image route: `<media-poster>` coordinates a child image the way `<picture>` does, instead of owning one in a shadow root beside a slot.

What carried over from here: the store-resolved poster on the metadata feature, `BaseVideoSkinProps` going away, `poster` moving to the provider, the React skins styled through an explicit `.media-poster`, and the ejected-skin fix for an inlined `` that `::slotted` never matched.

What #2039 does differently: no owned image, so no `part="img"` and no second image in the tree — which also means no `::part()` styling contract and no withheld `src` on a hidden download. The `data-loaded`-on-the-host reasoning from here holds either way and stayed.

The `alt=""` default from this PR is in #2039 too. Resolving `alt` through the store was considered and dropped: filling in a `src` has `<picture>` as precedent, describing an image does not.

---

Issue #1837, opened 2026-07-14
SPF-Backed `<mux-audio>` Element


decepulis:
Provide a `<mux-audio>` custom element backed by the SPF playback engine (`@videojs/spf/playback-engine`) for audio-only Mux playback, exposing the same `HTMLMediaElement`-compatible surface as its `<mux-video>` counterpart.

> [!NOTE]
> This is an AI-generated spec. Requirements should be independently confirmed.

mihar-22:
Closing as completed by #2045. The SPF-backed `<mux-audio>` implementation, platform exports, and tests are present on current main.

Issue #1836, opened 2026-07-14
SPF-Backed `<mux-video>` Element


decepulis:
Provide a `<mux-video>` custom element backed by the SPF playback engine (`@videojs/spf/playback-engine`), so Mux-hosted playback IDs work as a drop-in media element in the v10 player. Wires the SPF HLS engine and `SpfMedia` adapter behind the element's `HTMLMediaElement`-compatible surface.

> [!NOTE]
> This is an AI-generated spec. Requirements should be independently confirmed.

mihar-22:
Closing as completed by #2045. The SPF-backed `<mux-video>` implementation, platform exports, and tests are present on current main.

PR #2033, opened 2026-08-10
refactor(packages)!: relocate spf media facades

cjpillsbury:
Groundwork for the SPF-backed `<mux-video>` / `<mux-audio>` elements (#1836, #1837). No feature ships here — this is the package restructuring those elements need, landed on its own so the feature PR is only the feature.

## Why

SPF is unlike hls.js or dash.js in that it publishes no engine-shaped API for consumers to drive: the `@videojs/media` facade *is* how a composed engine is consumed. The code locations didn't reflect that — the adapters lived in `@videojs/spf` while the Medias composing them lived in `@videojs/media`, which depended on spf.

Inverting that mattered concretely. With the facade in a different package from the engine, every new Media capability that reads engine state has to promote engine internals to public API just to cross the package boundary — 21 symbols of live-window readers, presentation guards, and error copy in the version of this that went the other way. Facade and engine in one package, and that widening disappears.

## What changed

**`spf → media`, not the reverse.** `SimpleHlsMedia`, `SimpleHlsAudioOnlyMedia`, and the media-tracks projection now live in `packages/spf/src/playback/adapters/`. `@videojs/media` imports nothing from spf.

**Entry points follow one rule:** engines share one entry per format family; adapters get one each. So `@videojs/spf/hls` is the engines, and `@videojs/spf/simple-hls`, `/simple-hls-audio-only`, `/background-video` are the Medias. Driving an engine directly costs neither a Media nor `@videojs/media` — which is also what took `./hls` back under its size target, from 20.40 KB to 19.36 KB gzipped (it had been over, at 102%).

**`background-video` joins that rule.** Its adapter had been sitting in the engine layer, shipping both halves from one entry. The adapter moved to `adapters/`; the engine moved beside the other HLS engines as `engine-background-video.ts`, following `engine-audio-only` — also a subtractive variant of the full HLS engine, and always resident there. No new entry point. Deliberately no host-bound Media: nothing consumes one, and binding the mixin to `HTMLVideoElementHost` would give that entry a `@videojs/media` dependency it otherwise doesn't have.

**The Mux source split in two.** `@videojs/media/dom/mux/source` holds the engine-neutral layer — `MuxSourceBase`, the params types, the stream/poster/storyboard URL builders — so an SPF-backed Mux Media can consume it without acquiring hls.js. `mux/drm.ts` keeps the hls.js license-server shaping. The `@videojs/media/dom/mux` barrel is unchanged, so html, react, sandbox, and site needed no edits.

## ⚠️ Breaking

Both are invisible in an `exports` diff, which shows only additions. Worth reading carefully for that reason.

| Was | Now |
| --- | --- |
| `@videojs/media/dom/simple-hls` | `@videojs/spf/simple-hls` |
| `@videojs/media/dom/simple-hls-audio-only` | `@videojs/spf/simple-hls-audio-only` |
| `createBackgroundVideoEngine` from `@videojs/spf/background-video` | `@videojs/spf/hls` |

The first pair never had explicit `exports` keys — they matched `@videojs/media`'s `./dom/*` wildcard, so they simply stop resolving now that the dist output is gone. In the third, the entry name is unchanged but its contents are not: `@videojs/spf/background-video` is the adapter now, and the engine moved.

Two compatibility shims that only existed because spf sat below media are also gone: `SimpleHlsMediaError` extends `ErrorLike`, and `SimpleHlsMediaStreamType` *is* `MediaStreamType` rather than a structurally-compatible copy.

## Verification

`pnpm test` 22/22, `pnpm build` 19/19, `pnpm typecheck`, `pnpm check:workspace` 10/10.

Browser-verified in the sandbox, since unit tests resolve through vitest's src aliases and so exercise none of the entry-point wiring this PR changes: `html-simple-hls-video` (7 renditions / 7 audio / 14 text tracks), `html-simple-hls-audio-only`, `react-simple-hls-video`, `html-mux-video` (poster and storyboard URLs deriving from a bare `src`, storyboard track present), and `spf-background-video` (playing, looping, muted, audio side subtracted). E2E `spf-unsupported-source` + `spf-mse-recovery` 7/7 on chromium.

Also verified against the built artifacts rather than the source: no hls.js reference reaches `media/dist/**/mux/source/` or anywhere in `spf/dist/`, and `@videojs/spf/background-video` contains zero engine symbols.

## Notes for reviewers

**A full `pnpm build` is required** before the sandbox or e2e will run. Building only `media` and `spf` leaves `packages/html/dist/dev` importing the deleted path, and pages 500.

**Install weight, accepted not mitigated:** `@videojs/spf` now transitively installs hls.js / dashjs / @vimeo/player / mux-embed via `@videojs/media`. Nothing lands in any bundle. If it ever matters, the fix is optional peers on media, or splitting media's engine-free contracts and hosts into a light package.

**`./hls` headroom is now 651 B** against its 20 KB target (96.8%). Folding the background-video engine in cost a measured 281 B gzipped. The next engine-side addition should be measured before it lands.

**Unrelated flake:** `packages/html` → `media-button-element.test.ts` → "updates aria-label when html lang changes and provider has no explicit lang" fails intermittently under parallel load and passes in isolation. It's an i18n test; this PR's entire footprint in `packages/html` is two import lines.

---

> [!NOTE]
> **High Risk**
> Breaking import paths and a reversed spf/media dependency affect every Simple HLS and background-video consumer; incorrect entry-point wiring could pull engines or third-party engines into the wrong bundles.
> 
> **Overview**
> **Breaking package restructure** so SPF owns the HTMLMediaElement-shaped playback facades and `@videojs/media` no longer depends on `@videojs/spf`.
> 
> **SPF Medias and adapters** — `SimpleHlsMedia`, `SimpleHlsAudioOnlyMedia`, mixins, error surface, and media-tracks wiring move from `packages/media/src/dom/simple-hls*` into `packages/spf/src/playback/adapters/`. New public entry points: `@videojs/spf/simple-hls`, `@videojs/spf/simple-hls-audio-only`, and `@videojs/spf/background-video` (adapter only). `@videojs/spf/hls` keeps **engines only** (`createSimpleHlsEngine`, audio-only, background-video as `engine-background-video.ts`); adapters are no longer re-exported there so engine-only imports avoid `@videojs/media` and stay within the `./hls` size budget.
> 
> **Dependency flip** — `@videojs/spf` now depends on `@videojs/media`; `@videojs/media` drops `@videojs/spf`. Adapter code imports real `ErrorLike` and `MediaStreamType` instead of structural duplicates.
> 
> **Consumers** — `packages/html` and `packages/react` simple-hls components import Medias from the new spf paths. Sandbox background-video types come from `@videojs/spf/hls` for engine state; the element stays on `@videojs/spf/background-video`.
> 
> **Mux** — Shared Mux identity/URL helpers become `MuxSourceBase` under `@videojs/media/dom/mux/source`; hls.js-specific DRM shaping stays in `mux/drm.ts` for the hls.js-backed `MuxMedia`.
> 
> **Build/tests** — Media package drops simple-hls build entries and adds `./media-tracks`; spf vitest adds a `playback-adapters` browser project; design docs and `AGENTS.md` updated for the new layout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 37d7b72fb2f3546803de2002854adb6561a1e049. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/media/package.json
```diff
@@ -31,6 +31,11 @@
       "types": "./dist/dev/dom/*/index.d.ts",
       "development": "./dist/dev/dom/*/index.js",
       "default": "./dist/default/dom/*/index.js"
+    },
+    "./media-tracks": {
```
cjpillsbury, line 35: **NOTE**: Added for 3rd party implementors (which SPF can be thought of as one, albeit an official one)

packages/media/package.json
```diff
@@ -48,7 +53,6 @@
     "clean": "rimraf --glob dist types '*.tsbuildinfo'"
   },
   "dependencies": {
-    "@videojs/spf": "workspace:*",
```
cjpillsbury, line 51: **NOTE**: Dropping as part of the relocation effort so that we no longer have the circular dependencies issue (a primary motivation for creating the separate media package).

packages/media/tsdown.config.ts
```diff
@@ -16,11 +17,10 @@ const createConfig = (mode: PackageBuildMode): UserConfig => ({
     'dom/dash/index': './src/dom/dash/index.ts',
     'dom/hls-js/index': './src/dom/hls-js/index.ts',
     'dom/native-hls/index': './src/dom/native-hls/index.ts',
-    'dom/simple-hls-audio-only/index': './src/dom/simple-hls-audio-only/index.ts',
-    'dom/simple-hls/index': './src/dom/simple-hls/index.ts',
     'dom/vimeo/index': './src/dom/vimeo/index.ts',
     'dom/youtube/index': './src/dom/youtube/index.ts',
     'dom/mux/index': './src/dom/mux/index.ts',
+    'dom/mux/source/index': './src/dom/mux/source/index.ts',
```
cjpillsbury, line 23: **NOTE**: Adding for shareable and directly consumable types.

luwes (approved):
👍

PR #2039, opened 2026-08-10
feat(packages)!: resolve the poster in the store, and set src on img from it

### LINKED ISSUE #1798: Feature: Communicate Media Metadata to the UI

decepulis:
Closes #1798.

Stacked on #2176, which does the same thing for the title. The placeholder half is split out into #2063, which stacks on this.

Tag names are written without angle brackets throughout — this repo's PR body pipeline strips them, which is what mangled the earlier revisions of this description.

## Sandbox
- [React](https://v10-sandbox-git-claude-poster-picture-model-mux.vercel.app/react-mux-video/)
- [HTML](https://v10-sandbox-git-claude-poster-picture-model-mux.vercel.app/html-mux-video/)

## What is this?
1. The player resolves a `poster` out of the author's input and the media's content data, the same two tiers the title uses. This moved here from #1998 so the resolved value lands with the components that read it.
2. `media-poster` sets `src` on the image already inside it. As a side effect it extends `MediaElement` rather than `MediaUIElement` now, and drops its lazy, unused i18n chunk.
3. The packaged skins carry that image as poster-slot fallback content.
4. React's `Poster` resolves the URL from the player rather than taking it as a skin prop, and reports `data-loading` / `data-loaded` / `data-error` alongside `data-visible`.
5. Skins take `renderPoster` in place of the `poster` prop. The player owns the URL, so there is no string form.

## The store side
Following #2176, the poster reaches the store through the metadata feature's own config — the door the HTML provider element and the React player already used — and nothing else can write it.

- `setPoster` and `setDefaultPoster` are gone from the store, replaced by a private symbol action the config points at.
- `defaultPoster` is gone as a provider input. That tier existed so the config had somewhere to put a fallback, and nothing else read it.

Resolution is now author input, then whatever the media donates, then the empty string.

### Passing to the player
- React — `poster` on the player component: a preset's `VideoPlayer`, or the `Player` your own `createPlayer` returns
- HTML — the `poster` attribute on `video-player`

### Consuming in the image
An `img` with no source of its own is the one the element fills in. Give it a `src`, a `srcset`, or `source` candidates inside a `picture` and it is yours, left alone.

In React the component *is* the image: image attributes go straight on it, `render` swaps the element while keeping the resolved `src`, and leaving the source off lets the player fill it in.

### Overriding the skin image
- HTML — slot your own image; the plain one the skin carries is fallback content, so yours displaces it
- React — `renderPoster` draws your image in place of the skin's

## Merging main
Three commits on main changed shapes this branch touches, and this branch now follows them rather than working around them.

- **#2116 and #2154** renamed `createPlayer`'s `Provider` to `Player` and moved `Container` out of the factory. Every demo, sandbox template, generated e2e page, and ejected-skin template here moves the `poster` value onto the new player component.
- **#2094** replaced the React poster's `.media-poster` class hook with a structural `.media-default-skin > img` selector. This branch had added the class back; it no longer does, and the skin CSS files are untouched again. A `renderPoster` override that is not an `img` is therefore unstyled by the CSS skins, which matches what the HTML side already says about slotting a `picture`.

One gap turned up while folding #2094 in: the Tailwind skins only styled `::slotted(img)`, which reaches the image an author assigns but not the plain one this branch adds as slot fallback content — that one lives in the skin's own shadow root, where `::slotted` cannot see it. Without a rule for it the fallback rendered `static` and zero-sized. The CSS skins already carried the matching `media-poster img` rule; the Tailwind helper now does too.

## What this isn't
- The placeholder belongs to #2063 — its store field, its presentation, and its own setter removal. Here it keeps working exactly as it does on main — skin prop, `placeholdersrc` attribute, `::before` layer.
- No `posterAlt`. Filling in a `src` has `picture` as precedent; describing an image doesn't.
- `media-thumbnail` is untouched. Whether it should follow this model is its own question.

## Funny business
- *`media-poster` alone renders nothing.* Your layout supplies the image, and a `__DEV__` warning fires when a poster resolves without one. `picture` behaves the same way, though it's a platform element, so expectations differ.
- *A skin can only style an image you slot directly.* `::slotted(img)` doesn't reach inside a `picture` or a framework image component, and shadow CSS has no other way in. That sizing is yours.
- *A failed HTML poster shows the browser's broken-image glyph.* The React one is held transparent until `data-loaded`, so a failed poster there is simply invisible. #2063 removes that gate, because holding the image transparent would hide the placeholder painted behind it.
- *Ownership is settled once,* when an image becomes active. Handing one back means re-slotting it with a source, the way React settles controlled inputs. For the same reason, an image that rewrites its own `src` in place keeps `data-loaded` through the next fetch: we listen for `load` and `error` but don't watch the attribute.
- *Reading whether an image already loaded is only safe once.* `complete` is true for an image that omits both `src` and `srcset` whatever a parent `picture` is fetching, and a browser holds the previous request's pixels until the next one settles. Both bindings read `naturalWidth` and `complete` when they first see an image and never again; `load` and `error` cover everything after.
- *The tests set the poster through `setPlayerConfigValue`* and the feature's config entry, since that is the only door left and the same one a player uses.

---

> [!NOTE]
> **High Risk**
> Breaking public API across core metadata, HTML and React poster bindings, and all video skins, with nuanced image-ownership and load-state behavior that affects every integration path.
> 
> **Overview**
> **Breaking change:** poster is **player metadata** (author `poster` on `VideoPlayer` / `video-player`, then media `contentData.poster`), not a skin prop or direct store setters — `setPoster`, `setDefaultPoster`, and `defaultPoster` are removed.
> 
> **`<media-poster>` / React `Poster`** now **fill `src` only on an image the author supplies** (skin slot fallback or child ``), respect authored `src`/`srcset`/`<picture>`, and expose **`data-loading` / `data-loaded` / `data-error`** on the host alongside `data-visible`. Packaged skins ship a default `` inside the poster slot; React skins always render `<Poster>` and take **`renderPoster`** instead of a string `poster` prop.
> 
> Demos, sandbox templates, e2e page generation, site docs/examples, and ejected-skin scripts are updated to pass **`poster` on the player** and drop slotted poster images where the store drives the URL. Tailwind poster utilities also style the skin’s **fallback** `` (not only `::slotted`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea0ac025758f8b50cbf648ce20484886f4bf5329. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/e2e/scripts/generate-pages.ts
```diff
@@ -204,6 +204,7 @@ function reactVideoPage(media: string, resource: string, config: MediaTypeConfig
     ? `import { Video, VideoSkin, videoFeatures } from '@videojs/react/video';`
     : `import { ${reactMedia.component} } from '${reactMedia.importPath}';\nimport { VideoSkin, videoFeatures } from '@videojs/react/video';`;
 
+  // The poster URL is player configuration, not a skin prop.
```
decepulis: remove. I don't like comments that talk about historical shape. If people want archaeology, they can read the PR 😅

packages/core/src/core/ui/poster/poster-core.ts
```diff
@@ -1,24 +1,64 @@
...
+export type PosterMediaState = Pick<MediaPlaybackState, 'started'> &
+  Pick<MediaMetadataState, 'poster' | 'posterPlaceholder'>;
+
+/**
+ * How far the image the poster coordinates has got.
+ *
+ * Reported by the binding rather than derived here: only the binding holds the
+ * image, and its source may be one the author set rather than the resolved
+ * `src` below. `'none'` covers both no image and an image with no source.
+ */
```
decepulis, line 9: imo throughout this whole PR, you've been too verbose with comments. Double check with repo rules and conventions on comments. Consider only using them when they're used in the docs by our api reference generator or particularly needed

packages/html/src/define/live-video/minimal-skin.tailwind.ts
```diff
@@ -33,7 +33,10 @@ function getTemplateHTML() {
       <slot></slot>
 
       <media-poster class="${poster(true)}">
-        <slot name="poster"></slot>
+        <slot name="poster">
+          <!-- Replaced by an <img slot="poster"> of your own. -->
```
decepulis: don't introduce comments like this unless they were preexisting

packages/html/src/ui/poster/poster-element.ts
decepulis, line 1: This file is long and difficult to grok. simplify while maintaining adherence to the spec. If you don't remember the spec, ask

packages/media/src/core/state.ts
```diff
@@ -143,6 +143,17 @@ export interface MediaMetadataState {
...
   poster: string;
+  /**
+   * The resolved poster placeholder: a low-resolution stand-in shown behind the
+   * poster while it loads.
+   *
+   * This is a CSS `background-image` value, not a URL — it is written onto the
+   * poster image untouched, so a URL has to arrive wrapped: `url("…")`. Any blur
+   * has to be baked into the image itself, which is what `@mux/blurup` returns.
+   * Resolved independently of the poster, so a player can carry one without the
+   * other.
```
decepulis: WAY too verbose and specific
```suggestion
```

apps/sandbox/app/shared/sources.ts
```diff
@@ -263,9 +263,18 @@ export function getPosterSrc(source: SourceId): string | undefined {
   return id ? `https://image.mux.com/${id}/thumbnail.webp${imageQuery(source, 'poster')}` : undefined;
 }
 
-export function getPlaceholderSrc(source: SourceId): string | undefined {
+/**
+ * A CSS image, not a URL: `posterPlaceholder` reaches CSS untouched. A real
+ * integration would pass `@mux/blurup` output here, which arrives pre-blurred;
+ * this upscales a 20px thumbnail instead, and the browser's own smoothing does
+ * the blurring.
```
decepulis: Reconsider all your comments. I'll change a few specific ones, like this.
```suggestion
 * A CSS image. this upscales a 20px thumbnail instead,
 * and the browser's own smoothing does the blurring.
```

packages/core/src/core/ui/poster/poster-core.ts
```diff
@@ -1,24 +1,64 @@
...
+export type PosterImageStatus = 'none' | 'loading' | 'loaded' | 'error';
 
 export interface PosterState {
   visible: boolean;
+  /** Resolved poster URL, empty when nothing supplied one. */
+  src: string;
+  /**
+   * Resolved placeholder, empty when nothing supplied one. Reaches CSS as
+   * `--media-poster-placeholder` for a skin to paint, so it is a CSS image
+   * rather than a bare URL.
```
decepulis:
```suggestion
   * Reaches CSS as `--media-poster-placeholder` for a skin to paint
```

packages/core/src/core/ui/poster/poster-css-vars.ts
```diff
@@ -0,0 +1,13 @@
+/** CSS custom property names for poster visual state. */
+export const PosterCSSVars = {
+  /**
+   * The resolved poster placeholder, as a CSS image. Absent when nothing
+   * supplied one, so `var()` falls back.
+   *
+   * Set rather than painted: the component holds the value, a skin decides what
+   * to do with it. Painting it as the poster image's own `background-image` is
+   * the cheapest choice — one element covers both, and the poster arrives on top
+   * of it — but a layer of its own can cross-fade instead.
```
decepulis:
```suggestion
   * The resolved poster placeholder, as a CSS image. 
   *
   * It's easiest to pass to your poster's `background-image`,
   * but you could also use this in a separate layer and try some cross-fade animations.
```

packages/media/src/core/types.ts
```diff
@@ -522,6 +522,11 @@ export interface MediaContentData {
   readonly title?: MediaContentValue;
   /** URL of a still image representing the content. */
   readonly poster?: MediaContentValue;
+  /**
+   * A low-resolution stand-in shown behind the poster while it loads, as a CSS
+   * image — usually `url()` around a data URL with the blur already baked in.
+   */
```
decepulis: this is a good comment

packages/react/src/ui/poster/poster.tsx
```diff
@@ -11,18 +11,36 @@ import { renderElement } from '../../utils/use-render';
...
 /**
  * Displays the video poster image. Shows before playback starts, hides after.
  *
+ * Renders an `<img>`, so `srcset`, `sizes`, `loading`, and the rest of the
+ * image attributes are yours. Leave `src` off and the player's resolved
+ * `poster` fills it in, which is why the URL is set on the provider rather than
+ * here.
+ *
+ * The image is decorative unless you say otherwise. Describe a poster that
+ * carries meaning by passing your own `alt`.
```
decepulis, line 29: makes me wonder if we should store `poster-alt` in the store, too...

packages/react/src/ui/poster/poster.tsx
```diff
@@ -11,18 +11,36 @@ import { renderElement } from '../../utils/use-render';
...
+ *
+ * The player's resolved `posterPlaceholder` is set as
+ * `--media-poster-placeholder` for a skin to paint. The packaged skins make it
+ * this image's own `background-image`, so it shows until the poster covers it.
+ *
```
decepulis:
```suggestion
 *
 * The player's resolved `posterPlaceholder` is set as
 * `--media-poster-placeholder` for a skin to paint. 
 *
```
I don't want to promise this and for this comment to go stale

packages/react/src/ui/poster/poster.tsx
```diff
@@ -32,44 +50,83 @@ export const Poster = forwardRef(function Poster(
...
-      props: [elementProps, { 'data-loaded': loaded ? '' : undefined, onLoad: handleLoad }],
+      props: [
+        { alt: '' },
+        elementProps,
+        {
+          // An empty `src` requests the document URL, so leave the attribute off.
+          src: src === '' ? undefined : src,
+          // Offered, not painted: a skin decides what to do with it. On this
+          // element rather than a wrapper, since there is no wrapper. A `style`
+          // of your own is merged over this, key by key.
```
decepulis:
```suggestion
```

packages/skins/src/default/tailwind/components/poster.ts
```diff
@@ -1,32 +1,61 @@
...
+  '[&_img]:w-full',
+  '[&_img]:h-full',
+  '[&_img]:[object-fit:var(--media-object-fit,contain)]',
+  '[&_img]:[object-position:var(--media-object-position,center)]',
+  '[&_img]:rounded-(--media-video-border-radius)',
+  '[&_img]:[background-image:var(--media-poster-placeholder,none)]',
+  '[&_img]:bg-no-repeat',
+  '[&_img]:[background-size:var(--media-object-fit,contain)]',
+  '[&_img]:[background-position:var(--media-object-position,center)]',
+];
```
decepulis: Abstracting these out to variables makes this PR harder to read.

packages/skins/src/minimal/tailwind/components/poster.ts
```diff
@@ -1,32 +1,61 @@
...
+  '[&_img]:h-full',
+  '[&_img]:[object-fit:var(--media-object-fit,contain)]',
+  '[&_img]:[object-position:var(--media-object-position,center)]',
+  '[&_img]:rounded-(--media-video-border-radius)',
+  '[&_img]:[background-image:var(--media-poster-placeholder,none)]',
+  '[&_img]:bg-no-repeat',
+  '[&_img]:[background-size:var(--media-object-fit,contain)]',
+  '[&_img]:[background-position:var(--media-object-position,center)]',
+];
+
```
decepulis: Abstracting these out to a variable makes this PR harder to read. Minimize diff.

site/src/content/docs/reference/poster.mdx
decepulis, line 1: Make sure you've leveraged relevant skills to write this. `write-docs`, `write-references`, `write-guides.mdx`, `writing-style.md`, to name a few

site/src/content/docs/reference/poster.mdx
```diff
@@ -43,33 +41,179 @@ Import the component:
...
+
+The poster exposes it as the `--media-poster-placeholder` custom property; a skin decides what to do with it. The packaged skins paint it as the poster image's own `background-image`, so the poster covers it as it arrives — an image paints a background behind its content like anything else, which is why there is no placeholder component and no second element.
+
+The blur has to be part of the image you supply, not something a skin adds. [`@mux/blurup`](https://www.npmjs.com/package/@mux/blurup) returns exactly this: a tiny frame wrapped in an SVG that carries its own Gaussian blur.
+
+```ts
+import { createBlurUp } from '@mux/blurup';
+
+const { blurDataURL } = await createBlurUp(playbackId);
+```
```
decepulis: I don't want these docs to be mux specific; we'll figure out where to tell the blurup story separately

site/src/content/docs/reference/poster.mdx
```diff
@@ -43,33 +41,179 @@ Import the component:
...
+
+<FrameworkCase frameworks={["html"]}>
+```html
+<video-player poster="poster.jpg" poster-placeholder="url('data:image/svg+xml;charset=utf-8,...')"></video-player>
+```
+</FrameworkCase>
+
+The value is a CSS image, not a URL, so wrap a URL in `url()` yourself. Anything CSS accepts works, including a gradient.
+
+Mind the quoting. A blurred SVG carries double quotes of its own, so `url("...")` around one ends at the first of them and the browser drops the declaration silently — nothing paints, and there is no error. Use `url('...')`, as the examples above do. Written in markup the same value needs its double quotes as `&quot;`, so setting it from script is usually easier.
```
decepulis: Too specific
```suggestion
```

site/src/content/docs/reference/poster.mdx
```diff
@@ -43,33 +41,179 @@ Import the component:
...
+```
+
+A remote URL works too, but it costs a second request that can lose the race with the poster it is standing in for, and it will not be blurred.
+
+## Supply your own image
+
+The component fills in a source for you, but you never lose control of the image. `srcset`, `sizes`, `loading`, `<picture>`, and framework image components all stay available.
+
+<FrameworkCase frameworks={["react"]}>
+The component *is* the `<img>`, so image attributes go straight on it:
```
decepulis:
```suggestion
The component is the `<img>`, so image attributes go straight on it:
```

site/src/content/docs/reference/poster.mdx
```diff
@@ -43,33 +41,179 @@ Import the component:
...
+
+To render something else entirely, use `render`. It receives the resolved `src` along with the rest of the props:
+
+```tsx
+<Poster render={(props) => <Image {...props} alt="" fill />} />
+```
+</FrameworkCase>
+
+<FrameworkCase frameworks={["html"]}>
+`<media-poster>` renders no image of its own — it works like `<picture>`, filling in the `src` of the image you put inside it:
```
decepulis: link to mdn

site/src/content/docs/reference/poster.mdx
```diff
@@ -43,33 +41,179 @@ Import the component:
...
 
 <FrameworkCase frameworks={["html"]}>
 ```css
 media-poster:not([data-visible]) {
-  display: none;
+  opacity: 0;
+}
+```
+
+The image is an ordinary child, so style it as a descendant — there is no part to reach through:
```
decepulis:
```suggestion
The image is an ordinary child, so style it as a descendant
```

If there's no part to reach through, then why'd you mention it? You need to be way more minimal and efficient with your docs. Again, refer to the skills.

decepulis (commented):
A PR review for Claude!

generally, I'm not qualified to say if this PR follows the usual core/ui primitive boundaries. Double check that, too.

decepulis:
Thanks — all of it landed in three commits on top of the four already here, so your inline comments stay attached to the diff they were written against. Squash whenever you like.

## What changed

**Comments.** Cut throughout, and I took your specific suggestions as written. Kept only the class and component JSDoc the reference generator extracts. I also put back four comments I had deleted from the skin tailwind files without meaning to — they were yours, from before this branch.

**`media-poster`: 281 → 214 lines.** It was holding a `#status` field that was purely derived, a cache to avoid writing the same CSS property twice, and a release path that only ever ran on disconnect. The status is now a getter, the property is written where it is read, and "release" is just adopting `null`. Same behavior, same 22 tests.

**Tailwind.** The `slottedImage` / `childImage` locals are gone and the class lists are literal again, so those two diffs are additions rather than rewrites.

**Reference page.** Gone: the CSS-quoting essay, the `@mux/blurup` walkthrough, and the three separate places that re-explained where the placeholder gets painted. `<picture>` links to MDN now.

**One real bug, found on a second pass.** The React load handler recorded whatever `src` it found on the element, but the status compared that against the resolved one. An optimizing image component rewrites the attribute, so they never matched and the poster stayed `data-loading` forever after loading — which breaks exactly the `render` escape hatch the docs advertise. Fixed by recording the `src` the render asked for. There's a test, and I confirmed it fails without the fix.

## On core/ui boundaries

Mostly clean. `poster-css-vars.ts` matches seven siblings in `packages/core/src/core/ui/`. `PosterMediaState` composing two feature slices matches `LiveButtonMediaState`, which composes three. `PosterCore` touches no DOM, and the slot walking and attribute writes live in the bindings.

One thing does cross a line, and I'd like your call rather than my guess:

`packages/media/src/core` contains no other CSS. Every CSS constant in that package sits under `src/dom`. But `posterPlaceholder` now puts a CSS value in `src/core`, right next to `poster` and `storyboard`, which are plain URLs. I left it because you wrote "a CSS image" in two of your own suggestions and called the third comment a good one — but the alternative is storing a URL and wrapping it in each binding, which is what `main` did. That would drop the quoting footgun and lose gradients.

## Also worth naming

`main`'s skins blurred the placeholder themselves, with `filter: blur(var(--media-poster-placeholder-blur, 20px))` on a `::before` layer. Painting it as the image's own background means that layer is gone and so is `--media-poster-placeholder-blur`, so blur now has to be baked into the image. That follows from the design, but it removes a skin variable and should be a decision, not a side effect.

## Not acted on

An author's image that changes its own `src` in place, after it has already loaded, keeps `data-loaded` through the next fetch — we listen for `load` and `error` but don't watch the attribute. Closing it costs a `MutationObserver` per image. Given the direction of this review I left it alone; happy to add it if you'd rather have the guarantee.

## `poster-alt`

I think you're right, and not in this PR. The argument is the same one that moved the URL: what the poster depicts is a property of the content, so a media donor could supply it. `contentTitle` is already a localized content string in the store, so there's precedent. It would need the same "only when the author supplied none" rule the `src` has. Worth its own issue.

---

decepulis:
Absorbed — #1998 has `posterPlaceholder` now, and it's since moved on. Ready for you to rebase.

`151a69f2` landed **unchanged, at its original SHA**. I fast-forwarded rather than cherry-picked, since its parent was already this branch's tip. That should make your rebase drop the duplicate cleanly rather than relying on a patch-id match.

I left the two JSDoc comments alone, as asked — they still say nothing about URL vs CSS image.

## The part that matters for your rebase

The branch is no longer at `0414637`. It's now `7f4778b2`, which merged `main` up through #2040. That merge conflicted in `packages/media/src/dom/mux/media.ts` and the resolution may reach your work:

- #2033 split `mux/utils` into `mux/drm` and `mux/source`, and `MuxSource` is now declared in `media.ts` rather than imported
- `createMuxPosterURL` and `createMuxStoryboardURL` now come from `./source`
- #2014 changed the source setter's DRM handoff from `engine: withMuxDrm(source)` to a `...withMuxDrm(source)` spread

Only the imports actually conflicted; the setter body auto-merged, with `#refreshContentData` still running before `super.source =` and the dispatch still after.

## Verified on the merged branch

`pnpm build`, `pnpm typecheck`, `pnpm lint`, `pnpm check:workspace` clean. Tests: media 574, core 1412, html 307, react 394 — all passing. A full `pnpm build` is needed before `pnpm typecheck` after picking this up; the spf relocation leaves stale declarations that produce misleading `Cannot find module '@videojs/spf/simple-hls'` errors otherwise.

Title and body here are updated to cover the placeholder.

---

apps/sandbox/app/shared/sources.ts
```diff
@@ -270,9 +270,16 @@ export function getPosterSrc(source: SourceId): string | undefined {
   return id ? `https://image.mux.com/${id}/thumbnail.webp${imageQuery(source, 'poster')}` : undefined;
 }
 
-export function getPlaceholderSrc(source: SourceId): string | undefined {
+/**
+ * A CSS image. this upscales a 20px thumbnail instead,
+ * and the browser's own smoothing does the blurring.
```
decepulis: is this comment true?

site/src/content/docs/reference/poster.mdx
```diff
@@ -43,33 +41,167 @@ Import the component:
 
 The poster is visible before playback starts. Once the user plays or seeks, the poster hides permanently — pausing does not bring it back. The poster reappears when a new source is loaded.
 
+Set the URL on the player, not on this component:
```
decepulis: This isn't actually accurate, is it? In react, you can set a src directly. in html, you can set a src on a child. Setting it on the player is just one way of doing it. Users should be able to self-sort depending on their use case. "You can set the poster url on the player directly. This is useful if you're trying to change the poster inside a skin, or want to set all your media metadata in one place." Something like that.

site/src/content/docs/reference/poster.mdx
```diff
@@ -43,33 +41,167 @@ Import the component:
...
+```
+</FrameworkCase>
+
+<FrameworkCase frameworks={["html"]}>
+```html
+<video-player poster="poster.jpg" poster-placeholder="url('data:image/svg+xml;charset=utf-8,...')"></video-player>
+```
+</FrameworkCase>
+
+The value is a CSS image, not a URL, so wrap a URL in `url()` yourself. Anything CSS accepts works, including a gradient, and any blur has to be part of the image. The poster exposes it as the `--media-poster-placeholder` custom property for a skin to paint.
```
decepulis:
```suggestion
Anything CSS accepts works, including a gradient, and any blur has to be part of the image. The poster exposes it as the `--media-poster-placeholder` custom property for a skin to paint.
```

site/src/content/docs/reference/poster.mdx
```diff
@@ -43,33 +41,167 @@ Import the component:
...
+
+## Supply your own image
+
+The component fills in a source for you, but you never lose control of the image. `srcset`, `sizes`, `loading`, [`<picture>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture), and framework image components all stay available.
```
decepulis: on second thought, don't link to picture

site/src/content/docs/reference/poster.mdx
```diff
@@ -43,33 +41,167 @@ Import the component:
...
+
+<FrameworkCase frameworks={["react"]}>
+The component is the `<img>`, so image attributes go straight on it:
+
+```tsx
+<Poster srcSet="poster-480.jpg 480w, poster-1080.jpg 1080w" sizes="100vw" />
+```
+
+To render something else, use `render`. It receives the resolved `src`:
```
decepulis: God I hate how AI writes. "It receives the resolved `src`". Why should users know what "resolved" means in this context? Either define it or don't use it. Adhere to the jargon guidance in writing-style.md more diligently; consider the user arriving at this guide who doesn't have the same context you do.

site/src/content/docs/reference/poster.mdx
```diff
@@ -43,33 +41,167 @@ Import the component:
...
+  background-size: cover;
 }
 ```
+
+The property is set on `media-poster`, so a layer of your own can read it instead and cross-fade.
 </FrameworkCase>
 
 <FrameworkCase frameworks={["react"]}>
-React renders an `<img>` element. Add a `className` to style it:
+React renders the `<img>` itself. Add a `className` and use it as the selector:
```
decepulis:
```suggestion
React renders an `<img>` element. Add a `className` to style it:
```
Why did you rephrase a perfectly legible sentence and made it more confusing

decepulis (commented):
docs could use another pass. a few comments here or there.

decepulis:
Two commits: the docs pass, and one real bug Bugbot caught.

## Docs

Took your suggestions on 72 and 164 as written. The rest:

**Line 44 was wrong, and you were right about why.** "Set the URL on the player, not on this component" contradicts the section two headings below it. Now it says the player is one option and gives the reasons to pick it, close to your wording. The same false claim was in the `Poster` JSDoc — "which is why the URL is set on the provider rather than here" — so that went too.

**Jargon.** "Resolved" is gone from the prose in both places. The accessibility line leaned on it to make a point it could make plainly: a URL says nothing about what the image shows. Also dropped both MDN links on `<picture>`, not just the one you flagged — the only other reference page that links MDN links it once, so two on one page was mine, not the house style.

While in there: two sentences had one pronoun doing two jobs, and "hides permanently" was three sentences from "shows it again."

One "resolved" survives, in `poster-css-vars.ts`. That's your wording from the last round, so I left it.

## "Is this comment true?"

True in substance, not as English. The 20px thumbnail is real, and the upscaling really is what blurs it — no filter anywhere. But "instead" lost its other half when the blurup explanation came out, so it points at nothing. It also ran to two lines for a one-line fact. Now:

```ts
/** A 20px thumbnail, upscaled by the browser into a blur. */
```

## The bug

Bugbot found a real one and I'd missed it. `Poster` filled `src` from the store whenever no `src` prop was passed — including when you passed a `srcSet`. HTML's `hasSource` treats either attribute as yours, so the platforms disagreed. Worse, the file disagreed with itself: the load-status line already counted `srcSet` as a source.

It bites with `x` descriptors, where `src` joins the candidate set as the 1x entry. `<Poster srcSet="poster@2x.jpg 2x" />` would load the player's poster on a 1x display. The docs example uses `w` descriptors, where `src` is ignored outright, which is why nothing caught it.

There's a test, and I confirmed it fails without the fix.

## Checks

React 409, site 531, site build renders both framework variants. `pnpm typecheck` reports 12 errors in `packages/store/src/html/controllers/tests/snapshot-controller.test.ts` — identical with and without my changes, and CI's typecheck is green, so it looks local to this environment rather than the branch.

---

packages/react/src/ui/poster/poster.tsx
```diff
@@ -32,44 +48,78 @@ export const Poster = forwardRef(function Poster(
...
+      posterPlaceholder: metadata?.posterPlaceholder ?? '',
+    });
+  }
+
+  const { src: authoredSrc, srcSet: authoredSrcSet } = elementProps as { src?: string; srcSet?: string };
+
+  // Either authored source makes the image yours, matching `hasSource` in the
+  // HTML element. Filling `src` alongside a `srcset` would inject the player's
+  // poster into your candidate set as the 1x entry.
+  const src = authoredSrc ?? (authoredSrcSet === undefined && playback ? core.getState().src : '');
```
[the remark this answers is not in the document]
decepulis: Right, and I introduced it one commit earlier. Fixed in 73cee0f.

The outcome is now keyed on both halves of the source rather than `src` alone, so swapping a `srcSet` re-keys it. React sees that as a prop change, so it costs nothing here. The HTML element still can't see an author mutate an image in place — that limitation is unchanged and already called out in the description.

The test needed one piece of scaffolding worth flagging: happy-dom reports a `srcset`-only image as `complete`, so the mount effect that catches already-cached images fires and labels it. A browser fetching a new candidate reports `complete` as false, so the test stubs it to match. I confirmed the test fails without the fix.

---

packages/html/src/ui/poster/poster-element.ts
```diff
@@ -1,31 +1,214 @@
...
+          `<${this.localName}> resolved a poster but has no image to put it in. ` +
+            `Add one as a child: <${this.localName}><img alt=""></${this.localName}>`
+        );
       }
+      return;
+    }
+
+    if (!this.#owned) return;
+
+    if (src === '') {
```
luwes:
```suggestion
    if (!src) {
```

feels better to me, I know it probably can't be nullish but still. also a code ⛳

decepulis: Taken, in 77fe012 — and swept, since there were six of these across the two poster bindings and fixing only the flagged one would have left both files mixing the styles.

Two read better flipped to the positive branch rather than negated, so the placeholder writes now go `if (placeholder) setProperty` / `else removeProperty`, and React's style prop is `state.placeholder ? {...} : undefined`.

You're right that none of them can be nullish today — they're all `string`, either from `PosterCore.getState()` or defaulted with `?? ''`. The guard just stops caring if that ever widens.

---

luwes (approved):
LGTM

packages/react/src/presets/types.ts
```diff
@@ -10,7 +11,15 @@ export type BaseSkinProps<T = unknown> = PropsWithChildren<
...
+   * still comes from the player, so a framework image component is passed the
+   * resolved `src` along with the rest of the image props.
+   *
+   * @example
+   * ```tsx
+   * <VideoSkin renderPoster={(props) => <Image {...props} alt="" fill />} />
+   * ```
+   */
+  renderPoster?: RenderProp<Poster.State> | undefined;
 };
```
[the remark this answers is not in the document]
decepulis: Real, and confirmed by running the generator. `resolvePropsInterface` stripped `BaseSkinProps` and rewrote the skin's own props type, but left the alias in between — so the ejected file carried `type BaseVideoSkinProps = BaseSkinProps<T> & {…}` pointing at a type that was no longer there, and its props interface was missing a prop the component destructured.

Fixed in 83d700b: every `Base*SkinProps` alias is removed, and the members they added are carried into the flat interface, minus their JSDoc — those comments describe the skin component, not the player the ejected file exports. `RenderProp` and `Poster` were already resolving to public `@videojs/react` exports in the output. Two tests cover it, both failing without the fix.

---

packages/html/src/ui/poster/poster-element.ts
```diff
@@ -1,31 +1,214 @@
...
+    // An image that steps aside keeps downloading whatever we pointed it at.
+    if (this.#owned) this.#image?.removeAttribute('src');
+    this.#imageEvents?.abort();
+    this.#imageEvents = null;
+
+    this.#image = next;
+    this.#owned = next !== null && !hasSource(next);
+    // An image that finished before we started listening never reports again.
+    // Only a decoded one counts: a failed image is `complete` as well.
+    this.#outcome = next?.complete && next.naturalWidth > 0 ? 'loaded' : 'pending';
```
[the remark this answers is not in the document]
decepulis: Real. Reproduced with a test before fixing: an image that arrives already failed stayed at `data-loading` forever, since its `error` event was spent before we started listening. Reachable by re-slotting a cached failure, or by moving the element, which adopts again.

Fixed in c4d8fd1 — both outcomes are now read off a complete image, gated on it having a source, since an image with nothing to fetch is `complete` too. That matches what React already did.

Worth flagging: happy-dom never fetches, so it reports an image with a source as complete with nothing decoded — identical to a browser's failure. Two existing tests were passing only because the old code required `naturalWidth > 0` to conclude anything, so it read those as pending. The harness now marks author-slotted images as mid-fetch, which is what they are. The new test still fails without the source change, so the harness fix isn't what's carrying it.

---

packages/html/src/ui/poster/poster-element.ts
```diff
@@ -1,31 +1,217 @@
...
+    this.#image = next;
+    this.#owned = next !== null && !hasSource(next);
+    this.#outcome = 'pending';
+
+    if (!next) return;
+
+    // An image that finished before we started listening never reports again, so
+    // its outcome has to be read off it. Only one with a source counts: an image
+    // with nothing to fetch is `complete` as well.
+    if (next.complete && hasSource(next)) this.#outcome = next.naturalWidth > 0 ? 'loaded' : 'error';
```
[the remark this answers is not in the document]
decepulis: Right, and it's the case I waved off an hour ago. Per spec, `complete` returns true when both `src` and `srcset` are omitted — a check on the image's own attributes, with no regard for what a parent `<picture>` selected. So my `hasSource` gate let a mid-fetch picture poster through as a failure.

Fixed in ad36d15 — the adopt-time read is gated on the image's own source now, not on `hasSource`:

```ts
function reportsOwnRequest(img: HTMLImageElement): boolean {
  return !!img.getAttribute('src') || img.hasAttribute('srcset');
}
```

The picture test now asserts `complete === true` and `naturalWidth === 0` on that image directly, so it pins the spec behavior rather than relying on happy-dom agreeing by accident.

Known limit, worth stating rather than hiding: a picture whose candidate failed *before* adoption stays at `data-loading`. Nothing on the element distinguishes that from one still fetching — `complete`, `currentSrc`, and `naturalWidth` read identically in both cases — so it waits on `error`, which for that image has already been spent. Reachable only by re-slotting a picture whose failure is cached.

React had the same gap through a `render` override that sources the image itself, so it got the same guard and a test. Both new tests fail without their fix.

---

packages/html/src/ui/poster/poster-element.ts
```diff
@@ -1,31 +1,225 @@
...
+
+    this.#image = next;
+    this.#owned = next !== null && !hasSource(next);
+    this.#outcome = 'pending';
+
+    if (!next) return;
+
+    // An image that finished before we started listening never reports again, so
+    // its outcome has to be read off it while `complete` still means something.
+    if (next.complete && reportsOwnRequest(next)) this.#outcome = next.naturalWidth > 0 ? 'loaded' : 'error';
```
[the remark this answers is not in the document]
decepulis: Right — I over-corrected. `naturalWidth > 0` needs no interpretation: intrinsic dimensions mean something decoded, whatever `complete` claims about the request. Gating the whole read on the image's own source threw that away, so a cached picture candidate stuck at `data-loading`.

Fixed in 5e79824 by splitting the two directions, which aren't symmetric:

```ts
if (next.naturalWidth > 0) this.#outcome = 'loaded';
else if (next.complete && reportsOwnRequest(next)) this.#outcome = 'error';
```

Pixels prove success outright. Their absence only means failure where `complete` describes a request — the one thing a `<picture>` can't give us. React reads an image the same way now, so the two stay in step.

Tests on both sides, each failing without the change. The documented limit stays exactly where it was: a picture whose candidate failed before adoption, where every readable property matches one still fetching.

---

packages/react/src/ui/poster/poster.tsx
```diff
@@ -32,44 +49,90 @@ export const Poster = forwardRef(function Poster(
...
   useEffect(() => {
     const img = imgRef.current;
-    if (img?.complete && img.naturalWidth > 0 && img.getAttribute('src') === src) {
-      setLoadedSrc(src);
+    if (!img) return;
+
+    if (img.naturalWidth > 0) setOutcome({ src, srcSet: authoredSrcSet, status: 'loaded' });
+    else if (img.complete && (img.getAttribute('src') || img.hasAttribute('srcset'))) {
+      setOutcome({ src, srcSet: authoredSrcSet, status: 'error' });
     }
```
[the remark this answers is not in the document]
decepulis: Correct, and it's the staleness bug from earlier in this PR coming back through a different door. A browser keeps the previous request's pixels until the pending one settles, so re-reading the same node reported the old image's dimensions against the new source and skipped `data-loading`.

The real defect was re-reading at all. That read exists for one case — an image already settled before we could listen, which in React means hydrated from server markup — and that case only happens once. Fixed in c6493f4 by making it a mount effect; `load` and `error` cover every source after the first, which is what the HTML element does by reading an image once when it adopts it.

Two things fell out of that. The `ImageOutcome` keying still does the invalidating on a source change, so nothing else moved. And the `complete` stub in the srcSet test is now unnecessary — it was also naming happy-dom in a package that runs jsdom — so it's gone, and that test passes on the real behavior.

New test covers your exact scenario: load `a.jpg`, pin `naturalWidth` to 1280 the way a browser would, swap to `b.jpg`, expect `data-loading`. Fails without the change.

Four rounds on this function, three of them mine, all the same mistake in different clothes: treating an image property as if it described the current request. If a fifth turns up I'd rather lift this read into one shared helper than keep two copies in step — happy to do that here if you'd prefer it now, though it means a new `@videojs/core/dom` export.

---

mihar-22 (commented):
I'd recommend pulling "placeholder" feature out of #1998 and this PR #2039 into its own PR which can stack on poster work if needed. PRs are mixing concerns and it's just hard for me to review. 

I'd pull the "poster" work out of #1998 out into this PR. 

Keeping features isolated will help with reviewing and discussing. It also ends up producing a better changelog, otherwise the PR titles are quite unclear atm. 

---

Just a note on this title, "make `<media-poster>` act like `<picture>`". It's not exactly acting like it (depends I guess on framing). It is loose inspiration here but it's kind of the inverse, where `picture` treats `src` set on `` as fallback, and here we treat no `src` as meaning we own and control it. I'd probably focus the title on `feat(packages): media-poster now sets src on img` - something like that anyway

decepulis:
Split and renamed.

**The placeholder is out of here** — it's #2063 now, stacked on this branch. This PR keeps placeholder behavior exactly as it is on the base branch: the skin `placeholder` prop, the `placeholdersrc` attribute, and the `::before` layer with the image held transparent until it loads. So each PR stands on its own, and #2063 is the one that moves the value onto the player and paints it on the image instead.

**#1998 is untouched.** Darius's call — the store field stays there, since all three competing poster designs added it identically.

**You were right about the framing, and it wasn't only the title.** The same "works like `picture`" claim was in the element's JSDoc, the design record, and the reference page. All three now say what actually happens: `picture` treats the `src` on its `img` as the fallback, while here an image with no source of its own is the one we fill in. Title is `feat(packages)!: media-poster now sets src on img, from player state`.

I also rewrote this description from scratch. The earlier revisions were mangled — this repo's PR body pipeline strips tag names, so every `media-poster` and `picture` had become empty backticks, including in the title you quoted. Tag names are written bare now so they survive.

---

decepulis:
Heads up — I pushed to this branch and to #2063, so pull both before your next push.

At @decepulis's request the store work moved out of #1998, which is now the media half only: Mux, Vimeo, and the SPF Mux Medias donating `contentData`. Each resolved value follows the component that reads it.

**Here (`7e44f073`)** — `poster` on `MediaMetadataState`, the `poster` / `defaultPoster` tiers in `metadataFeature`, `setPoster` / `setDefaultPoster`, and the `PlayerFeatureConfig.action` JSDoc that lets config point at a public setter. Your `poster.tsx:58` read of `metadata?.poster` keeps working; it would have broken on the base merge without this.

**#2063 (`71c8f5ec`)** — the same shape for `posterPlaceholder`, plus that key rejoining `MediaContentData`. Two merge conflicts came up, in `packages/media/src/core/state.ts` and `types.ts`; I kept your JSDoc on both, since "as a CSS image" is more precise than what #1998 had.

Verified on each branch after the merge: typecheck, lint, `check:workspace`, and the core, media, react, html, and spf suites. If react/html tests fail for you with missing `poster` exports, it's stale declarations — run `pnpm build` first.

---

packages/core/src/core/ui/poster/poster-core.ts
```diff
@@ -1,24 +1,45 @@
...
+/** Reported by the binding, which holds the image. */
+export type PosterImageStatus = 'none' | 'loading' | 'loaded' | 'error';
 
 export interface PosterState {
   visible: boolean;
+  /** Resolved poster URL, empty when nothing supplied one. */
+  src: string;
+  loading: boolean;
+  loaded: boolean;
+  error: boolean;
```
mihar-22, line 22: Missing some jsdoc

packages/core/src/core/ui/poster/poster-data-attrs.ts
```diff
@@ -3,4 +3,7 @@ import type { PosterState } from './poster-core';
 
 export const PosterDataAttrs = {
   visible: 'data-visible',
+  loading: 'data-loading',
+  loaded: 'data-loaded',
+  error: 'data-error',
```
mihar-22, line 12: missing some jsdoc

packages/core/src/dom/store/features/metadata.ts
```diff
@@ -30,21 +38,33 @@ export const metadataFeature = definePlayerFeature({
       // another name there.
       html: { attribute: 'content-title' },
     },
+    poster: {
+      action: SET_USER_POSTER,
+      state: USER_POSTER,
+    },
```
mihar-22, line 44: would this be `content-poster` attribute in HTML?

packages/html/src/ui/poster/poster-element.ts
mihar-22, line 1: We shouldn't need shadow DOM here. It can observe and query the light DOM for the image.

packages/html/src/ui/poster/poster-element.ts
```diff
@@ -1,21 +1,130 @@
...
+    } else if (element instanceof HTMLImageElement) {
+      return element;
+    } else {
+      const img = element.querySelector('img');
+      if (img) return img;
+    }
+  }
+
+  return null;
+}
```
mihar-22: we can likely delete this if we're just querying for img or picture from light DOM

packages/html/src/ui/poster/poster-element.ts
```diff
@@ -1,21 +1,130 @@
...
+ * Renders no image of its own, so include one:
+ * `<media-poster><img alt=""></media-poster>`. Inside a skin, an
+ * `<img slot="poster">` of yours replaces the one the skin carries.
+ */
+export class PosterElement extends MediaElement {
   static readonly tagName = 'media-poster';
 
   static get observedAttributes(): string[] {
     // biome-ignore lint/complexity/noThisInStatic: intentional use of super
     return [...super.observedAttributes, 'placeholdersrc'];
```
mihar-22, line 77: Should we leave `placeholdersrc`?

packages/html/src/ui/poster/poster-element.ts
```diff
@@ -1,21 +1,130 @@
...
+ * Whether `complete` on this image describes a request. It is also true for one
+ * that omits both `src` and `srcset`, whatever a parent `<picture>` is fetching
+ * on its behalf, so only an image sourced from its own attributes can be read.
+ */
+function reportsOwnRequest(img: HTMLImageElement): boolean {
+  return !!img.getAttribute('src') || img.hasAttribute('srcset');
+}
+
+/** What the current source has reported so far. */
+type ImageOutcome = 'pending' | 'loaded' | 'error';
```
mihar-22: `ImageLoadState`?

packages/react/src/ui/poster/poster.tsx
```diff
@@ -11,18 +11,32 @@ import { renderElement } from '../../utils/use-render';
 
 export interface PosterProps extends UIComponentProps<'img', PosterCore.State> {}
 
+/** What the image reported, against the source it reported for. */
+interface ImageOutcome {
+  src: string;
+  srcSet: string | undefined;
+  status: 'loaded' | 'error';
+}
```
mihar-22: "reporting" is strange terminology appearing (prev files too), maybe `ImageSource` and it just happens to have a `state` which is `ImageLoadState` or something

packages/core/src/core/ui/poster/poster-core.ts
```diff
@@ -1,24 +1,45 @@
-import type { MediaPlaybackState } from '@videojs/media';
+import type { MediaMetadataState, MediaPlaybackState } from '@videojs/media';
+
+export type PosterMediaState = Pick<MediaPlaybackState, 'started'> & Pick<MediaMetadataState, 'poster'>;
+
+/** Reported by the binding, which holds the image. */
+export type PosterImageStatus = 'none' | 'loading' | 'loaded' | 'error';
```
mihar-22: `PosterImageLoadState`?

site/src/components/home/Demo/baseCode.ts
```diff
@@ -30,8 +30,8 @@ const Player = createPlayer({ features: videoFeatures });
 
 export function VideoPlayer() {
   return (
-    <Player.Provider>
-      <${skinComponent} poster="${VJS10_DEMO_VIDEO.poster}">
+    <Player.Provider poster="${VJS10_DEMO_VIDEO.poster}">
```
mihar-22: My PR should've updated and removed `Player.Provider` from all examples. It should be destructed `{ Player: VideoPlayer, usePlayer }`, just making sure we catch and update these.

site/src/content/docs/reference/poster.mdx
```diff
@@ -23,53 +23,153 @@ import basicUsageHtmlTs from "@/components/docs/demos/poster/html/css/BasicUsage
...
 ## Behavior
 
-The poster is visible before playback starts. Once the user plays or seeks, the poster hides permanently — pausing does not bring it back. The poster reappears when a new source is loaded.
+The poster shows before playback starts. Once the user plays or seeks, it hides; pausing does not bring it back. Loading a new source shows it again.
+
+You can set the poster URL on the player. This is useful when you want to change the poster from inside a skin, or keep all your media metadata in one place:
+
+<FrameworkCase frameworks={["react"]}>
+```tsx
+<Player.Provider poster="poster.jpg" />
```
mihar-22: No more `Player.Provider` 🙏

mihar-22 (approved):
Great! Few things to work through but overall pretty damn good 😎 

Let's see if we can weed out all the `Player.Provider` inconsistencies here or in a follow-up.

packages/core/src/core/ui/poster/poster-core.ts
decepulis: Claude, make it so.

packages/html/src/ui/poster/poster-element.ts
decepulis: Claude, get on it.

packages/html/src/ui/poster/poster-element.ts
decepulis: placeholder is outside the scope of this PR! This is purely posterville. If we decide in design to remove it, I'll repurpose #2063. Design doc coming soon.

packages/html/src/ui/poster/poster-element.ts
decepulis: Claude I trust you to make the right call here

packages/core/src/dom/store/features/metadata.ts
decepulis: Nope! `poster` means nothing on `video-player` so requires no change

packages/html/src/ui/poster/poster-element.ts
decepulis: Agreed, and done in f8ccee3.

The shadow root held only a slot and a `display: block` rule. The slot wasn't composing anything — the element renders nothing of its own — it was change detection, firing `slotchange` when children moved. A `MutationObserver` on `childList` does that job, so the element now walks its own children to find the image.

The walk still follows a skin's forwarding slot to whatever fills it, so an author's `img slot="poster"` still displaces the fallback image the skin carries.

One thing that went with it: `:host { display: block }`. Every skin and the docs demo position the element absolutely, which blockifies it anyway, so nothing moved — but a bare unstyled `media-poster` is now inline like the rest of the elements in the package.

Added a test for the case the shadow slot used to cover, an image appended as a plain child after mount. It fails with the observer removed.

Also picked up the two jsdoc comments in the same commit.

---

packages/core/src/core/ui/poster/poster-core.ts
decepulis: Done in `1131e73f`, along with the two related notes on the element and the React binding — they were one idea, so I did them together:

- `PosterImageStatus` → `PosterImageLoadState`, `setImageStatus` → `setImageLoadState`, `PosterCore.ImageStatus` → `ImageLoadState`
- `media-poster`: `ImageOutcome` → `ImageLoadState`, `#outcome` → `#imageLoadState`
- React: `ImageOutcome` → `ImageSource` carrying a `state`, as you suggested

On "reporting" — the type names were only half of it, so I took the prose with them. The one that earned its rename is `reportsOwnRequest` → `hasOwnSource`: it sits directly under `hasSource`, and the pair now names the difference instead of hiding it. `hasSource` counts a `<picture>`'s `<source>`; `hasOwnSource` counts only the image's own attributes, which is the distinction `complete` actually turns on.

Behavior is unchanged — core 1424, html 364, react 458.

Separately, both "missing some jsdoc" notes were already covered in `f8ccee3c`, which landed after you wrote them: `PosterDataAttrs` has a line per attribute, and `PosterCore` and the `PosterState` members are documented. I left the `PosterState` interface itself bare because none of its siblings document theirs.

---

packages/react/src/player/tests/create-player.test.tsx.orig
```diff
@@ -0,0 +1,513 @@
...
+      // Container should render
+      const containerEl = container.querySelector('[data-testid="container"]');
+      expect(containerEl).toBeTruthy();
+
+      // Video should render inside container
+      const videoEl = container.querySelector('[data-testid="video"]');
+      expect(videoEl).toBeTruthy();
+    });
+  });
+});
```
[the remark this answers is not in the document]
decepulis: Right, and mine — it came from a `patch` run while I was resolving the main merge, and I committed it along with the resolution in `889131c2`. Removed in `b50d0d50`, and `.gitignore` now covers `*.orig` and `*.rej` so the next one can't be staged by accident.

Nothing imported it, so it was dead weight rather than a live regression, but it was a stale copy of the real test file sitting in the tree at its pre-merge revision — which is exactly how it still had `contentTitle` on it.

---

packages/react/src/ui/poster/poster.tsx
```diff
@@ -30,44 +44,100 @@ export const Poster = forwardRef(function Poster(
...
-      props: [{ alt: '' }, elementProps, { 'data-loaded': loaded ? '' : undefined, onLoad: handleLoad }],
+      props: [
+        { alt: '' },
+        elementProps,
+        {
+          // An empty `src` requests the document URL, so leave the attribute off.
+          src: src || undefined,
+          onLoad: handleLoad,
+          onError: handleError,
+        },
```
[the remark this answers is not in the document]
decepulis: Checked this against `b0bb944d` and it doesn't hold — the empty string is scrubbed at the lines flagged here.

`poster.tsx#L134-139`, inside the cited range:

```ts
{
  // An empty `src` requests the document URL, so leave the attribute off.
  src: src || undefined,
  onLoad: handleLoad,
  onError: handleError,
}
```

`mergeProps` is last-wins for non-event keys, so that object also scrubs an author-passed `src=""`. With the poster unset, a `render`/`renderPoster` override receives the key `src` with value `undefined` (never `''`), and React omits the attribute — verified empirically, and pinned by `poster/tests/poster.test.tsx#L36`: *"leaves the src attribute off when nothing resolved one"*. The HTML side guards equivalently at `poster-element.ts#L212` (`img.removeAttribute('src')`).

On the mount change: `<Poster>` is indeed unconditional now, but that isn't a regression for this scenario. On `main`, `Poster` never read the store — `const src = (elementProps as { src?: string }).src` — so a render-prop override there was truthy, mounted, and got `src: undefined` *always*. This branch is strictly better.

Since `undefined` and `''` are both falsy, the guard doesn't change what a framework image component sees either way.

Resolving. Separately, the documented snippet in `poster.mdx` has a real (unrelated) typing problem — `RenderProp` resolves to `React.HTMLAttributes`, which has no `src` — which I'm raising with the author rather than folding in here.

---

Issue #1845, opened 2026-07-14
Feature: SPF Mux Data Integration


decepulis:
Wire Mux Data analytics into SPF-backed media. The hls.js path's `MuxData` component (`packages/core/src/dom/media/mux/mux-data.ts`) monitors playback by passing the `Hls` instance and `Hls` constructor to `mux-embed`. SPF has no `Hls` instance, so Mux Data needs an engine-agnostic (metrics-emitting) integration path for the SPF-backed `<mux-video>` (#1836) to reach analytics parity with the hls.js-backed element.

> [!NOTE]
> This is an AI-generated spec. Requirements should be independently confirmed.

decepulis:
We're not writing an hls.js-equivalent data integration. Just aiming for the low-hanging basics

cjpillsbury:
Moving out of the GA milestone (due 2026-09-30). This was scoped as lower priority relative to the rest of the GA cut and won't land in the remaining window; it stays open as post-GA SPF work. No change to scope.

Issue #1343, opened 2026-04-14
Bug: DashVideo Breaks SSR — Top-Level dashjs Import Reads `window`


decepulis:
> [!NOTE]
> **Triage estimate** — cycle priority and story points were assigned during an AI-assisted planning pass and are rough first guesses. Confirm or adjust when you pick this up.

## Problem

The React media components (`@videojs/react/media/*`) are not SSR-safe. Importing any of them server-side crashes with `ReferenceError: window is not defined`.

This was discovered while adding media element demos to the docs site (#1342) — Astro's SSR prerender fails when a React demo imports e.g. `DashVideo` from `@videojs/react/media/dash-video`.

## Root cause

The import chain pulls in browser-only libraries at module scope:

```
@videojs/react/media/dash-video
  → DashVideo (has 'use client', but that only helps React Server Components)
    → @videojs/core/dom/media/dash
      → import * as dashjs from 'dashjs'   ← accesses window at load time
```

Same pattern for all hls.js-based elements:

```
@videojs/core/dom/media/hls/index.ts:    import Hls from 'hls.js'
@videojs/core/dom/media/hls/hlsjs.ts:    import Hls from 'hls.js'
@videojs/core/dom/media/hls/preload.ts:  import Hls from 'hls.js'
@videojs/core/dom/media/dash/index.ts:   import * as dashjs from 'dashjs'
```

The `'use client'` directive on the React components prevents them from running as Server Components in React frameworks, but it does **not** prevent the module from being imported and evaluated server-side. Any SSR framework (Astro, Next.js, Remix) that imports these modules during build or server render will crash.

## Affected components

All media elements in `@videojs/react/media/*`:
- `HlsVideo`, `NativeHlsVideo`, `SimpleHlsVideo` (via hls.js)
- `DashVideo` (via dashjs)
- `MuxVideo`, `MuxAudio` (via hls.js)

## Expected behavior

Importing `@videojs/react/media/hls-video` (or any media component) in a Node.js / SSR environment should not throw. The component doesn't need to *render* anything server-side, but the module must be importable without crashing.

## Possible approaches

- Lazy-load the engine libraries inside the host class constructors via dynamic `import()` instead of top-level imports
- Guard the imports behind a browser environment check
- Use conditional exports (`"node"` vs `"browser"`) in the core package to provide SSR-safe stubs

luwes:
🤞 https://github.com/Dash-Industry-Forum/dash.js/pull/4929

[...]

PR #2040, opened 2026-08-10
fix(media): hook the media's actual playback engine in mux data

### LINKED ISSUE #2038: Bug: Mux Data Assumes media.engine Is an hls.js Instance
## Context

`MuxData` (`packages/media/src/dom/mux/mux-data.ts`) reads `media.engine` and hands it to
`mux-embed` as `hlsjs`, together with a statically imported `Hls` constructor. Its media
contract types the engine as hls.js only:

```ts
export interface MuxDataMedia extends EventTarget {
  readonly engine?: HlsJsMedia['engine'];
  readonly src: string;
}
```

Nothing enforces that assumption at the registration boundary. `MediaComponent.setMedia()`
takes any media host, so Mux Data can be — and is meant to be — placed beside any media:

- `<dash-video>` / `<DashVideo>` exposes `engine` as a `dashjs.MediaPlayerClass`. That
  instance is currently passed to `mux-embed` as `hlsjs`, so hls.js-shaped event wiring is
  applied to a dash.js player and engine-level telemetry never arrives.
- Native HLS and raw `<video>` hosts expose no JS engine, so no engine option is sent
  (correct), yet the hls.js integration is still configured and hls.js is still pulled in.
- SPF adapters expose `engine` as an SPF `Composition`, which is not an engine handle at
  all. The SPF integration itself is tracked separately in #1845.

## User impact

DASH playback monitored with Mux Data loses engine-level data (rendition switches, request
timing/throughput, engine errors) and can be attributed with hls.js-shaped assumptions.
Anyone adding Mux Data to non-hls.js playback also pays the hls.js bundle cost.

## Reproduction

1. Render a player with `<dash-video src="...mpd">` and a sibling `<mux-data>` (or the React
   `<DashVideo />` + `<MuxData />` equivalent).
2. Inspect the options passed to `Mux.monitor()`.
3. The dash.js `MediaPlayerClass` arrives as `hlsjs`, with the hls.js constructor as `Hls`;
   no dash.js integration is configured.

## Acceptance criteria

- Mux Data selects the `mux-embed` integration that matches the media it is registered
  with: hls.js engines use the hls.js path, dash.js engines use the dash.js path, and media
  with no JS engine are monitored from the media element alone.
- An engine Mux Data does not recognize (for example an SPF composition) is never passed to
  an engine-specific integration; that media falls back to element-only monitoring.
- The Mux Data media contract expresses the engines it supports instead of hls.js only, so
  adding another engine-backed media surfaces as a type change rather than silent
  misreporting.
- Tests cover hls.js-backed, dash.js-backed, and engine-less media.

## Out of scope

- SPF-backed Mux Data telemetry (#1845).

## Implementation notes

- `mux-embed`'s types (`mux-embed/dist/types/mux-embed.d.ts`) already describe both paths:
  `HlsOptions { hlsjs, Hls? }` and `DashOptions { dashjs }`, plus `addDashJS()` /
  `removeDashJS()`. dash.js needs no custom bridging.
- `packages/media/src/dom/mux/mux-data.ts` imports `{ Hls }` from `../hls-js`, which
  statically imports `hls.js`, so the hls.js coupling is a bundle concern as well as a
  behavioral one.
- Affected surfaces: `packages/media/src/dom/mux/mux-data.ts`,
  `packages/html/src/media/mux-data/mux-data-element.ts`,
  `packages/react/src/media/mux-data/mux-data.tsx`.

## Open question

- Should decoupling Mux Data from the static `hls.js` import land here, or as a follow-up?
  `mux-embed` accepts the `Hls` constructor for its hls.js path, so keeping that path
  working without a top-level `hls.js` import may need the media to supply it.

luwes:
Closes #2038
Refs #1845

## Summary

Mux Data assumed `media.engine` was an hls.js instance. It handed that engine to `mux-embed` as `hlsjs` and always sent the hls.js constructor along with it, so DASH playback was monitored through the hls.js integration and lost every engine-level signal. Engines are now matched by shape and routed to the integration that understands them.

## Changes

- hls.js engines keep the `hlsjs` path; dash.js players go through `mux-embed`'s own `dashjs` path, which was never being used.
- Media with no JavaScript engine (native HLS, a plain `<video>`) or an engine Mux Data cannot hook (an SPF composition) are monitored from the media element alone, with a one-time dev warning naming the data that will be missing. Previously an unrelated engine was passed to the hls.js option.
- Mux Data no longer imports hls.js. A `MuxData`-only import drops from **1,239,192 to 123,410 bytes**, and the DASH sandbox pages now bundle Mux Data with no hls.js chunk at all.
- `MuxDataMedia.engine` is typed `unknown`, since engine selection belongs to the resolver rather than the media contract.
- Both DASH sandbox templates compose `<mux-data>` / `<MuxData>` so the dash.js path is easy to exercise in a browser.

<details>

`toMuxDataEngineOptions()` recognizes engines structurally rather than with `instanceof`, so the module imports neither hls.js nor dash.js. That is what removes the bundle cost, and it also keeps dash.js's top-level `window` access (#1343) away from SSR. Two tests match against real `new Hls()` and `dashjs.MediaPlayer().create()` instances so library drift breaks the match loudly.

`mux-embed` reads hls.js's event names off the class and falls back to `window.Hls` when it is not given one. The class is taken from `engine.constructor`, which finds the bundled hls.js without importing it. If that class carries no `Events`, Mux Data monitors the element alone rather than letting `mux-embed` pick up an unrelated global.

SPF stays out of scope; #1845 still owns its telemetry. It now falls back to element-only monitoring instead of misreporting.

</details>

## Testing

1. `pnpm -F @videojs/media test` (565 passed, including new `toMuxDataEngineOptions` coverage for hls.js, dash.js, engine-less, and unrecognized engines)
2. `pnpm -F @videojs/html test src/media`, `pnpm -F @videojs/react test src/media`, `pnpm typecheck`, `pnpm check:workspace`
3. Manual: run the sandbox `dash-video` preset for HTML and React, and confirm Mux Data reports the view with dash.js data attached.

Note for reviewers: the sandbox templates carry a Mux Data environment key, which is client-visible by design, so the Akamai test streams can be attributed. `apps/sandbox` build currently also fails on unrelated stale scratch directories in the gitignored `src/`.

---

> [!NOTE]
> **Medium Risk**
> Changes analytics/telemetry wiring for all engine-backed playback; behavior is more correct but affects what Mux receives in production views.
> 
> **Overview**
> Fixes **Mux Data** treating every `media.engine` as hls.js, which meant DASH playback was wired through `mux-embed`'s hls.js integration and lost engine-level telemetry (renditions, segment timing, engine errors).
> 
> Monitoring now goes through **`toMuxDataEngineOptions()`**, which picks the right `mux-embed` hook by engine shape: dash.js → `dashjs`, hls.js → `hlsjs` + constructor from `engine.constructor` (no hls.js import), and everything else → element-only monitoring with a one-time dev warning. **`MuxDataMedia.engine`** is **`unknown`** to match that model.
> 
> **`hasMethods`** was added in `@videojs/utils/predicate` for structural checks. DASH HTML/React **sandbox** templates compose **`<mux-data>` / `<MuxData>`** with a sandbox env key so the dash path is easy to verify. Docs now describe hls.js and dash.js engine support.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce8febac5affe7afbf0a70a4a8b1d0d188a4a229. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/media/src/dom/mux/mux-data-engine.ts
+++ b/packages/media/src/dom/mux/mux-data-engine.ts
@@ -1,0 +1,75 @@
+import { isFunction, isObject } from '@videojs/utils/predicate';
+import type { MuxDataOptions } from './types';
+
+/** The `mux-embed` monitor options that hook a playback engine's own telemetry. */
+export type MuxDataEngineOptions = Partial<Pick<MuxDataOptions, 'Hls' | 'hlsjs' | 'dashjs'>>;
+
+type MuxDataHlsJsEngine = NonNullable<MuxDataOptions['hlsjs']>;
+type MuxDataHlsJsClass = NonNullable<MuxDataOptions['Hls']>;
+type MuxDataDashJsEngine = NonNullable<MuxDataOptions['dashjs']>;
+
+const warnedEngines = new WeakSet<object>();
+
+/**
+ * Pick the `mux-embed` integration for a media's playback engine.
+ *
+ * `mux-embed` monitors the media element on its own, and an engine integration
+ * is what adds engine-level data on top: rendition switches, request timing and
+ * throughput, and engine errors. It ships two — hls.js and dash.js — and each
+ * one is wired through a different option, so handing a dash.js player to the
+ * hls.js option leaves the view with element-level data only.
+ *
+ * Engines are matched by shape rather than by class so this module imports
+ * neither hls.js nor dash.js. Mux Data can be registered with any media without
+ * pulling an engine it will never touch into the bundle (dash.js in particular
+ * reads `window` on import), and a media whose engine has no integration — a
+ * raw `<video>`, native HLS, or an SPF playback engine — is monitored from the
+ * element alone instead of through the wrong integration.
+ *
+ * @returns Options to spread into a `Mux.monitor()` call. Empty when the engine
+ *   has no integration, which leaves element-level monitoring intact.
+ */
+export function toMuxDataEngineOptions(engine: unknown): MuxDataEngineOptions {
+  if (isDashJsEngine(engine)) return { dashjs: engine };
+
+  if (isHlsJsEngine(engine)) {
+    // `mux-embed` reads hls.js's event names off the class, falling back to
+    // `window.Hls` when it isn't given one. Take it from the instance so a
+    // bundled hls.js is always found, without importing hls.js here.
+    const Hls = toHlsJsClass(engine);
+    if (Hls) return { hlsjs: engine, Hls };
+  }
+
+  if (__DEV__ && isObject(engine) && !warnedEngines.has(engine)) {
+    warnedEngines.add(engine);
+    console.warn(
+      '[vjs-mux] Mux Data could not hook this playback engine, so the view is monitored from the media element alone. Engine-level data (rendition switches, request timing, engine errors) will be missing.',
+      engine
+    );
+  }
+
+  return {};
+}
+
+/** hls.js: loads a source into a media element and emits its own events. */
+function isHlsJsEngine(engine: unknown): engine is MuxDataHlsJsEngine {
+  return hasMethods(engine, ['loadSource', 'attachMedia', 'on', 'off']);
+}
+
+/** dash.js: attaches a source and a view, and exposes the metrics `mux-embed` reads. */
+function isDashJsEngine(engine: unknown): engine is MuxDataDashJsEngine {
+  return hasMethods(engine, ['attachSource', 'attachView', 'getDashMetrics', 'on', 'off']);
+}
+
+function hasMethods(value: unknown, methods: string[]): boolean {
+  if (!isObject(value)) return false;
+  return methods.every((method) => isFunction((value as Record<string, unknown>)[method]));
+}
+
+/** hls.js's own class, the only place its event names are published. */
+function toHlsJsClass(engine: object): MuxDataHlsJsClass | undefined {
+  const engineClass: unknown = engine.constructor;
+  if (!isFunction(engineClass)) return undefined;
+  const statics = engineClass as unknown as MuxDataHlsJsClass;
+  return isObject(statics.Events) ? statics : undefined;
+}
--- a/packages/media/src/dom/mux/mux-data-engine.ts
+++ b/packages/media/src/dom/mux/mux-data-engine.ts
@@ -1,4 +1,4 @@
-import { isFunction, isObject } from '@videojs/utils/predicate';
+import { hasMethods, isFunction, isObject, isString } from '@videojs/utils/predicate';
 import type { MuxDataOptions } from './types';
 
 /** The `mux-embed` monitor options that hook a playback engine's own telemetry. */
@@ -33,9 +33,9 @@ export function toMuxDataEngineOptions(engine: unknown): MuxDataEngineOptions {
   if (isDashJsEngine(engine)) return { dashjs: engine };
 
   if (isHlsJsEngine(engine)) {
-    // `mux-embed` reads hls.js's event names off the class, falling back to
-    // `window.Hls` when it isn't given one. Take it from the instance so a
-    // bundled hls.js is always found, without importing hls.js here.
+    // `mux-embed` reads hls.js's event names and error details off the class,
+    // falling back to `window.Hls` when it isn't given one. Take it from the
+    // instance so a bundled hls.js is always found, without importing it here.
     const Hls = toHlsJsClass(engine);
     if (Hls) return { hlsjs: engine, Hls };
   }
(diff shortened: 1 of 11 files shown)


packages/media/src/dom/mux/mux-data-engine.ts
```diff
@@ -0,0 +1,75 @@
...
+function isHlsJsEngine(engine: unknown): engine is MuxDataHlsJsEngine {
+  return hasMethods(engine, ['loadSource', 'attachMedia', 'on', 'off']);
+}
+
+/** dash.js: attaches a source and a view, and exposes the metrics `mux-embed` reads. */
+function isDashJsEngine(engine: unknown): engine is MuxDataDashJsEngine {
+  return hasMethods(engine, ['attachSource', 'attachView', 'getDashMetrics', 'on', 'off']);
+}
+
+function hasMethods(value: unknown, methods: string[]): boolean {
```
cjpillsbury: nitpick(blocking): Move this to the utils package. It looks generically handy.

packages/media/src/dom/mux/mux-data-engine.ts
```diff
@@ -0,0 +1,75 @@
...
+      '[vjs-mux] Mux Data could not hook this playback engine, so the view is monitored from the media element alone. Engine-level data (rendition switches, request timing, engine errors) will be missing.',
+      engine
+    );
+  }
+
+  return {};
+}
+
+/** hls.js: loads a source into a media element and emits its own events. */
+function isHlsJsEngine(engine: unknown): engine is MuxDataHlsJsEngine {
```
cjpillsbury, line 59: thought(non-blocking): This is very likely sufficient for now (maybe forever?) but consider using the APIs that mux-embed actually uses for its dash.js/hls.js engine monitors for the type checking.

packages/media/src/dom/mux/mux-data-engine.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (!isObject(value)) return false;
+  return methods.every((method) => isFunction((value as Record<string, unknown>)[method]));
+}
+
+/** hls.js's own class, the only place its event names are published. */
+function toHlsJsClass(engine: object): MuxDataHlsJsClass | undefined {
+  const engineClass: unknown = engine.constructor;
+  if (!isFunction(engineClass)) return undefined;
+  const statics = engineClass as unknown as MuxDataHlsJsClass;
+  return isObject(statics.Events) ? statics : undefined;
```
cjpillsbury: thought(non-blocking): Consider checking for `isString(statics.version)` as well.

packages/media/src/dom/mux/mux-data.ts
```diff
@@ -215,8 +225,7 @@ export class MuxData implements MuxDataProps {
       debug,
       ...(beaconCollectionDomain ? { beaconCollectionDomain } : {}),
       ...(disableCookies ? { disableCookies } : {}),
-      ...(hlsjs ? { hlsjs } : {}),
-      Hls,
+      ...toMuxDataEngineOptions(media.engine),
```
cjpillsbury, line 228: praise: 😎

cjpillsbury (approved):
I had one tiny nitpick blocking callout. Pre-approving on the assumption you'll move it before merging for expediency. 🚀

packages/media/src/dom/mux/mux-data-engine.ts
luwes: Done — it is `hasMethods` in `@videojs/utils/predicate` now, as a type predicate with its own tests. Agreed it is generally handy: recognizing a foreign object by the shape you need from it, without importing the library that defines it.

packages/media/src/dom/mux/mux-data-engine.ts
luwes: Took this on, and it turned up something worth knowing.

I pulled the member accesses out of `mux-embed`'s own monitors. The hls.js monitor uses `on`/`off` and reads `levels`, plus `Events`, `ErrorDetails`, and `version` off the class. The dash.js monitor uses `on`/`off`, `getCurrentTrackFor`, `getBitrateInfoListFor`, and `getRepresentationsByType`. The predicates now check exactly that, so a match means the integration has what it needs rather than what I guessed it needed.

The surprise: dash.js v5 dropped `getBitrateInfoListFor` (and `getQualityFor`) in favor of `getRepresentationsByType`, which is why the first version of this check failed against a real v5 player. `mux-embed` reads whichever is present, so the predicate accepts either and a comment records why. Worth flagging separately if any of `mux-embed`'s dash rendition reporting still assumes the v4 getter.

packages/media/src/dom/mux/mux-data-engine.ts
luwes: Added. The class check is now `Events` and `ErrorDetails` as objects plus `version` as a string, which is the full set `mux-embed`'s hls.js monitor reads off the class.

PR #2043, opened 2026-08-11
feat(packages): add chaptered time sliders

sampotts:
## Summary

Reworks the chaptered time slider based on the [Notion review](https://app.notion.com/p/mux/Chaptered-time-slider-review-3b897a7f89d081b68c3de4908382d3aa). It uses independently rendered HTML/CSS chapter segments rather than an SVG mask, while keeping the regular slider as a fallback until chapter cues are available.

## Changes

- Adds reusable chapter segment primitives for HTML and React, including custom rendering, normalized gaps, CSS variables, and interaction state.
- Updates the Default and Minimal CSS and Tailwind skins with segmented track, buffer, fill, highlighting, and chapter preview details.
- Adds an English chapter demo with a deliberate gap and coverage for the new core, HTML, and React behavior.

## Testing

- `pnpm -F @videojs/html test src/define/tests/registration.test.ts`
- `pnpm -F @videojs/react test src/ui/time-slider/tests/time-slider.test.tsx`
- `pnpm -F @videojs/skins build`
- `pnpm --dir apps/sandbox build`

---

> [!NOTE]
> **Medium Risk**
> Touches the core time-slider interaction and Default/Minimal skins across HTML and React, so visual and seek UX regressions are possible. Changes are additive with a regular-track fallback until chapter cues load.
> 
> **Overview**
> Adds **chaptered time sliders** that render independent HTML/CSS chapter segments instead of an SVG mask, with the regular track kept as a fallback until chapter cues are available.
> 
> **Core** introduces `SliderSegmentsCore` and `TimeSliderChaptersCore` to normalize cues (gaps, overlaps, clamping), expose per-chapter geometry/fill/buffer state, and drive chapter titles at the pointer or keyboard position.
> 
> **HTML and React** add `TimeSlider.Chapters` / `media-time-slider-chapters` plus chapter title parts, wired into Default and Minimal skins. Previews now show chapter titles alongside thumbnails/time, with highlight and gap styling.
> 
> Sandbox and docs demos include a chapters VTT source (with a deliberate gap) so the new segmented seek bar can be exercised end-to-end.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f0359e61603bb6a2947e08a9fdf73df3ddf462b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/react/src/presets/video/minimal-skin.tailwind.tsx
```diff
@@ -493,20 +489,32 @@ export function MinimalVideoSkinTailwind(props: MinimalVideoSkinProps): ReactNod
             </Time.Group>
 
             <TimeSlider.Root render={<SliderRoot />}>
-              <TimeSlider.Track render={<SliderTrack />}>
-                <TimeSlider.Fill render={<SliderFill />} />
-                <TimeSlider.Buffer render={<SliderBuffer />} />
-              </TimeSlider.Track>
+              <TimeSlider.Chapters
+                className={slider.chapters}
+                renderChapter={(props) => (
```
sampotts, line 494: I kept it `renderChapter` as it's not really rendering the element for `TimeSlider.Chapters` itself but I can change that if you preferred.

packages/core/src/core/ui/time-slider/time-slider-chapters/data-attrs.ts
mihar-22, line 1: we should move time slider chapters in a separate directory `time-slider-chapters` to follow conventions, separate component

packages/core/src/core/ui/time-slider/time-slider-chapters-core.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+/** CSS geometry and progress local to each chapter. */
+export const TimeSliderChapterCSSVars = {
+  start: '--media-slider-chapter-start',
+  end: '--media-slider-chapter-end',
+  width: '--media-slider-chapter-width',
+  fill: '--media-slider-chapter-fill',
+  buffer: '--media-slider-chapter-buffer',
+} as const;
```
mihar-22: Separate file to follow conventions `time-slider-chapters-css-vars.ts`. I think we can do a follow up to simplify these names to not include the component name, so just `data-attrs.ts`, `core.ts`, and `css-vars.ts`

packages/core/src/core/ui/time-slider/time-slider-chapters-core.ts
```diff
@@ -0,0 +1,68 @@
+import type { MediaTextCue } from '@videojs/media';
+import type { SliderSegmentRange } from '../slider/slider-segments-core';
+
+export interface TimeSliderChapterRange extends SliderSegmentRange {
+  /** Authored chapter cue, or `null` for an uncovered interval. */
+  cue: MediaTextCue | null;
+}
```
mihar-22: maybe move to a sibling `types.ts` - not important

packages/core/src/core/ui/time-slider/time-slider-chapters/data-attrs.ts
mihar-22: including the `core.ts` file too

packages/html/src/define/ui/compounds.ts
```diff
@@ -108,6 +110,8 @@ export function defineTime(): void {
 export function defineTimeSlider(): void {
   safeDefine(TimeSliderElement);
   defineSliderParts();
+  safeDefine(TimeSliderChaptersElement);
+  safeDefine(TimeSliderChapterTitleElement);
```
mihar-22: I think time slider chapters would be registered as a separate component, it's not a standard part like the rest - needs to be opt in

packages/html/src/define/video/minimal-skin.tailwind.ts
```diff
@@ -93,21 +93,31 @@ function getTemplateHTML() {
...
+                  <media-slider-fill class="${cn(slider.fill.base, slider.fill.fill)}"></media-slider-fill>
+                </media-slider-track>
+                <template>
+                  <div class="${slider.chapter.base}">
+                    <media-slider-track class="${slider.chapter.track}">
+                      <media-slider-buffer class="${cn(slider.fill.base, slider.fill.buffer)}"></media-slider-buffer>
+                      <media-slider-fill class="${cn(slider.fill.base, slider.fill.fill)}"></media-slider-fill>
+                    </media-slider-track>
+                  </div>
+                </template>
```
mihar-22: this is cool, you were able to keep both a single track and then render each individually... but why do we need buffer and fill in the "chapter" tracks (L102:107) if it's also outside (L97:100)?

packages/react/src/ui/time-slider/time-slider-chapters.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    return (
+      <SliderSegments
+        ref={ref}
+        {...props}
+        className={typeof className === 'function' ? (state) => className(getChapterState(state)) : className}
+        style={typeof style === 'function' ? (state) => style(getChapterState(state)) : style}
+        render={
+          typeof render === 'function' ? (renderProps, state) => render(renderProps, getChapterState(state)) : render
+        }
```
mihar-22: we have `isFunction` utility we can use here

mihar-22: i dont know for sure but isn't this normally abstracted into `core`

packages/utils/src/array/find-active.ts
```diff
@@ -0,0 +1,29 @@
+/** Finds the active item in a collection ordered by its start value. */
+export function findActive<Item>(
```
mihar-22: Name wasn't feeling clear, spiked with AI quickly to refactor into separate utils:

```ts
function findLastIndexAtOrBefore<Item>(
  items: readonly Item[],
  value: number,
  getValue: (item: Item) => number
): number {
  let low = 0;
  let high = items.length - 1;
  let index = -1;

  while (low <= high) {
    const mid = (low + high) >>> 1;

    if (getValue(items[mid]!) <= value) {
      index = mid;
      low = mid + 1;
    } else {
      high = mid - 1;
    }
  }

  return index;
}

export function findLastAtOrBefore<Item>(
  items: readonly Item[],
  value: number,
  getValue: (item: Item) => number
): Item | undefined {
  const index = findLastIndexAtOrBefore(items, value, getValue);

  return index < 0 ? undefined : items[index];
}

export function findRangeAt<Range>(
  ranges: readonly Range[],
  value: number,
  getStart: (range: Range) => number,
  getEnd: (range: Range) => number
): Range | undefined {
  const index = findLastIndexAtOrBefore(ranges, value, getStart);

  if (index < 0) return undefined;

  const range = ranges[index]!;
  const end = getEnd(range);
  const isLastRange = index === ranges.length - 1;

  if (value < end || (isLastRange && value === end)) {
    return range;
  }

  return undefined;
}

```

mihar-22 (commented):
Overall this is looking really good 😎 Great work! Thank you for taking in all the feedback and finding a workaround on the single track issue.

I did a AI review locally and it came up with these findings - might be worth having your agent review:

<details>
  <summary>Core Assessment</summary>

Core ownership is directionally correct:

- Text-track discovery remains in the existing DOM store feature.
- Cue normalization and generic segment geometry are framework-neutral.
- React and HTML use the same data attributes and CSS-variable names.
- Registration uses narrow `safeDefine` entry points; classes do not self-register.
- The segment collection is decorative while the existing slider thumb retains the interactive ARIA contract.

Core is not yet doing enough of the feature derivation. HTML and React still independently own chapter lookup, buffer localization, fallback rules, and normalization scheduling. In HTML, both the segment collection and title renormalize on slider updates, including pointer movement.

I would make `TimeSliderChaptersCore` own:

- normalized ranges and `hasChapters`;
- raw pointer-to-chapter lookup;
- local fill/buffer calculation;
- the final per-chapter state.

The platform layers could then limit themselves to store subscriptions and rendering. That would remove the React root-state workaround and make HTML/React parity easier to enforce.

The public documentation also needs a chaptered anatomy/example. The current [[TimeSlider reference](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/site/src/content/docs/reference/time-slider.mdx#L24-L46)](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/site/src/content/docs/reference/time-slider.mdx#L24-L46) does not describe `Chapters`, `ChapterTitle`, `renderChapter`, the HTML template/fallback contract, synthetic gaps, CSS variables, or data attributes.

</details>

<details>
  <summary>Local AI Review</summary>

1. **Medium — The final chapter disappears at the slider’s exact right edge**

   Both adapters publish `pointerValue` using `valueFromPercent()`, which clamps and then rounds to the slider step. With the demo’s `487.626s` duration and default one-second step, `100%` becomes `488`, outside the normalized chapter domain. Consequently, neither `findActive()` nor `SliderSegmentsCore` considers the last chapter active.

   I reproduced this in the deployed HTML and React previews: hovering at 80% showed “Track Performance”; pressing End or pointing at the right edge produced an empty chapter title.

   Relevant code: [[slider-core.ts](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/core/src/core/ui/slider/slider-core.ts#L145-L154)](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/core/src/core/ui/slider/slider-core.ts#L145-L154), [[React context](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/react/src/ui/time-slider/time-slider-root.tsx#L120-L125)](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/react/src/ui/time-slider/time-slider-root.tsx#L120-L125), [[HTML context](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/html/src/ui/time-slider/time-slider-element.ts#L173-L178)](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/html/src/ui/time-slider/time-slider-element.ts#L173-L178).

   Smallest fix: clamp again after rounding. Better for chapter hit-testing: expose/use `rawValueFromPercent()` so hover containment is continuous rather than quantized to the seek step. Add a test where `duration` is not divisible by `step`.
2. **Medium — An invalid HTML template silently removes the entire track**

   The HTML component hides the fallback as soon as chapter cues exist, then silently skips a chapter if the template does not contain exactly one `HTMLElement`. A template with two roots, an SVG root, or only text therefore results in no rendered chapters and no fallback.

   The one-HTMLElement constraint is also absent from the public JSDoc. Relevant ordering: [[time-slider-chapters-element.ts](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/html/src/ui/time-slider/time-slider-chapters-element.ts#L48-L70)](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/html/src/ui/time-slider/time-slider-chapters-element.ts#L48-L70).

   Smallest fix: validate the template once before hiding the fallback, retaining the fallback and emitting a development warning when invalid. Alternatively, support a fragment/multiple roots. Tests should cover missing, empty, multiple-root, and non-HTML templates.
3. **Medium — React’s collection-level state is arbitrarily the first chapter**

   `TimeSliderChaptersProps` exposes `className`, `style`, and `render` as functions of `TimeSliderChapterState`, but `SliderSegments` supplies the state of `geometry[0]` to those collection-level callbacks.

   That means:

   - The collection’s state always describes the first chapter.
   - Moving playback into another chapter does not make the collection callback describe that chapter.
   - The fallback path fabricates a pseudo chapter state.
   - There is no documented way for consumers to know why this particular chapter was selected.

   Relevant code: [[public props](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/react/src/ui/time-slider/time-slider-chapters.tsx#L18-L29)](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/react/src/ui/time-slider/time-slider-chapters.tsx#L18-L29), [[first-segment selection](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/react/src/ui/time-slider/slider-segments.tsx#L61-L70)](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/react/src/ui/time-slider/slider-segments.tsx#L61-L70).

   Smallest fix: give the root a collection-level state such as `{ hasChapters, chapters }`, or use the parent slider state. Reserve `TimeSliderChapterState` exclusively for `renderChapter`.
4. **Low — Chapter information remains pointer-only and is not coherently exposed to keyboard/screen-reader users**

   The preview is revealed only by `data-pointing`; keyboard focus and arrow/Home/End interaction do not reveal the chapter title. The slider’s `aria-valuetext` continues to contain only time. Meanwhile, because the opacity-zero title is not hidden from accessibility APIs, the browser accessibility tree includes the preview title as detached child text—potentially stale at the default pointer position.

   Relevant CSS: [[default slider.css](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/skins/src/default/css/components/slider.css#L117-L184)](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/skins/src/default/css/components/slider.css#L117-L184). The title is given `aria-live="off"` but no explicit association with the slider: [[React title](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/react/src/ui/time-slider/time-slider-chapter-title.tsx#L43-L50)](https://github.com/videojs/v10/blob/51262304935c5fb6f802f9c74d71523adb1b9d8c/packages/react/src/ui/time-slider/time-slider-chapter-title.tsx#L43-L50).

   Consider showing the preview on keyboard focus, resolving against current time when no pointer interaction exists, and deliberately including/associating the chapter title with the slider’s accessible value. Otherwise mark the visual preview hidden from accessibility APIs.

</details>

mihar-22 (approved):

packages/core/src/core/ui/time-slider/time-slider-chapters/data-attrs.ts
luwes: are you sure? it seems like `media-time-slider` is still the parent. or do you mean a new folder in the time-slider folder?

mihar-22: i was thinking originally top-level, but maybe you're right this is directly tied to time-slider so a subdir would make more sense 👍

mihar-22: Trying to differentiate between standard parts (track, fill, buffer, thumb) from opt-in stuff like chapters which are heavy

packages/html/src/define/video/minimal-skin.tailwind.ts
luwes: yes, also wondering this

sampotts: Removed. The original reasoning was:

> The explicit track is the progressive fallback. The chapter template is inert until the custom element upgrades, and real geometry cannot be generated until duration is available. Without the fallback, the timeline would initially be blank while the element upgrades and metadata/chapters load. The fallback is removed once authored chapters are available, so the extra nodes are only temporary.

...but it's a non-issue so I've ditched it.

packages/react/src/ui/time-slider/time-slider-chapters.tsx
sampotts: I've moved it to core 👍🏼

packages/html/src/define/ui/compounds.ts
sampotts: Done

Issue #1783, opened 2026-07-01
Feature: Cap Rendition to Player Size for <hlsjs-video> and hlsjs-powered <mux-video>


decepulis:
> [!NOTE]
> **Rough story-point estimate** from an AI-assisted planning pass — re-estimate when you pick this up.

hls.js / `<mux-video>` implementation of Cap Rendition to Player Size.

Don't request a rendition higher than the player's rendered dimensions. Saves bandwidth and cost. Dynamic — adjusts on player resize.

Maps to hls.js's `capLevelToPlayerSize` (with device-pixel-ratio handling).

## Prior art

- hls.js — [`capLevelToPlayerSize`](https://github.com/video-dev/hls.js/blob/master/docs/API.md)
- Mux Player — [`capRenditionToPlayerSize`](https://github.com/muxinc/elements/blob/main/packages/mux-player/REFERENCE.md): caps to player size with a 720p floor by default

Part of #1418.

decepulis:
see https://github.com/videojs/v10/issues/1866 for config conversation

PR #2061, opened 2026-08-12
feat(media): add maxAutoResolution cap to hls.js sources

### LINKED ISSUE #1777: Feature: Max Auto Resolution Cap for <hlsjs-video> and hlsjs-powered <mux-video>
> [!NOTE]

hls.js / `<mux-video>` implementation of the Max Auto Resolution Cap.

Client-side cap on ABR auto-selection — limits the highest rendition the player will auto-choose without removing renditions from availability. Manual selection still works.

Maps to hls.js's `autoLevelCapping`.

## Prior art

- hls.js — [`autoLevelCapping` + `capLevelToPlayerSize`](https://github.com/video-dev/hls.js/blob/master/docs/API.md)
- Mux Player — [`maxAutoResolution`](https://github.com/muxinc/elements/blob/main/packages/mux-player/REFERENCE.md): caps auto selection while keeping all renditions available for manual choice

Part of #1416.

R-Delfino95:
Closes #1777

## Summary

Adds `source.maxAutoResolution` to hls.js-backed medias, capping the highest rendition ABR selects on its own. Renditions above the cap stay in `videoRenditions` and remain selectable by hand. This is a ceiling on automatic selection, not a filter on availability.

* **Cap controller (`cap-level.ts`)** — `hls.autoLevelCapping` has a single writer: hls.js's own `CapLevelController` rewrites it every second. Setting the property from outside is undone within a second, so the cap is applied from a controller layered over hls.js's, overriding the one method every capped value passes through. Levels are ranked by pixel area rather than height. So an ultrawide rendition is judged on the pixels it actually costs and the largest one at or below the cap wins, falling back to the lowest when nothing fits.
hls.js's own player-size ceiling is kept. The more restrictive of the two wins.

* **Engine wiring** — the controller class is handed to the hls.js constructor, and it reads a mutable policy the media element owns. That split is what lets a cap change reach the running engine instead of rebuilding it: the key is deliberately kept out of `#engineConfigKey()`. A `capLevelController` passed through `source.engine.hlsJs` is extended rather than replaced.

* **`source.maxAutoResolution`** — read in `HlsJsMedia` and pushed down to the hls.js delegate, so `<mux-video>` and `<mux-audio>` inherit it unchanged.
Setting it applies synchronously rather than waiting up to a second for hls.js's next tick.

## Changes

- `MediaResolution` in `core/types.ts`, the `{height}p` union. `MuxResolution` becomes an alias so the two don't drift.
- New `hls-js/cap-level.ts`: the policy, the controller factory, and resolution → level-index resolution. Layers over whatever `capLevelController` the config already names, so a passthrough one keeps working.
- `HlsJsOnlyMedia` owns the policy and exposes `maxAutoResolution` as a live property; the setter asks the controller to re-evaluate instead of waiting for the next tick.
- `<mux-video>` and `<mux-audio>` inherit it; `MuxMedia`'s own `src` setter needed the key added so a URL change keeps the cap.
- Native HLS: no-op plus a `__DEV__` warning, since there is no equivalent lever there.
- Docs: the normalized-options section of `concepts/media-sources.mdx`, plus a note separating this from Mux's `playback.maxResolution` (server side).

## Testing

- `pnpm -F @videojs/media test` — 606 pass (+37: `cap-level.test.ts` for the arithmetic and controller, plus integration in `hls-media.test.ts` and `mux-media.test.ts`).
- `pnpm -F @videojs/html test` — 326 pass. `pnpm -F @videojs/react test` — 401 pass.
- `pnpm typecheck`, `pnpm check:workspace`, `pnpm -F site astro check` — clean.

Verified in the sandbox against a real Mux ladder (270/360/540/720): the cap applies at boot and on live changes without a new `Hls` instance, all four renditions stay listed, and picking 720p by hand while the cap sits at 360p plays 720p. A cap below the whole ladder falls back to the lowest rung; removing the key returns the ceiling to the player-size cap.

The second commit adds `?maxAutoResolution=` and `?preferPlayback=` to the hls.js and Mux sandbox templates, so the engine can be built with the cap already set rather than switched from the console. Both are absent unless named, so default sandbox behavior is unchanged.

For manual test, you can use the new query params in sandbox (local the deploy of this branch):
- [hls.js](https://v10-sandbox-git-fork-r-delfino95-feat-1777-max-auto-86ff39-mux.vercel.app/html-hlsjs-video/index.html?maxAutoResolution=360p)
- [mux-video](https://v10-sandbox-git-fork-r-delfino95-feat-1777-max-auto-86ff39-mux.vercel.app/html-mux-video/index.html?maxAutoResolution=260p)

If you want to test it with an existing player reproducing, you can set it using the console:
```js
const el = document.querySelector('hlsjs-video'); // or mux-video
el.source = { ...el.source, maxAutoResolution: '200p' };
```

## Notes

- hls.js abandons capping when the video element has no measurable size. A requested resolution ceiling doesn't depend on layout, so it's applied anyway — otherwise a hidden or not-yet-laid-out player streams uncapped.
- Audio-only streams never start capping, so the cap is inert there. Asserted rather than assumed.
- No DPR knobs exposed. `ignoreDevicePixelRatio` and `maxDevicePixelRatio` are still reachable through `source.engine.hlsJs`, and a normalized key can be added later without breaking anything. Flagging it so it reads as a decision, not an oversight.
- #1783 (`capRenditionToPlayerSize` and the floor) adds two more inputs to this same `getMaxLevel()` and needs no new plumbing.

---

> [!NOTE]
> **Medium Risk**
> Touches core hls.js ABR/cap-level behavior and source wiring across HLS and Mux media; mistakes could affect rendition selection or engine lifecycle, but changes are well-tested and native HLS is explicitly no-op with a dev warning.
> 
> **Overview**
> Adds **`source.maxAutoResolution`** so hls.js-backed players can cap the highest rendition adaptive bitrate selects automatically, without removing higher renditions from `videoRenditions` or forcing a manual-only ladder.
> 
> The hls.js path installs a **custom cap-level controller** (`cap-level.ts`) that intersects the requested ceiling with hls.js’s player-size cap via `getMaxLevel()`, using **pixel-area** matching (with a deliberate `480p` rounding tweak). The policy is mutated on the media element so cap changes **re-apply on the running engine** and are **excluded from `#engineConfigKey()`** so buffers and ABR state are not torn down. **`MuxMedia` / `<mux-video>`** inherit the option; **`playback.maxResolution`** still only affects the manifest URL.
> 
> Sandbox **hls.js and Mux HTML templates** accept **`?maxAutoResolution=`** and **`?preferPlayback=`** via `getInitialPlaybackOverrides()` so the first load is built with those settings. Docs cover the option and how it differs from Mux’s server-side max resolution.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5dcd76001fc4409394d1864f3acefaa2298e8b49. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/sandbox/app/shared/sandbox-listener.ts
+++ b/apps/sandbox/app/shared/sandbox-listener.ts
@@ -1,11 +1,18 @@
 import { SKINS } from '@app/constants';
 import { DEFAULT_SANDBOX_LOCALE, SANDBOX_LOCALE_TAGS, type SandboxLocaleTag } from '@app/shared/i18n/locale-meta';
 import type { Skin } from '@app/types';
+import type { MediaResolution } from '@videojs/media';
 import { DEFAULT_AUDIO_SOURCE, SOURCES, type SourceId } from './sources';
 
 export const PRELOAD_VALUES = ['none', 'metadata', 'auto'] as const;
 export type PreloadValue = (typeof PRELOAD_VALUES)[number];
 export const DEFAULT_PRELOAD: PreloadValue = 'metadata';
+
+// Any `{height}p`, not just the rungs `MediaResolution` names.
+const MAX_AUTO_RESOLUTION_PATTERN = /^\d+p$/;
+
+export const PREFER_PLAYBACK_VALUES = ['mse', 'native'] as const;
+export type PreferPlaybackValue = (typeof PREFER_PLAYBACK_VALUES)[number];
 
 const params = new URLSearchParams(window.location.search);
 
@@ -30,6 +37,18 @@ function readPreload(): PreloadValue {
   return PRELOAD_VALUES.includes(value as PreloadValue) ? (value as PreloadValue) : DEFAULT_PRELOAD;
 }
 
+function readMaxAutoResolution(): MediaResolution | undefined {
+  const value = params.get('maxAutoResolution');
+  if (!value || !MAX_AUTO_RESOLUTION_PATTERN.test(value) || Number.parseInt(value, 10) <= 0) return undefined;
+
+  return value as MediaResolution;
+}
+
+function readPreferPlayback(): PreferPlaybackValue | undefined {
+  const value = params.get('preferPlayback');
+  return PREFER_PLAYBACK_VALUES.includes(value as PreferPlaybackValue) ? (value as PreferPlaybackValue) : undefined;
+}
+
 let currentSkin = readSkin();
 let currentSource = readSource();
 let currentAutoplay = readBoolean('autoplay');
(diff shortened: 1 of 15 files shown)


luwes (approved):
LGTM 👍

PR #2062, opened 2026-08-12
feat(packages)!: add <mux-background-video> over the SPF background-video engine

cjpillsbury:
## Summary

`<mux-background-video>` and `MuxBackgroundVideo`: a muted, looping, chrome-less video from a Mux stream URL, over the SPF background-video engine. `src` is the whole surface — capping which rendition is fetched is a Mux URL param (`?max_resolution=720p`), so the manifest arrives already narrowed and the engine pins the largest track on offer.

## Changes

- **A standalone `mux-background-video` adapter and Media**, behind `@videojs/spf/mux-background-video`. It shares the engine with the generic background-video adapter but not the adapter itself: that one exposes a client-side `maxResolution` this has no use for, and binds no host, so it has no Media to hand an element. The two are otherwise the same today and say so in their docstrings.
- **The element, the React component, sandbox pages, and a CDN bundle**, so every platform the sandbox offers reaches it.
- **Nothing declares `loop` / `muted` / `autoplay` / `preload` any more** — on either background-video adapter. `attach` writes them onto the element instead, which is also what makes them readable back: a host-bound Media inherits all four and forwards to the element, where declaring them meant reads described an intention. The old suite asserted the element going unlooped while `media.loop` still said `true`.
- **The Media binds a narrow, local `<video>` host rather than the shared `HTMLVideoElementHost`.** A background video reaches four of that host's members and would ship the rest; this takes ~1.2 KB gzipped off every published entry and leaves `@videojs/media` out of the graph.
- **Fix: the CDN sandbox page rendered a plain `<video>` for any preset it doesn't claim.** Adding a preset makes it selectable on every platform, and that page falls through `getMediaTag`'s default when its switch has no case, so the element was never defined and no bundle loaded. Both background presets now answer one predicate rather than repeating a comparison in five places.

<details>

**Why a separate adapter rather than a layer over the generic one.** With `src`-only and no structured source there is nothing Mux-specific in it yet, so composing would have meant inheriting a `maxResolution` only to override it away. The seam exists for what is coming: screen-resolution and pixel-density capping is the named fast-follow, and it is the first thing that gives this adapter something of its own.

**Why the Media binds a one-off host.** The generic background-video adapter deliberately binds none — its docstring gives the reason as the `@videojs/media` dependency that `HTMLVideoElementHost` would add, which measured 1.26 KB gzipped. But an element needs a host-bound Media, so this one binds a local ~40-line host instead: `attach`, `detach`, and the four properties `attach` fixes.

That is the entire surface reachable here. The engine drives playback, the element exposes `src`, and the background player subscribes to no store features at all (`backgroundFeatures` is `[]`), against a shared host offering `currentTime`, `play()`, picture-in-picture, fullscreen, event forwarding, and media-component registration. Swapping it takes ~1.2 KB gzipped off each of the three published entries — spf 18,159 → 16,891, html 19,144 → 17,918, react 18,637 → 17,390 — and drops `@videojs/media` from the graph entirely. The Mux entry now costs 16,910 B against the generic one's 16,896: **14 bytes for a host-bound Media, where the shared host cost 1,286.**

The tradeoff is written on the host. A feature added to `backgroundFeatures` would read properties that are not there, as `undefined` rather than as a failure — that is the point at which this should compose over the shared host rather than replace it, and the reason the follow-up worth having is making `HTMLVideoElementHost` composable rather than repeating this per Media.

**`<background-video>` is untouched.** Rewiring it onto the engine would have silently broken its `nomuted` / `noloop` / `noautoplay` opt-outs, since `attach` force-writes all three.

</details>

## Breaking changes

- `BackgroundVideoMediaProps` is `src` and `maxResolution` only, and `BackgroundVideoMediaElement` no longer carries `loop` / `muted` / `autoplay` / `preload` at all — its base is a bare object, so nothing backfills them. Composing `BackgroundVideoMediaMixin` over a media host still does.

## Testing

```
pnpm -F @videojs/spf test      # 1485 passed, 11 skipped
pnpm -F @videojs/html test     # 317 passed
pnpm -F @videojs/react test    # 394 passed
pnpm typecheck && pnpm check:workspace
```

Browser smoke on the sandbox, HTML / React / CDN: autoplaying, looping, muted, `preload="auto"`, and `videoHeight` 720 on an asset whose ladder tops out at 1080p — the URL cap narrowed the manifest and the picker took the largest of what remained. The natively-played `<background-video>` was re-checked on both its own page and the CDN page as a regression.

### Smoke test it yourself

Branch preview: **https://v10-sandbox-git-feat-spf-mux-background-video-mux.vercel.app**

- **`<mux-background-video>`** — [HTML](https://v10-sandbox-git-feat-spf-mux-background-video-mux.vercel.app/?platform=html&preset=mux-background-video) · [React](https://v10-sandbox-git-feat-spf-mux-background-video-mux.vercel.app/?platform=react&preset=mux-background-video) · [CDN](https://v10-sandbox-git-feat-spf-mux-background-video-mux.vercel.app/?platform=cdn&preset=mux-background-video)
- **`<background-video>` regressions** — [HTML](https://v10-sandbox-git-feat-spf-mux-background-video-mux.vercel.app/?platform=html&preset=background-video) · [React](https://v10-sandbox-git-feat-spf-mux-background-video-mux.vercel.app/?platform=react&preset=background-video) · [CDN](https://v10-sandbox-git-feat-spf-mux-background-video-mux.vercel.app/?platform=cdn&preset=background-video)

Each background page renders one fixed source and takes no picker, so there is nothing to configure — it either autoplays silently on loop or it doesn't. The thing worth checking is that the Mux pages report `videoHeight` 720 on an asset whose ladder tops out at 1080p, which is the URL cap narrowing the manifest rather than a client-side pick.

---

> [!NOTE]
> **Medium Risk**
> Breaking change to `BackgroundVideoMediaProps` and playback-flag semantics on the generic background-video adapter, plus new HLS/SPF playback surface where wrong packaging (non-CMAF) fails at runtime.
> 
> **Overview**
> Adds **`<mux-background-video>`** and **`MuxBackgroundVideo`** for muted, looping hero video from a Mux **HLS** URL via the SPF background-video engine, with **`src` as the only control** (rendition caps via `?max_resolution=` on the URL). Ships HTML custom element, React wrapper, CDN bundle, sandbox presets (HTML/React/CDN), and a dedicated **CMAF/fMP4** demo asset constant.
> 
> Introduces **`@videojs/spf/mux-background-video`** (adapter + **`BackgroundVideoHost`**-bound Media) alongside the generic background-video adapter; the Mux path pins the largest offered rendition because the manifest is already narrowed server-side.
> 
> **Breaking:** both background-video adapters drop **`loop` / `muted` / `autoplay` / `preload`** from their public props/API — **`attach` force-writes** them on the `<video>` instead, and **`BackgroundVideoMediaProps`** is now **`src` + `maxResolution` only** for the standalone element. Shared shadow **template** extraction and **background skin CSS** (`:not(img, picture)`) so native and Mux background elements layout correctly; CDN sandbox treats both background presets uniformly so the new preset loads the right tag/bundle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit adda092f6c878188ef8a4d322c6fd05b8887aa7e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/sandbox/app/constants.ts
+++ b/apps/sandbox/app/constants.ts
@@ -14,6 +14,7 @@ export const PRESETS = [
   'dash-video',
   'audio',
   'background-video',
+  'mux-background-video',
   'vimeo-video',
   'youtube-video',
 ] as const;
--- a/apps/sandbox/app/shared/sources.ts
+++ b/apps/sandbox/app/shared/sources.ts
@@ -332,6 +332,19 @@ export const DEFAULT_DASH_SOURCE: SourceId = 'dash-1';
 
 export const BACKGROUND_VIDEO_SRC = 'https://stream.mux.com/Sc89iWAyNkhJ3P1rQ02nrEdCFTnfT01CZ2KmaEcxXfB008/low.mp4';
 
+/**
+ * The same clip as {@link BACKGROUND_VIDEO_SRC} over HLS, for the SPF-backed
+ * `<mux-background-video>`. Re-ingested from that asset's `high.mp4` rendition.
+ *
+ * Must be a **CMAF/fMP4** asset: SPF appends fMP4 segments directly and does no
+ * MPEG-TS transmuxing, so a TS-packaged playback ID surfaces the
+ * unsupported-container error instead of playing. Packaging follows the video
+ * quality tier — `premium` yields CMAF, while `plus`/`basic` (legacy
+ * `encoding_tier: smart`) yield MPEG-TS — which is why this is a separate asset
+ * rather than the `.m3u8` of the one above.
+ */
+export const HLS_BACKGROUND_VIDEO_SRC = 'https://stream.mux.com/JsDMLkGisX8lHq01wcVv32kQ2vIYvsrEXx007W15xDKJg.m3u8';
+
 export const VIMEO_VIDEO_SRC = 'https://vimeo.com/648359100';
 
 export const YOUTUBE_VIDEO_SRC = 'https://www.youtube.com/watch?v=aqz-KE-bpKQ';
--- a/apps/sandbox/app/shell/app.tsx
+++ b/apps/sandbox/app/shell/app.tsx
@@ -76,6 +76,8 @@ export function App() {
   const hlsPreset = preset === 'hlsjs-video' || preset === 'native-hls-video';
   const muxPreset = preset === 'mux-video' || preset === 'mux-audio';
   const muxSpfPreset = preset === 'mux-video-spf' || preset === 'mux-audio-spf';
+  // Both background presets render a fixed source and have no Tailwind skin.
+  const backgroundPreset = preset === 'background-video' || preset === 'mux-background-video';
   const availableSources =
     preset === 'audio'
       ? MP4_SOURCE_IDS
@@ -168,12 +170,12 @@ export function App() {
   // CDN, background video, and embed (Vimeo/YouTube) videos do not have a Tailwind skin variant.
   useEffect(() => {
     if (
-      (platform === 'cdn' || preset === 'background-video' || preset === 'vimeo-video' || preset === 'youtube-video') &&
+      (platform === 'cdn' || backgroundPreset || preset === 'vimeo-video' || preset === 'youtube-video') &&
       styling === 'tailwind'
     ) {
       setStyling('css');
     }
-  }, [platform, preset, styling]);
+  }, [platform, preset, backgroundPreset, styling]);
 
   const handleSourceChange = useCallback((value: string) => setSource(value as SourceId), []);
 
@@ -201,7 +203,7 @@ export function App() {
         locale={locale}
         onLocaleChange={setLocale}
         availableSources={availableSources}
-        isBackgroundVideo={preset === 'background-video'}
+        isBackgroundVideo={backgroundPreset}
         isSimpleHls={preset.startsWith('simple-hls-')}
         isMuxVideo={preset === 'mux-video'}
         isMuxAudio={preset === 'mux-audio'}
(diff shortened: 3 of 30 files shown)


luwes (approved):
👍 
just needs a type fix

heff:
> @videojs/spf/mux-background-video

Is that the right location? Other medias are under @videojs/media?

> It shares the engine with the generic background-video adapter but not the adapter itself: that one exposes a client-side maxResolution 

In our chat we concluded neither of them _needed_ a client side `maxResolution` to start and we were going to focus on a more automatic solution to capping. Not saying anything has to change here but confirming we heard the same thing.

cjpillsbury:
> > @videojs/spf/mux-background-video
> 
> Is that the right location? Other medias are under @videojs/media?
> 

Yes. TL;DR - SPF depends on media (not the other way around), because SPF (unlike other playback engines) uses the Media API for "double duty" as the engine composition's public api as well (unlike e.g. hls.js, dash.js, etc.).

> > It shares the engine with the generic background-video adapter but not the adapter itself: that one exposes a client-side maxResolution
> 
> In our chat we concluded neither of them _needed_ a client side `maxResolution` to start and we were going to focus on a more automatic solution to capping. Not saying anything has to change here but confirming we heard the same thing.

Yes I was trying to avoid changing the scope to include the hls/spf background-video. Code is identical wrt the engine composition. That said, I thought I said I wasn't going to change the non-mux version for the time being because that wasn't urgent, and I thought you agreed?

PR #2029, opened 2026-08-10
refactor(packages)!: simplify menus

mihar-22:
Closes #2089

Refs #1994
Refs #2022

## Summary

Continue the menu simplification exploration from #1994 while porting the useful state and composition ideas from #2022. Base menus now stay focused on generic menu behavior, while skins compose settings from menu and radio-group primitives and CSS transitions use the existing popover lifecycle.

## Changes

- Remove settings, i18n, value outlets, back-item types, and view abstractions from the base HTML and React menu APIs
- Compose captions, quality, playback-rate, and audio-track submenus directly in each skin, with selected labels supplied by the option hooks
- Split popover open requests from committed state so controlled React roots and cancelable HTML `open-change` events stay synchronized
- Replace the viewport/view transition machinery with nested menu presence hooks and active-panel height measurement
- Keep ending panels mounted for exit transitions, hide and inert covered root items, and reset nested menus when the root closes
- Add module-graph coverage that prevents base menus from regaining settings or i18n dependencies
- Update menu CSS variables, events, exports, skins, and sandbox examples for the simplified model

<details>
<summary>Implementation details</summary>

Submenu direction is inferred from local open state. Opening and closing reuse the popover transition lifecycle to expose starting and ending states, while a small shared height helper synchronizes the active panel height. There is no coordinator, viewport, view stack, direction channel, or settings controller.

This remains an exploratory draft; broader manual skin testing is intentionally left for review.

</details>

## Testing

- `pnpm -F @videojs/core test` — 91 files, 1,385 tests
- `pnpm -F @videojs/html test` — 41 files, 299 tests
- `pnpm -F @videojs/react test` — 39 files, 395 tests
- `pnpm -F @videojs/core build`
- `pnpm -F @videojs/html build`
- `pnpm -F @videojs/html build:cdn`
- `pnpm -F @videojs/react build`
- `pnpm -F @videojs/skins build`
- `pnpm -F @videojs/sandbox build`
- `pnpm typecheck`
- `pnpm check:workspace`
- `git diff --check`

---

> [!NOTE]
> **High Risk**
> Large breaking change to menu/popover/focus/a11y behavior across core, HTML, and consumers; incorrect adapter `syncOpen` or sizing could cause stuck overlays or wrong dimensions.
> 
> **Overview**
[...]

PR #2070, opened 2026-08-13
fix(packages): hide unavailable menu settings

mihar-22:
Refs #2089

## Summary

- give typed HTML menu settings native `hidden`, `data-hidden`, `data-disabled`, and `aria-disabled` semantics
- omit unavailable typed menu settings from the React tree
- remove skin rules that inferred visibility from raw `data-availability`
- preserve author-owned hidden state on regular menu items

This is complementary to the hidden-item navigation fix, but can be reviewed and merged independently.

## Test plan

- `pnpm build:packages`
- `pnpm typecheck`
- HTML menu-item tests: 2 passed
- React menu tests: 34 passed


mihar-22:
going to do this as follow-up to #2029

PR #2097, opened 2026-08-13
refactor(packages)!: rename the SPF background-video Media to hls-background-video

cjpillsbury:
Refs #873

## Summary

Renames the SPF background-video Media to `hls-background-video` and collapses `mux-background-video` into an alias over it, at every layer. The two SPF adapters were line-for-line the same, carrying a note that changes to one usually belonged in both until the Mux flavor grew something of its own. Screen-resolution-derived capping is that something — and it belongs to both, so there is nothing left to keep apart.

## Changes

- `@videojs/spf/background-video` → `@videojs/spf/hls-background-video`, named for the delivery format like the other HLS adapters. `@videojs/spf/mux-background-video` stays, re-exporting the same classes under Mux-flavored names.
- The surviving adapter is the Mux one: it gains the host binding and host-bound Media, and drops the generic one's client-side `maxResolution`. The picker pins the largest rendition on offer; narrowing that set is a delivery param on the URL, which keeps the excluded renditions out of the manifest rather than fetched-then-unpicked.
- `<hls-background-video>` and `HlsBackgroundVideo` are the platform-layer names; `<mux-background-video>` and `MuxBackgroundVideo` alias them.
- Sandbox gains an `hls-background-video` preset beside the Mux one, and the engine-level `spf-background-video` page becomes `spf-hls-background-video`.

**Breaking:** the `./background-video` entry point is gone, `BackgroundVideoMediaElement` is now `HlsBackgroundVideoMediaElement`, and its `maxResolution` property is removed. Nothing outside the sandbox consumed either — `packages/html` and `packages/react` were already on the Mux entry, which is unchanged.

Unrelated and untouched: the `<background-video>` tag / `BackgroundVideo` component, which play their source natively. That collision is half of why the SPF name was worth changing — `hls-` is what now says which one runs an engine.

<details>

**Why HTML can't alias by re-export.** A custom-element class holds one tag name; registering the same class twice throws. So `media/mux-background-video` re-exports the base, and the two `define/media` entries each subclass it with their own `tagName`. That also makes importing both a supported configuration — two names for one element, unlike the `mux-video` flavor tags where two engines compete for one name. `define/tests/media-aliases.test.ts` pins both halves.

SPF and React alias by plain re-export, so `MuxBackgroundVideoMedia === HlsBackgroundVideoMedia` and `MuxBackgroundVideo === HlsBackgroundVideo`. Their tests assert that identity rather than duplicating behavior coverage — if it's the same class, the canonical tests already cover it.

**Picker primitives stay put.** `maxResolutionToPixelArea` and `pickTrackUnderPixelArea` remain exported and tested in `media/primitives/select-tracks.ts`, unused by any adapter for now. The screen-resolution follow-up reads them, and now has one adapter to land in instead of two.

**Sandbox Reload button.** The engine-level page's Load button reassigned `src` to force a re-pick, but the setter early-returns on the URL already playing — so it was already a no-op, and with `maxResolution` gone there was nothing to re-pick anyway. It is now a real Reload: a full adapter rebuild.

</details>

## Testing

```
pnpm -F @videojs/spf test      # 1469 passed
pnpm -F @videojs/html test     # 343 passed
pnpm -F @videojs/react test    # 404 passed
pnpm typecheck
pnpm build                     # incl. the new hls-background-video CDN bundle
pnpm check:workspace
```

Manual: `pnpm dev`, then the `hls-background-video` and `mux-background-video` presets on both HTML and React, and `/spf-hls-background-video/` for the engine diagnostics. The Mux pages are the ones carrying `?max_resolution=720p`, so the diagnostic strip's selected rendition is what shows the capped manifest narrowing the pick.

---

> [!NOTE]
> **Medium Risk**
> Breaking public SPF export and API (`maxResolution` removal); wide but mechanical rename across html/react/spf with alias tests—playback behavior change is intentional (URL-based capping only).
> 
> **Overview**
> **Breaking rename and consolidation** of the SPF background-video stack: `@videojs/spf/background-video` becomes `@videojs/spf/hls-background-video`, with `BackgroundVideoMediaElement` renamed to `HlsBackgroundVideoMediaElement`. The duplicate generic adapter is removed; the surviving implementation drops **client-side `maxResolution`** and always pins the largest rendition in the manifest—capping is expected via URL params (e.g. Mux `?max_resolution=720p`).
> 
> **`mux-background-video` is now an alias** at every layer: `@videojs/spf/mux-background-video` re-exports the HLS classes; HTML registers `<hls-background-video>` as the canonical element with `<mux-background-video>` as a separate subclass tag; React/HTML export `MuxBackgroundVideo` as the same class/component.
> 
> **Platform and sandbox**: new `hls-background-video` preset and HTML/React sandbox pages; CDN bundle and preset wiring; engine demo renamed to `spf-hls-background-video` using `HlsBackgroundVideoMediaElement` without the max-resolution UI (Reload does a full adapter rebuild).
> 
> Native `<background-video>` / progressive MP4 playback is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c5f819bc386f00e5496cb1f1dec719d3e09c5f62. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


luwes (approved):
🚀

PR #2102, opened 2026-08-13
fix(core): restore the last selected subtitles track on toggle

### LINKED ISSUE #2032: Bug: `toggleSubtitles` action ignores track selection, turns on every subtitle/caption when media has multiple subtitle tracks

Media (@videojs/media)

Version 151.0.7922.76 (Official Build) (arm64)
MacOS 26.6

### Description

`toggleSubtitles` action bound to the built-in `c`/`s` `<media-hotkey>` bindings does not track "the currently showing track" or "the user's previously selected track." Instead it operates on *every* track of kind `subtitles`/`captions` on the media element at once:

```js
// dom/store/features/text-track.js
toggleSubtitles(forceShow) {
    const { media } = target();
    if (!isMediaTextTrackCapable(media)) return false;
    const subtitlesTracks = getTextTrackList(media, isCaptionOrSubtitleTrack);
    if (!subtitlesTracks.length) return false;
    const showing = subtitlesTracks.some((track) => track.mode === "showing");
    const nextShowing = forceShow ?? !showing;
    for (const track of subtitlesTracks) track.mode = nextShowing ? "showing" : "disabled";
    return nextShowing;
},
```

### Reduced test case

_No response_

### Steps to reproduce

1. Use a video with both english and german captions
2. Select english
3. Press c twice
4. Both english and german are now selected

### Current Behavior

On any file with more than one subtitle/caption language track, pressing `c` (or calling `toggleSubtitles(true)`) turns on *every* language track simultaneously

### Expected Behavior

I would expect it to toggle the last track the user had selected

### Errors

_No response_

10.0.0 beta 26

luwes:
Fixes #2032

## Summary

Toggling captions on a source with multiple subtitle/caption tracks turned on every language at once. `toggleSubtitles()` now remembers the track the user last had showing and restores just that one.

## Changes

- `toggleSubtitles()` shows at most one caption/subtitle track. Turning captions off records the track that was showing; turning them back on restores it.
- With nothing remembered (first toggle on a fresh source), it falls back to the first track in captions-menu order instead of enabling all of them.
- `toggleSubtitles(true)` while a track is already showing is now a no-op rather than enabling every other language, and it normalizes a multi-showing state down to one track.
- Selecting a track through the captions menu (`selectSubtitlesTrack`) updates the remembered track, so `c` off/on restores the menu choice.
- Documented the restore semantics on the `MediaTextTrackState.toggleSubtitles` contract.

<details>

The remembered track id lives in the `textTrack` feature state closure, so it is per player store and needs no new public state field. Both actions now share one helper that pairs caption/subtitle tracks with the same stable ids exposed through `textTrackList` (keeping `selectSubtitlesTrack` values valid) and orders them like the captions menu, plus a `showOnly()` helper that enables one track and disables the rest.

This also improves the hls.js path: `HlsJsMediaTextTracksMixin` resolves the active hls.js subtitle track from the first `showing` text track, which was ambiguous while every track was showing.

</details>

## Testing

`pnpm -F @videojs/core test` — new cases cover restoring the previously showing track, restoring a menu selection after `off`, `toggleSubtitles(true)` not enabling extra tracks, and the fallback when the remembered track is gone after a source change. Also ran the `@videojs/media`, `@videojs/html`, and `@videojs/react` suites plus `pnpm typecheck`.

Manual: load a source with two caption languages, select one, press `c` twice — only the selected language returns.

---

> [!NOTE]
> **Medium Risk**
> Changes core caption toggle and track-selection behavior used by keyboard shortcuts and the captions UI; logic is well-tested but affects hls.js active-subtitle resolution when multiple tracks were previously showing.
> 
> **Overview**
> Fixes **multi-language captions turning on at once** when using the `c` shortcut or `toggleSubtitles()`. The text-track feature now keeps **at most one** caption/subtitle track in `showing` mode and stores the last active track id in feature closure state.
> 
> **`toggleSubtitles()`** turns all caption/subtitle tracks off when disabling; when enabling, it restores the track that was showing, the one chosen via **`selectSubtitlesTrack()`**, or the **first track in captions-menu order** (captions before subtitles). **`toggleSubtitles(true)`** no longer enables every language—it leaves the current track alone or collapses a bad multi-showing state to a single track. If the remembered track disappears after a source change, it falls back to the first available track.
> 
> Shared helpers **`getSubtitlesTracks`** and **`showOnly`** align track ids with `textTrackList` and menu ordering. The **`MediaTextTrackState.toggleSubtitles`** contract documents the restore behavior. Tests cover restore, menu selection, forced show, and stale-track fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0752e6941abecf9a699f51b206398ffaeec58ed0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/dom/store/features/tests/text-track.test.ts
+++ b/packages/core/src/dom/store/features/tests/text-track.test.ts
@@ -183,7 +183,7 @@ describe('textTrackFeature', () => {
       ]);
     });
 
-    it('toggleSubtitles() enables and disables caption/subtitle tracks', () => {
+    it('toggleSubtitles() enables a single caption/subtitle track and disables them all', () => {
       const video = createVideo();
       const subtitlesTrack = createMockTrack('subtitles');
       const captionsTrack = createMockTrack('captions');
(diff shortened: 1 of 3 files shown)


mihar-22 (approved):
👌

PR #2116, opened 2026-08-13
refactor(react)!: simplify createPlayer and preset exports

mihar-22:
## Summary

- rename the React `createPlayer()` result from `Provider` to `Player`, alongside its typed `usePlayer` and `useMedia` hooks
- export each preconfigured preset player directly as `VideoPlayer`, `AudioPlayer`, `BackgroundVideoPlayer`, `LiveVideoPlayer`, or `LiveAudioPlayer`
- export a separate typed `usePlayer` hook from every preset path
- keep `createPlayer()` under the hood without turning preset players into runtime namespace objects
- preserve preset feature inference across player props and hooks
- let Next.js Server Components render the direct player component export

## Stack

- this is the base PR
- preset-first docs: #2152
- standalone containers: #2154

## Verification

- `pnpm -F @videojs/react test src/player/tests/create-player.test.tsx src/presets/tests/player.test.tsx`
- `pnpm typecheck`
- Next.js 16.3.1 App Router production builds with Turbopack and Webpack, rendering `<VideoPlayer>` from a Server Component and calling preset `usePlayer` from a nested Client Component

Refs #2004

---

> [!NOTE]
> **High Risk**
> This is a breaking public API rename across `@videojs/react` and all preset entry points; consumers must migrate from `Provider` to `Player` and update preset imports, though runtime behavior of the player store should be unchanged.
> 
> **Overview**
> **Breaking API change:** `createPlayer()` now returns a **`Player`** component (not `Provider`), with **`PlayerProps`** replacing `ProviderProps`. Context error messages and JSDoc examples use `<Player>` instead of `<Player.Provider>`.
> 
> Each preset path (`video`, `audio`, `background`, `live-video`, `live-audio`) now ships a dedicated **`VideoPlayer` / `AudioPlayer` / …** component and a separate typed **`usePlayer`** from new `player.ts` modules marked **`'use client'`**, so apps can import a ready-made player without destructuring `createPlayer` themselves.
> 
> Sandbox, e2e page generation, and ejected-skin codegen are updated to wrap skins in **`VideoPlayer`/`AudioPlayer`** (or preset players from `@app/shared/react/players`) instead of `Provider` + manual `createPlayer`. The metadata sandbox uses **`usePlayer`** from `createPlayer`’s return value rather than `Player.usePlayer`.
> 
> Tests and type tests cover the rename, preset exports, and that preset players are plain components (not `.Provider` namespaces).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1dcb67272e4d2d7d34525707631f981d832f7ae1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/react/src/player/create-player.tsx
+++ b/packages/react/src/player/create-player.tsx
@@ -31,14 +31,14 @@ export interface CreatePlayerConfig<Features extends AnyPlayerFeature[]> {
   displayName?: string;
 }
 
-export type ProviderProps<Config = object> = {
+export type PlayerProps<Config = object> = {
   [Key in keyof Config]?: Config[Key] | undefined;
 } & {
   children: ReactNode;
 };
 
 export interface CreatePlayerResult<Store extends PlayerStore> {
-  Provider: FC<ProviderProps<InferPlayerConfig<Store>>>;
+  Player: FC<PlayerProps<InferPlayerConfig<Store>>>;
   Container: typeof Container;
   usePlayer: UsePlayerHook<Store>;
   useMedia: () => Media | null;
@@ -50,7 +50,7 @@ export type UsePlayerHook<Store extends PlayerStore> = {
 };
 
 /**
- * Create a player instance with typed store, Provider component, Container, and hooks.
+ * Create a player instance with a typed Player component, Container, and hooks.
  *
  * @label Video
  * @param config - Player configuration with features and optional display name.
@@ -86,7 +86,7 @@ export function createPlayer(config: CreatePlayerConfig<AnyPlayerFeature[]>): Cr
     return store;
   }
 
-  function Provider(props: ProviderProps<any>): ReactNode {
+  function Player(props: PlayerProps<any>): ReactNode {
     const { children } = props;
     // Only inputs declared by selected features are forwarded to store actions.
     const configValues = pick(props, configKeys);
@@ -135,7 +135,7 @@ export function createPlayer(config: CreatePlayerConfig<AnyPlayerFeature[]>): Cr
   }
 
   if (__DEV__ && config.displayName) {
-    Provider.displayName = `${config.displayName}.Provider`;
+    Player.displayName = config.displayName;
   }
 
   function usePlayer<R>(selector?: (state: object) => R): AnyPlayerStore | R {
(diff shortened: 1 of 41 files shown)


packages/react/src/presets/audio/index.ts
```diff
@@ -1,7 +1,10 @@
 /** Audio-only player preset with playback and volume controls. */
+'use client';
```
sampotts, line 2: I'm not sure we need this in a barrel file. Only modules that use client hooks would need it I think. I'll do some testing this afternoon to validate that though.

sampotts (approved):
I think a lot of the `use client` decorations aren't needed but I'll validate that separately.

packages/react/src/presets/audio/index.ts
mihar-22: Cool, I'll leave the PR open. If you don't get to it, I can ask AI to validate.

sampotts: It's ok to just merge it I think and we can do some validation of that later. You're right in that any file calling `createPlayer` would need a `use client` decoration so this change will mean we can likely remove it from most examples once merged but still unclear what other files would need it. I've got Codex doing some testing with a Next instance at the moment.

mihar-22: It's good you mentioned it! I asked AI to test and confirm in Next.js. We do need `use client` on the exported module but it turns out the namespaced component `VideoPlayer.Provider` doesn't work. It was failing. So we have to use `VideoPlayer`, confirmed working.

packages/react/src/player/create-player.tsx
```diff
@@ -31,14 +31,14 @@ export interface CreatePlayerConfig<Features extends AnyPlayerFeature[]> {
...
-export type ProviderProps<Config = object> = {
+export type PlayerProps<Config = object> = {
   [Key in keyof Config]?: Config[Key] | undefined;
 } & {
   children: ReactNode;
 };
 
 export interface CreatePlayerResult<Store extends PlayerStore> {
-  Provider: FC<ProviderProps<InferPlayerConfig<Store>>>;
+  Player: FC<PlayerProps<InferPlayerConfig<Store>>>;
```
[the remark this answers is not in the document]
mihar-22: Fixed in `1dcb67272`: the ejected-skin generator now emits `export const { Player } = createPlayer(...)` and renders `<Player>` directly. Generated output was rebuilt and verified.

PR #2118, opened 2026-08-13
fix(packages): build iframe media embeds when the source arrives after attach

### LINKED ISSUE #1895: Bug: Error with Vimeo player in Sveltekit project

Media (@videojs/media)

Browser : Firefox 153.0.1

### Description

Hello, 

I usually don't write bug reports, but I have a problem while trying to implement Video.js with the Vimeo player in my Sveltekit project.

I have this VideoPlayer2 Svelte component (the src is hard-coded for testing) :

```<script lang="ts">
  import '@videojs/html/video/player';
  import '@videojs/html/video/skin';
  import '@videojs/html/media/vimeo-video';

	type Props = {
		src: string;
		title?: string;
		image?: string;
		description?: string;
		class?: string;
	};

	let { src, title = '', image, description = '', class: className }: Props = $props();
  
</script>

<video-player>
  <!--
    Skins contain the entire player UI and are easily swappable.
    They can each be "ejected" for full control and customization
    of UI components.
   -->
  <video-skin>
    <!--
        Media are players without UIs, handling networking
        and display of the media. They are easily swappable
        to handle different sources.
      -->

    <vimeo-video src="https://vimeo.com/1181503036"></vimeo-video>
  </video-skin>
</video-player>
```

I import it in one of my +page.svelte for testing, and I exported "export const ssr = false;" in the +page.svelte.ts to avoid server-side rendering. When loading the page, I get this error :

```
Uncaught Error: The player element passed isn’t a Vimeo embed.
    Player player.es.js:1476
    attach index.js:67
    #attachToTarget index.js:201
    CustomMedia index.js:189
    VimeoCustomMediaElement index.js:6
    MediaAttachElement media-attach-mixin.js:15
    VimeoVideo index.js:53
    VimeoVideoElement vimeo-video.js:4
```

While seeking for help, Claude AI proposed this explanation, I put it for reference:

"I traced this through the @videojs/html@10.0.0-beta.25 source (and @videojs/core).

CustomMediaElement's constructor calls #attachToTarget() synchronously, in the constructor:

```
constructor() {
  super();
  this.shadowRoot.innerHTML = ctor.getTemplateHTML(attrs); // attrs read from this.attributes
  this.#mediaHost = new MediaHost();
  this.#attachToTarget(); // <-- runs immediately
}

#attachToTarget() {
  const target = this.target; // internal iframe
  this.#mediaHost.attach(target);
}
```

And VimeoMedia.attach(target) in @videojs/core does:

```
attach(target) {
  this.#target = target;
  if (!target.src) {
    const initialSrc = buildVimeoIframeSrc(this.#src, ...);
    if (initialSrc) target.src = initialSrc;
  }
  this.#player = new VimeoPlayer(target); // throws if target has no valid src
}
```

getTemplateHTML(attrs) builds attrs from this.attributes at construction time. When a framework creates the element via document.createElement('vimeo-video') (Svelte, React, Vue all do this for client-side rendering), the custom element's constructor fires immediately on creation — before setAttribute('src', ...) is ever called. So attrs.src is empty, the internal iframe is built with no src, and new VimeoPlayer(iframe-without-src) throws "The player element passed isn't a Vimeo embed." because the iframe has no Vimeo URL yet.

The stack trace matches this exactly: attach → #attachToTarget → CustomMedia (constructor) → VimeoCustomMediaElement → MediaAttachElement → VimeoVideo → VimeoVideoElement — all constructor-time calls, before the framework gets a chance to set src.

Why it only shows up with frameworks

Static HTML parsed directly by the browser has all attributes present before the custom element upgrades, so this doesn't reproduce in plain HTML demos. Any framework that creates the element via the DOM API (createElement + setAttribute afterward) will hit this race. It likely affects any other iframe-backed media element in this package (not just Vimeo), since the same attach()-in-constructor pattern is shared.

Suggested fix

Defer attach() to connectedCallback (or at least to a microtask after construction) instead of calling it inside the constructor, so consumers have a chance to set src before the Vimeo player is instantiated."

### Reduced test case

_No response_

### Steps to reproduce

Load component containing Video.js Vimeo player in Sveltekit dev server.

### Current Behavior

Player output an error.

### Expected Behavior

Player doesn't output an error.

### Errors

Uncaught Error: The player element passed isn’t a Vimeo embed.
    Player player.es.js:1476
    attach index.js:67
    #attachToTarget index.js:201
    CustomMedia index.js:189
    VimeoCustomMediaElement index.js:6
    MediaAttachElement media-attach-mixin.js:15
    VimeoVideo index.js:53
    VimeoVideoElement vimeo-video.js:4

v10.0.0-beta.25

luwes:
Fixes #1895

## Summary

`vimeo-video` and `youtube-video` created their embed player during `attach()`, which for a custom element runs in its constructor — before any framework has set `src`. `@vimeo/player` throws for an iframe that isn't a Vimeo embed, so the element (and the surrounding Svelte/Vue render) broke outright; YouTube didn't throw but silently never became ready. Embed creation is now deferred until a source can actually be resolved.

## Changes

- Vimeo and YouTube medias attach to an iframe that has no embed yet, and build it on the first load that has a resolvable source, so create-then-set-`src` (Svelte, Vue, Angular, React) works like static HTML.
- The deferred build waits a microtask so every prop applied in the same task (`muted`, `controls`, `autoplay`, …) reaches the embed URL, regardless of the order a framework applies them.
- Emptiness is decided from the `src` attribute rather than the property, which resolves an empty attribute to the document URL — the case React rendered.
- An iframe that isn't a Vimeo embed now reports a `MediaError` and an `error` event instead of throwing out of a constructor.
- Embed elements forward attributes whose name diverges from their property (`muted` → `defaultMuted`, `playsinline` → `playsInline`) to the media host, which is what builds a late embed URL. Native `<video>`/`<audio>` routing is unchanged.
- React's Vimeo/YouTube components omit `src` instead of rendering `src=""`, which also silences React's empty-`src` warning.

<details>

`attach()` now only binds the target and delegates to a private `createPlayer()`, which builds the embed URL when the target has none and settles the load barrier when there is nothing to embed — otherwise `play()` and `requestFullscreen()` would wait on a load that can never complete. `load()` retries creation when a source arrives, guarded by load-barrier identity (Vimeo) and an in-flight flag (YouTube, whose API load spans more than a tick) so repeated source writes build a single player.

The `@vimeo/player` test mock accepted any target, which is why every existing test attached a src-less iframe and passed. It now rejects non-Vimeo embeds like the real library.

</details>

## Testing

`pnpm test`, `pnpm typecheck`, and Biome on the touched files. New coverage: deferred embeds, React's empty `src`, same-task prop ordering, single-build on repeated source changes, non-hanging `play()` while deferred, and the non-Vimeo iframe error — plus element- and component-level regressions in `@videojs/html` and `@videojs/react` against the real `@vimeo/player`.

Manual: render `<vimeo-video src="https://vimeo.com/…">` inside a SvelteKit (or any client-rendered) component; it no longer throws `The player element passed isn't a Vimeo embed.`

---

> [!NOTE]
> **Medium Risk**
> Touches core embed attach/load/play paths for Vimeo and YouTube across media hosts and React wrappers; behavior change for late-bound sources but heavily tested.
> 
> **Overview**
> Fixes **create-then-set-`src`** for Vimeo and YouTube embeds (Svelte, Vue, React, etc.), where `attach()` ran in the custom element constructor before frameworks assigned a source.
> 
> **Vimeo and YouTube media hosts** no longer build the iframe URL or instantiate the player in `attach()`. A deferred `#createPlayer()` runs from `load()` after a microtask so props like `controls`, `muted`, and `autoplay` land in the embed URL regardless of setter order. **Empty iframes** are detected via `getAttribute('src')` instead of the `src` property (empty attribute ≠ document URL). Load barriers settle when nothing can be embedded so `play()` does not hang; Vimeo catches invalid iframes and emits **`MediaError`** instead of throwing during construction.
> 
> **React** Vimeo/YouTube components omit `src` when there is no URL yet (avoids `src=""` warnings); the host builds the embed when props arrive.
> 
> Tests cover deferred embeds, empty React `src`, same-task prop ordering, single player on rapid source changes, and updated Vimeo mocks that reject non-Vimeo iframes like the real library.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3dfd800ef1a6d3a60d741c434b83b8a53d0ceb1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/html/src/media/tests/vimeo-video.test.ts
+++ b/packages/html/src/media/tests/vimeo-video.test.ts
@@ -39,11 +39,10 @@ describe('VimeoVideo', () => {
     const tag = defineVimeoVideo();
     const element = document.createElement(tag) as HTMLElement & { engine: unknown };
     element.setAttribute('src', 'https://vimeo.com/1181503036');
-    element.setAttribute('muted', '');
     element.setAttribute('controls', '');
     await flushDeferredEmbed();
 
-    expect(iframeSrc(element)).toContain('muted=1');
+    // Vimeo chrome is hidden with `controls=0` unless controls are asked for.
     expect(iframeSrc(element)).not.toContain('controls=0');
     expect(element.engine).not.toBe(null);
   });
--- a/packages/html/src/media/tests/vimeo-video.test.ts
+++ b/packages/html/src/media/tests/vimeo-video.test.ts
@@ -1,0 +1,50 @@
+import { describe, expect, it } from 'vitest';
+import { VimeoVideo } from '../vimeo-video/media';
+
+let tagCounter = 0;
+
+function defineVimeoVideo(): string {
+  const tag = `test-vimeo-video-${tagCounter++}`;
+  customElements.define(tag, class extends VimeoVideo {});
+  return tag;
+}
+
+function iframeSrc(element: HTMLElement): string {
+  return element.shadowRoot?.querySelector('iframe')?.getAttribute('src') ?? '';
+}
+
+/** Flush the microtask a deferred embed waits on so it sees every attribute. */
+async function flushDeferredEmbed(): Promise<void> {
+  await Promise.resolve();
+  await Promise.resolve();
+}
+
+describe('VimeoVideo', () => {
+  it('builds the embed for a src set after the element is created', async () => {
+    const tag = defineVimeoVideo();
+
+    // How every framework builds the element: created first, `src` set after, so
+    // the constructor runs with no attributes to build an embed from.
+    const element = document.createElement(tag) as HTMLElement & { engine: unknown };
+    expect(iframeSrc(element)).toBe('');
+
+    element.setAttribute('src', 'https://vimeo.com/1181503036');
+    await flushDeferredEmbed();
+
+    expect(iframeSrc(element)).toContain('https://player.vimeo.com/video/1181503036');
+    expect(element.engine).not.toBe(null);
+  });
+
+  it('builds the embed from attributes set after src in the same task', async () => {
+    const tag = defineVimeoVideo();
+    const element = document.createElement(tag) as HTMLElement & { engine: unknown };
+    element.setAttribute('src', 'https://vimeo.com/1181503036');
+    element.setAttribute('muted', '');
+    element.setAttribute('controls', '');
+    await flushDeferredEmbed();
+
+    expect(iframeSrc(element)).toContain('muted=1');
+    expect(iframeSrc(element)).not.toContain('controls=0');
+    expect(element.engine).not.toBe(null);
+  });
+});
--- a/packages/media/src/dom/custom-media-element/custom-media-element.ts
+++ b/packages/media/src/dom/custom-media-element/custom-media-element.ts
@@ -206,11 +206,6 @@ export function CustomMediaElement<T extends Constructor<MediaHost>>(
         if (prop in CustomMedia.prototype) continue;
 
         const attr = attribute ?? prop.toLowerCase();
-        // Embed hosts never mirror attributes onto their target, so an attribute
-        // whose name diverges from its property (`muted` → `defaultMuted`,
-        // `playsinline` → `playsInline`) has no other route to the mediaHost, and
-        // the host is what builds the embed URL for a source that arrives late.
-        if (!syncTargetAttributes && prop in MediaHost.prototype) mediaHostAttrToProp.set(attr, prop);
         Object.defineProperty(CustomMedia.prototype, prop, {
           get: function (this: CustomMedia) {
             return type === Boolean ? this.hasAttribute(attr) : this.getAttribute(attr);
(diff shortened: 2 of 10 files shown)


luwes:
Both Bugbot findings are addressed on this branch.

**Embed muted mapping shadows mute** — correct, and the mapping is gone. Registering `muted` → `defaultMuted` for embeds overwrote the existing `muted` → `muted` entry, so the mute control would have written the initial-embed flag instead of the live mute state. `custom-media-element.ts` is back to its original behavior (49f4ef805). The tradeoff is that an embed URL built after attach no longer picks up `muted`/`playsinline`; the player API applies both after load, so this is a URL-param nuance rather than a lost capability, and it can be revisited without shadowing.

**Deferred YouTube play does not wait** — also correct. The barrier `attach()` opens is settled when there is nothing to embed, so the later load that builds the embed now opens one of its own, with the same supersede check the rest of the file uses (f3dfd800e). Covered by a new test that fails without the barrier.

mihar-22 (approved):
👌

PR #2120, opened 2026-08-13
fix(media): keep the hls audio selection across group switches

### LINKED ISSUE #1856: Bug: HLS multi audio group doesn't persist selected variant

Core (@videojs/core)

All

### Description

Hey, there's an issue in the hls audio selection in v10, the hlsjs event `Hls.Events.AUDIO_TRACKS_UPDATED` is fired when the audio group is changed (for example in a multi-variant multi-quality audio playlist). When this event fires, the selected audio is reset to default that means any quality switch resets the selected audio language/variant.

### Reduced test case

_No response_

### Steps to reproduce

1. Load a multi-quality multi-variant audio hls playlist
2. Select a non-default variant (ex select the dutch language when english is the default)
3. Change quality (either via ABR or manually)
4. Notice that your audio variant as set back to english.

### Current Behavior

Audio selection is reset during `Hls.Events.AUDIO_TRACKS_UPDATED` events which is also fired during the playback of a file on group switch.

### Expected Behavior

Preserve audio selection. This can be done by not overriding audio selection on `Hls.Events.AUDIO_TRACKS_UPDATED` and instead setting it on `Hls.Events.AUDIO_TRACK_SWITCHING`. Hlsjs already handles audio selection preservation for us for this case.

I don't have the rights to make PR but i made a fix here:  https://github.com/videojs/v10/compare/main...zoriya:fix/hls-audio?expand=1

Feel free to make a PR out of this code, modify it or give me permissions to make a PR myself x)

### Errors

_No response_

@videojs/core@10.0.0-beta.25

luwes:
Fixes #1856

## Summary

hls.js emits `AUDIO_TRACKS_UPDATED` every time the active audio group changes, which a rendition switch can trigger mid-playback. The mixin rebuilt `audioTracks` on that event and seeded `enabled` from the manifest's `default` flag, so any quality change (manual or ABR) snapped the viewer's audio language back to the default.

## Changes

- Skip the `audioTracks` rebuild when hls.js re-announces the same track set, so the list and its selection survive an audio group switch
- Derive `enabled` from `AUDIO_TRACK_SWITCHING` / `AUDIO_TRACK_SWITCHED` instead of the manifest `default` flag; hls.js already matches the current track into the new group and re-announces it

<details>

hls.js reindexes `track.id` per audio group, so the "same set" check compares id, kind, label, and language positionally rather than object identity. When the set genuinely differs the list is rebuilt with nothing enabled, and the `AUDIO_TRACK_SWITCHING` that hls.js fires in the same tick supplies the selection. Because `change` events on the track list are coalesced per microtask, the reflection never round-trips a stale selection back into `engine.audioTrack`.

</details>

## Testing

`pnpm -F @videojs/media test` (582 passing) and `pnpm typecheck`. New cases in `packages/media/src/dom/hls-js/tests/media-tracks.test.ts` cover the group-switch sequence, engine-announced selection, and list rebuilds; both new selection tests fail against the previous implementation.

Manual: load a multi-quality, multi-language HLS stream, pick a non-default audio language, then change quality and confirm the language sticks.

---

Original fix by @zoriya in https://github.com/videojs/v10/issues/1856, adapted with the same-set guard and tests.

---

> [!NOTE]
> **Medium Risk**
> Changes playback-facing audio track mirroring and selection during live HLS events; behavior is localized to the hls.js mixin with new unit tests but affects user-visible language during ABR/rendition changes.
> 
> **Overview**
> Fixes audio language snapping back to the manifest default when quality changes during HLS playback.
> 
> The hls.js `HlsJsMediaMediaTracksMixin` no longer rebuilds `audioTracks` on every `AUDIO_TRACKS_UPDATED` if the announced set matches the current list (id, kind, label, language), so rendition-driven audio group switches preserve the user's language. **`enabled`** is now driven by `AUDIO_TRACK_SWITCHING` / `AUDIO_TRACK_SWITCHED` instead of seeding from the manifest **`default`** flag, aligning the DOM track list with hls.js's cross-group selection.
> 
> Tests cover engine-announced selection, same-set group switches, and full list rebuilds when tracks actually change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 28b156a76543d74e4e2d4940d9f8564817c674aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/media/src/dom/hls-js/media-tracks.ts
+++ b/packages/media/src/dom/hls-js/media-tracks.ts
@@ -34,6 +34,10 @@ function getLevelKey(level: HlsJsMediaTrackLevel): string {
   return `${level.url[0] ?? ''}|${level.width}x${level.height}|${level.videoCodec}|${level.bitrate}`;
 }
 
+function getAudioTrackKind(audioTrack: HlsJsMediaAudioTrack): string {
+  return audioTrack.default ? 'main' : 'alternative';
+}
+
 /**
  * Mirrors hls.js manifest levels and alternate audio into the media element's
  * `videoRenditions` / `audioTracks` lists, and wires user selection back to
@@ -56,6 +60,8 @@ export function HlsJsMediaMediaTracksMixin<Base extends Constructor<MediaTracksH
 
       engine.on(Hls.Events.MANIFEST_PARSED, this.#onManifestParsed);
       engine.on(Hls.Events.AUDIO_TRACKS_UPDATED, this.#onAudioTracksUpdated);
+      engine.on(Hls.Events.AUDIO_TRACK_SWITCHING, this.#onAudioTrackSwitch);
+      engine.on(Hls.Events.AUDIO_TRACK_SWITCHED, this.#onAudioTrackSwitch);
       engine.on(Hls.Events.LEVELS_UPDATED, this.#onLevelsUpdated);
       engine.on(Hls.Events.LEVEL_SWITCHED, this.#onLevelSwitched);
       engine.once(Hls.Events.DESTROYING, this.#teardown);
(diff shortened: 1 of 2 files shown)


mihar-22 (approved):
🍎

Issue #828, opened 2026-03-10
Feature: Include Distribution Files as a Release Asset


semiaddict:
### Should this be a feature proposal discussion?

- [x] This is not a substantial change

### Is there an existing issue for this?

- [x] I have searched the existing issues and discussions

### Which package(s) does this apply to? Select all that apply

HTML (@videojs/html)

### Motivation

I am trying to create a new version of the Video.js module for Drupal (https://www.drupal.org/project/videojs/), which uses composer to install the Video.js library.

### Description

It would be great to add distribution files as an asset in the GitHub releases as was already the case for previous Video.js versions.

### Expected Behavior

A distinct asset is added to each release with the distribution files.

### Alternatives and Workarounds

The use of the CDN files could be an alternative, but might not work for all projects.

### Examples in other libraries

The "video-js-8.23.6.zip" asset here https://github.com/videojs/video.js/releases/tag/v8.23.6

JPustkuchen:
Big +1 here! I agree that having to use a CDN is not a perfect solution, especially in terms of GDPR and local / intranet projects.
Having builds with minified and unminified JS lib would be perfect!

JPustkuchen:
Thanks @decepulis for adding this to the roadmap! Here's the related Drupal.org issue: https://www.drupal.org/project/videojs/issues/3601248

Drupal is very privacy-focused, so adding these files to the release (or any other resource to download the required library files for local usage) would be super helpful. Thank you!

Jerricho93:
Hi, thanks for submitting this, and sorry for the delayed response. We will definitely take this request into account, and we'll try to get to it sooner than later!

JPustkuchen:
@Jerricho93 FYI: We're NOW starting the Drupal implementation for VideoJS 10 so would be great to have the library available in a release.
Alternatively it should be available pre-built via https://asset-packagist.org/ or composer without the need to run a build.

There are different strategies taken by different libraries. I think a dist release would be most flexible.

Thanks!

PS: Maybe using the cdn folder from https://asset-packagist.org/package/npm-asset/videojs--html is a valid workaround, as proposed in https://git.drupalcode.org/issue/videojs-3577815/-/merge_requests/1

JPustkuchen:
Sorry made a typo! NOW, not "not" ;) Corrected!

JPustkuchen:
Thanks, waiting for the upcoming release to verify :)

JPustkuchen:
[...]

PR #2121, opened 2026-08-13
fix(build): make cdn bundles self-contained

luwes:
Refs #828

## Summary

Prod CDN bundles statically imported the i18n registry from a hard-coded `https://cdn.jsdelivr.net/...` URL, so a mirrored copy of `cdn/` still reached out to a public CDN at load. That blocks the offline and self-hosted deployments the "Self-host the player" guide already documents, and it blocks shipping a usable distribution archive for #828.

## Changes

- The i18n registry keeps its layers and subscribers on `globalThis`, so duplicate copies of `@videojs/core/i18n` converge on one set of translations instead of each holding their own. Public API is unchanged.
- CDN bundles now import the shared i18n entry through a relative specifier, so the build resolves entirely within its own origin and works offline.
- `pnpm -F @videojs/html check:cdn` asserts every module specifier in the built output is relative and resolves inside `cdn/`. CI runs it after building the bundles.

<details>

The absolute URL was a dedupe mechanism, not an addressing choice: the registry was module-scoped, so sharing one instance depended on every bundle naming the exact same URL. Duplicate instances are normal — separately loaded CDN bundles, a pinned and an unpinned URL for the same file, and two bundlers' output on one page all produce them. Moving the state to a realm-global symbol removes that constraint and lets the specifier stay relative.

The rewritten specifier is now derived from the chunk's own depth rather than special-casing `locales/`, which also corrects any chunk emitted deeper than one level (for example `media/mux-video/spf.js`).

This also fixes a latent hazard on jsDelivr. `how-to/i18n-register-locale` tells users to import the **unpinned** `.../@videojs/html/cdn/i18n.js` while bundles imported the **pinned** URL — different module-map keys, so `registerI18n` could be silently dropped. Those now converge.

The sandbox carried a Vite `transform` hook whose only job was rewriting that URL back to a local module; it is dead code once no bundle emits the URL, so it is removed.

</details>

## Testing

1. `pnpm -F @videojs/core test` (1434 passed) and `pnpm -F @videojs/html test` (343 passed)
2. `pnpm build:cdn && pnpm -F @videojs/html check:cdn` — 392 bundles, 943 specifiers, zero references to jsdelivr (was 59 files)
3. `pnpm typecheck`, `pnpm lint`, `pnpm check:workspace`

The four new registry tests load the module twice through a reset module cache to stand in for real duplication. Verified they are not vacuous: against the previous module-scoped implementation all four fail (`expected undefined to be 'Reproducir'`) while the existing eleven still pass. The `check:cdn` guard was likewise verified in both directions by injecting an absolute URL into a locale bundle.

---

> [!NOTE]
> **Medium Risk**
> Changes i18n runtime state (realm-global registry) and CDN load behavior for all self-hosted deployments; mitigated by unchanged API, new cross-instance tests, and automated CDN self-containment checks in CI.
> 
> **Overview**
> Production CDN bundles no longer pull the shared i18n registry from a hard-coded **jsDelivr** URL. The build rewrites those imports to **relative** paths to `i18n.js` / `i18n.dev.js`, computed from each chunk’s depth so nested outputs like `media/mux-video/spf.js` still resolve correctly. Mirrored or offline copies of `cdn/` can load without reaching an external CDN.
> 
> **`@videojs/core/i18n`** now stores registry layers and subscribers on **`globalThis`** via `Symbol.for('@videojs/i18n-registry')`, so multiple module instances (separate script tags, pinned vs unpinned URLs, mixed bundlers) share one translation store. The public API is unchanged; new tests cover cross-instance register, merge, notify, and reset.
> 
> CI runs **`pnpm -F @videojs/html check:cdn`** after `build:cdn` to fail on absolute URLs, bare specifiers, or imports that resolve outside `cdn/`. The sandbox Vite config drops the transform that rewrote jsDelivr URLs back to local files.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca2f0439e6f8218de140fd303d639a38c5d9497d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,3 +208,8 @@ jobs:
 
       - name: Build CDN bundles
         run: pnpm build:cdn
+
+      # Mirroring cdn/ onto another origin is supported, so the output must not
+      # reference an external CDN or resolve outside the directory.
+      - name: Check CDN bundles are self-contained
+        run: pnpm -F @videojs/html check:cdn
--- a/apps/sandbox/vite.config.ts
+++ b/apps/sandbox/vite.config.ts
@@ -11,7 +11,6 @@ const htmlCdnDir = resolve(__dirname, '../../packages/html/cdn');
 const htmlCdnI18nRegistry = normalizePath(resolve(htmlCdnDir, 'i18n.dev.js'));
 const cdnSandboxMainSrc = resolve(__dirname, 'src/cdn/main.ts');
 const cdnSandboxMainTemplate = resolve(__dirname, 'templates/cdn/main.ts');
-const jsdelivrCdnI18nPattern = /https:\/\/cdn\.jsdelivr\.net\/npm\/@videojs\/html@[^"']+\/cdn\/i18n\.js/g;
 
 /** True when this import should share the single CDN i18n registry module instance. */
 function resolvesToCdnI18nRegistry(source: string, importer?: string): boolean {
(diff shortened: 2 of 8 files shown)


mihar-22 (approved):

JPustkuchen:
@mihar-22 important regression: https://github.com/videojs/v10/issues/2526

PR #2122, opened 2026-08-13
feat(build): include distribution files as a release asset

### LINKED ISSUE #828: Feature: Include Distribution Files as a Release Asset

luwes:
Closes #828
Stacked on #2121 — review that first, and this diff will shrink to its own commits once #2121 merges.

## Summary

v8 shipped a `video-js-<version>.zip` release asset; v10 shipped no equivalent, so Composer-based projects such as [Drupal](https://www.drupal.org/project/videojs) had no way to install a browser-ready player. Each `@videojs/html` release now carries `videojs-html-<version>.zip` and `.tar.gz`.

## Changes

- `pnpm -F @videojs/html build:archive` assembles the archive from the production CDN bundles: every entry point, the shared chunks it imports, and nothing else. Sourcemaps and development builds stay out, so the zip is ~700 KB.
- The release workflow builds the archive and uploads it to the `@videojs/html` release, alongside a `SHA256SUMS` file.
- The release run verifies the CDN bundles are self-contained before publishing, so a regression fails the release instead of shipping an archive that cannot work offline.
- "Self-host the player" documents the archive and the Composer package declaration Drupal projects need.

<details>

Files are copied out of the existing `build:cdn` output rather than rebuilt, so the archive ships the same bundles the CDN serves and the release gains no extra build step.

Entry points come from the build config's exported entry list rather than a filename heuristic. `cdn/` holds the development and production builds side by side, and their content-hashed shared chunks are indistinguishable by name — 110 of the 280 non-`.dev.js` files there belong to the development build, so any glob-based filter silently ships them.

Excluding sourcemaps leaves dangling `sourceMappingURL` comments, so those are stripped and browsers do not request maps that are not there.

The archive is what self-hosters actually run, so the staged copy is re-verified before packing rather than trusting that the reachable set stayed self-contained. That check shares its graph traversal with `check:cdn`.

Archive layout, with one top-level directory because Composer strips it on extraction:

```
videojs-html-<version>/
  video.js  audio.js  background.js  …
  media/*.js  locales/*.js  i18n.js
  <shared chunks>
  LICENSE  README.md  VERSION
```

</details>

## Testing

1. `pnpm build:cdn && pnpm -F @videojs/html build:archive` — 194 bundles from 82 entries, 702 KB zip, 651 KB tar.gz
2. Verified the extracted archive in a real browser: served it over a local static server with **every off-origin request aborted**, loaded `video.js` and `locales/es.js` as separate module scripts, and got a booted player with `aria-label="Reproducir"` — zero blocked requests, zero page errors. That exercises both offline operation and the shared i18n registry across separately loaded bundles.
3. Confirmed the archive contains no sourcemaps, no development bundles, no `sourceMappingURL` comments, and no references to a public CDN.
4. `pnpm build:site` renders the updated guide; `pnpm lint`, `pnpm check:workspace`, and workflow YAML parsing all pass.

Note that the download URL contains the tag verbatim (`.../releases/download/@videojs/html@<version>/…`). Worth a sanity check on the first release that publishes an asset, since Composer will depend on that shape.

---

> [!NOTE]
> **Medium Risk**
> Changes the production release pipeline and artifact upload paths; mistakes could break Composer/Drupal installs or ship non-offline bundles, mitigated by check:cdn and archive verification.
> 
> **Overview**
> Restores a **v8-style downloadable player** for non-npm installs (Composer/Drupal): each `@videojs/html` release gets `videojs-html-<version>.zip`/`.tar.gz` plus `SHA256SUMS`, uploaded from CI after npm publish.
> 
> **`pnpm -F @videojs/html build:archive`** stages production CDN entry bundles and their shared chunks (from exported `entries` in `tsdown.cdn.config.ts`, not globs), strips `sourceMappingURL`, adds LICENSE/README/VERSION, and re-runs the self-contained import check on the staged tree. **`cdn-graph.ts`** centralizes specifier scanning so **`check:cdn`** and the archive use the same rules.
> 
> Release CD now runs **`check:cdn`** before publish and **`build:archive`** before **`gh release upload`**. Docs add a **“Download the release archive”** path and a Composer `drupal-library` example.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0fce2873ef46a983e5d70ec63b5005de900643e5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -74,6 +74,16 @@ jobs:
         if: ${{ steps.release.outputs.releases_created == 'true' }}
         run: pnpm build:cdn
 
+      # Fail before publishing rather than ship bundles that reach out to a public
+      # CDN at runtime, which would break self-hosted and archive installs.
+      - name: Check CDN bundles are self-contained
+        if: ${{ steps.release.outputs.releases_created == 'true' }}
+        run: pnpm -F @videojs/html check:cdn
+
+      - name: Build distribution archive
+        if: ${{ steps.release.outputs.releases_created == 'true' }}
+        run: pnpm -F @videojs/html build:archive
+
       # The site build emits per-framework markdown to site/dist/docs/framework/{html,react}/.
       # The @videojs/html and @videojs/react packages' prepack scripts (fired by `pnpm publish`
       # below) copy that subtree into each package's docs/ directory so it ships in the tarball.
@@ -92,6 +102,21 @@ jobs:
         run: pnpm -r publish --filter "./packages/*" --access public --provenance --no-git-checks
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      # Attach the archive to the @videojs/html release so installers outside npm
+      # (Composer, Drupal) have a stable, versioned download. The version comes from
+      # the manifest release-please just bumped, matching release-pr.yml.
+      - name: Upload distribution archive to the release
+        if: ${{ steps.release.outputs.releases_created == 'true' }}
+        run: |
+          VERSION=$(jq -r '."packages/html"' .github/release-please/.release-please-manifest.json)
+          gh release upload "@videojs/html@${VERSION}" \
+            "packages/html/archive/videojs-html-${VERSION}.zip" \
+            "packages/html/archive/videojs-html-${VERSION}.tar.gz" \
+            packages/html/archive/SHA256SUMS \
+            --clobber
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Update site/v10 branch
         if: ${{ steps.release.outputs.releases_created == 'true' }}
--- a/packages/html/.gitignore
+++ b/packages/html/.gitignore
@@ -1,1 +1,2 @@
 /cdn/
+/archive/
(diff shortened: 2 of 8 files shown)


mihar-22 (approved):
Awesome 😎

JPustkuchen:
Thanks @luwes - very very helpful. Very much looking forward to the upcoming release to see and try it! :)
Thank you!!

JPustkuchen:
@luwes FYI: This is currently broken since last release: https://github.com/videojs/v10/issues/2526
Any chance to trigger a fresh (fixed) release with the assets back again?

It's also still missing in RC1! We have a major regresion!

PR #2125, opened 2026-08-13
refactor(react): use audio track radio group in skins

mihar-22:
Refs #2148

## Summary

Adopt `AudioTrackRadioGroup` in every React video skin so the presets share the new audio selection plumbing while preserving their existing markup and styling.

## Changes

- Replace manual audio-track radio-group mapping in default and minimal skins
- Update both CSS and Tailwind skin variants
- Keep each skin-owned `Menu.RadioItem` root, labels, indicators, and classes

## Testing

Depends on #2124.


sampotts (approved):
Way better 👌🏼

Issue #1785, opened 2026-07-01
Feature: Cap Rendition to Player Size for SPF


decepulis:
> [!NOTE]
> **Rough story-point estimate** from an AI-assisted planning pass — re-estimate when you pick this up.

SPF implementation of Cap Rendition to Player Size.

Don't request a rendition higher than the player's rendered dimensions. Saves bandwidth and cost. Dynamic — adjusts on player resize.

SPF is the in-house HLS engine — there's no native hook, so this needs a resize observer, device-pixel-ratio handling, and a configurable floor (Mux defaults to 720p) built into SPF's rendition-selection logic.

## Prior art

- Video.js v8 (VHS) — [`limitRenditionByPlayerDimensions` + `useDevicePixelRatio`](https://github.com/videojs/http-streaming) (default `true`, no floor)
- Mux Player — [`capRenditionToPlayerSize`](https://github.com/muxinc/elements/blob/main/packages/mux-player/REFERENCE.md): caps to player size with a 720p floor by default

Part of #1418.

PR #2135, opened 2026-08-13
feat(spf)!: cap renditions to the screen and surface unplayable sources

cjpillsbury:
Refs #1785

## Summary

Two things the SPF background-video composition lacked: a rendition pick that respects the screen it will be shown on, and any way to say a source can't be played. Both arrive as composable pieces rather than variant-private logic — track selection is now the same rule chain the ABR path uses, which is a **breaking** change for anyone configuring `selectVideoTrack` / `selectAudioTrack` directly.

## Changes

**Selection by rule chain (breaking)**

- `SelectVideoTrackConfig.picker` / `SelectAudioTrackConfig.picker` / `BackgroundVideoEngineConfig.picker` are replaced by `rules` plus an optional `constraints` pre-pass — the model `switchVideoTrack` already ran. A rule written for one composes into the other unchanged; the two behaviors differ in *reactivity*, not in what a rule is.
- A rule now selects among real candidates, where a picker could return any id. An id the manifest never offered was never selectable, so that's the point rather than a limitation.
- `preferHighestResolution` ships as a shared rule, and the dead picker cluster it orphans is deleted (~230 lines from `media/primitives/select-tracks.ts`, none of it reachable from an entry point).

**Screen-size rendition cap**

- `screenResolutionCap` narrows candidates to the renditions whose *pixel area* fits the screen, composed ahead of the ranker in the background-video engine's default chain. Areas rather than `"1080p"`-style tiers, because a tier only describes a rendition once you assume its aspect ratio — the assumption that mis-measures an anamorphic ladder.
- Fed by `trackScreenResolution`, a new behavior over a new `media/dom/screen` reader/watcher. Device pixels by default; `undefined` where there's no screen to read, which the cap treats as "don't cap" rather than a cap of zero.
- A soft filter, never a constraint: an over-cap rendition is wasteful, not unplayable, so the cap can't make a source fail. Three fall-through paths (no signal composed, no screen, nothing fits) are all intentional.
- Player-size capping — the other half of #1785 — is **not** in this PR. Different signal source (`ResizeObserver`), different owning layer; `rendition-selection-caps.md` records the split.

**Unplayable sources stop being silent**

- Nothing about an unplayable source reaches the media element in this composition. Measured on Chromium and WebKit: an unsupported container, encryption with no EME, and an undecodable codec all leave `HTMLMediaElement.error` null with the element stalled at `readyState 0`, because nothing in MSE reports them either.
- The engine now reports SVTA-coded conditions onto `engine.state.errors` (owned per-source by `collectErrors`), and the Media promotes the first fatal one to `error` + an `'error'` event — the same surface, shape, and 99001 mapping the video and audio Medias already have.
- `<hls-background-video>` re-fires it as its own event and exposes `error`; the React component re-dispatches on the `<video>` so `onError` fires. A consumer of either holds the element, not the Media.
- Fatality is wider here than on the other adapters, and deliberately: only the *pinned* rendition's playlist is ever resolved, so a cause (1004 / 4008) can only concern the pick, and dropping the pick is final. A verdict-only list left MPEG-TS and DRM as silent stalls.
- Every fatal condition is explained on the console, not only the ones that substitute 99001 — otherwise a source with no video renditions arrives as a bare 2011. The sentence names no cause, since naming a missing feature would be wrong for that case; the structured conditions beside it carry the specifics.

<details>

- `applyRules` / `applyConstraints` / `SelectionRule` moved to `playback/primitives/selection-rules.ts` so both selection behaviors can share them. `selectVideoTrack` exists to tree-shake the ABR path out, so importing a composer from `track-switching.ts` would have dragged the bandwidth estimator back in.
- `selectVideoTrack` still *selects* only in `entry`. A sibling effect only ever **de**selects a pick the constraints later turn against — container and encryption are known only once a media playlist resolves, after the pick was made. Re-picking is what `switchVideoTrack` is for; doing it here would make this that behavior with extra steps.
- `reportAbsentTrackType` is a "constraint" that never constrains, composed at the *head* of the chain, where an empty input still means "the source offers none" rather than "the constraints ahead pruned them all". Opt-in per composition, so one that doesn't need it pays nothing.
- The screen watcher subscribes to four signals (`screen`'s own `change`, `resize`, orientation, and a `dppx` media query) and compares readings, so over-subscribing costs a discarded read rather than a spurious call. `Screen.prototype.onchange` is Chromium-only in a secure context — measured across all three engines — so it joins the others rather than replacing them.
- Bundle: `@videojs/spf/hls` sits at 19.99 KB gzipped against a 20 KB target. Composing `trackScreenResolution` is what consumed most of the remaining headroom.

</details>

## Preview

Sandbox on this branch — [v10-sandbox](https://v10-sandbox-git-feat-screen-resolution-cap-mux.vercel.app). The SPF background presets take the full HLS source list, so every shape below is reachable from the picker; these link straight to it.

| Source | Link | What to expect |
| :-- | :-- | :-- |
| 4K CMAF | [HTML](https://v10-sandbox-git-feat-screen-resolution-cap-mux.vercel.app/?platform=html&preset=hls-background-video&source=hls-4k) · [React](https://v10-sandbox-git-feat-screen-resolution-cap-mux.vercel.app/?platform=react&preset=hls-background-video&source=hls-4k) | Plays. The cap picks the largest rung at or under your screen — on a 3456x2234 display that's 2558x1440, not the 3838x2160 top rung |
| MPEG-TS | [HTML](https://v10-sandbox-git-feat-screen-resolution-cap-mux.vercel.app/?platform=html&preset=hls-background-video&source=hls-1) · [React](https://v10-sandbox-git-feat-screen-resolution-cap-mux.vercel.app/?platform=react&preset=hls-background-video&source=hls-1) | No video. `error.code === 99001`, console explains it with the 1004 condition attached, `video.error` stays null |
| DRM, unlicensed | [HTML](https://v10-sandbox-git-feat-screen-resolution-cap-mux.vercel.app/?platform=html&preset=hls-background-video&source=hls-drm-unlicensed) · [React](https://v10-sandbox-git-feat-screen-resolution-cap-mux.vercel.app/?platform=react&preset=hls-background-video&source=hls-drm-unlicensed) | Same 99001, over a 4008 in `engine.state.errors` |
| Audio-only | [HTML](https://v10-sandbox-git-feat-screen-resolution-cap-mux.vercel.app/?platform=html&preset=hls-background-video&source=hls-audio-only-cmaf) · [React](https://v10-sandbox-git-feat-screen-resolution-cap-mux.vercel.app/?platform=react&preset=hls-background-video&source=hls-audio-only-cmaf) | 2011, kept as-is — no cause behind it to substitute for. Still explained on the console, which is what makes a bare verdict actionable |

`preset=mux-background-video` is the same element and component under their Mux-flavored names, and should behave identically — [worth one pass](https://v10-sandbox-git-feat-screen-resolution-cap-mux.vercel.app/?platform=html&preset=mux-background-video&source=hls-4k) to confirm the alias.

To read the surfaced error from the console on either platform:

```js
// HTML
document.querySelector('hls-background-video').error
// engine sequence, both platforms (HTML shown)
document.querySelector('hls-background-video').getMediaTarget().engine.state.errors.get()
```

## Testing

```bash
pnpm -F @videojs/spf test          # 1512 passing
pnpm -F @videojs/html test
pnpm -F @videojs/react test
pnpm --dir apps/e2e exec playwright test spf-background-video \
  --project=vite-chromium --project=vite-webkit   # 20 passing
```

New e2e (`apps/e2e/tests/spf-background-video.spec.ts`) drives real manifests on both browser projects: fMP4 plays with nothing reported; MPEG-TS surfaces 99001 with the sequence holding a cause and *no* verdict behind it; the unlicensed DRM asset lands on the same code; an audio-only ladder keeps its own 2011; a source change back to fMP4 clears and plays; React's `onError` fires. The load-bearing one asserts the inner `<video>` stays at `readyState 0` with `error` null throughout — if a browser ever starts reporting these itself, that test is how we find out. The cap is covered through viewport × `deviceScaleFactor` (Playwright's `screen` option turns out not to apply — `window.screen` follows the emulated viewport in both projects).

Manual, Chromium against the sandbox's 4K preset on a 3456x2234 display: the pick moves from 3838x2160 to 2558x1440, playing at `readyState 4`. The sandbox's SPF background presets take the full HLS source list, so every failing shape is reachable there.

**Not verified on real Safari.** `vite-webkit` is Playwright's WebKit build, not the shipped Safari media stack; a `safaridriver` pass is still worth doing before release, since a CMAF rendition Safari's `isTypeSupported` rejects would now surface a fatal error rather than stalling quietly.

Found and filed along the way: #2162 (an empty `EXT-X-MAP` init URL fetches and appends the hosting page).

---

> [!NOTE]
> **High Risk**
> Breaking SPF selection config and wider fatal-error semantics for background video affect playback and consumer error handling; large cross-package change with real-network e2e.
> 
> **Overview**
> **Breaking:** track selection moves from `picker` callbacks to composable **`rules`** and optional **`constraints`**, shared with the ABR path; legacy pickers and related primitives are removed.
> 
> For **SPF background video**, the default chain **`[screenResolutionCap, preferHighestResolution]`** pins the largest rendition whose pixel area fits the screen, fed by new **`getScreenResolution` / `watchScreenResolution`** and **`trackScreenResolution`**. **`selectVideoTrack`** can deselect a pick when capability info arrives later; **`reportAbsentTrackType`** covers sources with no video renditions.
> 
> **Unplayable sources** are no longer silent: the background adapter promotes fatal SVTA conditions (including per-rendition causes in the pinned model) to **`error`** and **`'error'`**, with console logging. **`<hls-background-video>`** and the React wrapper forward that to consumers while the inner **`<video>`** stays at **`readyState 0`** with **`error` null**.
> 
> **Sandbox / e2e:** SPF background presets use a selectable HLS source list and **`withMuxMaxResolution`**; new fixtures and **`spf-background-video.spec.ts`** plus generated **`?src=`** background pages exercise playback, errors, recovery, and viewport caps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a8b47e6d6e3a9bdcd3c4a3ea25b73a9a3f5f0061. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/spf/src/media/dom/screen.ts
+++ b/packages/spf/src/media/dom/screen.ts
@@ -18,6 +18,8 @@
  */
 
 import { listen } from '@videojs/utils/dom';
+import { shallowEqual } from '@videojs/utils/object';
+import { isFunction } from '@videojs/utils/predicate';
 
 /** A screen's pixel dimensions. */
 export interface ScreenResolution {
@@ -34,8 +36,13 @@ export interface ScreenResolutionOptions {
    * Opt out for CSS pixels. Note that derating — a 3x phone rarely wanting 3x the
    * pixels of its layout — is a scale applied over this reading rather than a
    * reason to turn it off.
+   *
+   * ⚠️ Chromium and Gecko fold page zoom into `devicePixelRatio`, so with this on,
+   * zooming moves the reading even though the screen didn't change. WebKit holds
+   * the ratio independent of zoom and is unaffected. Whether a cap should track
+   * zoom is the cap's call; this flag is only what puts zoom in scope.
    */
-  useDevicePixelRatio?: boolean | undefined;
+  useDevicePixelRatio: boolean;
 }
 
 /**
@@ -52,13 +59,14 @@ export interface ScreenResolutionOptions {
  * cap should flap on rotation, or hold the larger budget across both — and
  * belongs to the cap rather than to the reading.
  */
-export function getScreenResolution(options: ScreenResolutionOptions = {}): ScreenResolution | undefined {
+export function getScreenResolution(
+  { useDevicePixelRatio }: ScreenResolutionOptions = { useDevicePixelRatio: true }
+): ScreenResolution | undefined {
   const screen = globalThis.screen;
   if (!screen) return undefined;
 
   // `|| 1` covers a missing or nonsense ratio: a CSS-pixel reading is still true
   // and still cappable, so it isn't worth failing the whole answer over.
-  const { useDevicePixelRatio = true } = options;
   const ratio = useDevicePixelRatio ? globalThis.devicePixelRatio || 1 : 1;
 
   // Rounded because device pixels are whole and a fractional ratio doesn't divide
@@ -78,8 +86,12 @@ export function getScreenResolution(options: ScreenResolutionOptions = {}): Scre
  * signal that implies one and compares readings to decide whether anything
  * actually moved. Comparing is what makes that safe: the signals overlap and
  * `resize` in particular is noisy, so over-subscribing costs a discarded read
- * rather than a spurious call. `onChange` fires only on a genuine change, and
- * never on subscribe — call {@link getScreenResolution} for the starting value.
+ * rather than a spurious call.
+ *
+ * `onChange` is called once on subscribe with the starting value — including
+ * `undefined` where there is no screen — and after that only on a genuine change.
+ * So a consumer gets its initial state from the watcher and never has to pair it
+ * with a separate {@link getScreenResolution} call.
  *
  * The signals, and what each one is here for:
  *
@@ -95,9 +107,16 @@ export function getScreenResolution(options: ScreenResolutionOptions = {}): Scre
  * - **`screen.orientation` change** — rotation, which swaps the axes without
  *   necessarily resizing the window.
  * - **a `(resolution: <ratio>dppx)` media query** — the device pixel ratio
- *   changing under a window that kept its size, from zoom or from moving to a
- *   display with a different ratio. Armed against the current ratio and re-armed
- *   when it moves, since the query only reports on the ratio it was built for.
+ *   changing under a window that kept its size, which is the cross-display drag
+ *   between displays of different density. Each query only answers about the ratio
+ *   it was built for, so it reports one change and its handler arms the next.
+ *
+ *   Worth keeping despite looking redundant, because it is the only coverage that
+ *   case has in WebKit and Firefox: neither implements `screen`'s change event,
+ *   and the drag doesn't resize the window. It is also a cleaner signal in Safari
+ *   than elsewhere — WebKit holds `devicePixelRatio` independent of page zoom, so
+ *   there it moves only on a real density change, where Chromium and Gecko fold
+ *   zoom into it as well.
  *
  * ⚠️ Known gap, on engines without `screen`'s change event: dragging a window
  * between two different-size displays that share a ratio, without the window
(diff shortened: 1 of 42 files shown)


packages/spf/src/media/dom/screen.ts
```diff
@@ -0,0 +1,196 @@
...
+    listen(
+      query,
+      'change',
+      () => {
+        watchRatio();
+        check();
+      },
+      { once: true, signal }
+    );
+  };
```
spuppo-mux, line 173: comment: I added a helper function for this. Non-blocking since I should probably adapt it on my PR after this merges

spuppo-mux (approved):
lgtm

packages/spf/src/media/dom/screen.ts
cjpillsbury: discussed out of band. @spuppo-mux will unify these use cases in a followup effort outside of this branch/PR.

Issue #1457, opened 2026-04-24
Feature: Media Element Providers


decepulis:
Build out the full set of media elements supported by [Media Chrome](https://www.media-chrome.org/docs/en/media-element) — third-party platform providers and playback engines exposed as `<video>`-compatible custom elements.

## Scope

Sub-issues below cover providers one at a time. Together they bring v10 to parity with the breadth of sources supported across [`muxinc/media-elements`](https://github.com/muxinc/media-elements), [Vidstack](https://vidstack.io/), [Plyr](https://github.com/sampotts/plyr), and the Video.js v8 plugin ecosystem.

## Out of scope

- `<mux-video>` — tracked in #977 "Feature: Mux Media Elements".
- `<hls-video>` / `<dash-video>` — already implemented under `packages/html/src/media/`.
- `<videojs-video>` (wrapping Video.js itself) — covered separately by the `<videojs-v8-video>` research issue in this epic.

## Children

See the parent-issue / sub-issues list on this issue.

PR #2169, opened 2026-08-14
feat(packages): add spotify audio media

### LINKED ISSUE #1462: Feature: `<spotify-audio>`
> [!NOTE]

Add a Spotify provider so Video.js 10 can play `https://open.spotify.com/...` tracks, albums, episodes, and shows via Spotify's embed iframe.

1. **Match Mux `<spotify-audio>` as a baseline** — Same `src` shape (`open.spotify.com/{type}/{id}`), same `config` passthrough for `startAt`, `theme`, `preferVideo`. Media Chrome already ships this and it's the closest thing to a standard.
2. **Audio-first defaults** — No poster, no fullscreen control in chrome, autoplay assumed to fail on mobile (Spotify's iframe cannot play until the user taps the Spotify play button on iOS).

## Design notes

- **No privacy-mode equivalent.** Unlike YouTube's `youtube-nocookie.com`, Spotify exposes no `dnt`/`nocookie` origin. The only privacy-shaped knob is `utm_source`, which is analytics attribution, not consent. GDPR messaging for embedders has to live outside the provider.
- **Branding is non-negotiable.** Spotify's embed terms require the Spotify logo and the in-iframe play button to remain visible — we can't hide iframe chrome the way a native `<audio>` skin does. Worth calling out so product doesn't promise a "clean" skin.
- **Iframe is the only playback surface.** Spotify does not offer a DRM-bypass or raw-stream API for third-party embedders outside Spotify Connect (which requires Premium + login). Playback must go through the iframe controller; our tech layer wraps its postMessage surface.
- **`preferVideo` turns an audio element into a video element.** Spotify video podcasts route through `/embed/.../video`. If we expose this, we need to decide whether `<spotify-audio>` vs. a sibling `<spotify-video>` is the right shape, or whether a single element switches based on `config.preferVideo`. Mux picked the latter.
- **Typed props vs. opaque `config` passthrough.** Same question as YouTube (#1434). Spotify's surface is small enough (`startAt`, `theme`, `preferVideo`) that typed props are tractable.

## Prior art

**Mux `<spotify-audio>` (media-elements)** — Custom element with an `<audio>`-compatible API. Accepts a `config` prop with `startAt`, `theme` (`'dark'` maps to `theme=0`; otherwise Spotify's default color-extraction theme), and `preferVideo` (swaps `/embed/` for `/embed/.../video`). Loads Spotify's iframe API script from `https://open.spotify.com/embed-podcast/iframe-api/v1` and wraps `createController`.
- Source: [`packages/spotify-audio-element/spotify-audio-element.js`](https://github.com/muxinc/media-elements/blob/main/packages/spotify-audio-element/spotify-audio-element.js)
- Package: [`muxinc/media-elements/packages/spotify-audio-element`](https://github.com/muxinc/media-elements/tree/main/packages/spotify-audio-element)

**Vidstack** — No Spotify provider. Its `audio` provider is HTML5-only (mp3/m4a/wav/flac/ogg/aac); embed providers are video-only (YouTube, Vimeo).
- Source: [`packages/vidstack/src/providers/audio/provider.ts`](https://github.com/vidstack/player/blob/main/packages/vidstack/src/providers/audio/provider.ts)

**Plyr** — No Spotify provider. Plyr's audio support is native HTML5 `<audio>` only; embed plugins exist only for YouTube and Vimeo.
- Plugins directory: [`src/js/plugins/`](https://github.com/sampotts/plyr/tree/master/src/js/plugins)

**react-player** — No Spotify support. Confirmed absent from its audio player list.
- Repo: [cookpete/react-player](https://github.com/cookpete/react-player)

**Video.js v8** — No maintained `videojs-spotify` plugin exists. Searching turns up only abandoned/demo code.

**Upstream docs**
- [Spotify Embeds overview](https://developer.spotify.com/documentation/embeds)
- [Using the iFrame API (tutorial)](https://developer.spotify.com/documentation/embeds/tutorials/using-the-iframe-api)
- [iFrame API reference](https://developer.spotify.com/documentation/embeds/references/iframe-api)

luwes:
Ports [`spotify-audio-element`](https://github.com/muxinc/media-elements/tree/main/packages/spotify-audio-element) from `muxinc/media-elements` into v10 as a first-class media host, reshaped to the architecture the Vimeo and YouTube ports already established.

> **Stacked on #2168.** Review the last two commits, or read the diff against `feat/cloudflare-video-media`.

## Motivation

Spotify is the one embed in `media-elements` that is audio-only, and it is the least `HTMLMediaElement`-shaped of the set. Bringing it in as a `Media` host exercises the embed-host contract against a player that reports playback as a periodic snapshot and exposes no volume control at all.

## What's here

`packages/media/src/dom/spotify` follows the YouTube port's file split:

- `props.ts` — the same ten-prop shape the other embed hosts use, with the inert fields documented
- `source.ts` — `SpotifySource`, embed options under `engine.spotify`, source parsing, and embed-URL building
- `iframe-api.ts` — controller typings and a one-shot loader for the Spotify iframe API
- `media.ts` — `SpotifyMedia`
- `tests/media.test.ts` — 54 cases

Plus `<spotify-audio>` in `packages/html` and `<SpotifyAudio>` in `packages/react`.

## Behavior worth reviewing

- **Ready-callback ordering.** The API script hands itself to `globalThis.onSpotifyIframeApiReady` as it evaluates, so the loader assigns that global *before* loading the script. A failed load clears the cached promise so a later host can retry.
- **`playback_update` is a snapshot, not an event stream.** `play`, `pause`, `waiting`, `playing`, `timeupdate`, `durationchange`, and `ended` are all derived from successive updates. That state machine is split into named private methods, each returning whether it consumed the update, so upstream's early-`return` chain stays legible.
- **Seeking.** The embed announces no seeks, so `currentTime` writes optimistically and fires `seeking`, then `seeked` on the first update reporting a different position. This replaces upstream's restore-the-old-time trick, which left `currentTime` stale.
- **Reloads use `loadUri`,** mirroring YouTube's `cueVideoById` shape, rather than upstream's iframe-`src` rewrite.
- **What Spotify cannot do.** No volume, no mute, no playback rate. `volume`/`muted` are fixed readers with inert setters; `playbackRate` is omitted entirely so the player's rate feature stays off instead of showing a dead control.
- **`src` forms.** `open.spotify.com` URLs for every embeddable entity — including localized (`/intl-de/`) and already-embedded (`/embed/`) forms — and `spotify:<type>:<id>` URIs.

## Verification

- `pnpm -F @videojs/media test src/dom/spotify` — 54 passed
- `pnpm test` — 5259 passed, 14 skipped
- `pnpm typecheck` — clean
- `pnpm -F @videojs/sandbox build` — clean; the new presets appear at `/html-spotify-audio/` and `/react-spotify-audio/`

## Notes

Split into two commits: the media host and platform wrappers, then the sandbox examples. The sandbox preset renders with the audio player and skin, and its fixed source is an episode rather than a track, since Spotify plays episodes in full for a signed-out listener and cuts tracks to a 30 second preview.

---

> [!NOTE]
> **Medium Risk**
> Touches shared volume/mute plumbing used by every player, plus a large async third-party iframe integration; risk is mitigated by extensive tests but regressions on mute-only or iOS volume edge cases are possible.
> 
> **Overview**
> Adds **Spotify** as a first-class audio embed (`SpotifyMedia`, `<spotify-audio>`, `<SpotifyAudio>`), wired through the Spotify iframe API with URL/URI parsing, embed options under `engine.spotify`, and playback driven from `playback_update` snapshots (play/pause/waiting/seek/end/loop). The embed is hidden when not using Spotify’s own `controls`, and omits volume/mute members so skins don’t show dead controls.
> 
> **Volume vs mute** is split: `MediaVolumeState` gains `mutedAvailability`, `isMediaMutedCapable`, and the player `volume` feature attaches when either volume or mute is supported. **Mute** UI hides when mute isn’t available (`MuteButtonCore` / React `isSupported`; HTML sets the `hidden` attribute).
> 
> Sandbox adds the **`spotify-audio`** embed preset (fixed episode source) with HTML/React demo pages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e8af35596d90281ff1b54d615e882edc5ddfb5e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Related issues

Closes #1462 — *Feature: `<spotify-audio>`*. Part of the #1457 *Media Element Providers* epic.

Against that issue's requirements:

1. **Matches Mux `<spotify-audio>` as a baseline** — same `src` shape (`open.spotify.com/{type}/{id}`), plus `spotify:{type}:{id}` URIs and the localized/embed URL forms. The `startAt` / `theme` / `preferVideo` options are all carried.
2. **Audio-first defaults** — an audio host with no poster and no fullscreen surface.

Two deliberate divergences from the issue text, both worth a look:

- **`startAt` is spelled `t`.** The issue describes Mux's `config.startAt`, which Mux maps to the `t` URL parameter. This port follows the convention the YouTube provider set — engine config is spelled exactly as the vendor spells it — so it is `engine.spotify.t`. Say the word if you would rather match Mux's name.
- **The API script is `open.spotify.com/embed/iframe-api/v1`,** not the `embed-podcast` path upstream still uses. The `embed` path is what Spotify's current docs publish; the `embed-podcast` one predates video podcasts.

On the issue's open design question of `<spotify-audio>` versus a sibling `<spotify-video>`: this follows Mux and keeps one element, with `engine.spotify.preferVideo` selecting the `/video` embed path.

The issue's non-negotiable branding note still holds — the Spotify logo and in-iframe play button stay visible, so this cannot be skinned like a native `<audio>`.

> Note: GitHub only auto-closes on merge into the default branch, so with this PR based on `feat/cloudflare-video-media` the keyword above will link rather than close until the stack lands on `main`.

### DIFF
--- a/packages/html/src/media/spotify-audio/media.ts
+++ b/packages/html/src/media/spotify-audio/media.ts
@@ -1,0 +1,59 @@
+import { CustomMediaElement } from '@videojs/media/dom/custom-media-element';
+import { buildSpotifyIframeSrc, SpotifyMedia } from '@videojs/media/dom/spotify';
+import { escapeHtml } from '@videojs/utils/string';
+import { MediaAttachMixin } from '../../store/media-attach-mixin';
+
+class SpotifyCustomMediaElement extends CustomMediaElement('iframe', SpotifyMedia) {
+  static override getTemplateHTML = (attrs: Record<string, string>): string => {
+    const initialSrc = buildSpotifyIframeSrc(attrs.src ?? '', templateAttrsToEmbedProps(attrs));
+    const srcAttr = initialSrc ? ` src="${escapeHtml(initialSrc)}"` : '';
+    return /*html*/ `
+      <style>
+        :host {
+          display: block;
+          min-width: 160px;
+          min-height: 80px;
+          position: relative;
+        }
+        iframe {
+          position: absolute;
+          inset: 0;
+          width: 100%;
+          height: 100%;
+          border: 0;
+        }
+        /*
+         * Without Spotify's own chrome the embed is a transport and nothing else:
+         * its player UI would otherwise show through whatever skin is drawn over
+         * it. Hidden rather than merely inert, and important so a consumer's own
+         * display rule cannot put it back on screen. An iframe in a hidden subtree
+         * still loads and plays its src.
+         */
+        :host(:not([controls])) {
+          display: none !important;
+        }
+      </style>
+      <iframe
+        part="iframe"
+        ${srcAttr}
+        allow="autoplay; clipboard-write; encrypted-media; fullscreen; picture-in-picture"
+        frameborder="0"
+        width="100%"
+        height="100%"
+        referrerpolicy="${escapeHtml(attrs.referrerpolicy ?? '')}"
+      ></iframe>
+    `;
+  };
+}
+
+function templateAttrsToEmbedProps(attrs: Record<string, string>) {
+  return {
+    autoplay: attrs.autoplay !== undefined,
+    loop: attrs.loop !== undefined,
+    controls: attrs.controls !== undefined,
+    playsInline: attrs.playsinline !== undefined,
+    preload: (attrs.preload as 'none' | 'metadata' | 'auto' | undefined) ?? 'metadata',
+  };
+}
+
+export class SpotifyAudio extends MediaAttachMixin(SpotifyCustomMediaElement) {}
--- a/packages/media/src/dom/spotify/media.ts
+++ b/packages/media/src/dom/spotify/media.ts
@@ -1,0 +1,760 @@
+// Adapted from `spotify-audio-element` from `muxinc/media-elements`,
+// ported to TypeScript and reshaped as a media host to fit the v10
+// media-host architecture (mirrors `dom/youtube`).
+//
+// Source: https://github.com/muxinc/media-elements
+// License: MIT
+
+import { createPublicPromise, type PublicPromise } from '@videojs/utils/function';
+import { deepEqual } from '@videojs/utils/object';
+import { isNumber } from '@videojs/utils/predicate';
+import { EMPTY_TEXT_TRACKS, EMPTY_TIME_RANGES } from '../../core/constants';
+import { MediaError } from '../../core/media-error';
+import type { Video } from '../../core/types';
+import { MediaPlayedRangesMixin } from '../media-played-ranges';
+import { createTimeRange } from '../utils';
+import {
+  loadSpotifyIframeApi,
+  type SpotifyControllerApi,
+  type SpotifyIframeApi,
+  type SpotifyPlaybackState,
+} from './iframe-api';
+import { spotifyMediaDefaultProps } from './props';
+import { buildSpotifyIframeSrc, parseSpotifySource, type SpotifySource } from './source';
+
+const SpotifyMediaBase = MediaPlayedRangesMixin(EventTarget);
+
+/**
+ * @fires sourcechange - Fired when `source` changes, either directly or by resolving a new `src`. Read `source` for the new value.
+ */
+export class SpotifyMedia extends SpotifyMediaBase implements Partial<Video> {
+  #target: HTMLIFrameElement | null = null;
+  /** The iframe `attach()` was handed, held while the controller's own stands in for it. */
+  #controller: SpotifyControllerApi | null = null;
+  /** The controller drops `loadUri` before it reports itself ready. */
+  #controllerReady = false;
+  /** A load was requested before the controller was ready; replay it on `ready`. */
+  #pendingLoad = false;
+  /** Controller creation is in flight; the API load makes it span more than a tick. */
+  #creatingController = false;
+  #loadComplete = createPublicPromise<void>();
+  /** Guards async controller creation across attach/detach cycles. */
+  #attachId = 0;
+
+  #src = spotifyMediaDefaultProps.src;
+  #autoplay = spotifyMediaDefaultProps.autoplay;
+
+  #loop = spotifyMediaDefaultProps.loop;
+  #controls = spotifyMediaDefaultProps.controls;
+  #playsInline = spotifyMediaDefaultProps.playsInline;
+  #preload = spotifyMediaDefaultProps.preload;
+  #poster = spotifyMediaDefaultProps.poster;
+  #source: SpotifySource | null = spotifyMediaDefaultProps.source;
+
+  #paused = true;
+  #ended = false;
+  #seeking = false;
+  #loaded = false;
+  #currentTime = 0;
+  #duration = Number.NaN;
+  #readyState = READY_STATE_HAVE_NOTHING;
+  #error: MediaError | null = null;
+  /** Playback started but is stalled; the embed reports it as buffering-while-paused. */
+  #waiting = false;
+  /** A pause was asked for, so the paused update it produces is not a stall. */
+  #pauseRequested = false;
+  /** The entity ran out; kept so the end of playback is acted on exactly once. */
+  #closeToEnded = false;
+
+  static PLAYER_SOFTWARE_NAME = 'spotify-audio';
+
+  /** Underlying Spotify iframe API controller (null until the API loads). */
+  get engine() {
+    return this.#controller;
+  }
+
+  /**
+   * The iframe holding the embed, which is not always the one `attach()` was
+   * given: the iframe API replaces it with one of its own, and that one is what
+   * everything from the embed URL down is written to afterwards.
+   */
+  get target(): HTMLIFrameElement | null {
+    return this.#target;
+  }
+
+  /**
+   * Bind the iframe hosting the embed. The iframe API and its controller follow
+   * as soon as an embed URL can be resolved, which is not always now: a framework
+   * that creates the element before setting `src` attaches an iframe with
+   * nothing to embed yet, and `load()` picks it up once a source arrives.
+   */
+  attach(target: HTMLIFrameElement | null): void {
+    if (!target || this.#target === target) return;
+    if (this.#target) this.detach();
+    this.#target = target;
+    this.#beginLoad();
+    this.#createPlayer();
+  }
+
+  detach(): void {
+    if (!this.#target) return;
+    this.#attachId++;
+    try {
+      if (this.#controller) {
+        // `destroy()` removes whatever `iframeElement` points at, which is the embed
+        // this host does not own — React renders it, and the custom element keeps it
+        // in its shadow root. Pointing the controller at a throwaway first leaves it
+        // with nothing to remove, so all `destroy()` does is unbind its listener.
+        this.#controller.iframeElement = createControllerPlaceholderFrame();
+        this.#controller.destroy();
+      }
+    } catch {
+      // The iframe API throws if the iframe was already removed.
+    }
+    this.#controller = null;
+    this.#controllerReady = false;
+    this.#pendingLoad = false;
+    this.#creatingController = false;
+    this.#target = null;
+    // Unblock callers awaiting load; they re-check `#controller` (now null) and no-op.
+    this.#loadComplete.resolve();
+    this.#resetState();
+  }
+
+  override destroy() {
+    this.detach();
+    super.destroy();
+  }
+
+  get src() {
+    return this.#src;
+  }
+  /** Spotify URL or URI. Setting it re-derives `source`, carrying its embed options over. */
+  set src(value) {
+    const { engine } = this.#source ?? {};
+    const next: SpotifySource = { ...(engine && { engine }), ...(value && { src: value }) };
+
+    // Everything happens in the `source` setter, so there is one path for storing
+    // it, deciding on a load, and dispatching `sourcechange`.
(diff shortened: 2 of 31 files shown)


packages/html/src/media/spotify-audio/media.ts
```diff
@@ -0,0 +1,53 @@
...
+function templateAttrsToEmbedProps(attrs: Record<string, string>) {
+  return {
+    autoplay: attrs.autoplay !== undefined,
+    defaultMuted: attrs.muted !== undefined,
+    loop: attrs.loop !== undefined,
+    controls: attrs.controls !== undefined,
+    playsInline: attrs.playsinline !== undefined,
+    preload: (attrs.preload as 'none' | 'metadata' | 'auto' | undefined) ?? 'metadata',
+  };
+}
```
mihar-22, line 57: saw this in the other PR, might be worth drying up later but no biggie

packages/media/src/dom/spotify/props.ts
```diff
@@ -0,0 +1,38 @@
...
+  muted: boolean;
+  loop: boolean;
+  controls: boolean;
+  /** Inert: the embed decides for itself how it plays back. */
+  playsInline: boolean;
+  preload: MediaPreloadType;
+  /** Inert: the embed renders its own artwork. */
+  poster: string;
+  source: SpotifySource | null;
+}
```
mihar-22: If we can pick or pull from our existing Media API types that'd be great 🙏

packages/react/src/media/spotify-audio/media.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const composedRef = useComposedRefs(attachRef, ref);
+  const [initialSrc] = useState(() =>
+    // `source.src` is the only other way to name an entity, so honor it when `src` is absent.
+    buildSpotifyIframeSrc(props.src || props.source?.src || '', { ...spotifyMediaDefaultProps, ...props })
+  );
+  const iframeProps = useSyncProps<SpotifyMediaProps, Record<string, unknown>>(media, props, spotifyMediaDefaultProps);
+
+  return (
+    <iframe
+      title="Spotify audio player"
```
mihar-22, line 36: Could this cause issues with hover?

mihar-22 (approved):

packages/media/src/dom/spotify/media.ts
```diff
@@ -0,0 +1,660 @@
...
+    this.#closeToEnded = false;
+    this.#currentTime = seconds;
+    if (this.#seeking) {
+      // The embed announces no seeks, so a position that is no longer the one we
+      // asked for is the only sign the seek landed.
+      this.#seeking = false;
+      this.dispatchEvent(new Event('seeked'));
+    }
+    this.dispatchEvent(new Event('timeupdate'));
+  }
```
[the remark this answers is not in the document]
luwes: Real, and in two directions. `#syncPosition` treated any differing position as the seek landing, so a seek while paused to a whole second made the snapshot equal the optimistic `#currentTime`, the early return fired, and `seeked` never came — leaving `#seeking` stuck true, which also permanently blocked end detection. In the other direction, a pre-seek snapshot still in flight reverted `currentTime` backwards and fired `seeked` early.

While seeking, a snapshot now only counts as the landing if it falls within `[Math.floor(target), target + SNAPSHOT_INTERVAL]` — `seek` truncates to whole seconds and snapshots arrive about once a second. Tests: `keeps the requested position until the embed reports the seek landed`, `completes a seek that lands on exactly the position it asked for`.

Known tradeoff: an unsatisfiable seek (past `duration`) now leaves `seeking` true rather than being cleared by the next unrelated snapshot.

packages/media/src/dom/spotify/media.ts
```diff
@@ -0,0 +1,660 @@
...
+
+    if (this.#waiting && !data.isPaused) {
+      this.#waiting = false;
+      this.#readyState = READY_STATE_HAVE_FUTURE_DATA;
+      this.dispatchEvent(new Event('playing'));
+      return true;
+    }
+
+    return false;
+  }
```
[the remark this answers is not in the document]
luwes: Real, and it contradicted the host's own documented contract that the embed reports a stall as buffering while still flagged paused. `#waiting` was only ever set on the paused-to-starting transition, so a stall *after* `playing` took the pause branch and emitted `pause`, then a fresh `play` on resume — the control bar flipping to paused during ordinary buffering, exactly as described.

The pause branch now distinguishes `isBuffering` (a stall, emitting `waiting`) from an idle pause. `pause()` records that the listener asked for it, so an explicit pause during a stall is still reported as `pause` rather than being swallowed — covered by a second test so the fix doesn't overshoot.

packages/media/src/dom/spotify/media.ts
```diff
@@ -0,0 +1,660 @@
...
+    const parsed = parseSpotifySource(this.#src);
+    if (!parsed) {
+      this.#error = new MediaError(`Unrecognized Spotify source: ${this.#src}`, MediaError.MEDIA_ERR_SRC_NOT_SUPPORTED);
+      this.dispatchEvent(new Event('error'));
+      // Unblock callers awaiting load so play() doesn't hang.
+      load.resolve();
+      return;
+    }
+    // The controller names entities by URI, whatever form `src` was written in.
+    this.#controller.loadUri(`spotify:${parsed.type}:${parsed.id}`);
```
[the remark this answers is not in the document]
luwes: Real. `load()` passed only `spotify:<type>:<id>` to `loadUri`, so a later `src` or `source` change lost the `t` start position and kept the old embed chrome — even though the `source` setter reloads *specifically* to apply engine options.

`load()` now compares the embed options the new source implies against what the iframe holds. Different options rebuild `target.src`, which is the only place `theme`, `preferVideo`, and undeclared params exist; an otherwise identical source keeps the cheap `loadUri` swap and applies the start position through the seek path. Tests cover the start position, an options-only change, and `preferVideo` switching to the video path.

packages/media/src/dom/spotify/media.ts
```diff
@@ -0,0 +1,660 @@
...
+    // no longer what is playing.
+    this.#closeToEnded = false;
+    this.#duration = seconds;
+    this.dispatchEvent(new Event('durationchange'));
+  }
+
+  #syncPosition(position: number) {
+    const seconds = position / 1000;
+    if (seconds === this.#currentTime) return;
+    this.#closeToEnded = false;
```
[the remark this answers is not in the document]
luwes: Real. `Math.ceil` infers `ended` up to a second early, so the embed still emits one trailing position before our `pause()` lands, and `#syncPosition` cleared `#closeToEnded` on that move while `#ended` stayed true — after which `#restartFromEnd` never ran and `play()` just `resume()`d at the tail, re-inferring `ended` immediately.

The flag is now only cleared for a position back inside the entity (`Math.ceil(seconds) < duration`); a position still at the end belongs to the run that just finished. The `currentTime` setter clears it too, since any seek starts a new run. Test: `restarts after the end even when a trailing update advances the position`.

packages/media/src/dom/spotify/props.ts
luwes: Done — `SpotifyMediaProps` now `extends Pick<Video, ...>` with `source` as the only Spotify-specific member, matching the Cloudflare host in #2168. Same change applied across all four new providers.

One deliberate note: it picks from `Video` rather than `Audio` even though this is an audio host, because it carries `playsInline` and `poster` (absent from `Audio`) so every embed host takes the same props and the platform wrappers stay uniform. `Pick` can't carry per-member JSDoc, so the notes about which props Spotify cannot honor moved to the interface doc.

packages/html/src/media/spotify-audio/media.ts
luwes: Agreed, leaving it. Six HTML hosts have their own copy now and they aren't identical — TikTok's returns four keys, Spotify's six — so a shared version needs a per-host key set, which is a cross-provider refactor touching five files none of these PRs own. Better as its own change.

packages/react/src/media/spotify-audio/media.tsx
luwes: Checked rather than guessed: no, it doesn't.

A native tooltip is shown for the innermost node under the pointer, and over a loaded cross-origin iframe that node lives in the child document — pointer events inside it go to the embedded document, so the parent never sees the hover. The only hoverable strip would be border/padding, and we render the frame with `border: 0` filling the host. It's also why hovering a YouTube embed shows no tooltip, despite YouTube's own embed snippet shipping `title="YouTube video player"`.

Worth keeping: `title` on an `iframe` is the WCAG-documented way to name a frame (technique H64, sufficient for 4.1.2 and 2.4.1), axe-core's `frame-title` rule flags an unnamed one, and a frame has no implicit role to label otherwise. All shipped React embed hosts already do this. It stays overridable too — `title` isn't in the props type, so a consumer-supplied one passes through and wins.

One genuine gap this surfaced: the *HTML* hosts don't set `title` on their template iframes, and neither do the existing Vimeo/YouTube ones. That's a cross-provider parity fix; happy to file it separately.

PR #2180, opened 2026-08-15
refactor(html)!: return PlayerElement from createPlayer

mihar-22:
Refs #2370

## Summary

Make `createPlayer()` return the configured player element users actually define, plus a controller already bound to that player's context. The `ContainerElement` foundation from #2280 is now on `main`, so this PR applies directly on top of it.

## Before

```ts
const { ProviderMixin, PlayerController, context, create } = createPlayer({
  features: videoFeatures,
});

class CustomPlayerElement extends ProviderMixin(UIElement) {}
new PlayerController(this, context, selectPlayback);
```

## After

```ts
const {
  PlayerElement: CustomPlayerElement,
  PlayerController,
  playerContext,
} = createPlayer({ features: videoFeatures });

customElements.define('custom-player', CustomPlayerElement);
new PlayerController(this, selectPlayback);
```

The result is exactly:

- `PlayerElement`: configured class that owns the store and attachment lifecycle
- `PlayerController`: typed controller already bound to `playerContext`
- `playerContext`: lower-level context escape hatch

`ProviderMixin`, `context`, and the standalone `create` store factory are removed from the result. Presets export their named player element and typed `PlayerController`. Guides, references, demos, generated snippets, JSDoc, and agent guidance use the same API throughout.

## Testing

- `pnpm exec vp run @videojs/html#build`
- `pnpm -F @videojs/html test` — 59 files, 416 tests
- `pnpm -F site test src/utils/installation/__tests__/codegen.test.ts` — 49 tests
- `pnpm exec vp run site#build`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm check:workspace`

---

> [!NOTE]
> **High Risk**
> Public HTML `createPlayer` surface is rewritten (removed mixin/factory/context args). Existing custom-element integrations will not compile or run without a migration.
>
> **Overview**
> **Breaking HTML API:** `createPlayer()` now returns a configured `PlayerElement` class (store + attach lifecycle) and a `PlayerController` already bound to that player's context. `ProviderMixin`, `context`, and the standalone `create()` store factory are gone.
>
> Custom players are defined directly (`customElements.define('…', PlayerElement)`). Controllers take only host + optional selector. Preset elements (`video-player`, audio, live, background) extend that class and re-export the bound controller.
>
> Docs, demos, and changelog follow: player reference is `/reference/player`, packaged skins make the player boxless, and layout CSS targets `media-container`.
>
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6bdfa2bf092d3f5b839aee754a73980ac7da235c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/html/src/define/audio/player.ts
```diff
@@ -1,18 +1,20 @@
...
   features: audioFeatures,
 });
 
-export class AudioPlayerElement extends ProviderMixin(MediaElement) {
+/** Player controller bound to the standard audio player store. */
+export const PlayerController = ConfiguredPlayerController;
+
+export class AudioPlayerElement extends PlayerElement {
   static readonly tagName = 'audio-player';
 }
```
sampotts: nit: Could we drop the Player prefix from the returned properties since they’re already scoped by createPlayer()? For example:

```ts
const {
  Element: PlayerElement,
  Controller: PlayerController,
} = createPlayer()
```

That feels pretty consistent with APIs that return a namespaced object and let consumers choose local names. For example, React’s createRoot() returns a root object with methods like render() / unmount(), and TanStack Query’s useQuery() returns generic properties like data, error, status, refetch, etc. rather than queryData, queryError, and so on.

The only slight wrinkle is Element itself being a DOM global, but aliasing on destructure avoids that nicely.

sampotts: GitHub mucked up - should associated to line 6.

mihar-22: I am torn here. 99% of the time I would agree and try to push simpler naming here since it's scoped by "createPlayer". There was two reasons of doing this: (1) we're trying to align the naming as much as we can with preset imports so there's more consistency jumping between the two worlds of preset/eject, and (2) encouraging AI to be consistent when naming and using these. You can imagine in some cases it could alias Element/Controller in very strange ways. 

That said I'm not fully convinced and I still prefer what you have there 😆 Maybe given HTML naming/expectations are different and these concepts being named out the box is clearer maybe we leave it...

sampotts (approved):

PR #2226, opened 2026-08-17
feat(ci): write changelog prose onto the release PR

decepulis:
Changelog prose now goes onto the release PR, before the cut, so it reaches production without a manual cherry-pick.

## The problem

`cd.yml` force-pushes `main` to `site/v10` as the last step of the release job. `changelog-prose.yml` was triggered *by* that release, so the prose PR didn't exist yet when production moved, and merging it later only touched `main`. videojs.org kept the raw release-please bullets until the next release swept them up — hence the manual cherry-pick onto `site/v10` after every release.

## The change

Dispatch the workflow with `target: release-pr` and it checks out `release-please--branches--main`, reads the pending version from the manifest, and commits the prose onto that branch. Merging the release PR then ships the polished changelog with the tag, the npm publish, and the `site/v10` push in one step. The release PR is the review surface, so the prose gets read alongside the release it describes.

## It can't block a release

- The job is `workflow_dispatch` only, so it's never a required check on the release PR.
- It only ever *adds* a commit. A failed, skipped, or rejected run leaves the raw release-please changelog exactly where it was and the release PR mergeable.
- It runs outside `cd.yml`, so it can't fail the publish.
- If the push is rejected because release-please rewrote the branch, the run stops and reports instead of forcing.

## Backstop

The `release: published` relay stays, dispatching `target: main` — the old behavior — for a release cut without prose. The guard now also skips when the file's `description` is already filled in, so the backstop stays quiet after a successful pre-cut run rather than opening a duplicate PR, and a re-dispatch can't rewrite prose from prose.

## Note on timing

`release-pr.yml` rewrites the raw changelog file on every push to the release branch, so anything merged to `main` after the prose run wipes it. That's deliberate — a stale changelog is worse than a regenerated one, and the release PR diff shows plainly when it's gone. Dispatch prose when you're ready to cut, then merge.

---

> [!NOTE]
> **Low Risk**
> CI/docs-only workflow and README changes; no runtime app code, and failed prose runs still leave releases mergeable with the raw changelog.
> 
> **Overview**
> **Changelog prose now defaults to the open release PR** so polished `site/src/content/changelog/*.mdx` ships with the same merge that tags, publishes, and force-pushes `main` → `site/v10`, avoiding the old lag where prose only hit `main` after production moved.
> 
> The **Changelog Prose** workflow gains a `target` input (`release-pr` vs `main`). For `release-pr`, it checks out `release-please--branches--main`, resolves the pending version from `.release-please-manifest.json`, shares the `release-pr` concurrency group with `release-pr.yml` (no racing raw changelog writes), and instructs Claude to **commit and push** to that branch instead of opening a PR. The `release: published` relay still dispatches with `target=main` for releases cut without prose.
> 
> Guards are tightened: skip when the raw file is missing **or** when `description` is no longer empty (so backstop and re-dispatch don’t duplicate or overwrite prose). `site/README.md` documents running the workflow with `target: release-pr` before merge and the cherry-pick fallback when prose runs post-release.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0fefd66f507613753f28a0d75ea47d0a91b6975c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/changelog-prose.yml
+++ b/.github/workflows/changelog-prose.yml
@@ -1,25 +1,43 @@
 name: Changelog Prose
 
+# Prose normally goes onto the open release PR, before the release is cut, so
+# that merging the release PR ships the polished changelog with the tag, the
+# npm publish, and the site/v10 force-push in one step. Nothing here gates the
+# release: this workflow only ever adds a commit, so a failed or skipped run
+# leaves the raw release-please changelog in place and the release PR mergeable.
+# The post-release path below stays as a backstop for a release that was cut
+# without prose.
 on:
   release:
     types: [published]
   # claude-code-action does not support the `release` event, so the dispatch
-  # job relays core releases into a workflow_dispatch run. Manual dispatch
-  # also lets us backfill prose for a release that was missed.
+  # job relays core releases into a workflow_dispatch run.
   workflow_dispatch:
     inputs:
-      tag_name:
-        description: 'Release tag (e.g. @videojs/core@10.0.0-beta.24)'
+      target:
+        description: 'Where the prose lands'
         required: true
+        default: release-pr
+        type: choice
+        options:
+          # Commit onto release-please--branches--main, before the cut.
+          - release-pr
+          # Open a PR against main, after the cut. Needs tag_name.
+          - main
+      tag_name:
+        description: 'Release tag, for target=main (e.g. @videojs/core@10.0.0-beta.24)'
+        required: false
         type: string
       release_url:
         description: 'Release URL (defaults to the tag release page)'
         required: false
         type: string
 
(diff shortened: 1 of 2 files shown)


decepulis:
Superseded by #2228, which takes the forward-port approach instead.

Writing prose before the cut meant triggering off pushes to the release branch, and that fires on every merge to `main` — ~8/day here, peaking at 20. That's an Opus run per merge, each one clobbered by the next, to produce a single changelog. #2228 leaves prose where it is and just automates the cherry-pick to `site/v10`.

PR #2242, opened 2026-08-18
feat(spf): Cap Rendition to Player Size

### LINKED ISSUE #1785: Feature: Cap Rendition to Player Size for SPF

spuppo-mux:
Closes #1785

## Summary

Automatically cap video rendition selection to the player's rendered size, so a small player stops pulling renditions larger than needed. Enabled by default. The cap is the *smallest tier that still covers the player*, plus everything below it.

## Changes

- `media/dom/element-size.ts` - Adds two observers which you can pass a onChange function to.
    - `observeElementSize`, which just passes the width and height to the callback (content box from the `ResizeObserver` entries)
    - `observeRenderedSize`, which also composes it with the device-pixel-ratio watcher.
    - Both functions return their cleanup function.
- `media/dom/device-pixel-ratio.ts` - `watchDevicePixelRatio`, via a re-armed `(resolution: …dppx)` query. Browser zoom or a move to another display changes the ratio without resizing the element, so the observer alone would miss it.
- `observePlayerSize` (`behaviors/dom/`) - a behavior which writes the measurement as distinct slots:
    - `playerWidth` / `playerHeight` (CSS pixels) and `playerScale`.
- `capToPlayerSize` - a new soft rule in `switchVideoTrack`'s chain, after the user filter (so a manual pick still wins) and before the ranker (which then chooses within the cap). Combines box x scale into a device-pixel area; renditions declaring no RESOLUTION are never capped out.
- Engine config:
     - `playerSizeCap`: `enabled` gates the measurement, 
     - `useDevicePixelRatio`: gates whether the cap scales it to device pixels.
- Sandbox: resizable video frame + a readout of the measurement and the cap it implies.

## Not handled yet

- **PiP and iOS native fullscreen.** In both, the in-page element keeps its layout box while the viewer watches a different surface, so the cap stays stuck at the inline size. Follow-up: go inert while the element is in a presentation mode we can't measure, then measure those surfaces properly (`PictureInPictureWindow`, `screen`).
- **Min cap / floor.** Deliberately left out. Can probably be added as a separate configurable rule.

AirPlay is unaffected: The receiver plays the native-HLS fallback under Safari's own ABR

## Testing

- `pnpm -F @videojs/spf test` (1234 passed)
- `pnpm typecheck`
- Biome on changed files
- Sandbox `spf-segment-loading`: dragged the frame corner across tier boundaries and
  watched the pick follow

---

> [!NOTE]
> **Medium Risk**
> Changes default video ABR selection (new cap in the switch chain). Wrong measurements or rounding could pick a lower/higher quality than intended, but the cap is a soft filter and stays inert without a size reading.
> 
> **Overview**
> Automatically caps ABR video selection to the player’s rendered size (on by default), so a small embed no longer pulls renditions larger than it can show. The cap is the **smallest ladder tier that still covers the player**, plus everything below it — not a hard at-or-below-area cut that would underserve a player sitting between tiers.
> 
> `trackPlayerResolution` measures the media element (device pixels by default) into `state.playerResolution`. `switchVideoTrack` then applies a new soft `playerResolutionCap` after the CDN scope and before bandwidth ranking. No measurement, a player larger than every rendition, or a manual pick leaves the cap inert. Tracks without RESOLUTION are never dropped.
> 
> Shared helpers land in `scaleResolution` / `smallestCoveringPixelArea`, and screen reading now uses the same `Resolution` type plus `watchDevicePixelRatio`. Engine config: `capRenditionToPlayerSize` and `useDevicePixelRatio`. The sandbox video is resizable and shows the implied cap.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c154ea521d40fb17c6233b770e634d70ae658d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/spf/src/media/dom/screen.ts
```diff
@@ -152,27 +152,7 @@ export function watchScreenResolution(
...
-      () => {
-        watchRatio();
-        check();
-      },
-      { once: true, signal }
-    );
-  };
-
-  watchRatio();
+  watchDevicePixelRatio(check, signal);
```
cjpillsbury, line 146: praise: 😎

packages/spf/src/playback/behaviors/dom/track-player-resolution.ts
```diff
@@ -0,0 +1,124 @@
...
+ * Project an observed CSS box into device pixels, or `undefined` when there is no
+ * measurement in it.
+ *
+ * Rounded because device pixels are whole and a fractional ratio doesn't divide a
+ * box evenly, the same as `getScreenResolution`. A zero axis is "not measurable"
+ * rather than a measurement of zero: an element that isn't being rendered reports
+ * a `0 × 0` box, and capping to an area of zero would pin every source to its
+ * smallest rendition.
+ */
+function toPlayerResolution(size: ElementSize & { scale?: number }): PlayerResolution | undefined {
```
cjpillsbury: nitpick(non-blocking): pull this into e.g. media/dom somewhere (I thought there was already a function like this for screen rez?) and change the names to something more generic, since the code is just about applying a scale and normalizing to integer units.

packages/spf/src/playback/behaviors/track-switching.ts
```diff
@@ -379,6 +398,44 @@ function filterByUserSelection<S extends SelectionKey, U extends UserSelectionKe
...
+ * means "don't cap" rather than a cap of zero, so the chain proceeds unnarrowed.
+ */
+function playerResolutionCap<S extends SelectionKey, T extends SwitchableTrack>(
+  tracks: readonly T[],
+  { state }: SelectionRuleDeps<PlayerResolutionCapStateMap<S>, AnySlotMap, TrackSwitchingConfig<S, T>>
+): readonly T[] {
+  const playerResolution = state.playerResolution?.get();
+  if (!playerResolution) return tracks;
+
+  const playerPixelArea = playerResolution.width * playerResolution.height;
```
cjpillsbury: thought(non-blocking): This could probably be rewritten a bit more simply. Also, like the aforementioned function, this looks like it could be broken into its own, more generic function.

packages/spf/src/playback/behaviors/track-switching.ts
```diff
@@ -379,6 +398,44 @@ function filterByUserSelection<S extends SelectionKey, U extends UserSelectionKe
...
+
+  const covering = tracks.map((track) => resolutionArea(track)).filter((area) => area >= playerPixelArea);
+  // Nothing covers the player — no opinion, fall through to the full set.
+  if (!covering.length) return [];
```
cjpillsbury: nitpick(blocking): While currently functionally equivalent, you should `return tracks` here (like above) to properly express what the rule is asserting. (add/update tests appropriately for this as well).

packages/spf/src/playback/behaviors/track-switching.ts
```diff
@@ -707,7 +764,7 @@ export const switchVideoTrack = defineBehavior({
         userSelectionKey: 'userVideoTrackSelection',
         getTracks: (presentation) => getTracksByType(presentation, 'video') as readonly VideoTrackCandidate[],
         constraints: [excludeFailedCdns, excludeUnplayableTracks],
-        rules: [filterByUserSelection, preferActiveCdn, rankByBandwidth],
+        rules: [filterByUserSelection, playerResolutionCap, preferActiveCdn, rankByBandwidth],
```
cjpillsbury: nitpick(blocking): Although an edge case, I think we want this to come after "prefer active CDN". Based on the way rules stack, this could _**hypothetically**_ cause a cdn switch if there's a (very nonstandard but technically in spec) mismatch between the video renditions' resolutions across CDNs. In your followup effort to pull this stuff into its own architecture, these are the sorts of things that would be good to add to tests (but consider not necessary for this PR).

cjpillsbury (approved):
A couple of very small nitpicks, but pre-approving on the assumption you'll resolve before merging. Overall, 👌

PR #2243, opened 2026-08-18
feat(media): cap hls.js renditions to the player size

### LINKED ISSUE #1783: Feature: Cap Rendition to Player Size for <hlsjs-video> and hlsjs-powered <mux-video>

R-Delfino95:
Closes #1783

Builds on the cap controller added in #2061.

## Summary

Adds `source.capRenditionToPlayerSize` and `source.minAutoResolution` to hls.js-backed medias. The player-size cap itself isn't new: `defaultHlsConfig` has shipped `capLevelToPlayerSize: true` since the beginning, undocumented and only reachable by rebuilding the engine. This turns it into a normalized option, makes it apply to the running engine, and adds the floor hls.js has no concept of.

* **The toggle (`cap-level.ts`)** — hls.js derives its player-size ceiling inside `getMaxLevel()`, the same call that filters out the levels `capLevelOnFPSDrop` has restricted. Returning early to skip the size cap would silently drop those restrictions too, and `restrictedLevels` is private, so there's no way to reapply them afterwards. The *measurement* is neutralized instead: `mediaWidth` and `mediaHeight` report `Infinity` while the cap is off, so hls.js's own arithmetic lands on the top allowed level and everything else it owns stays in the path. One seam for both behaviors rather than two divergent branches.

* **The floor** — `minAutoResolution` bounds the size-derived cap and nothing else. It's applied to that ceiling *before* the requested one narrows the result, so `maxAutoResolution: '360p'` alongside a `'720p'` floor still yields 360p. It isn't a quality minimum either: the cap is a ceiling, so ABR keeps selecting below it on a poor connection. And it's clamped to what the base controller allows without the measurement, so it can't hand back a rendition dropped frames have restricted.
Default `'720p'`. The low rungs of a ladder exist for bad network conditions, and capping a small player onto them looks worse than its size alone suggests. Name a lower rung to weaken the floor, or `'270p'` to lift it for any real ladder.

* **Device pixel ratio** — no code. hls.js already multiplies the measurement by `devicePixelRatio` (`ignoreDevicePixelRatio: false` by default), so a 640px player on a retina screen asks for 720p rather than 360p. The issue's DPR criterion is met by exposing the option; it's pinned with tests rather than implemented.

* **Wiring** — both keys ride the mutable policy #2061 introduced, so neither enters `#engineConfigKey()` and changing either never rebuilds the engine. No new plumbing.

## Changes

- `cap-level.ts`: the policy gains `capToPlayerSize` and `minAutoResolution`; `levelIndexAtOrAbove` is added as the complement of `levelIndexAtOrBelow` and reasoned the same way — `autoLevelCapping` is a ceiling on the *index*, so the answer is the first rung that reaches the floor, not the best match. `DEFAULT_MIN_AUTO_RESOLUTION` carries the rationale for `'720p'`.
- `hls-js-only.ts`: two more live properties. Defaults resolve inside the setters, so the getters report the cap actually in force rather than what a source happened to name. `capLevelToPlayerSize` stays `true` in `defaultHlsConfig`, now with a comment saying why: it runs the loop every cap here is evaluated on, `maxAutoResolution` included.
- `media.ts`: both keys on `HlsSource` with JSDoc. Carried through the `src` setter on definedness rather than truthiness — `false` is the value worth naming, and it's the falsy one. The `__DEV__` warning for native playback now names whichever keys were ignored.
- `mux/media.ts`: the same carry-over in its own `src` setter. The `source` JSDoc separates the client-side pair from Mux's server-side `playback.maxResolution` / `playback.minResolution`, which look alike and aren't related.
- Docs: a "Capping to the player's size" section in `concepts/media-sources.mdx`, plus a table of the four mechanisms that sound alike and a paragraph stating outright that `minAutoResolution` is not a quality floor.
- Sandbox: `?capRenditionToPlayerSize=` and `?minAutoResolution=` alongside the params #2061 added.

## Testing

- `pnpm -F @videojs/media test` — 955 pass (+38: the toggle, the floor, their interaction with `maxAutoResolution`, the `capLevelOnFPSDrop` edges, and DPR).
- `pnpm -F @videojs/html test` — 349 pass. `pnpm -F @videojs/react test` — 456 pass.
- `pnpm typecheck`, `pnpm lint`, `pnpm check:workspace`, `pnpm -F site astro check` — clean. Full `pnpm test` green across all 22 packages.

Verified in the sandbox against a real Mux ladder (270/480/720/1080) at 384px CSS on a DPR 2 screen — 768 device pixels, where the size cap alone lands on 480p:

| Source | Cap |
|---|---|
| *(defaults)* | 720p — the floor lifts the size cap |
| `minAutoResolution: '270p'` | 480p — the size cap on its own |
| `capRenditionToPlayerSize: false` | 1080p — size stops binding |
| `maxAutoResolution: '480p', minAutoResolution: '1080p'` | 480p — the floor doesn't outrank the ceiling |
| `maxAutoResolution: '270p', minAutoResolution: '720p'` | 270p — same, further apart |

The `Hls` instance is identical across all of them and all four renditions stay in `videoRenditions`. Re-checked with a level restricted by `capLevelOnFPSDrop`: the floor stops at the top allowed rung instead of re-admitting the restricted one.

For manual testing, the query params work on the sandbox deploy of this branch:

```
/html-hlsjs-video/?source=hls-2&minAutoResolution=270p
/html-hlsjs-video/?source=hls-2&capRenditionToPlayerSize=0
/html-hlsjs-video/?source=hls-2&maxAutoResolution=270p&minAutoResolution=720p
```

`source=hls-2` matters: its ladder reaches 1080p, so the 720p floor doesn't swallow the whole thing. Narrow the window until the player is under ~430px CSS on a retina screen, otherwise the size cap never binds and every row reads the same. Or from the console on a running player:

```js
const el = document.querySelector('hlsjs-video'); // or mux-video
el.source = { ...el.source, capRenditionToPlayerSize: false };
```

## Notes

- The `capLevelOnFPSDrop` interaction is resolved in both directions rather than left implicit. Switching the size cap off keeps the restrictions, and the floor can't lift the ceiling back onto a restricted level — a floor says "don't cap this small player that far down", not "decode what this device can't". Both are asserted.
- `source.engine.hlsJs.capLevelToPlayerSize: false` still stops hls.js's capping loop outright, which switches off every cap here and takes an engine rebuild. That's passthrough behaving as documented; the normalized key is the live one. Called out in the JSDoc.
- No `useDevicePixelRatio` key. DPR handling already works by default, `ignoreDevicePixelRatio` and `maxDevicePixelRatio` stay reachable through `source.engine.hlsJs`, and SPF's in-flight branch exposes one — adding a normalized key later is purely additive. Say the word if you want it now.
- `maxDevicePixelRatio` is left at `Infinity`. Nothing clamps it today — not hls.js, not Mux Player, not the SPF branch — so there's no prior decision to honor.
- No HTML attributes for either key. Both are normalized source options like `preferPlayback`; adding attributes later is additive.
- hls.js compares a rendition's largest dimension against the element's, deliberately, so an aspect-ratio change can't make capping oscillate. `maxAutoResolution` compares pixel area, matching SPF. The two heuristics can land on different renditions for the same player and ladder — documented in the JSDoc rather than unified here.

---

> [!NOTE]
> **Medium Risk**
> Changes automatic rendition selection for all default hls.js playback; behavior is heavily tested but misconfiguration could affect bandwidth and perceived quality.
> 
> **Overview**
> Adds **`source.capRenditionToPlayerSize`** (default on) and **`source.minAutoResolution`** (default **`720p`**) for hls.js-backed sources, turning the existing undocumented hls.js player-size cap into normalized options that **apply live** without rebuilding the engine.
> 
> The cap-level controller now **lifts or disables** the size-derived ceiling via reported dimensions (so **`capLevelOnFPSDrop`** restrictions stay intact), applies **`minAutoResolution`** as a floor on that size cap (not a quality minimum—**`maxAutoResolution`** still wins when stricter), and adds **`levelIndexAtOrAbove`** for floor math. **`HlsSource`**, **Mux** `src` carry-over, **sandbox** query params, and **docs** (including how these differ from Mux **`playback.*Resolution`**) wire the options through **HTML/React** tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4fa32ee8fb6ede77cdbb1864d05812e9be2ebde2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


luwes (approved):
LGTM 👍 
I was considering the naming a little but probably best to keep it consistent with mux-player

PR #2250, opened 2026-08-18
feat(html): add server-safe conditional exports

mihar-22:
Refs #1343

## Summary

Allow every public `@videojs/html` entry to be evaluated during SSR, prerendering, and server/edge module analysis without requiring DOM globals. Browser conditions continue to resolve the real custom-element implementation.

## Changes

- publish DOM-free facades for the default, React server, Edge, workerd, and worker conditions
- derive the `createPlayer` facade from the source API, so it stays compatible before or after the PlayerElement change
- preserve constructable classes, callable definitions/mixins, feature arrays, and safe dependency re-exports
- validate every public import and mixed condition order during the package build
- document Next.js, SSR, and browser-worker behavior

## Testing

- `pnpm -F @videojs/html build`
- 178 public imports across default, React server, Edge, workerd, worker, and browser conditions
- future stack merge (`PlayerElement` + pure preset exports): all 10 dependency builds and server checks pass

### DIFF
--- a/packages/html/package.json
+++ b/packages/html/package.json
@@ -9,7 +9,7 @@
     "url": "https://github.com/videojs/v10",
     "directory": "packages/html"
   },
-  "main": "dist/default/index.js",
+  "main": "dist/server/index.js",
   "module": "dist/default/index.js",
   "types": "dist/dev/index.d.ts",
   "sideEffects": [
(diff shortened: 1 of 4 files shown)


mihar-22:

PR #2286, opened 2026-08-20
fix(spf): report a verdict when the background ladder is undecodable

cjpillsbury:
Fast-follow to #2135, which merged without this fix.

## Summary

A background video whose whole ladder is undecodable in this environment — an HEVC or
AV1 source — stalls silently: no error is reported anywhere, and the element sits at
`readyState 0` with `error` null. This is the exact failure mode the unplayable-source
surface added in #2135 exists to close, left open for the one cause that is knowable
before playback starts.

Scoped to the background-video engine. `reportAbsentTrackType` is composed by no other
engine, so nothing else changes behavior; the remaining files are comment corrections.

## Changes

- `createBackgroundVideoEngine` now defaults its constraints to
  `[excludeUnplayableTracks, reportAbsentTrackType(2011)]` rather than the reverse. At the
  tail of the chain an empty input means "nothing playable here", which is what 2011
  already denotes — renditions all excluded as unplayable, *or* none offered to begin
  with. Only the second half was reaching a consumer.
- A container or encryption failure can now report `[1004, 2011]` rather than `[1004]`.
  `media.error` is unchanged: first-fatal-wins still surfaces the cause, and the 99001
  substitution is untouched. Only the logged sequence grows.
- Comments in `collect-errors`, `selection-rules`, and the background adapter argued for
  the old position, or claimed a verdict the pinned selection never emitted. They now
  describe what the code does.

<details>
<summary>Why codec is the only cause with nothing behind it</summary>

`reportAbsentTrackType` fires on an empty input at whatever position it holds, so at the
head of the chain it only ever saw the raw track list.

`CODECS` is in the **multivariant** playlist, so `excludeUnplayableTracks` can prune an
undecodable ladder *before* anything is picked. Nothing resolves, so
`reportUnsupportedTrackConditions` never runs and no cause exists — and the type was
still non-empty when the head-position reporter looked at it, so no verdict was emitted
either.

Container and encryption escape this only because neither is knowable until a media
playlist resolves: `parse-multivariant` labels every video track `video/mp4`, so those
renditions are always picked and resolved first and report 1004 or 4008 on the way down.

The order of `applyConstraints` cannot change *which* tracks survive, but this rule's
report is a side effect, and side effects are order-dependent. That distinction is now
written down where the old claim was.

</details>

## Testing

`pnpm -F @videojs/spf test` — 1519 passing / 11 skipped, including a new
`'reports for a ladder the environment cannot decode'` case (two-rendition HEVC ladder,
`canPlayTrack: () => false`), confirmed failing against the previous ordering.

`'makes no pick for an unplayable container'` is updated: it had codified the gap by
asserting `errors` stays `undefined`. The adapter needed no new test — its existing
`'explains a source with no video renditions too'` already drives the `[{code: 2011}]`
sequence through `media.error` and the console log, which is what this now produces.

Also green: `@videojs/html` and `@videojs/react` background-media suites, `pnpm typecheck`,
`pnpm check:workspace`.

---

> [!NOTE]
> **Medium Risk**
> Changes when the background engine emits fatal 2011 (including stacking it after 1004/4008). Adapter mapping and first-fatal-wins are unchanged, but error sequences and consumer-visible failures for codec-pruned sources now fire.
> 
> **Overview**
> Closes a silent stall on background video when the whole ladder is pruned by capability (e.g. HEVC/AV1) before anything resolves.
> 
> `createBackgroundVideoEngine` now defaults constraints to `[excludeUnplayableTracks, reportAbsentTrackType(2011)]`. At the **tail**, an empty set means “nothing playable” — no video offered *or* everything pruned — so 2011 covers codec-only failures that never hit `reportUnsupportedTrackConditions`. Container/DRM paths can now stack `[cause, 2011]`; first-fatal-wins still surfaces the cause, so `media.error` is unchanged.
> 
> Scoped to this engine (`reportAbsentTrackType` is not composed elsewhere). Tests cover an injected-undecodable HEVC ladder and expect 2011 after an unplayable-container prune.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da2fd798e0d5296e11d4a93565de879b538defc1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/design/spf/features/errors.md
+++ b/internal/design/spf/features/errors.md
@@ -226,7 +226,7 @@ DOM-free, so the codes are usable from any layer: `SvtaError`
 | `collectErrors` | `playback/behaviors/collect-errors.ts` | Owns the `errors` slot and its per-source lifecycle. No effects, no policy — a lifecycle owner, not an error handler. Same slot-owner-vs-writer split as `setupFailoverMonitor` / `failedCdns` |
 | `switchVideoTrack` / `switchAudioTrack` / `switchTextTrack` | `playback/behaviors/track-switching.ts` | Report the **verdict** (2011 / 2012) when a type has renditions but constraints pruned every one. Per-variant `noSupportedTrackCode`; text supplies none, since absent subtitles aren't a failure. Reports generically — it never reads a constraint's state, so it doesn't know *why* the set emptied |
 | `selectVideoTrack` | `playback/behaviors/select-tracks.ts` | The *pinned* variant. `entry` stays the only thing that ever **selects**; a sibling `effects` entry on the same `presentation-resolved` state only ever **de**selects, dropping a pick the constraints turned against. Keeping selection out of the reaction is what separates this from `switchVideoTrack` — a re-pick here would make it that behavior with extra steps — but it must *be* a reaction, since container and encryption are only known once a media playlist resolves, after `entry` ran. It reports nothing itself: whatever made the pick unplayable already reported its own cause (1004 / 4008), more specific than a verdict and already logged. A pick naming a track the manifest never offered is left alone, preserving the module's external-writes contract |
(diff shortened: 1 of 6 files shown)


spuppo-mux (approved):
Lgtm!

Issue #1776, opened 2026-07-01
Feature: SPF DRM


decepulis:
DRM for the SPF playback engine. Stub — details TBD.

PR #2291, opened 2026-08-20
feat(spf): add EME-based DRM support to the HLS engine

cjpillsbury:
Refs #1776

Lets the SPF HLS engine play DRM-protected content, using the browser's Encrypted Media Extensions (EME). In order: pick a content decryption module (CDM) the device actually has, fetch its certificate where one is needed, attach it to the video element, hold encrypted segments until that is done, fetch a license per key the stream declares, and report expiry or output restriction rather than showing a black frame. AirPlay gets its own path, because Safari hands the stream to the receiver and the receiver asks for keys itself.

Verified playing against real CDMs on all three key systems — Widevine, FairPlay, PlayReady — across Mux and third-party providers.

The full verification detail, the designs that were rejected, and the open questions are in `internal/design/spf/features/drm-support.md`. This description is meant as a guide to reviewing the diff.

## Where to look

106 files, but the reviewable surface is ~2,700 lines across nine. Suggested order:

| File | +lines | What to check |
|---|--:|---|
| `media/drm.ts` | 496 | The shape apps pass in (`DrmSystemsConfig`) — the public API this commits to, so worth the closest read |
| `media/dom/key-systems.ts` | 197 | Everything that differs between Widevine, PlayReady and FairPlay, as one value each. Supporting a new key system should mean adding a value here and changing nothing else |
| `media/dom/eme.ts` | 304 | Asking the browser for a CDM. Tries strongest security level first and steps down; if every level is refused, a second pass asks without naming one |
| `behaviors/dom/setup-media-keys.ts` | 346 | Picks a CDM, applies FairPlay's certificate, attaches the keys. The only writer of the three pieces of state everything else reads |
| `behaviors/dom/exchange-licenses.ts` | 195 | Gets the actual licenses: one CDM session per key the stream declares, then the request/response round trip per session |
| `media/dom/license-sessions.ts` | 257 | Talking to the license server — the fetch, retry and request/response rewriting shared by both DRM paths |
| `behaviors/dom/load-segments.ts` | +57/−13 | Where encrypted segments are held until keys attach. A small diff doing load-bearing work — the fix in `92508117d` is here |
| `behaviors/dom/setup-airplay-fairplay.ts` | 539 | Re-does all of the above for an AirPlay receiver, which asks for keys in a different format than the one negotiated for local playback. Self-contained — skippable if AirPlay is not your concern |
| `media/dom/fairplay-legacy.ts` | 256 | The Safari-only fallback, reachable only from the behavior above. See the last decision below for why it exists |

The remainder: ~2,300 lines of tests, plus sandbox, e2e, engine wiring and the design record.

## Surface beyond those nine files

Reviewer-relevant because it changes public API or reaches other packages:

- `parseMediaPlaylist` now reads the `#EXT-X-KEY` tags out of a playlist and exposes them as structured data (METHOD / URI / KEYFORMAT / KEYID / IV). The existing boolean `encrypted` flag is now derived from that rather than set separately.
- The `hls-video` and `hls-audio` Medias take a structured `source` (`src`, `type`, `drm`). Every composed key system is named once up front, and its settings are read from whichever source is current at the time — so switching sources does not rebuild the engine. The SPF `mux-video` Media derives Mux's three license servers from `drm.token`.
- Per-key-system `headers` — either fixed values or a function called at request time, which is how token auth is usually threaded — and `credentials`, plus per-source `licenseRequest` / `licenseResponse` / `certificateRequest` / `certificateResponse` transforms. Providers disagree about how license requests and responses are wrapped, so `detectFairPlayCkc`, `unwrapJsonLicense` and `formEncodeLicenseRequest` ship as optional helpers for the common variations; unused ones drop from the bundle.
- License and certificate fetches are retried and time-bound (`fetchWithRetry`: capped backoff on 5xx / network errors / first-byte timeouts, 4xx fatal) . This is the first place that retry policy exists; segment and manifest fetches can adopt the same helper later.
- A sandbox `spf-drm` page and DRM sources on the `hls-video` / `mux-video` presets, plus an opt-in HTTPS dev server (EME needs a secure context, so phones and TVs on the LAN need one) and a `?drm=` filter for machines with more than one CDM.

## Verifying this yourself

**Already automated, no setup:** a Clear Key suite runs on every PR and drives the whole path — pick CDM, attach, open session, exchange license, **decode** — against bundled Chromium, with only the license URL intercepted. The decode assertion fails under a wrong key, so it proves decryption rather than plumbing. Separately, `pnpm test:e2e:drm` negotiates all seven sandbox DRM sources against a local Chrome's real Widevine CDM and asserts every request names a security level; it skips itself if Chrome is missing, and is outside `test:all` because hosted CI cannot provision a CDM.

**Everything else is manual**, because a CDM is tied to a real browser on a real OS. The sandbox page exists for exactly this:

```
pnpm dev:sandbox          # http://localhost:5173/spf-drm/
pnpm dev:sandbox:https    # for phones/TVs — EME needs a secure context
```

**No checkout needed** — this branch deploys the sandbox to a public HTTPS URL, which is also what phones, tablets and TVs need, since EME only works in a secure context. Every link below is live right now and tracks the branch.

**Start here, on whatever machine you already have:**

- **Mux, all three key systems configured** — Widevine on Chrome, Firefox, Edge and Android Chrome; FairPlay on Safari and iOS. Expect a picture.
  https://v10-sandbox-git-feat-spf-drm-mux.vercel.app/spf-drm/?source=hls-drm

- **Must fail** — the same asset with no license available. Expect a reason per encrypted quality level, then one "nothing playable" verdict. A success here is the real bug.
  https://v10-sandbox-git-feat-spf-drm-mux.vercel.app/spf-drm/?source=hls-drm-unlicensed

- **Clear control** — no DRM at all. Run this first if something looks broken, to separate your setup from this PR.
  https://v10-sandbox-git-feat-spf-drm-mux.vercel.app/spf-drm/?source=hls-3

**Edge on Windows needs one extra step.** It ships both a Widevine and a PlayReady CDM, and Widevine wins by default — that is a deliberate preference order — see Decisions below — not a bug. To exercise PlayReady:

  https://v10-sandbox-git-feat-spf-drm-mux.vercel.app/spf-drm/?source=hls-drm&drm=playready

**Then vary the provider**, since license-server dialects differ more than the key systems do:

- **Shaka's demo server** — no auth, no Mux involvement anywhere.
  https://v10-sandbox-git-feat-spf-drm-mux.vercel.app/spf-drm/?source=hls-drm-widevine-cwip

- **Axinom** — third-party auth carried in a request header.
  https://v10-sandbox-git-feat-spf-drm-mux.vercel.app/spf-drm/?source=hls-drm-axinom

- **Axinom MultiKey** — a different key per quality tier. Read Known gaps before concluding anything from a failure here.
  https://v10-sandbox-git-feat-spf-drm-mux.vercel.app/spf-drm/?source=hls-drm-axinom-multikey

- **EZDRM** — a second FairPlay provider. Safari only; it has no Widevine or PlayReady configured.
  https://v10-sandbox-git-feat-spf-drm-mux.vercel.app/spf-drm/?source=hls-drm-ezdrm

`?drm=widevine|playready|fairplay` forces one key system on any of these, for machines with more than one CDM.

**What "working" looks like:** the page prints a live readout — `readyState`, `currentTime`, `framesDecoded`, `negotiatedKeySystem` and `errors`. Judge by `framesDecoded` climbing with `errors` empty, not by the picture: on hardware-backed Widevine the video region is legitimately black in any screen capture or remote-desktop session.

### Two things are expected to fail

Neither is caused by this PR. Worth reporting only if the *symptom* differs from what's described.

**Microsoft's PlayReady test asset never plays.** Bad key configuration in the asset itself, predating this branch. Windows/Edge only. PlayReady's real evidence here is Mux, verified on Edge/Windows.
https://v10-sandbox-git-feat-spf-drm-mux.vercel.app/spf-drm/?source=hls-drm-playready-msft

<details>
<summary>Why</summary>

Its content-key-type (`ckt`) value was invalid from the day it entered the catalog — copied from Shaka's demo asset list, which still carries the same value — and Microsoft's server answered HTTP 500 with `Invalid config data in ckt`. Correcting `ckt` gets a 200 and decode still never starts, the silent shape of a wrong content key. The asset's own license URL names a host that no longer resolves. Shaka can't be used as a cross-check: its demo asks for `com.microsoft.playready.recommendation`, which Edge refuses outright. Kept only in case it can be revived.

</details>

**EZDRM plays fine, but not while AirPlaying.** The source itself works in Safari and is worth testing. Only starting an AirPlay session on it fails — and it fails in Shaka too. Mux FairPlay over AirPlay *does* work, verified on a Roku.
https://v10-sandbox-git-feat-spf-drm-mux.vercel.app/spf-drm/?source=hls-drm-ezdrm

<details>
<summary>Why</summary>

On engaging AirPlay the receiver either never asks for a key, or asks and EZDRM's demo endpoint answers HTTP 500 — which it also returns for any malformed request including an empty body, so the status carries no information. It reproduces in Shaka Player's own demo on Shaka's native-HLS path, with no MediaSource involved anywhere, which is what rules out this engine as the cause. Tracked as unresolved in `drm-support.md`.

</details>

Already verified by hand on Chrome, Firefox, Safari/macOS, iOS Safari on a physical iPhone, Edge on Windows, Android Chrome on a real Widevine L1 device, and FairPlay over AirPlay to a Roku — each including the failure path. Per-environment detail is in `drm-support.md`.

## Decisions worth pushing back on

Each of these had a real alternative. Three SPF terms the reasoning leans on, since they are not obvious from outside:

- **Behavior** — the unit an engine is assembled from. `createHlsVideoEngine` lists them; each reacts to shared state and owns part of it.
- **Resolved** — a quality level whose own media playlist has been fetched and parsed. Only the selected one is resolved; the rest are entries in the top-level manifest until they are chosen.
- **Cause vs. verdict** — when something is unplayable, each individual reason is reported as a *cause*, and a single *verdict* is what surfaces to the app once nothing playable is left. Below, "4008 → 2011" means "a reason per encrypted quality level, then one unplayable verdict."

**DRM is built into the default HLS engine; there is no DRM-free build of it.**
Everyone importing `@videojs/spf/hls` carries the DRM code whether they use it or not. That is the objection and it is fair. The defence is that the cost is recoverable rather than fixed: an app that assembles its own engine without the DRM behaviors sheds ~90% of it, and naming fewer key systems sheds more (asking for Widevine alone drops PlayReady's XML challenge handling and `DOMParser` from the bundle). A type test fails if a behavior that is *always* composed starts touching DRM state — that is what keeps the escape hatch from silently closing. The AirPlay handoff can be dropped independently again.

**Encrypted segments are held back at the segment loaders, rather than by delaying MediaSource attachment.**
The rule is that no encrypted data reaches the decoder before its keys are attached. The tempting place to enforce it is to not attach the MediaSource until keys are ready — and that is wrong: attachment is also what produces duration, buffering and metadata, and none of hls.js, Shaka, dash.js or rx-player gate it. So the hold sits one layer down, on the code that dispatches segment fetches.
The piece of state it reads is called `segmentLoadingBlocked` — named for *what* is blocked rather than *who* is blocking. AirPlay already pauses those same loaders for an unrelated reason, so the loaders themselves never mention DRM.

**`exchangeLicenses` is listed before `setupMediaKeys` in the engine, although it runs after it.**
This reads backwards — licensing is listed ahead of the negotiation that makes licensing possible. The list order is *teardown* order: cleanups run in the order behaviors were registered, and license sessions must be closed before the MediaKeys they belong to are detached. Listing it first costs nothing at startup, because the licensing behavior simply waits until its inputs exist.

**Every key the manifest declares is licensed up front, rather than only the keys playback needs.**
Content with a different key per quality tier fires a license request per key before the first frame, including tiers that will never play. Providers cap concurrent sessions and some bill per license issued, so this is a real cost. It is deferred rather than defended: #2863 changes it, and a test pins today's behavior so that change shows up rather than sliding in. In practice it currently fans out *less* than that description implies — and the reason why is a bug, in Known gaps below.

**When a device has more than one CDM, the order of preference is FairPlay, then Widevine, then PlayReady.**
This decides Edge on Windows, which ships both Widevine and PlayReady: it gets Widevine. hls.js hardcodes the same order and Mux Player ships it unchanged. Where PlayReady is required instead — usually a hardware-DRM policy — `keySystems` overrides it.

**"No CDM could be negotiated" is recorded as the literal string `NO_KEY_SYSTEM`.**
Three outcomes have to be told apart: not negotiated yet, negotiated successfully, and tried-and-refused. The alternatives were a nullable type or a second piece of state. A reserved string keeps the type `string | undefined` and keeps the answer in one place; real key-system ids are reverse-DNS, so nothing can collide with it. The refused case earns its own value because that is the moment encrypted quality levels get dropped and the 4008 → 2011 sequence runs.

**A second, Safari-only DRM implementation ships alongside the standard one.**
`fairplay-legacy.ts` uses `WebKitMediaKeys`, the pre-standard API that EME replaced. It exists because Safari refuses the standard `generateRequest` call while the video is playing to an AirPlay receiver. That reproduces in Apple's own sample player and in Shaka Player, so it is WebKit's bug rather than this engine's — and Shaka, which carries no fallback, fails on content this plays. Nothing checks an OS version: the path is entered only on that one specific refusal, so it falls out of use by itself when WebKit fixes it, and can then be deleted outright.

**`/hls` is 28.05 kB brotli against a 20 kB target.**
Over budget, accepted for this iteration on the grounds that the cost is removable (first bullet) rather than inherent.

## Known gaps

- **Content with a different key per quality tier only licenses the tier it started on.** Only the playing quality level has its playlist fetched, so only that level's key is known when licensing runs — and licensing doesn't run again. Switching quality across a key boundary then has no license for the new key. Found by Cursor Bugbot on this PR and confirmed; tracked in #2863, where Axinom's test vector (keys split at 480→720) is a reproduction.
- Key expiry and output restriction (HDCP) are **reported but not acted on** — dropping or re-requesting affected quality levels is #2861. Keys that change mid-stream on a live stream are unsupported for Widevine and PlayReady (#2862); the same thing on VOD, and on FairPlay, does work. No offline or persistent licenses. `#EXT-X-SESSION-KEY` is not parsed. AES-128 encrypted content is refused with a specific error rather than played — that is a separate feature, not a DRM failure.


packages/spf/src/playback/behaviors/dom/load-segments.ts
```diff
@@ -192,22 +208,38 @@ function setupSegmentLoading<
...
+          const track = selectedTrack.get();
 
           segmentBoundarySignal.get();
+          const loader = peek(context[loaderKey]);
+          if (!track || !loader) return;
+
           const currentTime = peek(state.currentTime) ?? 0;
 
-          peek(context[loaderKey])!.send({
+          loader.send({
```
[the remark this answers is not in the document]
cjpillsbury: Confirmed, and fixed in 92508117d — thank you, this was a real defect.

The mechanism turned out to be different from the description, in a way worth recording. Registration order *does* hold: `setupMediaKeys` is composed at `engine.ts:561`, ahead of the loaders at 601, and the gate write lands first. The dispatch happened anyway.

Composing both real behaviors over shared signals and logging every gate write against every dispatch:

| scenario | sequence |
|---|---|
| clear source (control) | `LOAD(gate=undefined)` |
| gate up before anything resolves | *(no dispatch)* |
| **engine order** | `gate=true` → **`LOAD(gate=true)`** |
| loaders registered first | `LOAD(gate=undefined)` → `gate=true` |

`derivedStateSignal` does read the gate, but the machine commits to a dispatching state while the gate is still down and reaches the state handler after it is up. So the load is dispatched *behind* a raised gate. Composition order only decides which side of the handler the gate write lands on.

The handler bodies already re-read `track` and `loader` against exactly this hazard class — there's a comment on `'metadata-only'` explaining that a track can un-resolve and arrive before the transition lands. The gates now get the same treatment.

Existing coverage set the gate as *initial* state, which is the steady case and not the one production produces. The new test composes the two behaviors in engine registration order and carries a clear-source liveness control so it can't pass by never dispatching; it fails on the parent commit with `expected [ true ] to deeply equal []`.

packages/spf/src/playback/behaviors/dom/exchange-licenses.ts
```diff
@@ -0,0 +1,195 @@
...
+
+          // One session per manifest-carried init data of the negotiated
+          // system, projected by its own module. Empty when the module's keys
+          // carry none (FairPlay `skd://`), which routes to the fallback:
+          // active only when the manifest path opened nothing, because on
+          // manifest-licensed sources appends re-fire `encrypted` for content
+          // already being licensed, and reacting would double-license.
+          const declared = manifestInitData(presentation, module_);
+
+          for (const { initDataType, initData } of declared) open(initDataType, initData);
```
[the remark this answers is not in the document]
cjpillsbury: Confirmed. Not fixed on this PR — tracked in #2863, with the record updated in f0f17c8a6.

Verified the premise rather than taking it as given, and all three links hold:

1. `resolve-track` **gates on a selection** — it resolves the selected rendition's playlist and patches it back into `state.presentation`; other variants stay unresolved.
2. `declaredDrmKeys` counts keys on **resolved tracks only** (`isResolvedTrack`).
3. `exchangeLicenses` **cannot re-enter** — `derivedStateSignal` returns `'licensing'` for any resolved presentation, so gaining a resolved variant is not a state transition and `entry` does not run again.

So a ladder with a distinct KEYID per tier declares one key at entry, and a switch across a key boundary has no session. `#EXT-X-SESSION-KEY` would normally mask this by declaring every key up front; it's deliberately unparsed here, so nothing does.

Worth noting why the existing coverage doesn't catch it: Axinom's MultiKey vector plays because the smoke never left the initially-selected tier, and the fan-out pin test is handed a presentation with all four variants already resolved — a state the engine doesn't produce at entry. Both still pin what they were written for.

Not fixed here because the remedy — a reactive re-scan of `declaredDrmKeys` — is the same one mid-stream key rotation needs, and splitting it would mean designing the when-to-license question twice. #2863 now carries the mechanism, the reason the current tests miss it, and an acceptance criterion for the switch case.

packages/spf/src/media/drm.ts
```diff
@@ -0,0 +1,496 @@
...
+
+function resolveDrmValue<T>(value: DrmValue<T>): T | undefined {
+  if (typeof value !== 'function') return value;
+
+  try {
+    return (value as () => T | undefined)();
+  } catch {
+    return undefined;
+  }
+}
```
spuppo-mux, line 174: nitpick(non-blocking): this could be reused outside a drm context.

packages/spf/src/media/drm.ts
```diff
@@ -0,0 +1,496 @@
...
+      for (const track of switchingSet.tracks) {
+        if (!isResolvedTrack(track)) continue;
+
+        for (const key of getMediaPlaylistMetadata(track)?.keys ?? []) {
+          const identity = [key.method, key.uri, key.keyFormat, key.keyId, key.iv].join(' ');
+          if (seen.has(identity)) continue;
+
+          seen.add(identity);
+          keys.push(key);
+        }
```
spuppo-mux, line 341: thought(non-blocking): there already is some deduping logic in another spf module, may be worth it unifying it in the future to save some bytes

packages/spf/src/media/dom/license-sessions.ts
```diff
@@ -0,0 +1,257 @@
...
+
+/** A key id as lowercase hex, so the reported `data` names which key failed. */
+function keyIdHex(keyId: BufferSource): string {
+  const bytes =
+    keyId instanceof ArrayBuffer
+      ? new Uint8Array(keyId)
+      : new Uint8Array(keyId.buffer, keyId.byteOffset, keyId.byteLength);
+
+  return [...bytes].map((byte) => byte.toString(16).padStart(2, '0')).join('');
+}
```
spuppo-mux, line 48: nit: could use `bufferSourceBytes` from license-transforms

packages/spf/src/media/dom/eme.ts
```diff
@@ -0,0 +1,304 @@
...
+
+        const bucket = track.type === 'video' ? video : audio;
+
+        bucket.add(buildMimeCodec({ mimeType: track.mimeType, codecs: track.codecs }));
+      }
+    }
+  }
+
+  return { video: [...video], audio: [...audio] };
+}
```
spuppo-mux: thought(nb): (may need to review the logic)
May be good to add this helper
```
filterTracks(/* predicate */() => 
    (track.type === 'video' || track.type !== 'audio') &&
    (track.mimeType && track.codecs?.length > 0) continue;
}, /* groupBy */ (track) => {
    const bucket = track.type === 'video' ? video : audio;
    bucket.add(buildMimeCodec({ mimeType: track.mimeType, codecs: track.codecs }));
} )
```

packages/spf/src/network/retry.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export const DEFAULT_RETRY_POLICY: RetryPolicy = { maxRetries: 3, baseDelayMs: 1000, firstByteTimeoutMs: 8000 };
+
+/** 5xx is transient; a 4xx (bad token, malformed request) is the client's to fix, so it is fatal — no retry. */
+function isTransientStatus(status: number): boolean {
+  return status >= 500;
+}
+
+function abortError(): DOMException {
+  return new DOMException('Aborted', 'AbortError');
```
spuppo-mux, line 31: note(blocking): Just wanted to double-check this "DOM" reference is correct outside of a /dom package

packages/spf/src/playback/behaviors/dom/load-segments.ts
```diff
@@ -192,22 +208,50 @@ function setupSegmentLoading<
...
+          // very change that should be transitioning the machine out of this
+          // state, and would otherwise dispatch a load with no track.
+          const track = selectedTrack.get();
 
           segmentBoundarySignal.get();
+          const loader = peek(context[loaderKey]);
+          if (!track || !loader) return;
+
+          // Same concurrent-rise window as `'metadata-only'`'s entry; see there.
+          if (state.loadingSuspended?.get() || state.segmentLoadingBlocked?.get()) return;
```
spuppo-mux: nit(nb): this condition is repeated 3 times, may be worth refactoring into `isLoadingDormant(state)` or something like that _if_ we suspect it may change further.

packages/spf/src/media/dom/license-sessions.ts
cjpillsbury: Done.

packages/spf/src/media/drm.ts
cjpillsbury: The dedupe use cases work slightly differently unless I'm missing something. Gonna hold off + close this as won't do for now, but happy to revisit in a followup.

packages/spf/src/media/dom/eme.ts
cjpillsbury: Ok made a few changes in response to this. Check out updates to `/media/utils/tracks` for deets. Good callout. Had ripple effects that I would normally treat as scope creep just bc it adds more noise to the PR, but since you were down, I'm also down. 👯

packages/spf/src/playback/behaviors/dom/load-segments.ts
cjpillsbury: Done. (Did a closure-captured version for the time being).

cjpillsbury:
@spuppo-mux (for visibility) created https://github.com/videojs/v10/issues/2944 in case we want to do a research (and ideally enhancement) spike later.

spuppo-mux (approved):
LGTM! Tested on listed browsers and devices; only found the one noted AirPlay issue (already filed as follow up)

PR #2305, opened 2026-08-21
feat(packages): add wistia video media

### LINKED ISSUE #1459: Feature: Wistia Provider
Add a Wistia provider so Video.js 10 can play Wistia-hosted videos via the Wistia embed iframe / Player API.

1. **Embed options passthrough** — Expose enough of Wistia's embed options (`playerColor`, `controlsVisibleOnLoad`, `playbar`, `playButton`, `smallPlayButton`, `volumeControl`, `fullscreenButton`, `qualityControl`, `copyLinkAndThumbnailEnabled`, `endVideoBehavior`, `playsinline`, `muted`, `autoPlay`) that embedders can theme and configure the player without forking the adapter.
2. **Aurora vs. v1 script** — Wistia ships two embed stacks: the legacy `fast.wistia.com/assets/external/E-v1.js` + `_wq` bootstrap, and the newer Aurora `<wistia-player>` web component. Pick one as the integration target (or support both).
3. **ID vs. URL source shape** — Decide whether `src` accepts a Wistia media ID, a `https://<subdomain>.wistia.com/medias/<id>` URL, a `fast.wistia.com/embed/iframe/<id>` URL, or all of the above. Prior art diverges here.

## Design notes

- **Aurora is a moving target.** Wistia's current direction is the `<wistia-player>` custom element with attribute/property/`window.wistiaOptions` configuration — different from the legacy `_wq.push({ options })` shape. Targeting Aurora aligns with Wistia's roadmap but is less documented in prior art.
- **Subdomain-per-account.** Wistia URLs embed the account subdomain (`<account>.wistia.com/medias/<id>`). The adapter does not need the subdomain to load the iframe (Wistia resolves by media ID), but parsers should not require it.
- **Analytics is load-bearing for Wistia customers.** Wistia embeds fire view/heatmap tracking by default; any adapter that swaps in a different UI (Video.js skin) should be clear about whether Wistia-side analytics still runs. Plyr-style "skin over the iframe" preserves it; a provider that replaces the chrome does not.

## Prior art

**Mux `<wistia-video>` (media-elements)** — `<video>`-compatible custom element. Loads `fast.wistia.com/assets/external/E-v1.js`, bootstraps through `_wq.push({ id, onReady, options })`. Maps native attrs to Wistia options: `autoPlay`, `preload` (default `'metadata'`), `playsinline`, `endVideoBehavior: 'loop'` when `loop` is set, `chromeless: !controls`, `playButton: controls`, `muted`. Parses IDs from `wistia.com/(medias|embed)/<id>` or `wi.st/...`.
- Source: [`packages/wistia-video-element/wistia-video-element.js`](https://github.com/muxinc/media-elements/blob/main/packages/wistia-video-element/wistia-video-element.js)
- Docs: [media-chrome.org `<wistia-video>`](https://www.media-chrome.org/docs/en/media-elements/wistia-video)

**Video.js v8 (`ryanpatrickcook/videojs-wistia`)** — Community `Tech` implementation; broader embed-option passthrough via URL query params. Parses `playerColor`, `playbar`, `playButton`, `smallPlayButton`, `volumeControl`, `fullscreenButton`, `controlsVisibleOnLoad`, `chromeless`, `autoplay`, `muted`, `volume`, `endVideoBehavior` from the `src` URL and forwards them to `_wq`. Last push 2020; not actively maintained but illustrates the v8-era parameter surface.
- Source: [`vjs.wistia.js`](https://github.com/ryanpatrickcook/videojs-wistia/blob/master/vjs.wistia.js)
- Repo: [ryanpatrickcook/videojs-wistia](https://github.com/ryanpatrickcook/videojs-wistia)

**Upstream — Wistia** — Aurora Player is a web component (`<wistia-player>`); options can be set as HTML attributes, JS properties, or a global `window.wistiaOptions` object. Legacy iframe/JS embeds use the `_wq` queue with an `options` object.
- Aurora Player API overview: [docs.wistia.com/docs/player-embed-api](https://docs.wistia.com/docs/player-embed-api)
- JavaScript Player API (legacy handle pattern): [docs.wistia.com/docs/javascript-player-api](https://docs.wistia.com/docs/javascript-player-api)
- Embed options catalog: [docs.wistia.com/docs/embed-options-and-plugins](https://docs.wistia.com/docs/embed-options-and-plugins)

luwes:
Closes #1459

## Summary

Adds `<wistia-video>` and `<WistiaVideo>` over Wistia's Aurora player. Wistia ships its own web component and React wrapper, so rather than wrap them in a media host, each platform uses Wistia's own component and normalizes it in place — the player a consumer holds and the media the store drives are the same object.

## Changes

- `<wistia-video>` extends Wistia's `WistiaPlayer` and adds the attribute surface a media element is written with (`src`, `controls`, `loop`, `muted`, `playsinline`, `poster`, `preload`, `autoplay`).
- `<WistiaVideo>` wraps `@wistia/wistia-player-react`, translating the same props and normalizing the element it renders.
- `source` carries Wistia's own options — `mediaId`, `playerColor`, `qualityMin`, and anything else the player understands — straight through, so the components don't have to know about each one. `src` accepts a Wistia URL or hashed id and resolves it to a `mediaId`.
- A shared normalizer in `@videojs/media/dom/wistia` gives the element the media contract: the members `HTMLMediaElement` has that Wistia names differently or not at all, and the events it spells in kebab case.
- Sandbox preset for both platforms, and a CDN bundle.

<details>

**Why normalize in place rather than host.** Wistia's player is close enough to a media element that a host would mostly be forwarding. Patching the element means one object, and both platforms share the mapping so they cannot drift.

**The awkward corners of the mapping**, each of which cost a round of debugging:

- `controls` is left alone. Wistia already has a `controls` and it means the player's control instances, which its internals read. Chrome is eight separate switches, applied as a group.
- `paused` is `state !== 'playing'`. Wistia answers `state === 'paused'`, which is `false` in the `beforeplay` state a player opens in — so a play/pause toggle would pause a player that had never started.
- `seeking` doesn't exist as a property, and Wistia doesn't reliably follow a `seeking` event with a `seeked`. It's tracked from the events, and the first playhead report after a seek closes it and dispatches `seeked` — otherwise the flag latches on and the store stops syncing the playhead for good.
- `durationchange` comes from three events. `duration` reads `0` until `api-ready`, media data lands on `loaded-media-data`, and `loaded-metadata` is documented as possibly delayed until the viewer interacts — so a media nobody clicks would never report a duration from that alone. `loadedmetadata` is held to once per source.
- `timeupdate` comes from both `time-update` and `second-change`, so the clock runs at whatever resolution turns up.
- `seekable` returns an empty range of its own rather than the shared `EMPTY_TIME_RANGES`, which the store reads as "no buffer surface at all" and skips the buffer feature for good.

These last few are all the same failure mode: the store gates each feature on a predicate that asks only whether members are defined, once, at attach. A missing one skips the feature outright rather than degrading it, so the tests assert the predicates directly.

**Styling.** A chromeless player gets `pointer-events: none` so the skin receives the clicks, and reads the skin's `--media-video-border-radius` with `overflow: hidden` to crop itself — Wistia paints into child elements, so the radius alone won't clip them. Wistia's own `roundedPlayer` is squared so the player isn't rounded twice.

**Build.** `@wistia/*` and `preact` are external to `packages/html`'s declaration build: Wistia's published types import a file the package omits and pull in `preact`'s CommonJS declarations, neither of which bundle.

</details>

## Testing

- `pnpm -F @videojs/media test` — 39 tests covering the normalizer: source and id parsing, every remapped member, the event aliases, and the capability predicates the store gates on.
- `pnpm -F @videojs/html test` / `pnpm -F @videojs/react test` — attribute and prop translation on each platform.
- Manual: `pnpm dev`, pick the **Wistia Video** preset on HTML, React, and CDN. Check playback from a click on the display, that the clock and scrubber run, duration appears, mute toggles, and the corners are cropped by the skin.

Not verified against a real browser by me — the normalizer is exercised against a stub, so the event names come from Wistia's docs and typings rather than observed traffic.

---

> [!NOTE]
> **Medium Risk**
> New third-party player integration with SSR global shims and store-contract mapping; playback bugs are possible, but it does not touch auth or data handling.
> 
> **Overview**
> Adds **Wistia video** as an embed media: HTML `<wistia-video>` and React `<WistiaVideo>` use Wistia’s own `<wistia-player>` instead of wrapping it, then normalize that element so the store and consumer share one object.
> 
> `@videojs/media/dom/wistia` maps Wistia’s API onto the media contract (`src`/`source`, events, `paused`, seeking, duration, chrome via the eight control switches). `src` accepts a hashed id or Wistia URL (including `wtime`); extra Wistia options pass through `source`. A server shim lets the package evaluate without browser globals.
> 
> Sandbox gets a **Wistia Video** preset (HTML, React, CDN). Chromeless players disable pointer events and clip to the skin radius so Video.js chrome stays in control.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d872399383475c9a7615e6f313d1686d2c016a92. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/media/src/dom/wistia/media.ts
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * The tag Wistia registers its player element under, and it has to be declared here: this is the module whose
+ * evaluation registers the element and the only thing React takes from it, so naming it somewhere that pulls
+ * none of Wistia in would leave a bundler free to drop this module and React rendering an undefined tag.
+ */
+export const WISTIA_PLAYER_TAG = 'wistia-player';
+
+/** Wistia's own player element: the type of the node React renders and hands back. */
+export type { WistiaPlayer };
```
mihar-22, line 17: I think we might want to namespace this to prevent being confused with our own "Player" concepts? Maybe something like...

```ts
export namespace WistiaLib {
  export type Player = WistiaPlayer;
}
```

That's just one idea. Open to however we can reduce that mix up.

packages/media/src/dom/wistia/media.ts
```diff
@@ -0,0 +1,113 @@
+// Keep this import first: it lends `@wistia/wistia-player` the browser globals a server lacks and the package
+// reads while it evaluates. See `server-shim.ts`.
+import './server-shim';
```
mihar-22: Feel so so about this but i get it. Might just need some testing in a SSR'd app. I think in the near future we can have a deeper look at how we ship media in the server and think about how we bundle/expose it.

packages/react/package.json
```diff
@@ -152,6 +152,7 @@
     "@types/react": "^19.2.17",
     "@videojs/icons": "workspace:*",
     "@videojs/skins": "workspace:*",
+    "@wistia/wistia-player": "^0.7.12",
```
mihar-22, line 161: This supposed to be here? Other media libraries don't seem to be

mihar-22 (approved):

packages/media/src/dom/wistia/media.ts
luwes: updated it to be like dash media

packages/react/package.json
luwes: this is needed for the `@wistia/wistia-player` stub in the test.

PR #2319, opened 2026-08-21
fix(html): restore popup APIs after reconnect

mihar-22:
Refs #2366

Depends on #2318 (stacked on `fix/html-player-rebinding`, which supplies the idempotent `SnapshotController.track()` this change builds on).

## Summary

Re-implementation of the original branch on top of the current `main` (the previous version referenced menu internals and an `AlertDialogElement` implementation that no longer exist). Popup, tooltip, menu, and dialog roots now own their imperative APIs strictly per DOM connection: each connection creates the API, both `disconnectedCallback` and `destroyCallback` tear it down through the same private dispose method, and a reconnect silently replays the open (or current error) state without emitting a duplicate `open-change`.

Bugs fixed, all confirmed on `main`:

- `PopoverElement` recreated `createPopover()` on every connect but only destroyed it in `destroyCallback`, so every reconnect leaked a live API and its listeners.
- `TooltipElement`, `MenuElement`, `DialogElementBase` (`DialogElement`, `AlertDialogElement`), and `ErrorDialogElement` destroyed per connection but had no `destroyCallback`, so `destroy()` while connected left the API and its trigger/document listeners live.
- None of them restored `open` (or the current error) after a reconnect: `willUpdate` only syncs when `changed.has('open')`, and reconnecting schedules no update.

## Changes

- `packages/store`: `SnapshotController.hostDestroyed()` releases the subscription permanently and ignores later `hostConnected()`/`track()` calls. `DestroyMixin.destroyCallback()` already invokes `hostDestroyed()` on tracked controllers, so elements no longer release snapshot controllers by hand.
- `packages/html/src/ui/restore-guard.ts`: small shared `RestoreGuard` that suppresses change notifications while an element replays state it already holds into a recreated API. Popover, tooltip, and dialog roots check `active` inside `onOpenChange` and wrap the replay in `run()`. The menu commits through `MenuApi.syncOpen()`, which already bypasses the cancelable open-change request, and the error dialog reopens from `willUpdate`, so neither needs the guard.
- Each root gets one `#disposeConnection()` (trigger cleanup, lock release, abort controllers, API destroy) called from `disconnectedCallback` and `destroyCallback`; the dispose is idempotent because a destroyed element may still be removed from the DOM afterwards.
- `connectedCallback` replays the held open state after `hasUpdated` and calls `requestUpdate()` so attributes, the Popover API show/hide, trigger ARIA, and positioning are refreshed for the new connection.

## Testing

- `pnpm exec vp run @videojs/store#build` and `pnpm exec vp run @videojs/html#build`, then `pnpm typecheck` from a clean `tsgo --build`: pass (0 errors)
- `pnpm -F @videojs/store test`: 134 passed
- `pnpm -F @videojs/html test`: 462 passed; `create-i18n.test.ts > updates media-text when html lang changes` fails intermittently (2 of 3 isolated runs) in a file this branch does not touch and whose code uses neither `SnapshotController` nor `track()`, so it is unrelated
- New coverage: rapid remove/reappend cycles, destroy-while-connected, and open-state restore with no duplicate `open-change` for `PopoverElement`, `TooltipElement`, `MenuElement`, `DialogElement`, and `ErrorDialogElement`; `hostDestroyed()` for `SnapshotController`. Swapping the element sources back to the base branch makes 8 of the 10 new element tests fail (the error-dialog restore already passes on base when a player provider is present because its context consumer requests an update on reconnect).
- `pnpm lint:fix:file` on touched files: no errors; remaining anti-slop notes are pre-existing casts in the test files and one justified `as unknown as` in the new error-dialog test provider

### DIFF
--- a/packages/html/src/player/player-controller.ts
+++ b/packages/html/src/player/player-controller.ts
@@ -1,8 +1,8 @@
 import type { PlayerStore } from '@videojs/core/dom';
-import type { ReactiveController, ReactiveControllerHost } from '@videojs/element';
+import type { ReactiveControllerHost } from '@videojs/element';
 import { ContextConsumer } from '@videojs/element/context';
 import type { InferStoreState, Selector } from '@videojs/store';
-import { StoreController } from '@videojs/store/html';
+import { SnapshotController } from '@videojs/store/html';
 
 import type { PlayerContext } from './context';
 
@@ -31,12 +31,12 @@ export type PlayerControllerHost = ReactiveControllerHost & HTMLElement;
  *   }
  *   ```;
  */
-export class PlayerController<Store extends PlayerStore, Result = Store> implements ReactiveController {
+export class PlayerController<Store extends PlayerStore, Result = Store> {
   readonly #host: PlayerControllerHost;
   readonly #selector: Selector<InferStoreState<Store>, Result> | undefined;
 
-  #consumer: ContextConsumer<PlayerContext<Store>, PlayerControllerHost>;
-  #store: StoreController<Store, Result> | null = null;
+  readonly #consumer: ContextConsumer<PlayerContext<Store>, PlayerControllerHost>;
+  #snapshot: SnapshotController<object, Result> | null = null;
 
   /**
    * @param host - The host element that owns this controller.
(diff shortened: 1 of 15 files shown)


mihar-22:
Superseded by #2591. GitHub refuses to change this PR's base because it is part of a stack under #2317, and the branch is now stacked on #2318 (`fix/html-player-rebinding`). Same branch and commit; only the base differs.

Issue #1679, opened 2026-06-11
Architecture: Revisit Core State Projection in React Hooks for React Compiler Compatibility


decepulis:
> [!NOTE]
> This issue is AI-generated

Our React hooks mix two models that React Compiler can't reconcile. They subscribe to store state declaratively (`usePlayer` / `useSyncExternalStore`), but then project UI state *imperatively during render* — keeping a long-lived `*Core` instance in `useState`, mutating it with `core.setMedia(media)`, and reading derived state back out with `core.getState()`. When the React Compiler auto-memoizes such a hook, it can reuse a stale snapshot and skip the re-projection on later renders, so the hook keeps returning old derived state even though the store and DOM are correct.

This bites hardest when store data changes *after* the first render — e.g. captions, where `textTrackList` is populated asynchronously after the `textTrack` feature attaches. The hook's first projection sees no tracks (`availability: 'unavailable'`), the store later updates, but the memoized hook output stays stale, so the Captions row never appears in the menu.

The pattern is not isolated to one hook — nearly every `*Core`-backed hook and component in `@videojs/react` mutates and reads its core instance during render (`useCaptionsOptions`, `usePlaybackRateOptions`, and the `*Root` components for slider, time, thumbnail, controls, dialogs, indicators, etc.). Today it only surfaces in environments that run the React Compiler (the Astro docs site enables `babel-plugin-react-compiler`); plain React and our Vitest suites don't exercise it. As more consumers adopt the React Compiler, this is likely to crop up again. We should revisit how React hooks bridge core state into the render flow so the result stays correct under auto-memoization, without per-hook escape hatches.

## Context

- Discovered on the docs site, where React Menu demos intermittently dropped the Captions option. Sam tracked it to render-time projection conflicting with the React Compiler and shipped a targeted workaround in #1677: a `'use no memo'` directive on `useCaptionsOptions` that opts that one hook out of auto-memoization so each render re-runs `setMedia` / `getState`. That fix is deliberately narrow — it unblocks the docs site but doesn't address the underlying pattern shared by the other hooks.
- Follow-up requested in the [PR review](https://github.com/videojs/v10/pull/1677#pullrequestreview-4478708989): "open a follow-up issue so we can clean this up for good! I'm sure more and more of our consumers will be using React Compiler."

sampotts:
Thanks for opening this! 👍🏼

PR #2322, opened 2026-08-21
fix(react)!: acquire media instances after commit

mihar-22:
Refs #1679

Depends on #2316.

## Summary

Acquire engine-bearing media instances only after React commits, then initialize and publish them before paint. This prevents SSR and abandoned mounts from constructing playback engines and makes readiness explicit through a nullable useMediaInstance return.

## Changes

- Change useMediaInstance to return Instance | null until its layout-effect acquisition commits
- Initialize before player publication and own detachment/destruction on class swap or unmount
- Coordinate DOM attachment teardown so ownership clears, the target detaches once, and only then the instance is destroyed
- Keep destroyed acquisitions unavailable across Suspense hide/reveal until a fresh committed instance exists
- Attach acquired media to an existing host through one stable ref without cycling forwarded refs
- Allow prop synchronization to wait for a nullable target
- Guard the few adapters that directly consume the media instance
- Cover SSR, abandoned render, StrictMode recreation, setup failure, pre-passive readiness, class replacement, integrated teardown order, and Suspense reconnection

## Breaking change

External useMediaInstance callers must handle null until commit readiness. The built-in adapters already do so.

## Testing

- pnpm -F @videojs/react test: 525 tests
- pnpm build --filter=@videojs/react
- pnpm typecheck
- pnpm -F site api-docs
- Biome and git diff --check

### DIFF
--- a/packages/react/src/media/hls-background-video/media.tsx
+++ b/packages/react/src/media/hls-background-video/media.tsx
@@ -58,6 +58,7 @@ export const HlsBackgroundVideo = forwardRef<HTMLVideoElement, HlsBackgroundVide
   // typed for, and anything else listening on the node — the consumer's own
   // `addEventListener`, a testing-library assertion — sees the same failure.
   useEffect(() => {
+    if (!media) return;
     const forward = () => videoRef.current?.dispatchEvent(new Event('error'));
     media.addEventListener('error', forward);
     return () => media.removeEventListener('error', forward);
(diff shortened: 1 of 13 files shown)


mihar-22:
Closing after re-reviewing against current main.

The stated motivation was that render-phase construction of media instances builds playback engines during SSR and abandoned mounts. That no longer holds: `HlsJsMedia` and friends construct lazily, building the engine only on the `source`/`load` path, and #2316 moves that write into a layout effect. What remains render-phase is a garbage-collectable `EventTarget` allocation, which React permits.

Against that, this PR made `useMediaInstance` nullable (breaking, with reference docs not updated), added a synchronous second render to every media component, and introduced two new abstractions to replace a ten-line ref callback.

If a concrete leak survives #2316, e.g. a `Media` subclass whose constructor allocates external resources, the fix belongs in that constructor rather than in the hook contract. Refs #1679.

PR #2327, opened 2026-08-21
refactor(react)!: separate slider render and motion state

mihar-22:
Refs #1679

Depends on #2325.

## Summary

- Split React slider data into semantic render state and an explicit high-frequency motion channel.
- Add `useSliderMotion()` for pointer-driven custom content while keeping ordinary render, className, and style callbacks stable.
- Keep Thumb ARIA synchronized with live drag motion.
- Give React ownership of fill and buffer CSS variables while the imperative binder owns only `--media-slider-pointer`.
- Make slider projection cores render-local and keep retained drag/event machinery commit-synchronized.
- Observe geometry only for edge-aligned thumbs.
- Document the new hook and migration path. Core and HTML slider state remain unchanged.

## Breaking changes

- `useSlider()` returns `motion` instead of `input`.
- `computeState` receives discrete `SliderInteractionState` rather than the full input object.
- React slider callback state is now `SliderRenderState` and omits high-frequency pointer motion.
- Pointer-driven custom UI should call `useSliderMotion()` inside a slider root.

## Verification

- React: 67 files, 547 tests passed.
- Focused slider lifecycle/motion/variants: 5 files, 72 tests passed.
- React build and workspace typecheck passed.
- API generation and API builder: 225 tests passed.
- Astro check: 0 errors.
- Independent stacked-branch review found no actionable issues.

### DIFF
--- a/packages/core/src/core/ui/slider/slider-segments-core.ts
+++ b/packages/core/src/core/ui/slider/slider-segments-core.ts
@@ -85,7 +85,11 @@ export class SliderSegmentsCore {
     });
   }
 
-  getState(segment: SliderSegmentGeometry, slider: SliderState, pointerValue: number): SliderSegmentState {
+  getState<Slider extends Pick<SliderState, 'value' | 'pointing' | 'dragging' | 'interactive'>>(
+    segment: SliderSegmentGeometry,
+    slider: Slider,
+    pointerValue: number
+  ): SliderSegmentState {
     const { last, ...geometry } = segment;
     const contains = (value: number): boolean =>
       value >= segment.start && (value < segment.end || (last && value === segment.end));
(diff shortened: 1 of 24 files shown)


mihar-22:
Closing in its current form after re-reviewing against current main. The semantic/motion split is still the right direction and `useSliderMotion` is a reasonable public hook, but this branch cannot be rebased safely:

- A mechanical rebase silently reverts the press-locking fix from #2505 (`36bb83600` moved the controls lock from drag start/end to press start/end; this PR rewrote that block on older code).
- 26 of its 36 conflicts are inherited from the #2316 → #2320 → #2325 chain.
- `alignState`/`getAlignmentSizes` reimplement `adjustForAlignment` from core in React while HTML keeps the core path, which duplicates geometry math across packages.
- Renaming `input` → `motion` and changing `computeState` break the documented `useSlider` surface with no RFC.

Plan: once the #2325 utility layer lands, reopen as two PRs on main. First a non-breaking PR for React-owned fill/buffer CSS variables and edge-only geometry observation, then `useSliderMotion` with a deprecated `input` alias and a short RFC for the `computeState` change, with edge-alignment measurement moved into `SliderApi` so core stays the single owner. Refs #1679.

PR #2357, opened 2026-08-24
fix(packages): improve menu behavior and performance

### LINKED ISSUE #2302: Bug: menu max-height/max-width clamp reads popover CSS vars a menu never sets

`@videojs/core`, `@videojs/html` (default video **and** audio skins)

Confirmed in Google Chrome `151` on macOS. Not browser-specific — it is a CSS custom-property name mismatch.

### Description

`.media-menu`'s `max-height` / `max-width` clamp is inert: it reads custom properties that a menu never sets.

`MenuElement` positions itself with `MenuPositioningCSSVars`, which spreads `PopoverCSSVars` but deliberately overrides the two available-space names to the menu namespace:

```js
// packages/core/src/dom/ui/menu/create-menu.ts
const MenuPositioningCSSVars = {
  ...PopoverCSSVars,
  availableWidth: MenuCSSVars.availableWidth,   // --media-menu-available-width
  availableHeight: MenuCSSVars.availableHeight, // --media-menu-available-height
};
```

The skin still reads the popover names:

```css
/* default video skin, and identically in the audio skin */
.media-default-skin .media-menu {
  max-width: var(--media-popover-available-width, none);
  max-height: min(var(--media-popover-available-height, var(--menu-max-height)), var(--menu-max-height));
}
```

`--media-popover-available-*` is never written on a `<media-menu>`, so `var()` always takes the fallback and the effective clamp is a flat `--menu-max-height` regardless of how much room the menu actually has.

Read off a live menu in a 260px-tall player:

```
--media-menu-available-height:     192px     <- set by MenuElement, correct
--media-popover-available-height:  <unset>   <- what the skin reads
computed max-height:               320px     <- fallback; clamp never engaged
```

This is a regression. At `10.0.0-beta.26` `MenuElement` used `PopoverCSSVars` directly (it imports `PopoverCSSVars` and reads `PopoverCSSVars.availableWidth`), so the names matched and the clamp worked. The split appears to come from "core: centralize popup positioning" (#1904); the skin CSS was not moved with it.

It is mostly invisible at the default `--menu-max-height` of `calc(var(--spacing) * 56)` (224px), which fits most players — so the menu looks bounded when it is only ever being bounded by the constant. It becomes visible as soon as a consumer raises the cap, or presumably on any player shorter than 224px.

### Reduced test case

No hosted page, but it reproduces on the default skin with no custom code:

```js
const menu = document.querySelector('.media-menu--settings');
menu.openMenu('imperative-action');
const cs = getComputedStyle(menu);
cs.getPropertyValue('--media-menu-available-height');    // e.g. "192px"
cs.getPropertyValue('--media-popover-available-height'); // "" — what the skin reads
cs.maxHeight;                                            // 320px / the flat --menu-max-height
```

### Steps to reproduce

1. Render a default-skin video player in a short container (e.g. 460x260).
2. Open the settings menu and read the computed styles above.
3. Optionally set `--menu-max-height` on `.media-menu` to something larger than the container (e.g. `320px`) and open a menu whose content is that tall — the menu grows past the player instead of clamping to the available space.

### Current Behavior

`max-height` resolves to `--menu-max-height` in all cases; `max-width` resolves to `none`. The menu can render taller than the space between its trigger and the positioning boundary, overflowing the player.

### Expected Behavior

The menu clamps to the space the positioner measured, i.e. `--media-menu-available-height` / `--media-menu-available-width`.

Either point the skin at the menu names:

```css
max-width: var(--media-menu-available-width, none);
max-height: min(var(--media-menu-available-height, var(--menu-max-height)), var(--menu-max-height));
```

or have `MenuPositioningCSSVars` publish both namespaces so existing skin CSS keeps working.

### Errors

None — no console output; purely a layout result.

`10.0.0-beta.31` (regressed somewhere after `10.0.0-beta.26`)

sampotts:
Closes #2302

## Summary

Improve settings-menu performance, sizing, and submenu highlight behavior across the HTML and React implementations and the Default and Minimal skins. This keeps animated menu navigation smooth in Safari, restores available-space constraints, and prevents stale or missing highlights as users move between menu levels.

## Changes

- Avoid remeasuring and repositioning root menus on every animated menu-host resize while preserving trigger, boundary, window resize, and scroll tracking
- Use the menu-owned available width and height variables in the CSS and Tailwind skins
- Keep menu item labels on one line and reduce the settings-menu minimum width from 48 to 44 spacing units
- Clear the parent item highlight when a submenu opens and restore the highlight when focus returns
- Cancel pending initial highlights when menu state explicitly clears them
- Add focused core, HTML, React, and E2E regression coverage

<details>
<summary>Benchmarking and Safari investigation</summary>

The benchmark added 5,000 unrelated DOM nodes and repeatedly navigated between settings-menu panels while sampling `requestAnimationFrame` intervals.

### Local Safari

| Configuration | FPS | p95 frame |
| --- | ---: | ---: |
| Existing behavior | ~13.0 | ~131 ms |
| Menu host removed from resize tracking | ~60.1 | ~18 ms |
| Host tracking active, panel/group tracking disabled | ~13.1 | ~129 ms |

The animated menu host produced 119 resize callbacks during the sample. Removing only that host from popup-positioner tracking restored the display's 60 FPS ceiling while the panel/group menu-size observers continued firing. This isolates the expensive path to repeated popup geometry reads and positioning writes rather than the CSS transition alone.

### Sauce Labs Safari 26

The same harness ran in Safari 26.4 on macOS 26/Apple Silicon through Sauce Connect:

| Configuration | FPS | p95 frame |
| --- | ---: | ---: |
| Width and height transitions | 7.8 | 269 ms |
| No size transition | 29.1 | 249 ms |

The control itself was capped around 29 FPS and Sauce video capture was enabled, so those are not reliable absolute frame-rate measurements and were not used to override the local causal A/B result. The run did verify the Safari 26 target and showed that its remote environment also penalized animated menu geometry.

</details>

## Testing

- `pnpm -F @videojs/core test src/dom/ui/popover/tests/popup-positioner.test.ts`
- `pnpm -F @videojs/core test src/dom/ui/menu/tests/create-menu.test.ts` (79 passed)
- `pnpm -F @videojs/html test src/ui/menu/tests/menu-element.test.ts` (23 passed)
- `pnpm -F @videojs/react test src/ui/menu/tests/menu.test.tsx` (44 passed)
- `pnpm -F @videojs/skins build`
- Core, React, and HTML package builds
- `pnpm --dir apps/e2e exec playwright test tests/menu-scroll.spec.ts --project=vite-chromium`
- `pnpm --dir apps/e2e exec playwright test tests/menu-scroll.spec.ts --project=vite-webkit`
- Biome and `git diff --check`

`pnpm typecheck` still reports the existing SPF `MediaPlaylistMetadata` errors in `capabilities.ts`, `report-track-conditions.ts`, and the HLS video adapter.

---

> [!NOTE]
> **Medium Risk**
> Touches menu positioning, resize observation, and nested highlight/focus sync—easy to regress Safari animation or keyboard navigation, but the change is localized and well tested.
> 
> **Overview**
> Stops root menus from being remeasured on every animated host resize, which was tanking Safari FPS during settings-panel transitions. `PopupPositioner` now accepts `trackResize`; HTML and React menus pass `false` so trigger, boundary, window, and scroll tracking remain.
> 
> Skins clamp with `--media-menu-available-*` instead of popover vars, keep item labels on one line, and drop settings min-width from 48 to 44 spacing units.
> 
> Submenu interaction is synchronized: opening a child clears the parent highlight, focus restores it, pending initial highlights are cancelled when cleared, and items highlight on focus. Core, HTML, React, and E2E tests cover the new behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 74830fb43497c04968434c6ba29c0c502474b4c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sampotts:
@mihar-22 I'll close this for now but we need to resolve the Safari menu issues caused by ResizeObserver before the next release.

PR #2403, opened 2026-08-25
fix(site): scope component props by framework

### LINKED ISSUE #2271: docs(site): stale docs from #2234
## Summary
API changes in #2234 (refactor(compiler): add declarative component registries) may have made the following documentation stale.

## Triggering PR
https://github.com/videojs/v10/pull/2234

## API Changes

**`Poster`**
- Gained a `src: string` prop (previously `"props": {}`), sourced from `PosterProps` in `packages/core/src/core/ui/poster/poster-core.ts:9`.
- The prop has **no `frameworks` field**, so `ApiPropsTable.astro:39` renders it for *every* framework — including HTML.
- The prop has **no description**, so the generated table renders an empty Description cell.
- Gained a component description: "Framework-neutral poster component props."

**`Menu.Root`**
- The `isSubmenu` prop was removed from `Root.props`. It still exists in `Root.state` (`isSubmenu: boolean`) and as the `data-submenu` attribute, so the styling/state surface is unchanged.
- No documentation, demo, or README references `isSubmenu` by name — a repo-wide grep only hits package source. No stale docs found for this change.

### High Confidence
| File | Line(s) | Issue | API Change |
|---|---|---|---|
| `site/src/content/docs/reference/poster.mdx` | 105 | `<ComponentReference component="Poster" />` now renders a `src` prop row in the **HTML** framework variant, but `<media-poster>` does not support a `src` attribute — `PosterElement` only observes `placeholdersrc` (`packages/html/src/ui/poster/poster-element.ts:13`). The generated reference now advertises an attribute that does nothing. | `Poster.src` added with no `frameworks` scoping |
| `site/src/content/docs/how-to/migrate-from-media-chrome.mdx` | 176–182 | Prose states "Where `media-poster-image` took its own `src`, `media-poster` wraps the image you bring" — directly contradicted by the new `src` row that the HTML reference page now shows. | `Poster.src` added with no `frameworks` scoping |

### Medium Confidence
| File | Line(s) | Issue | API Change |
|---|---|---|---|
| `site/src/content/docs/reference/poster.mdx` | 34–40, 68 | HTML anatomy puts `src` only on the child ``, and the Styling section says "You control the child `` — this means `srcset`, `sizes`, `loading="lazy"` … work naturally." Needs reconciling with the prop table now rendered on the same page. | `Poster.src` added with no `frameworks` scoping |
| `site/src/content/docs/reference/poster.mdx` | 105 | The generated `src` row has an empty Description cell because `PosterProps.src` carries no JSDoc. Per `AGENTS.md`, API-reference exports need richer JSDoc. | `Poster.src` added without a description |

### Low Confidence
| File | Line(s) | Issue | API Change |
|---|---|---|---|
| `site/src/content/docs/how-to/migrate-from-video-js-8.mdx` | 370 | `<media-poster></media-poster>` — correct today, but inconsistent with the reference page if `src` is intended as a real HTML attribute. | `Poster.src` added with no `frameworks` scoping |
| `site/src/components/docs/demos/poster/html/css/BasicUsage.html` | 9–14 | Demo places `src` on the child `` rather than on `<media-poster>`; same consistency question as above. | `Poster.src` added with no `frameworks` scoping |

## Recommended Actions
- Decide which side is correct, then fix the source of truth rather than the page:
  - If `src` is React-only, scope it in the extractor output (`frameworks: ["react"]`) so it stops rendering in the HTML variant of `/reference/poster`. Note that `PosterCore` never reads `src`, and React's `Poster` already gets `src` from `UIComponentProps<'img'>` — so `PosterProps.src` may be dead surface that should not be extracted at all.
  - If `<media-poster src>` is meant to be supported, implement it in `PosterElement` and update `poster.mdx` (anatomy + styling) and `migrate-from-media-chrome.mdx:176–182`.
- Add JSDoc to `PosterProps.src` in `packages/core/src/core/ui/poster/poster-core.ts:9` so the generated table has a description.
- Consider whether `placeholdersrc` / `--media-poster-placeholder` (HTML) and `data-loaded` (React) should also appear in the Poster reference — neither is currently extracted or documented. Pre-existing, not caused by this PR.
- No action needed for the `Menu.Root.isSubmenu` prop removal; `isSubmenu` remains available as state and via `data-submenu`, and no docs referenced the prop.

decepulis:
## Summary

- Filter extracted component props by their React and HTML availability.
- Keep adapter-internal props out of framework surfaces that do not expose them.

## Verification

- `CI=true pnpm -F site test scripts/api-docs-builder/src/tests/e2e.test.ts`
- `CI=true pnpm typecheck`

Closes #2271

### DIFF
--- a/packages/core/src/core/ui/poster/core.ts
+++ b/packages/core/src/core/ui/poster/core.ts
@@ -24,6 +24,7 @@ export interface PosterState {
 
 /** Framework-neutral poster component props. */
 export interface PosterProps {
+  /** Image URL for the React poster. HTML authors set the source on the child image instead. */
   src?: string | undefined;
 }
 
--- a/site/scripts/api-docs-builder/src/pipeline.ts
+++ b/site/scripts/api-docs-builder/src/pipeline.ts
@@ -412,6 +412,12 @@ function buildSingleComponentReference(source: ComponentSource, program: OxcProj
   };
 
   if (htmlData) {
+    for (const [name, prop] of Object.entries(result.props)) {
+      if (!htmlData.properties.includes(name)) prop.frameworks = ['react'];
+    }
+  }
+
+  if (htmlData) {
     result.platforms.html = { tagName: htmlData.tagName };
   }
 
@@ -450,6 +456,12 @@ function buildMultiPartReference(
         cssCustomProperties: cssVarsData ? buildCSSVars(cssVarsData) : {},
         platforms: { react: {} },
       };
+
+      if (htmlData) {
+        for (const [name, prop] of Object.entries(partRef.props)) {
+          if (!htmlData.properties.includes(name)) prop.frameworks = ['react'];
+        }
+      }
 
       if (!partRef.description) delete partRef.description;
 
--- a/site/scripts/api-docs-builder/src/tests/e2e.test.ts
+++ b/site/scripts/api-docs-builder/src/tests/e2e.test.ts
@@ -171,6 +171,7 @@ describe('Component pipeline (end-to-end)', () => {
       expect(ref.props.onPressedChange).toMatchObject({
         type: 'function',
         description: 'Callback when pressed state changes.',
+        frameworks: ['react'],
       });
       expect(ref.props.onPressedChange!.detailedType).toBeDefined();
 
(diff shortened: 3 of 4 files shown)


mihar-22 (commented):
Interesting problem. Don't think this is the fix but going to think about this more and how we're building our components.

mihar-22:
Closing so stack can continue but this is important. Keeping the issue tab open so I can figure this out. It looks like some compiler work created this problem.

PR #2404, opened 2026-08-25
feat(site): generate component event references

### LINKED ISSUE #1426: Docs: Document Emitted Events for Each Component
Migrated from this [media-chrome issue](https://github.com/muxinc/media-chrome/issues/998). Update the docs for each individual component to clarify which events the component emits

decepulis:
## Summary

- Extract typed custom-element event metadata for component references.
- Render component event tables alongside generated props, states, and CSS APIs.

## Verification

- `CI=true pnpm -F site test scripts/api-docs-builder/src/tests/e2e.test.ts`
- `CI=true pnpm typecheck`

Closes #1426

### DIFF
--- a/packages/html/src/ui/alert-dialog/alert-dialog-element.ts
+++ b/packages/html/src/ui/alert-dialog/alert-dialog-element.ts
@@ -2,7 +2,11 @@ import { AlertDialogCore, AlertDialogDataAttrs } from '@videojs/core';
 
 import { DialogElementBase } from '../dialog/dialog-element';
 
-/** A modal dialog with alert semantics. */
+/**
+ * A modal dialog with alert semantics.
+ *
+ * @fires open-change - Fired when the dialog's open state changes.
+ */
 export class AlertDialogElement extends DialogElementBase {
   static readonly tagName = 'media-alert-dialog';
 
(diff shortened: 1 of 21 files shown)


packages/html/src/ui/alert-dialog/alert-dialog-element.ts
```diff
@@ -15,6 +15,7 @@ import { alertDialogContext } from './context';
 
 let idCounter = 0;
 
+/** @fires open-change - Fired when the dialog's open state changes. */
```
decepulis: This is fine, but it's also configuration > convention. Do we have no way of detecting events emitted through some consistent typescript detection? If not, this is the next best thing.

decepulis: Confirmed: literal `emit('...')` and `dispatchEvent(new Event/CustomEvent('...'))` calls are already discovered conventionally across the element hierarchy. `@fires` only enriches descriptions. I added a builder E2E regression proving an untagged literal event is still emitted name-only while the tagged event keeps its description; no new event configuration is required.

mihar-22 (approved):
Would be nice to explore a events interface in core for this. Approve for now but we can clean it up later.

decepulis:
@mihar-22
> Would be nice to explore a events interface in core for this.

totally agreed!

PR #2420, opened 2026-08-25
docs(site): add native Video and Audio references

### LINKED ISSUE #1758: Docs: Video Media Provider
> [!NOTE]

As [discussed](https://mux.slack.com/archives/C096HT3S8F4/p1782279858207559), add docs for the `<video>` / `<Video>` media provider and child elements. 

[Edit:] and audio

decepulis:
## Summary

- Add hand-authored Video and Audio reference pages with React and HTML demos.
- Document native element composition, player registration, source and track children, provider precedence, and native APIs.

## Verification

- `CI=true pnpm -F site astro check`
- `CI=true pnpm -F @videojs/react test src/media/tests/video.test.tsx src/media/tests/audio.test.tsx`
- `CI=true pnpm -F @videojs/html test src/player/tests/create-player.test.ts`

Closes #1758

### DIFF
--- a/site/src/content/docs/reference/audio.mdx
+++ b/site/src/content/docs/reference/audio.mdx
@@ -71,7 +71,13 @@ Place a native `<audio>` anywhere inside `<audio-player>`. It becomes that playe
 
 Video.js does not replace the native audio API. Use `src` for one URL or add child [`<source>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/source) when the browser should choose between candidates. See <DocsLink slug="concepts/media-sources">Media sources</DocsLink> when you need an HLS, Mux, Spotify, or other playback adapter.
 
-React forwards native attributes, event handlers, and children to the `<audio>`, and its `ref` resolves to the `HTMLAudioElement`.
+<FrameworkCase frameworks={["react"]}>
+`Audio` forwards native attributes, event handlers, and children to the underlying `<audio>`. Its `ref` resolves to the `HTMLAudioElement`.
+</FrameworkCase>
+
+<FrameworkCase frameworks={["html"]}>
+`<audio>` is the native element itself. Set its attributes and properties, listen for native media events, and access it through an `HTMLAudioElement` reference.
+</FrameworkCase>
 
 See MDN for the complete [`<audio>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/audio), [`HTMLAudioElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAudioElement), and shared [`HTMLMediaElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement) APIs.
 
--- a/site/src/content/docs/reference/audio.mdx
+++ b/site/src/content/docs/reference/audio.mdx
@@ -24,10 +24,10 @@ import basicUsageHtmlTs from "@/components/docs/demos/audio/html/css/BasicUsage.
 ## Import
 
 <FrameworkCase frameworks={["react"]}>
-Import `Audio` and the matching feature bundle from the audio preset:
+Import `Audio` from the audio preset:
 
 ```tsx
-import { Audio, audioFeatures } from '@videojs/react/audio';
+import { Audio } from '@videojs/react/audio';
 ```
 </FrameworkCase>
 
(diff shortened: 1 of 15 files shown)


site/src/content/docs/reference/audio.mdx
```diff
@@ -0,0 +1,117 @@
...
+import basicUsageHtmlTs from "@/components/docs/demos/audio/html/css/BasicUsage.ts?raw";
+
+## Import
+
+<FrameworkCase frameworks={["react"]}>
+Import `Audio` and the matching feature bundle from the audio preset:
+
+```tsx
+import { Audio, audioFeatures } from '@videojs/react/audio';
+```
```
decepulis: imo every reference should have clear guidance on how to import something. add another PR to the stack that adds that near the top of the guide somehow. Open to the author for the best way to do that. Anatomy? A new section called import? Rely on the existing examples? What makes sense?

site/src/content/docs/reference/audio.mdx
```diff
@@ -0,0 +1,117 @@
...
+<audio>
+  <source />
+</audio>
+```
+</FrameworkCase>
+
+## Player registration
+
+<FrameworkCase frameworks={["react"]}>
+Place `Audio` inside the `Player` returned by `createPlayer({ features: audioFeatures })`. On mount, its composed ref registers the native `HTMLAudioElement` as that Player's media provider. Replacing or unmounting it detaches the previous element. Outside a Player, `Audio` still renders a standalone native `<audio>` without creating or attaching a Video.js store.
```
decepulis: I'm wary of this guidance. Many users won't be using `createPlayer`; they'll be using a precomposed provider.

site/src/content/docs/reference/audio.mdx
```diff
@@ -0,0 +1,117 @@
...
+</FrameworkCase>
+
+<FrameworkCase frameworks={["html"]}>
+Place a native `<audio>` anywhere under `<audio-player>`. The player observes its full descendant subtree and attaches the first native `<video>` or `<audio>` in document order. Adding, removing, or replacing native media causes the provider to update and detach the previous element when necessary.
+
+A registered custom media component takes precedence over a descendant native element. If the custom media unregisters, the player can fall back to the first native media element still in its subtree.
+</FrameworkCase>
```
decepulis: Much of this is implementation detail. What do users actually need in a reference? How to use it. Re-read the references you wrote through this lens.

site/src/content/docs/reference/audio.mdx
```diff
@@ -0,0 +1,117 @@
...
+<FrameworkCase frameworks={["react"]}>
+`AudioProps` extends React's `AudioHTMLAttributes<HTMLAudioElement>`. Native attributes, event handlers, and children pass through to the `<audio>`, and `ref` resolves to its `HTMLAudioElement`. Use that ref for native properties and methods.
+</FrameworkCase>
+
+<FrameworkCase frameworks={["html"]}>
+The element is an ordinary `HTMLAudioElement`; Video.js does not wrap or replace its native DOM API.
+</FrameworkCase>
+
+See MDN for the complete [`<audio>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/audio), [`HTMLAudioElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAudioElement), and shared [`HTMLMediaElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement) APIs.
+
```
decepulis: Consider leaning on MDN even more. We don't _need_ to maintain docs for video and audio; they already exist. We should be more interested in the differences between our video/audio and the native ones, if any exist

site/src/content/docs/reference/video.mdx
decepulis, line 1: apply comments from audio here

site/src/content/docs/reference/audio.mdx
decepulis: Added a separate top-of-stack draft, #2430, with a dedicated Import section across all generated component references. It uses one framework-aware helper for React named exports and HTML registration modules, plus an authoring rule and regression test.

site/src/content/docs/reference/audio.mdx
decepulis: Replaced the low-level `createPlayer` path with the precomposed `AudioPlayer` path; the page now presents `Audio` as the native provider slot inside the normal preset.

site/src/content/docs/reference/audio.mdx
decepulis: Re-read both pages through that lens and removed the observer, precedence, replacement, and detach implementation details. The reference now focuses on import, anatomy, normal preset usage, source children, and the one Video.js-specific attachment behavior users need to know.

site/src/content/docs/reference/audio.mdx
decepulis: The Native behavior section now delegates the full element API to MDN and only calls out the Video.js differences: automatic player attachment plus React prop/children/ref forwarding.

site/src/content/docs/reference/video.mdx
decepulis: Applied the same reader-focused rewrite to Video: precomposed `VideoPlayer`, no `createPlayer` plumbing, implementation details removed, and native behavior delegated to MDN.

mihar-22 (approved):

PR #2428, opened 2026-08-25
fix(packages): make html entries server importable

mihar-22:
## Summary

Make HTML custom-element entry points safe to evaluate in server runtimes that do not provide DOM globals.

## Changes

- Use inert HTMLElement fallbacks for reactive, background-video, and icon element class definitions
- Run SSR import coverage in a real Node environment
- Cover package root, player/skin, HLS, background media, element, and icon imports

## Testing

- vp run @videojs/html#build
- pnpm -F @videojs/element test
- pnpm -F @videojs/icons test
- pnpm -F @videojs/html test
- pnpm typecheck
- Built HTML entry import sweep in bare Node; DASH and Shaka are intentionally handled separately

### DIFF
--- a/packages/html/src/media/hls-background-video/media.ts
+++ b/packages/html/src/media/hls-background-video/media.ts
@@ -6,10 +6,12 @@ import type { Constructor } from '@videojs/utils/types';
 import { MediaAttachMixin } from '../../store/media-attach-mixin';
 import { getTemplateHTML } from '../background-video/template';
 
+const HTMLElementBase = globalThis.HTMLElement ?? class {};
+
 // `MediaAttachMixin` is typed as returning its base, so its `disconnectedCallback`
 // isn't visible for `super` to reach. `CustomElement` declares the lifecycle
 // callbacks this element overrides.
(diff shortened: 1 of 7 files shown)


decepulis (approved):

packages/html/src/media/hls-background-video/media.ts
```diff
@@ -6,10 +6,12 @@ import type { Constructor } from '@videojs/utils/types';
 import { MediaAttachMixin } from '../../store/media-attach-mixin';
 import { getTemplateHTML } from '../background-video/template';
 
+const HTMLElementBase = globalThis.HTMLElement ?? class {};
```
sampotts, line 9: We repeat this 4 times. Should we just export it from somewhere?

sampotts (approved):

packages/html/src/media/hls-background-video/media.ts
mihar-22: I think for this particular type of thing I'm okay with duplication, prefer that then running into weird class declaration and module loading issues. Makes/keeps the SSR boundary explicity.

PR #2429, opened 2026-08-25
fix(media): make engine entries server importable

### LINKED ISSUE #1343: Bug: DashVideo Breaks SSR — Top-Level dashjs Import Reads `window`

mihar-22:
## Summary

Keep DASH and Shaka media entries importable when Astro, Nuxt, SvelteKit, or another server runtime evaluates them without browser globals.

## Changes

- Route the DASH package export to an inert media host on servers and retain the real dash.js host under the browser condition
- Add the explicit .js extension required for Node to resolve Shaka Player published ESM
- Exercise built package exports in fresh Node processes and verify browser resolution still selects the DASH implementation

## Testing

- vp run @videojs/media#build
- pnpm -F @videojs/media test (44 files, 1,048 tests)
- pnpm typecheck
- pnpm check:workspace
- vp run @videojs/html#build
- vp run @videojs/react#build
- Bare Node imports of the built HTML and React DASH/Shaka entries

Closes #1343

This is independent of #2428, which handles the HTML custom-element DOM globals.

### DIFF
--- a/packages/media/package.json
+++ b/packages/media/package.json
@@ -33,6 +33,14 @@
       "development": "./dist/dev/dom.js",
       "default": "./dist/default/dom.js"
     },
+    "./dom/dash": {
+      "types": "./dist/dev/dom/dash/index.d.ts",
+      "browser": {
+        "development": "./dist/dev/dom/dash/index.js",
+        "default": "./dist/default/dom/dash/index.js"
+      },
+      "default": "./dist/default/dom/dash/server.js"
+    },
     "./dom/*": {
       "types": "./dist/dev/dom/*/index.d.ts",
       "development": "./dist/dev/dom/*/index.js",
--- a/packages/media/src/dom/dash/server.ts
+++ b/packages/media/src/dom/dash/server.ts
@@ -1,0 +1,44 @@
+import { MediaTracksMixin } from '../../core/media-tracks';
+import { HTMLVideoElementHost } from '../video-host';
+import type { DashMediaProps, DashSource } from './media';
+
+export type { DashEngineConfig, DashMediaProps, DashSource } from './media';
+
+export const dashMediaDefaultProps: DashMediaProps = {
+  src: '',
+  source: null,
+};
+
+const DashMediaHost = MediaTracksMixin(HTMLVideoElementHost);
+
+/** An inert DASH host used when the package is evaluated outside a browser. */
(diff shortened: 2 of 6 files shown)


decepulis (approved):
you can decide if this is in scope for this PR. Maybe you can restore these examples to the shaka and dash react component references now that they're ssr safe

sampotts (approved):

PR #2440, opened 2026-08-26
fix(packages): improve menu behavior and sizing

sampotts:
## Summary

Improve settings-menu sizing and submenu highlight behavior across the HTML and React implementations and the Default and Minimal skins. The current Safari benchmark no longer reproduces the earlier performance issue, so this keeps normal popup resize tracking and avoids adding a special-case positioning option.

## Changes

- Use menu-owned available width and height variables for CSS and Tailwind size clamps
- Keep menu item labels on one line and reduce the settings-menu minimum width
- Match Tailwind content positioning to CSS so menu padding is measured once
- Clear the parent item highlight when a submenu opens and restore it when focus returns
- Cancel pending initial highlights when menu state explicitly clears them
- Add focused core, HTML, React, and E2E regression coverage

## Testing

- `pnpm build:packages`
- `pnpm -F @videojs/core test src/dom/ui/popover/tests/popup-positioner.test.ts` (5 passed)
- `pnpm -F @videojs/html test src/ui/menu/tests/menu-element.test.ts` (14 passed)
- `pnpm -F @videojs/react test src/ui/menu/tests/menu.test.tsx` (44 passed)
- `pnpm -F @videojs/skins test` (20 passed)
- Local WebKit two-item geometry check: CSS and Tailwind both measure 74px with the same 5px bottom gap
- Local Safari benchmark: branch and `origin/main` both remain at approximately 61 FPS
- Formatting and `git diff --check`


decepulis (approved):
Approved after review of the menu sizing, submenu highlight, and intentional Safari resize-tracking changes.

PR #2446, opened 2026-08-26
fix(ci): publish matching html release archives

mihar-22:
## Summary

Make self-hosting archive examples version-aware and ensure release CI builds and uploads artifacts from the commit that release-please tagged.

The beta.32 release job checked out the triggering beta.31 commit, so it rebuilt and uploaded beta.31 while beta.32 remained without archive assets. The docs use the verified beta.31 archive for that known gap and will follow the package version automatically for future releases.

## Changes

- Check out the `@videojs/html` release SHA before building, publishing, and updating `site/v10`
- Fail release CI if the checked-out package version does not match release-please output
- Upload to the exact release tag and verify the ZIP, tarball, and checksum assets exist
- Render archive examples from one package-derived placeholder instead of repeated beta literals
- Preserve a beta.32 to beta.31 fallback until the missing beta.32 assets are backfilled
- Make `curl` fail on HTTP errors

## Testing

- `pnpm -F site test src/utils/tests/satteriCdnVersion.test.ts`
- `pnpm check:workspace`
- `pnpm build:site`
- Parsed `.github/workflows/cd.yml` as YAML
- Verified the rendered self-hosting page contains beta.31 URLs and no unresolved placeholders
- Verified the beta.31 archive URL returns a successful HTTP response

### DIFF
--- a/site/src/consts.ts
+++ b/site/src/consts.ts
@@ -20,6 +20,13 @@ export const BLOG_PAGE_SIZE = 10;
 export const VJS10_VERSION = htmlPackage.version;
 export const VJS10_HTML_CDN_BASE = `https://cdn.jsdelivr.net/npm/@videojs/html@${VJS10_VERSION}/cdn`;
 
+// The beta.32 release job checked out the commit before its version bump, so its
+// archive upload rebuilt beta.31 instead. Future releases use the current package
+// version; keep the latest available archive in docs until beta.32 is backfilled.
+const VJS10_HTML_ARCHIVE_VERSION_OVERRIDES = new Map([['10.0.0-beta.32', '10.0.0-beta.31']]);
+
+export const VJS10_HTML_ARCHIVE_VERSION = VJS10_HTML_ARCHIVE_VERSION_OVERRIDES.get(VJS10_VERSION) ?? VJS10_VERSION;
+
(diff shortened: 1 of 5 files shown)


decepulis (commented):
We've got that feature where we inject CDN versions into our docs. Imo we should use that here.

site/src/consts.ts
```diff
@@ -20,6 +20,13 @@ export const BLOG_PAGE_SIZE = 10;
 export const VJS10_VERSION = htmlPackage.version;
 export const VJS10_HTML_CDN_BASE = `https://cdn.jsdelivr.net/npm/@videojs/html@${VJS10_VERSION}/cdn`;
 
+// The beta.32 release job checked out the commit before its version bump, so its
+// archive upload rebuilt beta.31 instead. Future releases use the current package
+// version; keep the latest available archive in docs until beta.32 is backfilled.
+const VJS10_HTML_ARCHIVE_VERSION_OVERRIDES = new Map([['10.0.0-beta.32', '10.0.0-beta.31']]);
```
decepulis, line 26: lol

decepulis (approved):
sure!

JPustkuchen:
@decepulis did you see https://github.com/videojs/v10/issues/2526?

PR #2461, opened 2026-08-27
docs(site): document imperative control in the migration guides

decepulis:
The Mux Player guide explains how to script the player directly; the Plyr, Media Chrome, and Video.js 8 guides did not. Completes the Plyr guide's imperative section, adds a "Control the player imperatively" section to the Media Chrome guide, and augments the Video.js 8 guide's "Drive playback" section with framework cases mapping old ref-to-`videojs()` habits.

After review, all four guides lead with the same story: control everything through the player's store — actions like `play()`, `seek()`, `setPlaybackRate()` — with direct `media.*` calls as the escape hatch, not the default. All four state the architectural point: player state derives from native media events, so scripting the media element never desyncs the UI. Every referenced API verified against `packages/` source; also corrects `enterFullscreen` → `requestFullscreen` in the Video.js 8 table.

**Stack 5/13**. Base: `claude/docs-stack-04-mux-guide`.

https://claude.ai/code/session_01BGK1DZagw6XWLAX6L3LH47

---
_Generated by [Claude Code](https://claude.ai/code/session_01BGK1DZagw6XWLAX6L3LH47)_

---

> [!NOTE]
> **Low Risk**
> Documentation-only edits to site migration MDX; no runtime or API behavior changes.
> 
> **Overview**
> Aligns **Plyr**, **Media Chrome**, **Video.js 8**, and **Mux Player** migration guides on how to drive playback imperatively in v10.
> 
> Each guide now treats the **player store** as the primary model: store actions such as `play`, `seek`, `setVolume`, `toggleFullscreen`, and `toggleSubtitles`, reached via `player.store.state` in HTML or `usePlayer` in React. **Drive playback** / imperative API tables list those actions first and still document direct `media.*` calls, with shared wording that player state follows native media events so the UI stays in sync.
> 
> **Media Chrome** gains a full **Control the player imperatively** section (new for that guide). **Plyr** and **v8** gain expanded framework notes on refs, `useMedia`, and `PlayerController`. The **v8** table fixes fullscreen naming (`requestFullscreen` instead of `enterFullscreen`). **Mux** only tightens the existing drive-playback table and intro paragraph to match the same story.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce16716b0174424b4b69ad6e4611b00c6dc28ffb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/site/src/content/docs/how-to/migrate-from-media-chrome.mdx
+++ b/site/src/content/docs/how-to/migrate-from-media-chrome.mdx
@@ -300,10 +300,10 @@ React skins do not use Shadow DOM. Use `render` props when you build individual 
 
 ## Control the player imperatively
 
-Media Chrome taught you to script the slotted media element directly, and that habit carries over unchanged. The media is a plain child of the player, and Video.js derives player state from the native media events, so `video.play()` or `video.currentTime = 30` never leaves the controls out of sync. What changes is where everything else lives: Media Chrome had no player object, while Video.js puts the state the browser doesn't own — fullscreen, captions, quality — on the player.
+Control everything through the player's store. Media Chrome had no player object; Video.js gives you one, with an action for every operation — `play`, `togglePaused`, `seek`, `setVolume`, `toggleFullscreen` — so one mental model covers playback and the state the browser doesn't own, such as fullscreen, captions, and quality.
 
 <FrameworkCase frameworks={["html"]}>
-Query the media element and drive it as before. Player-owned actions live on the player element's store:
+Hold a reference to the player element and call actions on its store:
 
 ```js
 const player = document.querySelector('video-player');
@@ -313,7 +313,7 @@ player.store.state.toggleFullscreen();
 </FrameworkCase>
 
 <FrameworkCase frameworks={["react"]}>
-Put a `ref` on the media component; its value is the rendered `HTMLVideoElement`, so there's no `media` slot to reach through. Player-owned actions come from the same <DocsLink slug="reference/use-player">`usePlayer`</DocsLink> hook you read state with:
+Select an action with the same <DocsLink slug="reference/use-player">`usePlayer`</DocsLink> hook you read state with:
 
 ```tsx
 import { usePlayer } from '@videojs/react/video';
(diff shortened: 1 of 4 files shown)


decepulis (commented):
I kinda want to lead with "control everything through the store" since that's a simple mental model that covers all situations. And then, "you can still access the media, though, if you want" as an escape hatch. What do you think? Does that solve the underlying notion issue? Does that fit in with how we communicate about imperative control elsewhere?

decepulis:
I think store-first is the better lead, and it holds up on all three counts:

**Does the model cover all situations?** Yes, literally. I checked the feature sources: the store exposes `play()`/`togglePaused()` (playback), `seek()` (time), volume, rate, fullscreen, captions, quality, PiP — every row in these guides' mapping tables has a store action. So "control everything through the store" isn't a simplification that breaks down later; it's true.

**Does it fit how we communicate elsewhere?** Better than the current framing does. The `usePlayer` reference already says store access is "for imperative actions like play, pause, and volume changes", and #2447 (typed store state on player elements) is about to make the store the front door for HTML too — `player.play()` straight off the element. The current sections lead with an ownership split (browser-owned → media element, player-owned → store) that the reader has to internalize before they can do anything; store-first needs no such table.

**Does it solve the underlying Notion issue?** Yes, more directly. The item is migrators asking "where did `player.play()` go?" — one answer ("the player's store") beats "it depends which half of the API you mean."

Two things I'd keep in the escape-hatch paragraph rather than lose: (1) the "player state derives from native media events, so scripting the media element never desyncs the UI" line — for Media Chrome and Plyr migrators that habit is the whole reason the section exists, and the reassurance is the load-bearing part; (2) the media element / `useMedia` path stays the *only* route for a few things — swapping `src`, replacing the media component, native events, and the `engine` hatch — so the escape hatch should name those as its legitimate uses, not just "if you want."

I'll rewrite the four guides' sections with that inversion once your review pass is done and the edit hold lifts.

---

decepulis:
Store-first inversion pushed in b7b610a. All four guides' imperative sections now lead with "control everything through the player's store" — the Plyr table maps each Plyr call to its store action (`play()`, `seek(10)`, `setVolume(0.5)`, `toggleSubtitles()`…), the vjs8 and Mux tables list the player action first with the media property second, and Media Chrome opens with the store before the escape hatch. Direct media scripting stays as the escape-hatch paragraph, keeping the can't-desync reassurance and its exclusive jobs (swapping `src`, replacing the media component, native events, `engine`).

---

mihar-22 (approved):

PR #2472, opened 2026-08-27
fix(packages): preserve fullscreen after pointer activation

sampotts:
## Summary

Keep fullscreen active when a pointer-activated fullscreen button is followed by Space. Pointer activation now returns focus to the player container, while keyboard activation preserves button focus.

## Changes

- Share the pointer-focus decision through fullscreen button core
- Apply matching focus behavior in HTML and React
- Cover pointer and keyboard activation plus the fullscreen/Space browser flow

## Related

- [Spacebar exits fullscreen before controls receive focus](https://app.notion.com/p/mux/Spacebar-exits-fullscreen-before-controls-receive-focus-839a5ec569b74707b1bdbd8e35562cbb?v=3bc97a7f89d08023a21d000c69cb1285&source=copy_link)

## Testing

- Core fullscreen button: 19 tests
- HTML fullscreen button: 2 tests
- React fullscreen button: 2 tests
- `pnpm lint`
- `pnpm typecheck`

### DIFF
--- a/apps/e2e/tests/keyboard.spec.ts
+++ b/apps/e2e/tests/keyboard.spec.ts
@@ -211,6 +211,16 @@ for (const entry of PAGES as readonly PageEntry[]) {
         await expect(player.muteButton).toBeFocused();
       });
     } else {
+      test('keeps fullscreen after pointer activation followed by Space', async ({ page }) => {
+        await player.fullscreenButton.click();
+        await expect(player.fullscreenButton).toHaveAttribute(DATA_ATTRS.fullscreen, '');
+        await expect(player.playerRoot).toBeFocused();
+
+        await page.keyboard.press('Space');
+        await expect(player.fullscreenButton).toHaveAttribute(DATA_ATTRS.fullscreen, '');
+        await expect(player.playButton).not.toHaveAttribute(DATA_ATTRS.paused);
+      });
+
       test('operates captions with Enter and Space when a track is available', async ({ page }) => {
         await page.evaluate(() => {
           const video = document.querySelector('video');
--- a/packages/core/src/dom/ui/button.ts
+++ b/packages/core/src/dom/ui/button.ts
@@ -1,7 +1,9 @@
 import type { UIEvent, UIKeyboardEvent } from './event';
 
+export type ButtonActivationSource = 'pointer' | 'keyboard' | 'virtual';
+
 export interface ButtonOptions {
-  onActivate: (event: UIEvent) => void;
+  onActivate: (event: UIEvent, source: ButtonActivationSource) => void;
   isDisabled: () => boolean;
 }
 
@@ -28,7 +30,7 @@ export function createButton(options: ButtonOptions): ButtonProps {
         return;
       }
 
-      onActivate(event);
+      onActivate(event, (event.detail ?? 0) > 0 ? 'pointer' : 'virtual');
     },
 
     onPointerDown(event) {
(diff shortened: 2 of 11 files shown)


packages/core/src/core/ui/fullscreen-button/core.ts
```diff
@@ -64,6 +64,10 @@ export class FullscreenButtonCore {
     };
   }
 
+  shouldFocusContainer(detail?: number): boolean {
+    return Boolean(detail);
+  }
```
mihar-22: This doesn't make much sense to me. How does any `detail: number` determine if the container should be focused?

mihar-22: I see. It's trying to check if to see if this is a pointer event (mouse/touch) opposed to key event because `details` lives on those events to differentiate types (e.g., left vs. right click). I think it's a little too obscure and models core stuff according to a DOM-specifics. Intent is lost.

I was initially thinking something like `isPointerActivation(event)` but did some digging into Base UI. I modelled something with AI here that might help:

<details>
  <summary>🗒️  Click see proposal</summary>

## Button activation source design

`createButton` in `@videojs/core/dom` should translate browser events into a semantic activation source. Framework packages consume that source; framework-neutral component cores remain unaware of events and focus.

```ts
export type ButtonActivationSource =
  | 'pointer'
  | 'keyboard'
  | 'virtual';

export interface ButtonOptions {
  onActivate: (
    event: UIEvent,
    source: ButtonActivationSource
  ) => void;
  isDisabled: () => boolean;
}
```

Definitions:

- `pointer`: mouse, touch, or pen click.
- `keyboard`: Enter or Space handled directly by `createButton`.
- `virtual`: pointerless `click`, such as assistive-technology or programmatic activation.

`createButton` owns the browser heuristic:

```ts
onClick(event) {
  onActivate(
    event,
    event.detail > 0 ? 'pointer' : 'virtual'
  );
}

onKeyDown(event) {
  if (event.key === 'Enter') {
    event.preventDefault();
    onActivate(event, 'keyboard');
  }
}

onKeyUp(event) {
  if (event.key === ' ') {
    onActivate(event, 'keyboard');
  }
}
```

The second callback argument is backward-compatible: existing consumers can continue ignoring it.

Framework responsibilities:

```ts
if (
  focusContainerOnPointerActivation &&
  source === 'pointer'
) {
  container.focus({ preventScroll: true });
}
```

- React: add `focusContainerOnPointerActivation?: boolean` to `createMediaButton`; fullscreen opts in.
- HTML: `FullscreenButtonElement` checks `source === 'pointer'` and focuses its container.
- Core: do not add `FullscreenButtonCore.shouldFocusContainer()`.

Tests should cover:

- `click.detail > 0` → `pointer`
- Enter/Space → `keyboard`
- `click.detail === 0` → `virtual`
- Fullscreen focuses the container only for `pointer`
- Pointer fullscreen followed by Space toggles playback without exiting fullscreen

Base UI uses the same “virtual click” terminology and encapsulates `detail === 0` behind `isVirtualClick`, including additional browser/assistive-technology handling. That is useful prior art for keeping the heuristic at the interaction boundary rather than exposing `detail` to component logic. [Base UI source](https://github.com/mui/base-ui/blob/master/packages/react/src/floating-ui-react/utils/event.ts#L13-L22)

</details>

mihar-22 (changes requested):

mihar-22 (approved):
Thank you!

PR #2474, opened 2026-08-27
fix(packages): isolate focused slider hotkeys

sampotts:
## Summary

Let focused sliders own their keyboard input before player hotkeys run, and make focused volume movement use the same 5% step as volume hotkeys.

## Changes

- Handle slider keydown during capture in HTML and React
- Preserve native pointer-capture event names in the shared element prop adapter
- Define the 5% volume step in every HTML and React skin
- Keep time-slider stepping unchanged

## Related

- [Focused sliders also trigger player hotkeys](https://app.notion.com/p/mux/Focused-sliders-also-trigger-player-hotkeys-c9f0f2b16a57427eb9671cc563a56f23?v=3bc97a7f89d08023a21d000c69cb1285&source=copy_link)

## Testing

- Core slider and element props: 66 tests
- HTML slider and time slider: 42 tests
- React slider, time slider, and volume slider: 80 tests
- Ejected-skin generator: 32 variants
- Site ejected-skin tests: 16 tests
- `pnpm lint`
- `pnpm typecheck`

### DIFF
--- a/packages/core/src/dom/ui/slider.ts
+++ b/packages/core/src/dom/ui/slider.ts
@@ -59,7 +59,7 @@ export interface SliderRootStyle extends Record<string, string> {
 }
 
 export interface SliderThumbProps {
-  onKeyDown: (event: UIKeyboardEvent) => void;
+  onKeyDownCapture: (event: UIKeyboardEvent) => void;
   onFocus: () => void;
   onBlur: () => void;
 }
@@ -270,7 +270,7 @@ export function createSlider(options: SliderOptions): SliderApi {
 
   // --- Thumb props ---
   const thumbProps: SliderThumbProps = {
-    onKeyDown(event) {
+    onKeyDownCapture(event) {
       if (options.isDisabled()) {
         if (event.key !== 'Tab') event.preventDefault();
 
--- a/packages/html/src/presets/audio/minimal-skin.ts
+++ b/packages/html/src/presets/audio/minimal-skin.ts
@@ -5,6 +5,7 @@ import { SkinElement } from '../skin';
 
 import styles from '../../define/audio/minimal-skin.css?inline';
 
+const VOLUME_STEP = 5;
 const SEEK_TIME = 10;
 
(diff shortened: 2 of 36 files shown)


packages/html/src/presets/audio/minimal-skin.ts
mihar-22, line 1: Should we just make these the defaults for volume slider and hotkey `volumeStep`? Avoids configuring in all the skins and seems sensible to me?

packages/core/src/dom/ui/slider.ts
```diff
@@ -59,7 +59,7 @@ export interface SliderRootStyle extends Record<string, string> {
 }
 
 export interface SliderThumbProps {
-  onKeyDown: (event: UIKeyboardEvent) => void;
+  onKeyDownCapture: (event: UIKeyboardEvent) => void;
```
mihar-22, line 62: A comment for why we're using "capture" would help. Assuming to beat player hotkeys?

---

~I think `event.stopPropagation()` makes more sense here since child is stopping parent? Parent here being the `Container`.~ Scratch that, won't work because of React synthetic system.

mihar-22 (approved):

packages/core/src/dom/ui/slider.ts
sampotts: I asked Codex the same thing: 

CleanShot 2026-08-27 at 14 32 47@2x

sampotts: I'll get a comment in though for sure.

packages/html/src/presets/audio/minimal-skin.ts
sampotts: https://github.com/videojs/v10/pull/2484

packages/core/src/dom/ui/slider.ts
sampotts: https://github.com/videojs/v10/pull/2484

PR #2475, opened 2026-08-27
fix(core): suppress repeated volume boundary feedback

sampotts:
## Summary

Suppress repeated volume-boundary feedback while a volume key remains held, then allow a fresh transition after keyup.

## Changes

- Carry the keyboard repeat flag into input actions
- Preserve existing boundary state during repeated keydown events
- Allow the next non-repeated keydown to start fresh feedback

## Related

- [Volume bound shake loops while key is held](https://app.notion.com/p/mux/Volume-bound-shake-loops-while-key-is-held-12289ce259cf4fb0bfb46c8ae1f7899a?v=3bc97a7f89d08023a21d000c69cb1285&source=copy_link)

## Testing

- Core volume indicator and input action: 11 tests
- `pnpm lint`
- `pnpm typecheck`

### DIFF
--- a/packages/core/src/core/ui/input-action/input-action.ts
+++ b/packages/core/src/core/ui/input-action/input-action.ts
@@ -19,6 +19,7 @@ export interface InputActionEvent {
   value?: number | undefined;
   source?: InputActionSource | undefined;
   key?: string | undefined;
+  repeat?: boolean | undefined;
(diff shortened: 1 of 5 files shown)


mihar-22 (approved):
Nice, feels better 👌

PR #2517, opened 2026-08-28
fix: scale seek-bar thumbnails to fill their box

sampotts:
## What

Seek-bar thumbnail previews stayed at their native tile size when the player entered fullscreen. Two related fixes to the preview:

1. **Scale to fill (the reported bug).** `ThumbnailCore.calculateScale` returned early on `maxRatio < 1`, so it only ever scaled a sprite tile *down*. Constraints were already re-read on every update, so when a container query widened the preview in fullscreen the new `max-width` was seen and then ignored. It now fills the max constraints in either direction, and min constraints win over max as they do in CSS — that branch was previously unreachable whenever a max constraint was finite.

2. **No spinner flash on a failed sheet.** `updateSrc` cleared `error` on every URL change, so scrubbing back and forth across a sheet boundary restarted the loading state for a sheet already known to 404, flashing the skin's spinner shell under the pointer on each crossing. Failed sheet URLs are now remembered. It stays recoverable rather than permanently sticky: the renderer assigns the `src` either way, so a sheet that failed on a network blip and succeeds later clears itself on load.

## Why this shape

The first version of this branch reworked sizing onto `width: 100%` + `aspect-ratio` + a `ResizeObserver`. That was more machinery than the bug needed and introduced three regressions of its own: non-16:9 tiles were cover-fitted and cropped (~44% of a 9:16 tile, since every skin clamps the preview with `max-height`), a failed sprite rendered a full-size empty black box instead of collapsing, and the element silently required a wrapper with a definite width — breaking third-party skins built on the previous shrink-wrap pattern. That approach is reverted here; the element sizes itself intrinsically from its CSS constraints as before.

## Verification

Measured in Chromium: the thumbnail grows from **190px → 334px** on entering fullscreen (against a 192px → 336px `max-width`), wrapper hugging it, aspect ratio held at 1.79 throughout.

- e2e `sandbox-skin-styling.spec.ts`: 30 passed, including a new fullscreen scaling test across all 8 skin/renderer combinations (html/react × default/minimal × css/tailwind).
- Unit: core 1475, skins 25, html 447, react 536.
- `pnpm typecheck`, `pnpm lint`, `pnpm check:workspace`.

## Reviewer notes

- The new e2e assertions use `offsetWidth`, not `getBoundingClientRect()`: the preview scales in over 150ms, so a bounding rect is transform-skewed mid-animation.
- No ARIA changes. The preview host is already `aria-hidden="true"`, so an accessibility review confirmed no naming or status-message requirement in any state, and that a live region here would race the seek slider's own `aria-valuetext` during a scrub.
- **Forced colors is deliberately out of scope and will follow separately.** Worth knowing what's there: forced-colors mode drops every `box-shadow`, and the preview's only edge comes from one — the surface shadow in the default skin, the `::after` ring in the minimal skin — so it renders as an unbounded `Canvas` rectangle over the video. The same trap makes the vjsc skin's existing forced-colors borders no-ops (`surface.styles.ts` builds them from `ring-*` and inset shadows), leaving every vjsc menu, tooltip, and feedback popup borderless in High Contrast mode. Its e2e test doesn't catch it because it only diffs CSS against Tailwind output, and two equally-empty outputs match. The fix in both cases is `outline`, which survives forced colors and follows `border-radius`.

---

> [!NOTE]
> **Medium Risk**
> Changes core thumbnail sizing math and load/error state during scrubbing, which affects every skin’s seek preview; risk is mitigated by broad e2e and unit coverage.
> 
> **Overview**
> Seek-bar thumbnail previews no longer stay at native tile size when fullscreen widens their CSS `max-width` via container queries. **`ThumbnailCore.calculateScale`** now sizes tiles to **fill** the max constraint box (scaling up as well as down) while letting min constraints override max, matching prior “fit only when too large” behavior.
> 
> The DOM thumbnail controller **remembers failed sprite sheet URLs** so scrubbing back onto a known-bad sheet keeps `error` without toggling `loading` (no spinner flash); a later successful `load` clears the failure cache.
> 
> **Playwright** adds a per-skin/renderer test that hovers the seek bar, enters fullscreen, and asserts larger layout size, stable aspect ratio, and sprite coverage. **Unit tests** cover the new scale math, failed-sheet revisits, resize observer notifications, and observer teardown. **Docs** describe fill-up scaling for growing preview boxes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 44402e487194891b29bd4fa88e38058e4cbfb556. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/core/ui/thumbnail/core.ts
+++ b/packages/core/src/core/ui/thumbnail/core.ts
@@ -69,12 +69,13 @@ export class ThumbnailCore {
   }
 
   /**
-   * Calculate a uniform scale factor that fits `tileWidth × tileHeight` within the given CSS min/max constraints while
+   * Calculate a uniform scale factor that sizes `tileWidth × tileHeight` to the given CSS min/max constraints while
    * preserving aspect ratio.
    *
-   * - Scales down when the tile exceeds max constraints.
-   * - Scales up when the tile is smaller than min constraints.
-   * - Returns `1` when no scaling is needed.
+   * - Fills the max constraints, scaling the tile up as readily as down. A box that grows — entering fullscreen widens it
+   *   through a container query — has to take the tile with it rather than leave it at its native size.
+   * - Raises that to meet min constraints, which win over max as they do in CSS.
+   * - Returns `1` when unconstrained.
    */
   calculateScale(tileWidth: number, tileHeight: number, constraints: ThumbnailConstraints): number {
     const { minWidth, maxWidth, minHeight, maxHeight } = constraints;
@@ -82,18 +83,13 @@ export class ThumbnailCore {
     const maxRatio = Math.min(maxWidth / tileWidth, maxHeight / tileHeight);
     const minRatio = Math.max(minWidth / tileWidth, minHeight / tileHeight);
 
-    // Scale down if exceeding max constraints.
-    if (Number.isFinite(maxRatio) && maxRatio < 1) return maxRatio;
+    const scale = Number.isFinite(maxRatio) ? maxRatio : 1;
 
-    // Scale up if below min constraints.
-    if (Number.isFinite(minRatio) && minRatio > 1) return minRatio;
-
-    return 1;
+    return Number.isFinite(minRatio) && minRatio > scale ? minRatio : scale;
   }
 
   /**
-   * Compute container and image dimensions for the current thumbnail, scaled to fit within the element's CSS min/max
-   * constraints.
+   * Compute container and image dimensions for the current thumbnail, scaled to the element's CSS min/max constraints.
    *
    * The container clips the sprite sheet via `overflow: hidden`, and the image is positioned with `transform:
    * translate()` to show the correct tile.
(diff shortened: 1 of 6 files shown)


packages/core/src/core/ui/thumbnail/core.ts
```diff
@@ -69,31 +69,27 @@ export class ThumbnailCore {
...
 
-    // Scale down if exceeding max constraints.
-    if (Number.isFinite(maxRatio) && maxRatio < 1) return maxRatio;
+    const scale = Number.isFinite(maxRatio) ? maxRatio : 1;
 
-    // Scale up if below min constraints.
-    if (Number.isFinite(minRatio) && minRatio > 1) return minRatio;
-
-    return 1;
+    return Number.isFinite(minRatio) && minRatio > scale ? minRatio : scale;
```
sampotts, line 88: This is the bulk of the change. The rest is tests and handling failed src a little better.

mihar-22 (approved):

PR #2534, opened 2026-08-28
fix(spf): never adopt the UA-default preload on attach

### LINKED ISSUE #2532: Bug: Attach-Time Preload Read Clobbers Author Intent With the UA Default on Shadow-Video Medias
## Context

`syncPreload`'s DOM→state read (documented "most-recent-wins on attach") copies `mediaElement.preload` into `state.preload` on `context.mediaElement` swap. The SPF Medias hand the engine a shadow-DOM `<video>` that never carries the host's `preload` attribute — and the `preload` IDL property returns a W3C value even when no attribute is set: the **UA default**, which differs per browser (measured 2026-08-28 on bare elements: `'metadata'` Chromium, `'auto'` WebKit).

Result: `<hls-video preload="metadata">` forwards correctly — host attribute → `attributeChangedCallback` → adapter `preload` setter → `state.preload.set('metadata')` — and the attach-time read then overwrites it with the inner element's UA default. The read cannot distinguish author intent from UA policy because it consults the property, not the attribute.

- On Chromium the clobber is invisible: the UA default happens to equal the page's value.
- On WebKit `state.preload` flips to `'auto'`, and the loaders run full-range exactly as designed: a paused player buffers the entire stream under `preload="metadata"`.

Verified chain: with `<hls-video preload="metadata">`, paused, the adapter's IDL mirror reads `'metadata'` on both browsers (the setter demonstrably ran), the shadow video's `preload` **attribute** is null on both, and `engine.state.preload` reads `'metadata'` on Chromium / `'auto'` on WebKit; WebKit reaches `readyState 4` with the full window buffered while Chromium buffers nothing.

Related (same host→media forwarding family, different mechanism): #1893.

## User impact

`preload="metadata"`'s bandwidth restraint is browser-dependent on the SPF Medias: WebKit users pay a full download for a paused player the author asked to load metadata only.

## Direction

- SPF should **never defer to the media element's UA-default preload**. Absent author intent, SPF's own default is `'metadata'` (the `defaultPreload` backfill `syncPreload` already carries).
- The DOM→state read direction is the source of this bug class. **Dropping the bidirectional syncing altogether is on the table** (discussed out of band) if it's proving not worth the complexity or edge cases — leaving `state.preload` fed only by author intent (host attribute/property, engine config) over the SPF default.

## Acceptance criteria

- An author-set `preload` on a Media host governs `state.preload` across browsers; with no author value, `state.preload` is SPF's `'metadata'` default — the element's UA default is never a source.
- The DOM→state read is either removed (with the rest of the bidirectional sync, if the simplification holds up) or constrained so it cannot mistake a UA default for author intent.
- Regression coverage pins `state.preload` for a host-attribute'd Media — and for a no-attribute Media — on a browser whose UA default differs from `'metadata'`.

cjpillsbury:
Fixes #2532

## Summary

`syncPreload`'s attach-time DOM→state read consulted the `preload` IDL property, which reports a browser-dependent UA default (`'metadata'` Chromium, `'auto'` WebKit) even when no attribute was authored. The SPF Medias attach a shadow `<video>` that never carries the host's `preload` attribute, so the read clobbered author intent with UA policy — on WebKit a paused `<hls-video preload="metadata">` buffered the full stream. Chromium's default coincidentally matching SPF's masked the bug there.

## Changes

- The DOM→state read now consults the `preload` property only when the element actually carries a `preload` content attribute, so a UA default is never adopted as author intent.
- Absent author intent, `state.preload` stays on SPF's own `'metadata'` default; its doc no longer claims to match the element's implicit (UA-defined) default.
- `MediaElementLike` gains an optional `hasAttribute` for the reflection surface; a plain `MediaElementLike` without attribute reflection keeps its property authority (it has no UA to inject a default).
- Regression coverage in `sync-preload.test.ts` (authored-attribute gate, attach/swap/url-change paths) and `adapter.test.ts` (author value forwarded via the adapter survives attach).

<details>
<summary>Design choice: constrain the read vs drop the sync</summary>

The recorded direction on #2532 ruled out deferring to the UA default and left two shapes open: drop `syncPreload`'s DOM→state read entirely, or keep it constrained. This PR constrains it to authored intent: attribute-present reads keep the IDL property's spec normalization, and most-recent-wins on attach still holds for authored elements. Dropping the bidirectional sync altogether remains on the table if the attribute gate isn't worth its edge cases.

</details>

## Testing

- `pnpm -F @videojs/spf test src/playback/behaviors/tests/sync-preload.test.ts src/playback/adapters/hls-video/tests/adapter.test.ts` — 91 pass, no type errors (this branch is based on `main`, which has no WebKit unit tier; the divergence itself was diagnosed and measured in #2532).
- Manual cross-browser check: paused `<hls-video preload="metadata">` on WebKit should no longer buffer the stream; behavior now matches Chromium.

## Review context

This diff was produced end-to-end by a context-free agent session working from the recorded direction on #2532, as rehearsal 2 of the SPF automation proving-out plan (`internal/design/spf/automation/proving-out.md`). Review is deliberately of the cold diff; the embedded design choice above is one of the graded touchpoints.

---

> [!NOTE]
> **Medium Risk**
> Changes core playback loading behavior on media attach and URL changes; wrong gating could still mis-sync preload vs author intent, but scope is limited to sync-preload and the fix is well-tested.
> 
> **Overview**
> Fixes **#2532**: attach-time DOM→state preload sync no longer treats the shadow `<video>`’s browser-dependent `preload` IDL default as author intent (notably WebKit `'auto'` overriding `<hls-video preload="metadata">` and over-buffering).
> 
> **`syncPreload`** now reads `mediaElement.preload` only when the element has a `preload` content attribute (via optional **`hasAttribute`** on **`MediaElementLike`**); without an authored attribute it keeps existing **`state.preload`** or backfills SPF’s **`'metadata'`** default. Authored attributes still win on attach (most-recent-wins). **`DEFAULT_PRELOAD`** documentation is updated to state it is SPF policy, not the element’s UA implicit default.
> 
> Regression tests cover the attribute gate in **`sync-preload.test.ts`** and attach paths in **`adapter.test.ts`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3caa1ed3e32a8aeaf6605a307ad02abdf24f630. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/spf/src/playback/behaviors/sync-preload.ts
+++ b/packages/spf/src/playback/behaviors/sync-preload.ts
@@ -3,10 +3,13 @@
  *
  * Two effects:
  *
- * - **Read (DOM → state)** — on `context.mediaElement` swap or `state.presentation.url` change, copies
- *   `mediaElement.preload` into `state.preload` if it's a W3C value and `state.preload` isn't holding an extended
- *   (non-W3C) value. When the DOM has no W3C opinion and `state.preload` is undefined, backfills from
- *   `config.defaultPreload` (default-default `'metadata'`) so `state.preload` is never undefined in steady state.
+ * - **Read (DOM → state)** — on `context.mediaElement` swap or `state.presentation.url` change, copies the media
+ *   element's _authored_ preload into `state.preload` if it's a W3C value and `state.preload` isn't holding an extended
+ *   (non-W3C) value. Author intent only: on a DOM element the `preload` IDL property reports a browser-dependent UA
+ *   default even when no attribute was set (`'metadata'` Chromium, `'auto'` WebKit), so the property is consulted only
+ *   when the element carries a `preload` content attribute — a UA default is never a source. When the element has no
+ *   authored W3C opinion and `state.preload` is undefined, backfills from `config.defaultPreload` (default-default
+ *   `'metadata'`) so `state.preload` is never undefined in steady state.
  * - **Write (state → DOM)** — on `state.preload` change or `context.mediaElement` swap, writes `state.preload` back to
  *   `mediaElement.preload` if the value is W3C.
  *
@@ -63,12 +66,21 @@ function syncPreloadSetup({
     // Extended sticky: leave non-W3C values (e.g. 'canplay') alone.
     if (current !== undefined && !isStandardPreload(current)) return;
 
-    const target: StandardPreload | undefined =
-      mediaElement && isStandardPreload(mediaElement.preload)
+    // Author intent only: a DOM element's `preload` property reports a UA
+    // default when no attribute was authored (the shadow <video> under an SPF
+    // Media never carries one), so gate the property read on attribute
+    // presence. An element without attribute reflection has no UA defaults —
+    // its property is authored by construction.
+    const authoredPreload =
+      mediaElement && (!mediaElement.hasAttribute || mediaElement.hasAttribute('preload'))
         ? mediaElement.preload
-        : current === undefined
-          ? defaultPreload
-          : undefined;
+        : undefined;
+
+    const target: StandardPreload | undefined = isStandardPreload(authoredPreload)
+      ? authoredPreload
+      : current === undefined
+        ? defaultPreload
+        : undefined;
     if (target === undefined || target === current) return;
 
     state.preload.set(target);
--- a/packages/spf/src/media/types/index.ts
+++ b/packages/spf/src/media/types/index.ts
@@ -41,6 +41,13 @@ export interface AddressableObject {
  */
 export interface MediaElementLike {
   preload: string;
+  /**
+   * DOM attribute-reflection surface, present on real elements. Where it exists, reflected IDL properties like
+   * `preload` report a browser-dependent UA default even when no attribute was authored — so consumers that need author
+   * intent must gate property reads on attribute presence. A non-DOM implementation may omit it; its properties carry
+   * no UA defaults and are authored by construction.
+   */
+  hasAttribute?(qualifiedName: string): boolean;
 }
 
 // =============================================================================
--- a/packages/spf/src/media/utils/preload.ts
+++ b/packages/spf/src/media/utils/preload.ts
@@ -10,7 +10,8 @@ export function isStandardPreload(value: unknown): value is StandardPreload {
 
 /**
  * Default `preload` value used as the fallback across behaviors (`syncPreload`, `resolvePresentation`,
- * `isBlockingPreload`). Matches the `<video>`/`<audio>` element's implicit default.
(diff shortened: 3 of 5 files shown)


cjpillsbury:
## Verification: before/after measured on the preview deployments

> **Correction:** an earlier version of this comment predicted a buffering/network difference on this sandbox page. That was wrong — I hadn't run the steps before posting them. On the shipped `hls-video` Media, preload **gating** keeps paused segment loading off on this VOD page regardless of the clobbered value, so the network waterfall looks identical before and after. The honest observable is the engine state slot the bug clobbers: `state.preload`. (The buffered-window manifestation that #2532 was diagnosed from lives in the live/reduced-composition e2e context, which isn't in this PR's tree.)

### Measured results (state.preload, paused, authored `preload="metadata"`)

| Browser | control (`main`) | fixed (this PR) |
| --- | --- | --- |
| Real Safari 26.6.2 (safaridriver) | **`"auto"`** — author intent clobbered by the UA default | `"metadata"` ✓ |
| Playwright WebKit 26.4 | **`"auto"`** (also for an injected attribute-less `<hls-video>`) | `"metadata"` ✓ |
| Playwright Chromium 147 | `"metadata"` (UA default coincides — the masking) | `"metadata"` ✓ no-op |

### Reproduce it yourself (30 seconds, any browser)

1. Open the **control**: https://v10-sandbox-git-main-mux.vercel.app/html-hls-video/?preload=metadata — don't press play, wait a few seconds, then run in the console:
   ```js
   document.querySelector('hls-video').getMediaTarget().engine.state.preload.get()
   ```
   Safari/WebKit: `"auto"` (the bug — the authored attribute says `metadata`). Chrome: `"metadata"` (coincidence).
2. Open the **fixed preview**: https://v10-sandbox-git-fix-2532-preload-ua-default-mux.vercel.app/html-hls-video/?preload=metadata — same snippet reads `"metadata"` everywhere.
3. Sanity: press play on the fixed preview — playback starts normally; `?preload=auto` still reads `"auto"` (authored intent honored, not dropped).

packages/spf/src/playback/behaviors/sync-preload.ts
spuppo-mux, line 99: thought: this is probably no longer needed. If we remove it we can keep whatever is on media element as strictly author intent. However, I'm also thinking about whether it would be worth to reuse this logic for other standard media attributes and I'm noticing that this particular way to track author preference would not work for disableRemotePlayback.

spuppo-mux (approved):
non blocking thought

PR #2553, opened 2026-08-30
fix(packages): stabilize slider keyboard input

sampotts:
## Summary

Fix shared slider keyboard interaction so held keys advance monotonically, remain within range, settle on release, and keep player controls visible for the full press. The volume slider is the browser regression because delayed `volumechange` events exposed the general controlled-slider timing problem.

## Changes

- Track in-flight keyboard percentages and rebase them when pointer input takes over.
- Treat a held key as one slider press, releasing the controls lock on key-up or blur.
- Prevent React pointer-only style synchronization from overwriting keyboard-driven values.
- Disable position transitions while sliders are focused across the Default, Minimal, Tailwind, and VJSC skins.
- Synchronize the volume adapter immediately because `volumechange` can be asynchronous; the shared slider remains media-agnostic.
- Add browser coverage for uninterrupted boundary holds, release stability, overflow, and controls visibility.

## Testing

- `pnpm build:packages`
- `pnpm typecheck`
- `pnpm -F @videojs/core test src/dom/ui/tests/slider.test.ts src/dom/store/features/tests/volume.test.ts`
- `pnpm -F @videojs/html test src/ui/time-slider/tests/time-slider-element.test.ts`
- `pnpm -F @videojs/react test src/ui/slider/tests/slider.test.tsx src/ui/volume-slider/tests/volume-slider.test.tsx`
- `pnpm -F @videojs/skins test`
- `pnpm lint`
- Chromium sandbox E2E: repeated slider keys and held-key controls visibility (2 passed)

---

> [!NOTE]
> **Medium Risk**
> Touches core slider input, React style sync, and volume state timing—high user-visible surface area but scoped changes with broad test coverage.
> 
> **Overview**
> Fixes **held arrow-key** interaction on sliders (especially volume) so values step monotonically, stay in range, and don’t visually bounce when media state updates lag behind the UI.
> 
> The shared DOM slider now tracks **`lastKeyPercent`** and uses it for **`event.repeat`** steps instead of re-reading a stale controlled percent; pointer drags refresh that baseline. **`UIKeyboardEvent`** gains an optional **`repeat`** flag. React **`useSlider`** only applies pointer-driven CSS var sync while **dragging/pointing**, so keyboard steps aren’t overwritten by the previous value. Skins disable slider position transitions under **`group-focus-within/slider`** so keyboard moves aren’t animated into the wrong place.
> 
> **Volume** **`setVolume`** immediately **`set`s** store volume/muted because **`volumechange`** can arrive late. **Controls** treat **`keydown`** on the container like activity (not just **`keyup`**), so auto-hide doesn’t fire mid key-repeat.
> 
> Adds **Playwright** coverage (held keys, boundaries, release stability, delayed **`volumechange`**, controls stay visible) plus unit tests for repeat stepping and controls idle behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b870a37738d9bf24c44e007815e0be97ac931291. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/e2e/shared/fixtures/volume-slider.ts
+++ b/apps/e2e/shared/fixtures/volume-slider.ts
@@ -1,0 +1,89 @@
+import type { Locator, Page } from '@playwright/test';
+
+interface VolumeKeyResult {
+  beforeRelease: number;
+  frames: number[];
+  gaps: number[];
+  release: number[];
+}
+
+function getThumbPercent(slider: Locator): Promise<number> {
+  return slider.evaluate((element) => {
+    const thumb = element.querySelector<HTMLElement>('.media-slider__thumb, media-slider-thumb, [role="slider"]')!;
+    const sliderRect = element.getBoundingClientRect();
+    const thumbRect = thumb.getBoundingClientRect();
+
+    return 100 - ((thumbRect.top + thumbRect.height / 2 - sliderRect.top) / sliderRect.height) * 100;
+  });
+}
+
+export async function holdVolumeKey(
+  page: Page,
+  slider: Locator,
+  key: 'ArrowDown' | 'ArrowUp'
+): Promise<VolumeKeyResult> {
+  await slider.evaluate((element) => {
+    const frames: number[] = [];
+    const gaps: number[] = [];
+    const record = () => {
+      const thumb = element.querySelector<HTMLElement>('.media-slider__thumb, media-slider-thumb, [role="slider"]')!;
+      const sliderRect = element.getBoundingClientRect();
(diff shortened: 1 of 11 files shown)


mihar-22 (changes requested):
Temporary block. Same reasons as https://github.com/videojs/v10/pull/2558#pullrequestreview-5062588204 since it's touching the skins package.

sampotts:
All good! I can port it over when the time comes.

mihar-22 (approved):
great!

PR #2565, opened 2026-08-31
fix(media): keep one Mux Data view per viewing across loadstarts

### LINKED ISSUE #2562: Bug: Mux Data Splits One Viewing Into Duplicate Views and Sessions: Every Loadstart Destroys and Re-Creates the Monitor
## Context and impact

The Mux Data media component (`packages/media/src/dom/mux/mux-data.ts`) tears down and re-creates its `mux-embed` monitor on every `loadstart`, and generates a fresh `view_session_id` per `monitor()` call. Medias fire `loadstart` for many same-video reasons (remote playback engage, engine rebuilds, MSE re-attach), so a single continuous viewing is reported to Mux Data as multiple unrelated players/views, each landing in its own session. QoE metrics — startup time, rebuffering, watch time, errors — are split across the fragments and attributed to views that never logically existed, and views that belong to one session can never be grouped.

## Mechanism

1. `MuxData` listens for `loadstart` and runs `#reinitialize` (`packages/media/src/dom/mux/mux-data.ts:197-204`): `target.mux.destroy()` followed by a fresh `MuxDataSdk.monitor(target, …)`. The contract doc states the assumption directly — Mux Data needs "a `loadstart` to re-monitor on" (`mux-data.ts:34-45`) — and the test suite pins it ("re-monitors with the new engine when the media fires loadstart", `packages/media/src/dom/mux/tests/mux-data.test.ts:93`).
2. `mux-embed`'s own typings say the monitor/`viewinit` path "should only be emitted for the first view within a player; for a change of videos within the same player, `videochange` should be used" (`mux-embed@5.18.1`, `dist/types/mux-embed.d.ts`, `VIEW_INIT` doc). Destroy + re-monitor manufactures a brand-new player and view where the SDK expects one continuous player emitting `videochange` (or `programchange` for live linear).
3. `view_session_id` handling (`mux-data.ts:228-231`) makes the fragments uncorrelatable, with two opposite behaviors depending on whether a `metadata` prop happens to be set:
   - No `metadata` prop → the destructuring default `{}` is per-call, so every re-monitor generates a fresh UUID; views of one session never share a session id.
   - `metadata` prop set → the generated id is written into the **caller's** object as a side effect and is sticky for that object's lifetime, even across genuinely distinct videos/sessions later.

## `loadstart` without a new video (verified paths)

- **Remote playback engaged:** `load()` dispatches a synthetic `loadstart` whenever `remote.state === 'connected'` (`packages/media/src/dom/hls-js/media.ts:359-362`), and the Google Cast provider calls `load()` on session start, disconnect, and source handoff (`packages/media/src/dom/google-cast/google-cast-provider.ts:130,138,289`; `packages/media/src/dom/google-cast/media.ts:156`). Engaging AirPlay continues the same viewing on the same element, yet splits the view. (A Cast handoff may legitimately warrant its own monitoring decision, but it should be deliberate, not fallout.)
- **Engine rebuild with an unchanged source:** an engine-config change reruns `load()`, which always dispatches `loadstart` when a delegate exists (`packages/media/src/dom/hls-js/media.ts:416-419`).
- **MSE lifecycle:** `mediaElement.load()` runs on MediaSource attach, detach reset, and post-AirPlay rebuild (`packages/spf/src/media/dom/mse/mediasource-setup.ts:143,163`); also the native-HLS DRM path (`packages/media/src/dom/native-hls/drm.ts:202`). This shapes the eventual SPF integration (#1845) too.

Startup also runs two monitor generations back to back: `attach()` monitors immediately, then the initial `loadstart` destroys that monitor and creates another.

## Latent concurrent-monitor leak

`attach()` overwrites `#target` before `#reinitialize` runs (`mux-data.ts:73-76`), so on a target swap `#reinitialize` destroys the *new* target's (nonexistent) monitor and the old target's monitor keeps beaconing. `HTMLMediaElementHost.attach` doesn't detach components from a previous target either (`packages/media/src/dom/media-host/media-host.ts:51-63`). No current host swaps targets while attached, so this is latent rather than active — but monitor ownership belongs to the same contract.

## Acceptance criteria

- One logical viewing produces one Mux Data view. A video change within the same player emits `videochange` on the existing monitor instead of destroying and re-creating it.
- Same-video `loadstart`s — remote playback engage/disengage, engine rebuild, MSE re-attach, explicit `load()` — do not create new views or players.
- An engine instance swap re-hooks engine telemetry on the live monitor (`mux.addHLSJS`/`removeHLSJS`/`addDashJS`/`removeDashJS` on `MuxOnVideoElement`) rather than destroying it.
- `view_session_id` is honored when caller-supplied; otherwise the integration either omits it or generates one stable id per session lifetime (a deliberate decision, documented) — never a fresh UUID per `monitor()` call, and never mutated into the caller's `metadata` object.
- A monitor is never left live on a target the component no longer owns (target swap destroys or `swapElement`s first).
- Tests cover: same-src `load()` keeps the monitor; src change emits `videochange` with an updated `video_id`; engine-swap re-hook; target swap; session-id stability and no caller-object mutation.

## Related

- #1845 — the SPF Mux Data integration should land on the corrected contract rather than inheriting the `loadstart` assumption.
- #2038 / #2040 — prior engine-detection fix in this same component.
- #977 — the original Mux media elements feature this integration shipped under.

cjpillsbury:
Fixes #2562
Refs #1845

## Why

A single continuous viewing shows up in Mux Data as several unrelated players and views. Medias fire `loadstart` for many same-video reasons — remote playback engaging, an engine rebuild, a MediaSource re-attach — and the integration destroyed and re-created its `mux-embed` monitor on every one. Each fragment became its own view under a freshly generated `view_session_id`, so startup time, rebuffering, watch time, and errors are attributed to views that never logically existed and can never be grouped back into one session. When a caller supplied a `metadata` object, the generated id was also written into it as a side effect. The SPF integration (#1845) would have inherited all of this.

## What changed

`loadstart` is now a hint: the media's current `src` and `engine` are compared against what the monitor already reflects, and only real changes get a reaction.

- A same-video `loadstart` is a no-op; the monitor and its view stay alive.
- A video change emits `videochange` on the live monitor instead of destroy + re-`monitor()`, per the SDK's own contract (`viewinit` is first-view-only).
- A cleared source is ignored without being tracked, so a video loaded after it still emits `videochange` (and re-setting the same video stays a no-op).
- An engine swap re-hooks telemetry on the live monitor (`addHLSJS`/`removeHLSJS`/`addDashJS`/`removeDashJS`) rather than restarting it.
- A monitor started before the first source names its pending view via `updateData` rather than emitting a spurious `videochange`.
- `view_session_id` is generated once per component so a player's views group into one session; a caller-supplied id always wins, and caller metadata is never mutated.
- `attach()` destroys the old target's monitor before switching targets, closing a latent leak where a live monitor kept beaconing on an abandoned target.
- Full destroy + re-monitor remains only for options baked into `monitor()` itself (SDK, beacon domain, debug, cookies).

The rationale for the live-monitor model, with links to the SDK contract and prior art, is recorded in [`internal/decisions/media/mux-data-monitor-lifecycle.md`](https://github.com/videojs/v10/blob/fix/mux-data-view-continuity/internal/decisions/media/mux-data-monitor-lifecycle.md).

<details>

Syncing follows the media-state pattern (cf. `sourceFeature`): events trigger a deferred reconcile that reads current state and cross-compares against tracked identity, so redundant hints are free. This also makes the fix robust to how often a media fires `loadstart`, which varies by flavor — the hls.js-backed media synthesizes one per `load()`, while shadow-video medias forward every native `loadstart` raw. Identity compares use the facade `src` (the element's `currentSrc` is a `blob:` URL under MSE).

</details>

## Testing

`pnpm -F @videojs/media test` (1141 passing, including 17 rewritten/new `MuxData` cases covering same-src loads, `videochange` — including after a cleared source — engine re-hooking, first-source naming, session-id stability, caller-metadata immutability, and target swaps). The html and react `mux-data` suites pass. `pnpm typecheck` is clean for `packages/media`.

---

> [!NOTE]
> **Medium Risk**
> Changes production Mux Data view/session attribution and beacon lifecycle; behavior should align with the SDK contract but analytics dashboards may shift compared to the prior split-view bug.
> 
> **Overview**
> Fixes **#2562** by changing how `MuxData` talks to `mux-embed`: instead of tearing down and calling `monitor()` on every `loadstart`, it keeps **one monitor per video target** and reconciles media state when `loadstart` fires.
> 
> **Same `src` / engine rebuilds** (remote playback, MediaSource re-attach, etc.) are no-ops. **Real source changes** emit `videochange` on the live monitor. **Engine swaps** re-hook hls.js/dash.js via `addHLSJS`/`removeHLSJS` (and dash equivalents) without restarting the monitor. **Cleared sources** are ignored for tracking so the next real load still compares against the last known video. **First source after attach** uses `updateData` instead of a spurious `videochange`. **`view_session_id`** is generated once per `MuxData` instance (caller metadata is no longer mutated). **`attach()`** to a new target destroys the previous element’s monitor.
> 
> Full destroy + re-`monitor()` remains only when monitor-level options change (SDK, beacon domain, debug, cookies) or the target changes. Rationale is recorded in `internal/decisions/media/mux-data-monitor-lifecycle.md`, with tests expanded for the new sync behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 04dbd69e3c8cf5c1e494dd76574560514a9f04d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/decisions/README.md
+++ b/internal/decisions/README.md
@@ -41,6 +41,7 @@ Add alternatives, consequences, or source links only when they materially explai
 
 | Area | Decisions |
 | --- | --- |
+| `media/` | Media contracts, hosts, and third-party integrations such as Mux Data |
 | `player/` | Provider, container, media discovery, and player composition |
 | `spf/` | Stream-processing ownership and coordination |
 | `store/` | State-management contracts |
--- a/internal/decisions/media/mux-data-monitor-lifecycle.md
+++ b/internal/decisions/media/mux-data-monitor-lifecycle.md
@@ -1,0 +1,28 @@
+---
+status: decided
+date: 2026-09-01
+---
+
+# Mux Data Keeps One Monitor Per Target
+
+## Decision
+
+`MuxData` starts one `mux-embed` monitor per target and keeps it alive for the target's lifetime. A source change is reported to the live monitor as `videochange`, and an engine swap re-hooks engine telemetry on the live monitor. The monitor is destroyed and re-created only when the target changes or when an option baked into `monitor()` itself changes (SDK, beacon domain, debug, cookies).
+
+## Why
+
+Destroying and re-running `monitor()` per source, the shape Mux Player's `playback-core` uses, is the outlier among Mux integrations and the cause of #2562. mux-embed's documented contract for playing several videos in one player is `videochange`: the SDK ends the current view, resets every `video_*` and `view_*` field, and arms `viewstart` on the next `play`. Every Mux-maintained integration (the Video.js plugin, THEOplayer, Kaltura, Akamai, React Native Video) follows it. A fresh `monitor()` per source instead re-runs first-view setup, so each fragment of one viewing lands as a separate player and view and can never be grouped back together. `playback-core` rediscovered this the hard way: its FairPlay retry has to force `muxDataKeepSession` and re-hook hls.js on the live monitor to avoid duplicate views, which is the same mechanism this decision makes the default.
+
+Three properties of the contract shaped the details:
+
+- The reset is complete, so `videochange` carries the full video metadata every time, not a diff.
+- `viewstart` waits for `play`. That is safe on `HTMLMediaElement` hosts because the load algorithm sets `paused` on every `load()`, so a source change always yields a fresh `play`. Integrations whose pipeline can swap sources without a paused transition (THEOplayer) synthesize `play` themselves. An SPF media (#1845) must either preserve that transition or do the same.
+- A cleared source is not a video change. mux-embed's element monitor does not listen to `emptied` or `abort`; the load algorithm's `pause` simply idles the view. So a cleared source is neither reported nor tracked, and the next real source is compared against the last video the monitor was told about.
+
+## Sources
+
+- Mux Data, Monitor HTML5 Video Element, "Changing the video": https://www.mux.com/docs/guides/monitor-html5-video-element
+- mux-embed SDK behavior (`videochange` resets the view and arms `viewstart` on `play`; the element monitor's event list), as shipped in the `mux-embed` npm package: https://www.npmjs.com/package/mux-embed
+- `playback-core` destroy-per-source and its `muxDataKeepSession` workaround: https://github.com/muxinc/elements/blob/e1a32eb3fb147fcbd42f810ecf54101f4582f868/packages/playback-core/src/index.ts#L671-L679 and https://github.com/muxinc/elements/blob/e1a32eb3fb147fcbd42f810ecf54101f4582f868/packages/playback-core/src/index.ts#L697-L740
+- Bug and fix: https://github.com/videojs/v10/issues/2562, https://github.com/videojs/v10/pull/2565. Inheriting work: https://github.com/videojs/v10/issues/1845
+- Implementation: [`mux-data.ts`](/packages/media/src/dom/mux/mux-data.ts), [`mux-data.test.ts`](/packages/media/src/dom/mux/tests/mux-data.test.ts)
--- a/packages/media/src/dom/mux/mux-data.ts
+++ b/packages/media/src/dom/mux/mux-data.ts
@@ -1,7 +1,7 @@
 import Mux from 'mux-embed';
 
 import { getPlayerVersion } from './env';
-import { toMuxDataEngineOptions } from './mux-data-engine';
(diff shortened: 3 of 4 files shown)


luwes (approved):
LGTM 👍

PR #2567, opened 2026-08-31
refactor(packages)!: playback adapters packages

mihar-22:
## What this does

Splits every playback engine and extension out of `@videojs/media` into a package named for the media it plays. `@videojs/media` keeps only shared code.

## Packages

| Adapters | Extensions | Stays built in |
| --- | --- | --- |
| `@videojs/hlsjs-video` | `@videojs/mux-data` | `@videojs/spf` (HlsVideo, HlsAudio, background) |
| `@videojs/mux-video`, `@videojs/mux-audio` | `@videojs/google-cast` | `@videojs/native-hls-video` |
| `@videojs/dash-video`, `@videojs/shaka-video` | | |
| `@videojs/vimeo-video`, `@videojs/wistia-video`, `@videojs/youtube-video` | | |
| `@videojs/cloudflare-video`, `@videojs/twitch-video`, `@videojs/tiktok-video`, `@videojs/spotify-audio` | | |

## Rules

- Adapter and extension packages export the class, its props, and defaults. Nothing else.
- Each pins its engine (`hls.js`, `dashjs`, `shaka-player`, `@vimeo/player`, `@wistia/wistia-player`, `mux-embed`) as an exact dependency.
- `@videojs/html` and `@videojs/react` keep every façade entry point and list the packages as optional peers at exact versions. Users install the framework package plus the media they play.
- `@videojs/spf` and `@videojs/native-hls-video` are regular dependencies of the framework packages, so HLS and native HLS need no extra install.
- `@videojs/media` has two entry points: `@videojs/media` and `@videojs/media/dom`. The `dom/*` subpaths are gone.
- Mux identity helpers live in a private `@videojs/mux` workspace package, bundled into `@videojs/mux-video`, `@videojs/mux-data`, and `@videojs/spf` (code and declarations). Whether to publish it is a later call.

## For users

```bash
pnpm add @videojs/react @videojs/mux-video
```

```tsx
import { MuxVideo } from '@videojs/react/media/mux-video'; // unchanged
```

## Also in this PR

- Site install generator, CLI docs output, package READMEs, and reference-page Import sections name the package to install.
- The Mux SPF adapters live with their packages: `@videojs/mux-video/spf` (SPF-backed `MuxVideoMedia` plus the shared `MuxMediaMixin`) and `@videojs/mux-audio/spf` (`MuxAudioMedia`). Both depend on `@videojs/spf`; spf no longer depends on `@videojs/mux` or, through it, hls.js. The unused `@videojs/spf/mux-background-video` alias entry is gone (html and react alias `<mux-background-video>` themselves).
- `@videojs/mux-audio` exports `MuxAudioMedia`, a class extending `@videojs/mux-video`'s Media, so audio players install by the media they play. Audio extending video is upside down; a `TODO(mux)` marks the shared base to extract.

## Stack

#2598 (cdn) → #2600 (Extension naming) → #2601 (MediaExtension) → #2602 (Adapter naming) → #2603 (createMediaElement) → #2604 (createMediaComponent)

## Validation

`pnpm build:packages`, `pnpm typecheck`, `pnpm lint`, `pnpm check:workspace`, `pnpm test`, site and CLI tests.

---

> [!NOTE]
> **High Risk**
> Removes public `@videojs/media/dom/*` entry points and changes the install/peer-dependency model for every playback integration—apps must add adapter packages or builds will fail at import time.
> 
> **Overview**
> **Breaking change:** playback engines and extensions that lived under `@videojs/media/dom/*` are now standalone npm packages (`@videojs/hlsjs-video`, `@videojs/mux-video`, `@videojs/dash-video`, embed adapters, `@videojs/mux-data`, `@videojs/google-cast`, etc.). `@videojs/media` is narrowed to engine-neutral contracts, hosts, and `@videojs/media/dom` only—those `dom/*` subpath exports are removed along with bundled third-party runtimes.
> 
> `@videojs/html` and `@videojs/react` still expose the same façade import paths but depend on the matching adapter as an **optional peer**; install docs, CLI `generateReactInstallCode`, and tests now tell users to add e.g. `@videojs/hlsjs-video` or `@videojs/mux-audio` beside the framework package. `@videojs/native-hls-video` stays a direct framework dependency so native HLS does not require a separate install.
> 
> Mux is reorganized: private `@videojs/mux` holds shared URL/source helpers (bundled into public Mux packages), SPF-backed Mux Medias move to `@videojs/mux-video/spf` and `@videojs/mux-audio/spf` (removed from `@videojs/spf` exports), and `@videojs/mux-audio` introduces `MuxAudioMedia` extending the video adapter. Release-please, CI matrices, pkg-pr-new, commitlint, sandbox, and registry e2e setup are updated to build and test the new packages; e2e overlays explicitly pull in `@videojs/hlsjs-video` where registry tests need it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 92b5dc039c020f3d2b7ad845c8e4eb4d0944958f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/mux/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@videojs/mux",
```
mihar-22, line 2: Should call out this popped up as it contained core types and stuff across mux integrations. I've kept it private for now and just let the consuming packages bundle it. We can decide later what we call it and whether it needs to be public or not (unlikely yet).

site/src/utils/installation/codegen.ts
```diff
@@ -105,6 +105,24 @@ function getSkinFile(skin: Exclude<Skin, 'none'>): 'skin' | 'minimal-skin' {
...
+  const packages: Partial<Record<Renderer, string>> = {
+    dash: '@videojs/dash-video',
+    hls: '@videojs/hlsjs-video',
+    'mux-audio': '@videojs/mux-audio',
+    'mux-video': '@videojs/mux-video',
+    vimeo: '@videojs/vimeo-video',
+  };
+
+  return packages[renderer] ?? null;
+}
```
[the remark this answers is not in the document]
mihar-22: Follow-up should be fine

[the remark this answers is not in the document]
mihar-22: Fixed in 11ff2a253: `getAdapterPackage` is now a complete `Record<Renderer, string | null>`, so YouTube, Cloudflare, Spotify, TikTok, and Twitch install their adapter package, and adding a renderer without deciding what it installs is a type error. Tests cover every embed renderer and the built-in ones.

packages/google-cast/README.md
```diff
@@ -0,0 +1,28 @@
+# @videojs/google-cast
+
+The Google Cast extension for Video.js. It adds a Chromecast remote-playback route to whichever media the player is
```
cjpillsbury: nitpick(non-blocking): We may not want to use the term "route" in our docs if we're considering reworking how that works?

packages/hlsjs-video/src/globals.d.ts
```diff
@@ -0,0 +1,5 @@
+declare global {
+  const __DEV__: boolean;
```
cjpillsbury, line 2: question(non-blocking): should this be in a more shared location? (can happen later/as a followup if so)

packages/html/README.md
```diff
@@ -10,6 +10,22 @@
...
+
+```bash
+pnpm add @videojs/html @videojs/hlsjs-video
+```
+
+```ts
+import '@videojs/html/media/hlsjs-video';
+```
+
+Each media has its own package: `@videojs/hlsjs-video`, `@videojs/mux-video`, `@videojs/mux-audio`, `@videojs/dash-video`,
```
cjpillsbury: praise: 👌

packages/html/README.md
```diff
@@ -10,6 +10,22 @@
...
+`@videojs/shaka-video`, `@videojs/vimeo-video`, `@videojs/wistia-video`, `@videojs/youtube-video`,
+`@videojs/cloudflare-video`, `@videojs/twitch-video`, `@videojs/tiktok-video`, and `@videojs/spotify-audio`. Extensions
+follow the same pattern with `@videojs/mux-data` and `@videojs/google-cast`. The SPF-backed `HlsVideo`, `HlsAudio`,
```
cjpillsbury: nitpick(non-blocking): I believe we've leaned away from "SPF-backed" jargon for our docs unless/until it's relevant? That said, this is more "dev docs", so maybe isn't a thing to fret about?

packages/media/src/dom/mux/index.ts
```diff
@@ -1,20 +0,0 @@
-export type { DrmSystemConfig, DrmSystemsConfig, KeySystem } from '../../core/drm';
```
cjpillsbury, line 1: question(non-blocking): is the idea for dropping this entirely instead of migrating to the new package that we don't expect folks to be direct consumers here, so this is a bit redundant?

packages/spf/README.md
```diff
@@ -9,6 +9,24 @@
 `@videojs/spf` is a lightweight, bundle-size-optimized streaming engine for Video.js 10. It provides
 HLS playback with adaptive bitrate switching, WebVTT captions, and MSE support.
 
+## Installation
+
+Install SPF with the framework façade that exposes the HLS media component:
+
+```bash
+pnpm add @videojs/html @videojs/spf
```
cjpillsbury: nitpick(blocking): looks like the LLM did too much pattern matching here, since SPF isn't a peer dep. Easy to fix though!

packages/spf/vite.config.ts
```diff
@@ -7,6 +7,11 @@ import { cachedTaskInputs, packageTestTask, workspaceTaskDependencies } from '..
 
 const createPackConfig = (mode: PackageBuildMode): PackUserConfig => ({
   ...packageBuildConfig(mode, 'neutral'),
+  // `@videojs/mux` is a private workspace package: inline it (code and declarations) rather than depend on it.
+  deps: {
+    alwaysBundle: ['@videojs/mux'],
```
cjpillsbury: question(blocking): Did we just make SPF have a dependency on mux which has a dependency on hls.js? I think we did, and that may require some unwinding/moving. Maybe the mux "flavors" of SPF-backed adapters need to move to the new mux package? 😓 Hope I'm wrong, but this may have been an oversight in our discussion.

cjpillsbury (changes requested):
I have one blocking question/concern that we might have missed in our discussion. Otherwise this looks great (with one nitpick blocker as well).

packages/hlsjs-video/src/globals.d.ts
mihar-22: Right now we have these global types per package, but it would be nice if we could share a base! Leave it as a follow-up!

packages/google-cast/README.md
mihar-22: Will address!

packages/html/README.md
mihar-22: Going to remove this section!

packages/spf/README.md
mihar-22: Will fix!

packages/spf/vite.config.ts
mihar-22: [I left comment higher up](https://github.com/videojs/v10/pull/2567#discussion_r3921360507) that there seems to be shared Mux types/utilities across packages atm (spf, mux-video, mux-audio), and in order to move them out of `@videojs/media` I temporarily created a private internal package called `@videojs/mux` (for those utilities/types). It's bundled and not a shipped dependency.

We can decide in the future if we want to make this package public, or what we even call it, just filling a gap for now. It's only dependencies are `@videojs/media` and `@videojs/utils`.

I do imagine to avoid this dependency on SPF we'll need to move the Mux SPF adapters over to the relevant `mux-video` and `mux-audio` packages. Not sure if that creates any complexity for you.

mihar-22: Let me know what feels like the right play. Should we move those Mux adapters out of SPF now, leave as-is, or something else?

cjpillsbury: We can do this or we can migrate the Mux-specific SPF-backed adapters to the mux package - defer to you on final call there.

packages/media/src/dom/mux/index.ts
mihar-22: Yes. Nothing outside the workspace imported `@videojs/media/dom/mux` (the html and react façades were the only consumers, and they now import `@videojs/mux-video`), and we are pre-1.0, so a re-export shim would only keep the old path alive without a consumer. Dropped rather than migrated.

packages/spf/README.md
mihar-22: Fixed in 0e87c2e67: the spf README is back to its `main` version.

packages/spf/vite.config.ts
mihar-22: Done in 0e87c2e67: the Mux SPF adapters moved out of spf into their packages. `@videojs/mux-video/spf` exports the SPF-backed `MuxVideoMedia` plus the shared `MuxMediaMixin`, and `@videojs/mux-audio/spf` exports `MuxAudioMedia`; both packages depend on `@videojs/spf`. spf no longer depends on `@videojs/mux` (or hls.js through it), and the unused `@videojs/spf/mux-background-video` alias entry is gone since html and react alias `<mux-background-video>` themselves.

packages/google-cast/README.md
mihar-22: Fixed in 0e87c2e67: the README is now just "The Google Cast extension for Video.js."

packages/html/README.md
mihar-22: Removed in 0e87c2e67, along with the same list in the react README.

cjpillsbury (approved):
🚢 🇮🇹

packages/mux-audio/src/spf/media.ts
```diff
@@ -4,8 +4,8 @@
 // audio-only. Same shape as `hls-audio` importing its error surface
 // from `../hls-video`.
 
-import { HlsAudioMedia } from '../hls-audio/media';
-import { MuxMediaMixin } from '../mux-video/adapter';
+import { MuxMediaMixin } from '@videojs/mux-video/spf';
+import { HlsAudioMedia } from '@videojs/spf/hls-audio';
```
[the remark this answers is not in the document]
mihar-22: This is how it was before, we have some unwinding to do later.

PR #2573, opened 2026-08-31
Discovery: SPF-backed media on capability-manifest hosts

cjpillsbury:
Refs #2535

## Summary

Stacked exploration on #2535 asking what capability-manifest hosts buy SPF-backed media in practice: two host recompositions with measured sizes, plus demos of the two seams (element attributes, React defaults) that need updating before "dropped from the API" is fully true. Not for merge as-is — each commit records a finding for the media-api-lock conversation.

## Changes

- **Slim SPF video host** — composes the hls-video host *without* `streamType`, `live`, and `error` (the capabilities `HlsVideoMediaMixin` fully implements itself). All 1565 spf tests and typecheck pass with zero adapter changes; the entry moves 26.22 → 26.17 kB brotli (−50 B, hls.js external). The byte case for engine-backed media is ~nil; the win is one implementation per member instead of a shadowed base forwarder that can drift.
- **Composed background host** — rebuilds `BackgroundVideoHost` as `createMediaHost([seek, volume, autoplay, source])`, the closest available composition to its four properties. Its 34 tests pass, but the entry moves 15.88 → 16.53 kB brotli (+650 B), carries fourteen members the background player never reads, and re-adds `@videojs/media` to an entry deliberately free of it. Capability granularity is contract-sized, not member-sized — subtractive Medias need finer descriptors or a member-pick form.
- **Element seam** — `CustomMediaElement` now dev-warns when a declared attribute has no backing media member instead of silently degrading to plain reflection (the `disableRemotePlayback`-precedence bug shape). Full elements also trip it for `autoPictureInPicture`/`controlsList`/`loading`, which would want a `reflectOnly` marker in a real change. Tests pin the slim-host element surface.
- **React seam** — `deriveMediaDefaultProps` derives a defaults object from descriptor fallbacks, and its test pins a drift that already exists: `sourceCapability` says `preload` defaults to `'metadata'`, `hlsVideoMediaDefaultProps` says `''`. Derivation would make the drift impossible, but requires adapters to declare their owned members as descriptors (manifest-as-metadata, not only manifest-as-implementation).

<details>

- The slim video host copy-pastes `HTMLVideoElementHost`'s unmanifestable remainder (webkit presentation API, document-derived fullscreen/PiP) — it needs a reusable export from `@videojs/media` before any slim video host is real.
- `HlsVideoMediaMixin<Base extends Constructor<any>>` accepted the slim host without complaint: migration is frictionless, but typing a mixin's base requirement against capabilities is its own follow-up.
- Sizes measured the way the CI harness does: esbuild minify + brotli max, `hls.js`/`mux-embed` external, `__DEV__: false`.

</details>

## Testing

- `pnpm -F @videojs/spf test` — 1565 passed, no type errors (slim host + composed background host swapped in)
- `pnpm -F @videojs/media test` — 1161 passed with the attribute-backing warning in place
- New suites: `dom/custom-media-element/tests/attribute-backing.test.ts`, `spf .../hls-video/tests/derived-defaults.test.ts`


cjpillsbury:
The exploration has grown past the original three commits — a map of what's here now, in order:

1. `e4993c519` / `bac74e6ee` — the two measured recompositions the PR body describes (slim hls-video host −50 B, composed background host +650 B with the granularity finding).
2. `32f7d8cb3` — seam demos: `CustomMediaElement` dev-warns on declared attributes with no backing member; `deriveMediaDefaultProps` pins the existing `preload` defaults drift.
3. `40da8cdf4` / `0a88e2ce1` — the composition **replaces** the hand-rolled `BackgroundVideoHost` (per review), then drops `volumeCapability`: background video is video-only by design, and `muted` was the autoplay-policy workaround the adapter writes directly on the element — a worked example of "aren't ever" API subtraction.
4. `9eee818a3` — `PropertyConfig` gains `source: 'attribute' | 'host'`: the attribute-as-store assumption is now parameterized. Under the default, a write the attribute can't represent (`disableRemotePlayback = false` with the attribute absent) never reaches the host setter — the general shape of the precedence bug. No defaults flipped.
5. `35f6557d6` — `CustomMediaElement(tag, Host, { properties })`: injected config instead of `static properties` overrides, which are order-sensitive (first subclass to be defined freezes the shared factory prototype for its siblings — pinned in a test).
6. `302657dfe` — `MediaHostBase.bridgesTargetEvents`: target-event bridging is an assumption the hand-rolled background host deliberately lacked; the composition silently un-subtracted it and one engine error became a 424-call feedback loop with the React component's node re-fire (masked until spf's dist was rebuilt). The background host opts out. Side finding: `MediaHostConstructor` erases base statics, so the opt-out can't carry `override`.
7. `801e2a36b` — **end to end for hls-video**: the adapter declares its owned surface as descriptors (with an adapter-refined `source` descriptor encoding the `preload` deviation once), `hlsVideoMediaDefaultProps` derives from them (values pinned identical → React unchanged transitively), `HlsVideoMedia` merges the declarations into its manifest (`supportsMediaCapability` covers adapter-owned capabilities; `getMediaCapabilityEvents` gains first consumers; `engine` deliberately off-manifest as not-yet-official), and `<hls-video>` injects its property table with `source: 'host'` for `disableRemotePlayback`/`streamType`.

Suites: media 1165, spf 1571, html 452, react 538, core 1488 — all green against rebuilt dists.

PR #2581, opened 2026-09-02
refactor(skins): consolidate design system tokens and utilities

mihar-22:
Stacked on #2580.

## Summary

A review pass over the skins design system after the generated-skin parity work. Every value that varies per theme or preference now flows through a `--media-*` token with a matching Tailwind theme key, repeated selector patterns became shared utilities and variants, and the parts of the pipeline that kept those from reaching consumers were fixed along the way.

## Changes

- **Tokens.** Motion, shadow, backdrop filter, and geometry values are declared once in the theme files, aliased as Tailwind keys such as `duration-media-fast`, `scale-media-hidden`, and `rounded-media-controls`, and named after those keys. Reduced motion, reduced transparency, and forced colors override tokens in a `base.preferences` sublayer instead of per-rule variants. That also fixes scrims, dialogs, menu blur, slider previews, and the seek indicator, which still animated or blurred under those preferences.
- **Utilities and variants.** `surface-media`, `focus-ring-media`, `mask-media-volume`, `clip-media-*`, `transition-media-popup`, and friends replace repeated recipes. `media-compact`, `media-wide`, `media-highlighted`, `media-transitioning`, and `media-anchored` replace raw container, hover, starting-style, and `@supports` prefixes. Menu items gate hover highlighting on hover-capable pointers.
- **Registry.** The shared Tailwind source exports as theme keys, functional utilities, and custom variants. The html registry shares one theme stylesheet across skins, the theme item ships generated docs for public tokens, keys, utilities, and variants, and style assets emit in dependency and composition order.
- **Dev tooling.** The vjsc plugin writes a Tailwind candidate manifest into the Vite cache, aliases it as `vjsc:candidates`, keeps it watched so the dev CSS recompiles as modules record, and carries candidates across restarts. The skins playground fully reloads when workspace dist rebuilds reach its entry instead of hot swapping into a broken player context. The sandbox scans its generated Tailwind skins.
- **Fixes surfaced by the review.** The live skins' captions menu trigger has its tooltip back, the popup entry distance scales with the player, the Minimal caption offset applies from the video element, and the dead Tailwind entry, color alias, and dependencies are gone.
- **Docs.** The skins README walks the package structure, style layers, Tailwind entries, build outputs, and commands. The component skill records the shared utility and token rules.

## Testing

- `pnpm -F @videojs/skins test`: unit, build, and registry contracts, including a test that compiles every shipped Tailwind class against the exported registry theme and one that every aliased internal token is declared.
- `pnpm -F vjsc test`, `pnpm typecheck`, `pnpm check:workspace`, and `vp run validate:shadcn` for the three registries.
- Manual in the playground and sandbox: Tailwind and CSS modes across skins, the candidate manifest recompile on a cold start, the captions tooltip in the live skins, the sandbox Tailwind pages, and a single full reload when dist files rebuild.

---

> [!NOTE]
> **Medium Risk**
> Broad visual and build-output changes across all skins and Shadcn registry consumers; mitigated by contract tests and token-driven preferences, but regressions in theming or registry install order remain possible.
> 
> **Overview**
> This PR **recenters skins styling on a shared token and Tailwind design system**, then rewires builds, registries, and skin modules to use it instead of duplicated literals and ad hoc breakpoints.
> 
> **Tokens and CSS layering** move motion, shadows, radii, and hidden-state values onto `--media-*` variables with `@theme inline` aliases (`duration-media-*`, `scale-media-hidden`, etc.). Preference handling shifts into a `base.preferences` layer so reduced motion, transparency, and forced colors win without sprinkling `motion-reduce` / `opaque` on every rule. Preset-specific bases split into `base.video.css` and `base.audio.css`, and bundling picks the right entry via `skinBaseStylesheet`.
> 
> **Shared Tailwind recipes** land in `tailwind.css` as flat `@utility` and `media-*` variants (`surface-media`, `media-compact` / `media-wide`, `media-transitioning`, `clip-media-*`, …). Style modules and per-skin controls are trimmed to deltas; common chrome moves into layout modules (`controls.styles`, `audio-controls.styles`, `time.styles`, `indicator.styles`).
> 
> **Registry and packages** ship one **`_style-theme`** item (documented from `vars.ts` / `utilities.ts`); HTML skins import theme CSS before scoped `skin.css` and no longer embed full base bundles. A new registry test asserts exported theme utilities/variants match the source and compile every Tailwind class in shipped TSX.
> 
> **Dev and docs**: playground gains CSS vs Tailwind compare, RTL, richer copy report, and `tailwind.dev.css` + vjsc candidate manifest scanning; sandbox adds a generated-components `@source`. README and the create-vjsc-component skill document the pipeline. Smaller fixes include captions menu `ButtonTooltip`, live-audio controls deduplication, and token-driven play-button hidden icons.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 159089a379c8ee78f9fcceaa8cf35be33be2af67. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sampotts (commented):
Just a quick thought on this - I've not done a super thorough review... 

I’m concerned we may be sharing too much styling between the skins. While this works well for the two existing skins, it could make future skins harder to build, particularly if they need to break the mould, as we’d have to untangle those shared assumptions.

My preference would be to keep each skin largely self-contained, sharing only clear base or reset styles. For example, something like shadow-media-thumb introduces a specific visual choice into the shared layer. If a future skin uses a different thumb shadow, that abstraction starts to break down.

I can see why sharing these styles makes sense given the similarities between the current skins, but I worry it could constrain us in unexpected ways later.

packages/skins/src/styles/feedback/buffering-indicator.styles.ts
```diff
@@ -6,9 +6,9 @@ export default styles({
     root: {
       className: 'media-buffering-indicator',
       utilities: [
-        'pointer-events-none absolute inset-0 hidden place-content-center text-white',
-        'before:absolute before:inset-0 before:bg-black/35 before:backdrop-blur-sm',
-        'not-data-visible:[--media-spinner-animation:none] data-visible:grid motion-reduce:[--media-spinner-animation:none]',
+        'pointer-events-none absolute inset-0 hidden place-content-center text-media-controls-foreground',
+        'before:absolute before:inset-0 before:bg-media-scrim/35 before:backdrop-filter-media-scrim',
```
sampotts, line 10: nit: we should probably be consistent and stick with "backdrop" rather than "scrim"

decepulis (approved):

PR #2599, opened 2026-09-02
chore(cd): prepare the 10.0.0-rc.1 release

decepulis:
Moves the release train from beta to a release candidate. This PR is the release tooling plus the package READMEs that ship inside the npm tarballs; everything that deploys on merge (site copy, root README, blog post) is in #2658 and waits for rc.1 to be live.

## What this does

**1. Cuts the next release as `10.0.0-rc.1` instead of `10.0.0-beta.33`.**
release-please's `prerelease` strategy only increments the trailing number of the existing prerelease, so switching `prerelease-type` to `rc` alone would still produce beta.33. This PR sets a top-level `"release-as": "10.0.0-rc.1"`; with the `linked-versions` plugin every package (27, including the new `@videojs/cdn` and adapter packages) lands on rc.1, and the changelog range stays anchored to the beta.32 tags.

**`release-as` is temporary.** It is hard-coded for this one release and #2658 removes it once rc.1 has shipped. While it remains, every release PR proposes rc.1.

The March alpha→beta switch (#846, #850) instead reset the manifest and package versions to `beta.0`. That works, but because no `beta.0` tag existed release-please replayed history into the beta.1 changelog and emitted a broken `beta.0...beta.1` compare link, so I avoided repeating it.

`prerelease-type` is also flipped to `rc` on every package so the config reads truthfully; it has no effect while a prerelease version is already current.

**2. Fixes the npm `next` dist-tag.**
`pnpm publish` only sets `latest`, so `next` has pointed at `10.0.0-alpha.11` for the whole beta channel (and `@videojs/cli` / `@videojs/media` have no `next` at all). New `.github/workflows/npm-dist-tag.yml` moves a dist-tag onto a published version of every public workspace package, including the nested `packages/adapters/*` and `packages/extensions/*` buckets. `cd.yml` calls it after each release for `next`, and it is also `workflow_dispatch`-able.

**3. Swaps "Beta" for "Release Candidate" in the 9 package READMEs.** These are the READMEs npm shows for each package, so they have to be on main before the release PR merges.

The cd.yml YAML fix and the GitHub App attribution that used to be part of this PR landed separately in #2616 and #2632.

## After merging

1. release-please rewrites the release PR (#2617) as `chore: release main` proposing 10.0.0-rc.1 across all packages. Check that its diff is only version bumps and changelogs before merging it (see #2629 for what a stale regeneration looked like).
2. Merge the release PR. cd.yml publishes and moves `next`.
3. Merge #2658 promptly. It removes `release-as` and publishes the announcement. Until it lands, any push to main makes release-please open a release PR proposing rc.1 again; don't merge that one.

## Not done here

- The GitHub repo description still says "[Beta]"; that is a repo setting.

## Verification

- `actionlint` on `cd.yml` and `npm-dist-tag.yml`: clean.
- Both workflows and the release-please config parse.
- The dist-tag loop enumerates the same 25 public packages as cd.yml's `./packages/**` publish filter.
- Behavior of `release-as` + `linked-versions` confirmed against release-please v17 source (`strategies/base.ts` `buildNewVersion`, `plugins/linked-versions.ts` `preconfigure`).
- The `Sandbox` CI failure is pre-existing on main (same assertion, same value); see the comment below.

https://claude.ai/code/session_01JZ5RTqfnRMiiASH1T7tWVj


packages/cli/README.md
```diff
@@ -2,7 +2,7 @@
 
 [![package-badge]][package]
 
-> **⚠️ Beta** Experimental adoption in real projects.
+> **⚠️ Release candidate** Experimental adoption in real projects.
```
decepulis: Noticed you didn't change this line

packages/spf/README.md
```diff
@@ -2,7 +2,7 @@
 
 [![package-badge]][package]
 
-> **⚠️ Beta** Experimental adoption in real projects.
+> **⚠️ Release candidate** Experimental adoption in real projects.
```
decepulis: This, too

site/src/content/docs/concepts/browser-support.mdx
```diff
@@ -12,7 +12,7 @@ Support starts with an evergreen browser baseline and extends to the rendering e
 Video.js 10 currently targets **evergreen browsers**: the latest two stable versions of Chrome, Firefox, Safari, and Edge.
 
 <Aside type="note">
-  Video.js 10 is in beta. The browser support matrix may expand as we approach GA based on community feedback and usage data.
+  Video.js 10 is a release candidate. The browser support matrix may expand as we approach GA based on community feedback and usage data.
 </Aside>
```
decepulis, line 16: remove this aside

site/src/content/docs/how-to/installation.mdx
```diff
@@ -22,8 +22,8 @@ import LLMCase from '@/components/docs/LLMCase.astro';
 import CopyMarkdownButton from '@/components/CopyMarkdownButton.tsx';
 import { TabsRoot, TabsList, TabsPanel, Tab } from '@/components/Tabs.tsx';
 
-<Aside type="caution" title="Beta Software">
-Video.js v10 is currently in _beta_. The API may evolve with [feedback&#x1F64F;](https://github.com/videojs/v10/issues). See the [Changelog](https://github.com/videojs/v10/blob/main/CHANGELOG.md) for recent updates and the <DocsLink slug="concepts/v10-roadmap">Roadmap</DocsLink> for more details on what's coming.
+<Aside type="caution" title="Release Candidate">
+Video.js v10 is a _release candidate_. The API may still change with [feedback&#x1F64F;](https://github.com/videojs/v10/issues) before the stable release. See the [Changelog](https://github.com/videojs/v10/blob/main/CHANGELOG.md) for recent updates and the <DocsLink slug="concepts/v10-roadmap">Roadmap</DocsLink> for more details on what's coming.
 </Aside>
```
decepulis: We don't intend on the API changing anymore. Maybe like...


```suggestion
<Aside type="caution" title="Release Candidate">
Video.js v10 is a _release candidate_. This means we feel the API is stable and we're ready to start building with it. With your [feedback](https://github.com/videojs/v10/issues), a few docs tweaks, and maybe a bug fix or two, we're looking forward to shipping 10.0.0 very soon.

[Read the Release Candidate blog post here](todo)
</Aside>
```

decepulis: maybe make this aside not "caution", too. Something milder

decepulis:
The failing **publish** check (Preview Releases / pkg-pr-new) is not this PR's. It fails identically on `main` for #2586 and #2600: `apps/sandbox/scripts/prepare-template.ts` rewrites `apps/sandbox/package.json` (drops `@videojs/icons`, pins `vite-plus` to `0.2.8`) and pkg-pr-new's frozen `pnpm install` then rejects the lockfile mismatch. This branch only inherits that through the merge from `main`; nothing here touches the sandbox or the template script. No fix for it exists on `main` yet, so I'm not porting one.

---

.github/workflows/npm-dist-tag.yml
```diff
@@ -0,0 +1,59 @@
...
+      # they have nothing to tag.
+      - name: Move dist-tag on every public package
+        run: |
+          for manifest in packages/*/package.json; do
+            if [ "$(jq -r '.private // false' "$manifest")" = "true" ]; then
+              continue
+            fi
+            name=$(jq -r '.name' "$manifest")
+            npm dist-tag add "${name}@${VERSION}" "$TAG"
+          done
```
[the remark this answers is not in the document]
decepulis: Confirmed and fixed in 2e8ab269: the loop now globs `packages/*/package.json packages/*/*/package.json`, matching the `./packages/**` publish filter in cd.yml. Locally that enumerates the same 25 public packages cd.yml publishes.

---

.github/release-please/release-please-config.json
```diff
@@ -4,6 +4,7 @@
   "tag-separator": "@",
   "include-component-in-tag": true,
   "include-v-in-tag": false,
+  "release-as": "10.0.0-rc.1",
```
luwes, line 7: this shouldn't be needed unless you want to pin it to a specific version. if we keep this we'll need to update it each release I believe

decepulis:
**CI note: the `Sandbox` failure on 2dbc7ad6 is not this PR's.**

`suites/sandbox/tests/sandbox-skin-styling.spec.ts:93` ("react minimal … scales thumbnails in fullscreen", tailwind-from-registry and css-from-package variants) fails the aspect-ratio assertion at line 162 with `Received: 0.786` against a `< 0.02` tolerance. The [same job on main's tip 2c84bbaf](https://github.com/videojs/v10/actions/runs/34177115306/job/101909052332) fails on the same assertion with the same value. This PR touches no skin, thumbnail, or sandbox code (release-please config, two workflows, package READMEs).

No targeted fix is open yet. #2633 reworks the React thumbnail parts but is a feature change, not a fix for this assertion, so I have not ported it here. Everything else on this head is green.

---

.github/release-please/release-please-config.json
decepulis: Ah. I meant to put this in the PR body. I'm going to hard-code it for this release and take it out afterwards.

.github/workflows/cd.yml
```diff
@@ -18,6 +18,9 @@ jobs:
   release:
     runs-on: ubuntu-latest
     environment: production
+    outputs:
+      released: ${{ steps.release.outputs.releases_created }}
+      version: ${{ steps.release.outputs['packages/html--version'] }}
```
luwes, line 23: why is this needed? we already have access to those outputs without these lines

decepulis: Within the `release` job, yes: `steps.release.outputs.*` is available to every later step, and the existing steps use it that way. The new `dist-tag` job at the bottom of the file is a separate job, because it `uses:` the reusable `npm-dist-tag.yml` workflow and a reusable workflow can only be called at the job level, not as a step. Cross-job, the only way to read a step's output is `needs.release.outputs.*`, and that requires the producing job to declare `outputs:`. That is all these three lines do.

---

decepulis: lol sorry Wes. My Claude is eager to respond. 

To add additional context: we have a new `dist-tag` job to catch `@next` up to our leading edge and keep it there (it's been pointing at alpha this whole time)

luwes (approved):

Issue #1224, opened 2026-04-06
Bug: FrameworkCase Renders Non-Matching Content to the DOM


decepulis:
## Problem

`FrameworkCase.astro` renders **all** framework variants to the DOM on every page, hiding non-matching content with the `hidden` attribute. For example, visiting `/docs/framework/react/concepts/overview/` produces HTML that includes the full HTML-framework content in hidden `<div>` wrappers.

This means:
- Extra DOM nodes on every docs page that will never be seen
- Hydrated islands inside hidden FrameworkCase blocks still download and execute JS
- Wrapper `<div class="contents" hidden>` elements pollute the DOM tree
- Requires `data-search-ignore` and `data-llms-ignore` attributes to prevent downstream pipelines from picking up the hidden content

## Why it's like this

The original implementation tried `{shouldRender && <slot />}` for true conditional rendering, but this triggered an Astro bug where `client:*` hydration directives inside conditional slots caused hydration to break for the entire page. This may be related to Astro's hydration script deduplication mechanism (withastro/astro#8222, withastro/astro#5032).

The `hidden` attribute workaround was adopted to keep hydration working while still excluding content from search and LLM pipelines.

## Expected behavior

Non-matching framework content should be completely absent from the built HTML output — not rendered and hidden.

decepulis:
## Update: `Astro.slots.render()` approach reverted

#1223 attempted to fix this with `Astro.slots.render('default')` + `set:html`, bypassing the `<slot />` hydration bug. While it worked in initial testing (build passed, DOM was clean, islands count halved), it silently broke hydration on **HTML framework pages only** — React pages were unaffected.

Symptoms (reported by @mihar-22 in #1316):
- `client:*` islands fail to hydrate silently
- Framework switcher freezes the page
- Style selector renders empty
- File selector in demos breaks
- Affects all browsers, not just Safari

Reverted in #1317. The `hidden`-attribute workaround remains the only known safe approach.

### What we know now

- `{shouldRender && <slot />}` — breaks hydration (Astro consumes hydration scripts even when slot isn't rendered)
- `Astro.slots.render()` + `set:html` — breaks hydration on HTML pages only, works on React pages. Root cause unclear — may be related to how the HTML framework content uses `client:*` directives differently
- `hidden` attribute — works but pollutes the DOM with non-matching framework content

mihar-22:
[...]

PR #2645, opened 2026-09-07
feat(site): redesign docs site and reorganize doc sections

mihar-22:
## Summary

Redesigns the docs site and reorganizes its information architecture: a darker dark theme with semantic surface tokens and a swappable accent, a flat sidebar with a Guides / Components / API switcher, a type eyebrow and "Copy page" button on every page, Base UI–based pickers with real logos on the installation page whose picks live in the URL, and a simplified docs URL scheme with two content folders, `guides/` and `reference/`, with every old address redirecting from `netlify.toml`.

## Changes

- **URLs**: every hand-written page lives under `guides/` (how-to and concept alike) and every builder-backed page under `reference/components/` or `reference/api/`. Guide slugs are short nouns (`captions`, `poster`, `live-streams`, `self-hosting`, `content-security-policy`, `user-preferences`). The old `how-to/`, `concepts/`, and flat `reference/` URLs, and their `.md` twins, all redirect; deleted API pages end at an intentional 404. The CLI `docs` command and its README follow the new slugs.
- **Redirects**: `netlify.toml` is the single source of truth. The 136 rules for moved pages sit ahead of the section-wide wildcards so a page that changed section and slug lands in one hop; 33 rules that pointed at deleted pages were removed. The writing guide and `write-docs` skill say to add a rule when a slug changes.
- **Sidebar**: flat sections, no folders (the writing guide records why); framework and style as segmented controls with logos; a Guides / Components / API switcher with an animated active pill (plus a dev-only Writing Style entry). Guides are grouped into Getting Started, Frameworks, Environment, Concepts, Playback, Migrate; UI reference into Layout, Skins, Buttons, Sliders, Menus, Display, Dialogs, Behavior. Noun labels replace the "How to" phrasing. Pages may be `hidden` or absent from the sidebar. Restoring the sidebar scroll no longer moves the window.
- **Page header**: a type eyebrow ("Guide › Playback", "Reference › Buttons") derived from the content folder, with a split "Copy page" menu (View as Markdown, Open in ChatGPT, Open in Claude) in the same row. Hover anchors on prose headings. Prev/next cards show direction and title only.
- **Stability**: per-API `stability` is `unstable` in frontmatter, schema, badge, and skill reference, and any page carrying it opens with a note explaining that the API works but its shape may still change.
- **Authoring guardrails**: a content test guards the guide/reference boundary (no literal "How to" title, no how-to sections or custom-UI note on reference pages) and prints task-verb headings on Concepts pages as warnings. The writing guide and the `write-docs`, `review-docs`, and `review-branch` skills describe concept and how-to as two shapes of guide and ask for the shape before drafting and during review.
- **Agent exports**: `llms-markdown` refactor with typed code fences, flattened tabs, absolute links, `<link rel="alternate" type="text/markdown">`, per-framework `llms-full.txt`, and a dev-server `.md` middleware. The Build with AI guide documents all of it.
- **Theme**: soft dark default with a deep tone option. `--color-accent` is the swappable token; gold, red, and magenta derive from the brand colours with a lightness nudge per theme (relative `oklch`), and the appearance menu lists gold, orange, red, magenta with orange the default. The home page, caution asides, and the footer easter egg stay on literal orange. Heading tokens stay on Eurostile's licensed bold; Instrument Sans keeps semibold. Squircle corners on buttons, selects, cards, and code frames. Thin sidebar scrollbar; dark mode keeps a dark thumb so `color-scheme: dark` is respected. Theme-colour hexes live in one constant.
- **Legacy banner**: once it has scrolled out of view, later pages this session start without it, decided before first paint, so the nav and sidebar hold one position across navigations, back, and reload.
- **Search**: the DocSearch trigger looks like an input with placeholder and ⌘K / Ctrl K chips. The chips are painted before hydration from a platform flag set in the head, so hydration changes no pixels and nothing flashes between pages.
- **Code frames and API tables**: header and code share one frame with a hairline divider; API detail rows animate open (using the shared `--ease-soft` curve) and are inert when closed. Inline code has one root component, `MarkdownCode`, which also highlights when given `code`. A generic `Grid` backs link cards and related links and is available in MDX.
- **Installation page**: card pickers for framework (React, HTML, Vue, Svelte), use case, skin, and media source (URL detection, brand logos, Mux upload zone). Picks live in the URL using the CLI's vocabulary (`?preset=live-video&skin=minimal&media=hls&install-method=npm&source-url=…`) so reload, back, shared links, and framework switches keep them; defaults are omitted. The store fits skin and media to the preset and keeps the skin tier across media types. Islands hydrate against the server defaults and update, so URL picks and early clicks never cause hydration mismatches. The URL input commits on a debounce, paste, blur, or Enter, and the preview only loads complete URLs. Vue and Svelte installation pages are reachable from the picker but hidden from the sidebar.
- **Icons**: `lucide-react` replaced with local SVGs; brand logos from SVG Logos / Simple Icons (CC0), licences noted beside the assets.
- **Home page**: hero video hydrates on load; the hero reveal waits for the poster; the download-speed chart uses measured bundle sizes; the color-block easter egg is revealed by overscroll.
- **Content**: Content Security Policy, Mux Data, CDN, and the former concept pages are filed as guides; the custom element lifecycle page is folded into Build your own UI component; the Features page has noun headings; eleven guides gained Related sections; the beta blog post links to the site itself.

<details>

- Semantic tokens are runtime CSS variables bridged into Tailwind through a second `@theme inline` block; dark mode derives borders from `manila-light` at 10 percent. `--ease-soft` is the reveal curve used by the hero, the API detail panel, and the overscroll spring.
- The docs collection loader derives `type` (`guide` or `reference`) from the first path segment; there is no frontmatter override. `getDocTypeFromId` throws for pages outside a typed folder.
- `src/utils/docs/diataxis.ts` holds the content rules; its test sweeps `src/content/docs`.
- Sidebar top-level sections are the switcher; `getSectionsForGuide` returns the tab first, so `data-category` uses the section inside the tab. The `hidden` flag keeps a page buildable and in breadcrumbs / llms.txt while omitting it from the list and prev/next.
- The sidebar tab morph uses cross-document view transitions: a named pill plus per-label view-transition names with z-index stacking. `DocsSidebarRestoration` centres the active link by adjusting the aside's own `scrollTop` rather than `scrollIntoView`.
- `BannerInit` hides the legacy banner before paint when it was dismissed or when a session flag says it has scrolled out of view; `LegacyBanner` sets that flag from an IntersectionObserver.
- `PlatformInit` writes `data-platform` on `<html>` before paint; `docsearch.css` renders the key chips as pseudo-elements sharing the real key styles until DocSearch hydrates.
- `stores/installation.ts` reads the URL once when the first picker subscribes and rewrites it in place on change (`utils/installation/url-state.ts` parses and serialises with the CLI names). A `useCase` listener fits skin and media to the preset. `useSelection` wraps `useSyncExternalStore` with the default as the server snapshot. The HTML install tabs are driven by the store.
- `THEME_COLORS` in `consts.ts` is shared by `ThemeInit`, the appearance store, and the default meta tag, since ThemeInit runs before the stylesheet is available.
- `MediaSourceCase` gates installation prose on the selected media source through the store; `FrameworkCase` keeps its `hidden` approach because Astro 7 still drops the island bootstrap when the first framework block on a page is a skipped conditional slot (#1224, reproduced on this branch).

</details>

## Testing

- `pnpm -F site test` (631 tests, 44 files), `pnpm -F @videojs/cli test` (70), `pnpm -F site astro check` (0 errors), `pnpm build:site`, `pnpm -F site check:anchors`, `pnpm check:workspace`.
- Every moved page's old URL and every former `how-to/` address was simulated through `netlify.toml` against the built output: all land on a built page in one hop; deleted API pages end at the 404 guard.
- Installation URL persistence verified in Google Chrome, Chromium, and WebKit for HTML and React: picks to URL, reload, away and back, direct links, framework switch, and preset changes, with no page errors.
- Search trigger compared pixel-for-pixel before and after hydration in Chromium and WebKit (zero differing pixels); banner and sidebar positions traced across navigations in both engines.
- Verified in Chromium and WebKit (light and dark, desktop and mobile widths) via Playwright screenshots of the installation page, reference pages, moved guides, the sidebar, and the home page with a non-default accent.
- Not exercised: the DocSearch modal against a refreshed index and the live Mux upload flow (need network / auth).

---

> [!NOTE]
> **Medium Risk**
> Large URL and redirect surface plus new client URL/hydration behavior on installation; mistakes would break bookmarks and external links rather than core playback, but regression testing of redirects and pickers is critical.
> 
> **Overview**
> **Redesigns the docs site** and **reorganizes information architecture**: hand-written pages move to `guides/` (how-to and concept together); API reference splits into `reference/components/` and `reference/api/`, with **`netlify.toml`** carrying hundreds of 301s (plus `.md` twins) and intentional 404s for retired API pages. The CLI **`docs`** command and bundled slugs follow **`guides/installation`**.
> 
> **Authoring and automation** point at `writing-style/write-guides.mdx`, require Diátaxis shape checks (`pnpm -F site test diataxis`), and update review/write skills, changelog context, and API-reference sync paths. Feature doc slugs from the api-docs builder now use **`reference/api/feature-*`**.
> 
> **LLM/agent exports** refactor `llms-markdown`: shared HTML→Markdown pipeline, dev-server **`.md`** middleware, **`llms-full.txt`**, richer turndown (tabs, code fences, absolute URLs), and sitemap entries for the full indexes.
> 
> **Site UX** (per PR scope): flat sidebar with Guides/Components/API switcher, theme/accent tokens, installation pickers synced to URL query params, DocSearch trigger styling, related-links MDX folding via **`satteriRelatedLinks`**, stability badge **`unstable`**, and **`lucide-react`** dropped for local Octicon/Fluent SVGs plus CC0 brand logos. Adds homepage bundle-size snapshot tooling and minor styling guidance in **`site/AGENTS.md`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fcce1dff5d8a85e882db2ec7b09c4773ac5f6eb5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


site/netlify.toml
```diff
@@ -43,21 +43,813 @@
...
+# Docs sections were renamed: how-to → guides, reference → components. Keep every old URL (and its .md twin) alive.
+[[redirects]]
+  from = "/docs/framework/:framework/how-to/*"
+  to = "/docs/framework/:framework/guides/:splat"
+  status = 301
+
+[[redirects]]
+  from = "/docs/framework/:framework/reference/*"
+  to = "/docs/framework/:framework/components/:splat"
+  status = 301
```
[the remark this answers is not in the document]
mihar-22: Fixed in 329a5411f: explicit rules for `how-to/write-guides` and `reference/write-references` (plus their `.md` twins) now sit ahead of the wildcards and point at `writing-style/`. Note these pages are dev-only and are not emitted in production builds, so the old URLs will 404 at the correct address rather than a misleading one.

.agents/skills/write-api-reference/references/mdx-structure.md
```diff
@@ -12,14 +12,14 @@ title: MuteButton              # PascalCase component name
 frameworkTitle:
   html: media-mute-button      # HTML custom element tag name
 description: A button component for muting and unmuting audio playback
-stability: beta                  # Optional; marks an individual API as beta
+stability: experimental          # Optional; marks an individual API as less settled than the rest
```
decepulis: should we have the stability tag match what's in the sidebar?

.agents/skills/review-docs/SKILL.md
```diff
@@ -8,7 +8,7 @@ description: Review Video.js documentation without editing it. Use for accuracy,
 Treat implementation, types, tests, schemas, and generated output as factual sources.
 
 1. Identify the audience, intended outcome, artifact type, and owning source files.
-2. For site guides and concepts, read `site/src/content/docs/how-to/write-guides.mdx`. For generated reference pages, read `site/src/content/docs/reference/write-references.mdx`. Use neighboring docs for other artifact types.
+2. For site guides and concepts, read `site/src/content/docs/writing-style/write-guides.mdx`. For generated reference pages, read `site/src/content/docs/writing-style/write-references.mdx`. Use neighboring docs for other artifact types.
```
decepulis: going to disregard slugs for this review. We're still talking about that in slack.

mihar-22:
## Latest round: URL simplification and authoring guardrails

35 commits since the last push, 122 files.

**URLs**
- Two folders only: `guides/` for every hand-written page, `reference/` for builder-backed pages. The `concepts/` folder is gone.
- 17 guide slugs shortened to nouns: `captions`, `poster`, `controls`, `live-streams`, `casting`, `fullscreen`, `quality`, `thumbnails`, `preferences`, `playback-errors`, `keyboard-shortcuts`, `picture-in-picture`, `background-video`, `internationalization`, `mux-data`, `self-hosting`, `content-security-policy`.
- Renamed: `overview` → `architecture`, `security` → `content-security-policy`, `use-videojs-with-vue` → `vue`, `use-videojs-with-svelte` → `svelte`.
- Removed: the custom element lifecycle page. Its cleanup advice moved inline into Build your own UI component, and the old URL redirects there.
- Every old URL on the whole branch was simulated against the redirect rules. 172 land on a live page, 16 deleted API pages end at the intentional 404 guard, nothing loops or dead-ends. 33 dead `netlify.toml` rules removed.

**Sidebar**
- Order: Getting Started, Frameworks, Environment, Concepts, Playback, Migrate.
- Getting Started: Installation, Customize skins, Architecture, Why Video.js?, Browser support, Build with AI, Roadmap, Contribute. Build a component is hidden but still built and linked.
- Tooling renamed Environment: TypeScript, Bundlers, CDN, Self-hosting, Content Security Policy.
- Frameworks: Vue, Svelte, HTML only. Migration labels dropped "From". Customize section dissolved.
- Installation guide gained a "Coming from another player?" tip linking the four migration guides.

**Authoring and Diátaxis**
- Page type derives from the folder and shows as the eyebrow: "Guide › Playback", "Reference › Buttons".
- `redirectFrom` on a sidebar entry generates per-framework redirects for the page and its `.md` twin. Used for every move above.
- A test guards the guide/reference line: no literal "How to" title, no how-to sections or `CustomUiNote` on reference pages, every page in the sidebar, no redirected slug still live. It also prints, as warnings only, task-verb headings on Concepts pages (seven today).
- Writing guide and skills describe concept and how-to as two shapes within guides. `write-docs` asks for the shape before drafting, `review-docs` checks shape against prose, `review-branch` asks for that on docs changes.

**Content**
- Features concept got noun headings. Eleven guides gained Related sections. Inbound links, two changelog entries, and two CI scripts updated for every move.

.agents/skills/write-api-reference/references/mdx-structure.md
decepulis: nit(non-blocking): is it weird that it's `experimental` in code but `unstable` in UI?

site/integrations/docs-redirects.ts
decepulis, line 1: suggestion(non-blocking): I'm not sure I'm a fan of maintaining redirects in the sidebar. I'd prefer a single source of truth -- netlify.toml

Really, though, matter of taste. I see it both ways.

decepulis: fwiw, I haven't worried too much about maintaining redirects during the beta stage. Doing what I can to keep us from having to maintain a massive redirect table. Think we need to care during RC?

site/integrations/llms-markdown.ts
decepulis, line 1: simply lovely!

site/scripts/snapshot-bundle-sizes.mjs
decepulis, line 1: hah! love it.

site/src/assets/icons/paintbrush.svg
decepulis, line 1: nit(non-blocking): you know, this is the one icon that bothers me. Not because it's a bad icon. but because it's an unfilled icon next to two filled icons in the navbar. (and because we have a functional button next to two nav links, but tbh I don't have any better ideas)

site/src/components/blog/BlogPostCard.astro
```diff
@@ -27,7 +27,7 @@ const { post, authors } = Astro.props;
       </h2>
       {
         authors.length > 0 && (
-          <div class="mb-4 font-bold">
+          <div class="mb-4 font-semibold">
```
decepulis, line 30: question: Do we import the semibold font? I guess I'll see as I keep reading this PR.

site/src/components/Code/inlineHighlighter.ts
decepulis, line 1: it makes me feel warm and fuzzy to see that my code grows up nicely to this use case. happy to see `serverHighlighter` still living its best life.

site/src/components/docs/api-reference/AnchorLink.astro
```diff
@@ -0,0 +1,22 @@
...
+  name: string;
+}
+
+const { id, name } = Astro.props;
+---
+
+<a
+  href={`#${id}`}
+  aria-label={`Link to ${name}`}
+  class="ml-1 inline-flex size-5 items-center justify-center rounded align-middle text-muted opacity-0 transition-opacity group-hover:opacity-100 focus-visible:opacity-100 intent:text-accent"
```
decepulis, line 19: suggestion(non-blocking): I have a state called `intent` which works for both `hover` _and_ `focus-visible`. I'd love it if your agent did a quick run through to see whether it's used in the correct places in the correct tways.

site/src/components/docs/api-reference/DetailRow.astro
```diff
@@ -1,106 +1,113 @@
...
+    hasDetail && "cursor-pointer intent:bg-surface/60 dark:intent:bg-surface",
+  ]}
 >
-  <slot />
-  <Td class="align-top">
+  <tr
+    class:list={[
+      "*:px-3 *:pt-3 *:align-top sm:*:px-4",
+      description ? "*:pb-1" : "*:pb-3",
+    ]}
```
decepulis, line 41: nitpick(non-blocking): selectors like this are un-tailwindish, requiring an author to look in multiple places to understand where styles come from.

But I get that sometimes they're unavoidable. I'd prefer we relocate these styles onto the children, but, I'm not dogmatic about it.

site/src/components/docs/api-reference/DetailRow.astro
```diff
@@ -1,106 +1,113 @@
...
-              </>
-            )}
+<style>
+  [data-apiref-detail] {
+    display: grid;
+    grid-template-rows: 0fr;
+    opacity: 0;
+    transition:
+      grid-template-rows 280ms cubic-bezier(0.2, 0.8, 0.2, 1),
+      opacity 200ms ease;
```
decepulis, line 97: question: Do we have any shared animation curve tokens? I'm worried about finding this kind of definition, inline, for if/when we eventually do a brand pass with an animation opinion.

site/src/components/docs/api-reference/InlineCode.astro
decepulis, line 1: suggestion(non-blocking): can this practically be made more generic? We already have inline code concepts in the typography folder -- this component even draws on them. Could we make changes there at the root instead of here at the leaves?

site/src/components/docs/api-reference/SimpleRow.astro
```diff
@@ -0,0 +1,27 @@
...
+  id?: string;
+  name: string;
+  description?: string;
+}
+
+const { id, name, description } = Astro.props;
+---
+
+<tbody id={id} class="border-t border-line">
+  <tr class="*:px-3 *:py-3 *:align-top sm:*:px-4">
```
decepulis: ditto similar comment elsewhere about tailwind-ish idioms

site/src/components/docs/DocsLinkCards.astro
decepulis, line 1: suggestion(non-blocking): something I've found enormously helpful in other projects? a generic grid component that you can use for multiple purposes in markdown, like [this](https://github.com/decepulis/portfolio-svelte-site/blob/87384c813e94bf6d533b0249cbe35edce8740f4f/src/lib/markdoc/components/Grid.svelte). Then we could use it for, say, images! Neat! 

You'd still have to solve how to unwrap with `contents`, but... if there's a quick, elegant way of doing that while still keeping around a generic grid component, I'm on board.

site/src/components/docs/DocsNavigation.astro
```diff
@@ -45,17 +54,17 @@ function getGuideUrl(framework: SupportedFramework, slug: string): string {
...
       <a
         href={getGuideUrl(framework, next.slug)}
-        class="grid items-center gap-x-2 rounded-xs border border-manila-dark p-4 text-right dark:border-warm-gray intent:bg-manila-75 dark:intent:bg-warm-gray"
+        class="grid items-center gap-x-2 rounded-lg corner-squircle border border-manila-dark p-4 text-right dark:border-line intent:bg-manila-75 dark:intent:bg-warm-gray"
         style="grid-template-areas: 'direction empty' 'title icon';grid-template-columns: minmax(0,1fr) auto; grid-template-rows: auto minmax(0,1fr);"
       >
-        <div class="text-p3" style="grid-area: direction;">
-          Next
+        <div class="text-p4 text-muted" style="grid-area: direction;">
+          {getSection(next.slug) && <>{getSection(next.slug)} · </>}Next
```
decepulis: nit(non-blocking): 

idk. sometimes, i feel like section here just adds visual noise. (Especially with the mental hurdle of the logical but confusing next/previous placement.)

other times, I see its value. 

What do you think?

site/src/components/docs/DocsSidebar.astro
```diff
@@ -1,30 +1,145 @@
...
 type Props = {
   framework: SupportedFramework;
   docTitles: Map<string, string>;
   docStabilities: Map<string, DocStability | undefined>;
   class?: string;
+  /**
+   * Give the active link's indicator a view-transition name so it slides between pages of the same section. Only one
+   * rendered sidebar per page may opt in; duplicate names would cancel the transition.
+   */
+  animateIndicator?: boolean;
```
decepulis, line 23: question: Do we have pages with multiple sidebars...? or is this premature abstraction?

site/src/components/docs/DocsSidebarRestoration.astro
```diff
@@ -28,32 +28,25 @@ const { docsSidebarId } = Astro.props;
...
-            document
-              .querySelectorAll(`#${docsSidebarId} details[id]`)
-              .forEach((details) => {
-                const savedState = state.detailsState[details.id];
-
-                if (savedState !== undefined) {
-                  details.open = savedState;
-                }
-              });
-          }
```
decepulis, line 42: suggestion(non-blocking): if we're killing folders, let's make that a more-durable decision with some reasoning in a skill or something. I don't want to thrash back and forth next time we think of a reason to add them back.

site/src/components/docs/RelatedLinks.astro
decepulis, line 1: suggestion(non-blocking): if you take my other suggestion of adding a `Grid` component, most of this could be achieved with that. Sure you'd lose some of the special headers, but... might be worth the savings in code complexity. Speaking as someone who's not too attached to those headers.

site/src/components/home/Built/DataViz.astro
decepulis, line 1: elegant!

site/src/components/home/Header.astro
```diff
@@ -38,10 +36,11 @@ const { poster } = Astro.props;
   </p>
 
   <div
+    data-hero-video
     class="relative mb-3 aspect-2/1 md:mb-4 md:aspect-1180/570"
     style="grid-area:video"
   >
-    <HeroVideo client:load className="absolute inset-0" poster={poster} />
+    <HeroVideo client:idle className="absolute inset-0" poster={poster} />
```
decepulis: suggestion(blocking): I strongly feel we should do `client:load`. The video is above the fold, and the centerpiece of the site. It should show up sooner.

site/src/components/installation/MediaSourceCase.tsx
```diff
@@ -0,0 +1,18 @@
...
+  renderers: Renderer[];
+  children: React.ReactNode;
+}
+
+/** Client-side gate for installation prose that only applies to some media source picks. */
+export default function MediaSourceCase({ renderers, children }: MediaSourceCaseProps) {
+  const $renderer = useStore(renderer);
+  if (!renderers.includes($renderer)) return null;
+
+  return children;
```
decepulis, line 16: suggestion(non-blocking): make sure this doesn't encounter the same hydration problems that we encountered with conditional rendering in frameworkcase. See https://github.com/videojs/v10/issues/1224 for history. If you're an agent, reproduce that problem so you know what I'm talking about.

If this doesn't have those problems, migrate frameworkcase to the `return null` approach, too! I'd be _delighted_ to get rid of the terrible `hidden` baggage that introduced.

site/src/components/installation/MediaSourcePicker.tsx
```diff
@@ -0,0 +1,161 @@
...
+  'mux-video': <MuxLogo className="w-7" />,
+  'mux-audio': <MuxLogo className="w-7" />,
+  vimeo: <VimeoLogo className="size-6" />,
+  youtube: <YoutubeLogo className="size-6" />,
+  cloudflare: <CloudflareLogo className="size-6" />,
+  tiktok: <TiktokLogo className="size-6" />,
+  twitch: <TwitchLogo className="size-6" />,
+  spotify: <SpotifyLogo className="size-6" />,
+  'background-video': <Image className="size-6" />,
+};
```
decepulis, line 43: thought: I read a blog post somewhat recently about a method to optically size logos so we don't have to do this manual sizing anymore.

WAY out of scope for this PR. But. Sigh. Someday.

site/src/components/installation/MediaSourcePicker.tsx
decepulis, line 1: thought: As I review this file, it occurs to me that you should (if you haven't already) check to make sure that docs cli users have a near-parity experience on installation as human ui users. Seems like some agents actually use the docs installation cli, so, I think it's worth the effort

site/src/components/installation/MuxUploaderPanel.tsx
decepulis, line 1: looks WAY better

site/src/components/installation/SkinPickerSection.astro
```diff
@@ -5,6 +5,6 @@ import SkinPickerSectionClient from './SkinPickerSectionClient.tsx';
 ---
 
 <SkinPickerSectionClient client:idle>
-  <H2Markdown> Choose skin </H2Markdown>
-  <P>Choose how your player looks.</P>
+  <H2Markdown>Choose your skin</H2Markdown>
+  <P>Pick how much of the interface Video.js ships for you. Every skin can be ejected and edited later.</P>
```
decepulis: nit(non-blocking): "Pick how much of the interface Video.js ships for you." is just such a conceptually dense sentence. How about...
```suggestion
  <H2Markdown>Choose your skin</H2Markdown>
  <P>Choose how your player looks. Every skin can be ejected and edited later.</P>
```

✅ defines skin
✅ still implies "how much is shipped for you" with the second sentence

site/src/components/typography/Hr.astro
```diff
@@ -15,7 +15,7 @@ const { as: Tag = 'hr', maxWidth = true, class: className, ...props } = Astro.pr
 <Tag
   class={twMerge(
     clsx(
-      "my-8 border-t border-manila-75 dark:border-warm-gray",
+      "my-8 border-t border-manila-75 dark:border-line",
```
decepulis: question: is `line` a new color? Or is it a semantic class that communicates that all lines are this color? if the latter, then this should be `my-8 border-t border-line`, right? And you should check all usages of `-line` for this.

site/src/components/AppearanceMenu.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Swatch colours are fixed brand values rather than theme tokens, since the accent token itself changes with the
+ * selection and a swatch must keep showing its own colour.
+ */
+const ACCENT_SWATCHES = {
+  orange: { label: 'Orange', color: '#ff6200' },
+  gold: { label: 'Gold', color: '#e08a00' },
+  magenta: { label: 'Magenta', color: '#cc3566' },
+  red: { label: 'Red', color: '#eb3132' },
```
decepulis: suggestion(non-blocking): is it possible to access these colors via `var` so we have a single source of truth for color tokens -- tailwind?

site/src/components/Aside.astro
```diff
@@ -36,37 +39,42 @@ const renderTitle = title ?? defaultTitles[type];
...
+    <div class="mt-1 shrink-0">
       <Icon
-        size={18}
         className={clsx(
+          "size-4",
           clsx(
             type === "note" && "text-magenta",
             type === "tip" && "text-magenta",
-            type === "caution" && "text-orange",
+            type === "caution" && "text-accent",
```
decepulis: suggestion(blocking): oop this was a mistake. This is supposed to be orange, not accent

decepulis: appears a few times in this file

site/src/components/FooterEasterEgg.astro
```diff
@@ -1,22 +1,29 @@
...
     className,
   ]}
   style="grid-template-rows: 80fr 60fr 45fr 20fr 10fr"
+  {...rest}
 >
   <div class="bg-bright-yellow dark:bg-gold"></div>
-  <div class="bg-gold dark:bg-orange"></div>
-  <div class="bg-orange dark:bg-red"></div>
+  <div class="bg-gold dark:bg-accent"></div>
+  <div class="bg-accent dark:bg-red"></div>
```
decepulis: suggestion(blocking): this is supposed to be orange, not accent

site/src/components/Tabs.tsx
decepulis, line 1: thought: it took me a huge amount of work to make sure this worked consistently across all of astro's quirks. in my brief testing, it seems like tabs still work in your PR. but. I hope you made these changes intentionally and cautiously.

site/src/components/Tabs.tsx
```diff
@@ -278,9 +273,27 @@ interface TabsPanelProps {
   className?: string;
   variant?: TabsVariant;
 }
+
+/**
+ * Collapsed height for long code. Content that only slightly exceeds the cap is shown in full, since a "Show more"
+ * button that reveals a couple of lines is more annoying than the extra height.
+ */
+const COLLAPSED_MAX_HEIGHT = 512;
+const COLLAPSE_SLACK = 96;
```
decepulis, line 282: nit(non-blocking): is this a `tabs` concern? I guess it could be, considering we use `tabs` as our code frame...

site/src/content/blog/2026-03-10-videojs-v10-beta-hello-world-again.mdx
decepulis, line 1: suggestion(non-blocking): this is out of the scope of your PR but
1. check out my recent PR for framework-agnostic docs URLs
2. noticing some use of our netlify url; we should change those to relative URLs

site/src/content/docs/concepts/custom-element-lifecycle.mdx
decepulis, line 1: ha

site/src/content/docs/guides/architecture.mdx
```diff
@@ -194,4 +194,11 @@ You can use it from `@videojs/react/background`.
...
 
-<DocsLinkCard slug="concepts/presets">Learn more about presets</DocsLinkCard>
+<DocsLinkCard slug="guides/presets">Learn more about presets</DocsLinkCard>
+
+## Related guides
+
+- <DocsLink slug="guides/installation">Installation</DocsLink>
+- <DocsLink slug="guides/features">Features</DocsLink>
+- <DocsLink slug="guides/skins">Skins</DocsLink>
+- <DocsLink slug="guides/presets">Presets</DocsLink>
```
decepulis, line 204: thought: One reason I've always been hesitant about these kinds of sections in docs is because they're HARD to maintain. I'm tempted to (separately) maybe generate these with semantic similarity or... something like that...?

site/src/content/docs/reference/components/dash-video.mdx
```diff
@@ -3,7 +3,7 @@ title: DashVideo
 frameworkTitle:
   html: dash-video
 description: DASH video element powered by dash.js for adaptive bitrate streaming
-stability: beta
+stability: experimental
```
decepulis: suggestion(non-blocking) We should {programmatically,manually} add a block to the top of unstable apis to explain what we mean by `unstable`. In the case of dash-video and shaka-video, the underlying playback engines are very stable. We're just still iterating on API. I'd like that communicated.

Non-blocking because this is beyond the scope of this PR

site/src/content/docs/writing-style/write-guides.mdx
```diff
@@ -29,14 +29,18 @@ Oh hey by the way. A lot of this knowledge is encoded in the `write-docs` and `r
...
 
 <Aside type="note">
 Diátaxis has a fourth mode — learning-oriented **tutorials**. We don't have a tutorials section yet — something we've intentionally chosen to omit, for now.
 </Aside>
 
+Concept pages and how-to guides share one folder and one URL prefix, `guides/`. The difference is the shape of the page, not its address. Reference has its own folder because the api-docs-builder owns its shape. The site derives that split from the path and shows it in the eyebrow above the title (for example "Guide › Playback").
+
+A unit test, `pnpm -F site test diataxis`, guards the line between the two folders: no title starts with "How to", no reference page carries how-to sections or a `<CustomUiNote />`, every page is in the sidebar, and no redirected slug still exists. Whether a guide reads as a concept or a how-to is a judgement you make while writing and reviewing, with the compass below. The test helps by printing, as warnings only, any heading on a Concepts-section page that opens with a task verb such as "Add" or "Configure".
```
decepulis: suggestion(non-blocking)

> every page is in the sidebar

I actually had this vague intention that one could create a page that's not in the sidebar but still accessible. like https://www.mux.com/docs/guides/spaces-to-livekit

I don't remember if I implemented this tbh. If I did, we should remove this phrase

site/src/content/docs/writing-style/write-references.mdx
```diff
@@ -148,12 +148,12 @@ title: PlayButton
...
   html: media-play-button
 description: A button component for playing and pausing media playback
-stability: beta
+stability: experimental
 ---
 ```
 
 Use `frameworkTitle` to show the HTML custom element tag name when the HTML framework is selected.
-Add `stability: beta` when the individual API is available for testing but may still change. The page header and sidebar display the status automatically.
+Add `stability: experimental` when the individual API is available for testing but may still change. The whole of v10 is in beta, so "beta" on one page says nothing; "experimental" marks the APIs that are less settled than the rest. The page header and sidebar display the status automatically.
```
decepulis: thought: this line will go stale quickly as we exit beta

site/src/layouts/Base.astro
```diff
@@ -153,18 +161,30 @@ const fullTitle =
   <body
     class:list={[
       "relative",
-      "font-sans text-p2 text-pretty antialiased",
+      "font-sans text-p2 text-pretty",
```
decepulis, line 167: nit(non-blocking): to my taste, if we want a heavier font, I prefer `font-weight: 450` + `antialiased` over removing antialiasing. But. I'd want Sam in the loop to make that decision since he's the professional designer in the room who made this call 👀

site/src/layouts/Base.astro
```diff
@@ -178,5 +198,65 @@ const fullTitle =
...
+
+                clearTimeout(releaseTimer);
+                releaseTimer = setTimeout(release, 180);
+              },
+              { passive: true },
+            );
+          })();
+        </script>
+      )
+    }
```
decepulis, line 264: thought(non-blocking): I wonder how many react components we could convert to astro+script. I initially made a rule that all client-side functionality should be astro for maintainability, assuming that human contributors would be maintaining the code. Now that that's not true, I don't mind relaxing that restriction.

Out of scope.

site/src/layouts/Docs.astro
```diff
@@ -33,19 +33,24 @@ const docsSidebarId = 'docs-sidebar';
...
   description={`${doc.data.description} — Video.js ${seoCategory}.`}
+  easterEgg="thin"
+  easterEggClass="md:left-70 lg:left-75"
 >
   <Fragment slot="priority-head">
     <StyleInit />
   </Fragment>
   <Fragment slot="head">
     <DocsSidebarRestoration docsSidebarId={docsSidebarId} />
+    <MarkdownAlternate />
```
decepulis, line 44: question: is `MarkdownAlternate` uniformly applied on all pages that offer markdown alternates? My DRY sense is tingling.

site/src/layouts/Docs.astro
```diff
@@ -33,19 +33,24 @@ const docsSidebarId = 'docs-sidebar';
...
     <slot name="head" slot="head" />
   </Fragment>
   <div
     class:list={[
       "grid min-h-screen grid-cols-1",
       "md:grid-cols-(--md-grid-cols) lg:grid-cols-(--lg-grid-cols)",
```
decepulis, line 50: thought(non-blocking): I used to feel strongly that we should reduce the use of tailwind arbitrary variants with this pattern. A generic classname and the specifics in style.

Sam and I recently did some performance testing though and found that this pattern isn't worth the squeeze.

Out of scope. Just saying things out loud so there's a chance we come back to them.

site/src/pages/blog/[...slug].astro
```diff
@@ -54,6 +55,7 @@ const jsonLdSchema = createBlogPostingSchema({
   image={post.data.ogImage}
   twitterImage={post.data.twitterImage}
 >
+  <MarkdownAlternate slot="head" />
```
Comment (decepulis) [inline site/src/pages/blog/[...slug].astro:58]: Ah. I take back my DRY title comment from earlier

site/src/pages/blog/[...slug].astro
```diff
@@ -67,15 +69,18 @@ const jsonLdSchema = createBlogPostingSchema({
...
         <FormattedDate date={post.data.pubDate} />
         <span class="hidden sm:inline">&bull;</span>
         <span data-search-ignore data-llms-ignore
           >{remarkPluginFrontmatter.minutesRead}</span
         >
       </div>
-      <h1 class="mb-2 font-display text-h2 leading-tight uppercase sm:text-h15">
+      <h1
+        class="mb-3 font-display text-h2 leading-tight uppercase text-balance sm:text-(length:--blog-title)"
+        style="--blog-title: 2.75rem"
```
Comment (decepulis) [inline site/src/pages/blog/[...slug].astro]: suggestion(non-blocking): feels too big to me. imo bring back the old size

site/src/stores/appearance.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export type ThemePreference = (typeof THEME_PREFERENCES)[number];
+export type Accent = (typeof ACCENTS)[number];
+export type Tone = (typeof TONES)[number];
+
+/** Page background per resolved theme and tone, mirrored into the `theme-color` meta tag for browser chrome. */
+const THEME_COLORS = {
+  light: '#ebe4c1',
+  deep: '#151414',
+  soft: '#1e1d1d',
```
decepulis: suggestion(non-blocking): similar `var` comment to earlier. would rather all design tokens live in one place, to the extent that's possible. 

Seeing this is related to `theme-color`, so, maybe not possible. But we should leave a comment so the connection to the source of truth is clear.

site/src/styles/globals.css
```diff
@@ -118,32 +120,32 @@
   --text-h1: 3.75rem; /* 60px */
   --text-h1--line-height: 1.1;
   --text-h1--letter-spacing: var(--tracking-tightest);
-  --text-h1--font-weight: var(--font-weight-bold);
+  --text-h1--font-weight: var(--font-weight-semibold);
```
decepulis: suggestion(blocking): I know for a fact we didn't buy this font in semibold. We should make sure that css adheres to only weights that are loaded

site/src/styles/globals.css
```diff
@@ -187,6 +194,74 @@
...
+}
+
+.dark {
+  --line: color-mix(in srgb, var(--color-manila-light) 10%, transparent);
+  --line-strong: color-mix(in srgb, var(--color-manila-light) 18%, transparent);
+  --surface: var(--color-soot);
+  --surface-raised: var(--color-warm-gray);
+  --muted: color-mix(in srgb, var(--color-manila-light) 60%, transparent);
+  --hover: var(--color-warm-gray);
+}
```
decepulis, line 233: suggestion(non-blocking): Big change but I'm on board! As I've mentioned elsewhere, we should just make a quick pass to make sure that these are uniformly applied

site/src/styles/globals.css
```diff
@@ -250,6 +325,43 @@
   }
 }
 
+.dark * {
+  scrollbar-color: var(--line-strong) transparent;
+
+  &::-webkit-scrollbar-thumb {
+    background: var(--line-strong);
+  }
+}
```
decepulis, line 347: suggestion(non-blocking): It's too bad you didn't like my manilla scroll bars in dark mode! I'll miss them.

site/src/styles/globals.css
```diff
@@ -250,6 +325,43 @@
...
+    --radius-sm: 0.35rem;
+    --radius-md: 0.5rem;
+    --radius-lg: 0.7rem;
+    --radius-xl: 1rem;
+    --radius-2xl: 1.375rem;
+    --radius-3xl: 2rem;
+
+    corner-shape: squircle;
+  }
+}
```
decepulis, line 376: niiiice

site/src/styles/globals.css
```diff
@@ -281,7 +393,7 @@
   }
 }
 
-@utility horizontal-bars-orange {
+@utility horizontal-bars-accent {
   background-image: url("/horizontal-bar-orange.svg");
 }
```
decepulis, line 411: suggestion(non-blocking): imo it's more correct to keep these utilities named `orange`

site/src/styles/shiki-transformers.css
```diff
@@ -89,3 +89,19 @@
...
   border-radius: 0;
 }
+
+/*
+ * Inline snippets highlighted by `highlightInline` carry both palettes; pick the dark one under `.dark`. They render
+ * at the table text size rather than the relative inline-code size so the token colors stay legible.
+ */
+[data-code-inline] {
+  font-size: var(--text-p3);
+}
```
decepulis: suggestion(non-blocking) I prefer to size code with `em`s so the utilities survive more situations. Maybe you'd get the same visual intent with a more-durable `font-size: 1em`?

site/src/styles/globals.css
```diff
@@ -102,6 +102,8 @@
   --color-bright-yellow: #ffca18;
   --color-gold: #ffa81b;
   --color-orange: #ff6200;
+  /* The accent follows the appearance menu (see `html[data-accent]` below); orange stays the literal brand colour. */
+  --color-accent: var(--color-orange);
```
decepulis, line 106: suggestion(blocking): hmm some color pairings on the home page look pretty bad. I think we should use orange on the home page, instead of the accent color... docs should be safer tho


Image

site/src/utils/docs/diataxis.ts
decepulis, line 1: huh. I don't know what to think of this! Ready to try it out, though! And thanks for adding it.

site/src/utils/satteriRelatedLinks.ts
decepulis, line 1: this already partially addresses my concern about related link section maintainability. clever!

site/src/content.config.ts
```diff
@@ -82,11 +82,12 @@ const docs = defineCollection({
       // Get updatedDate from git history
       const updatedDate = entry.filePath ? await gitService.getLastModifiedDate(entry.filePath) : null;
 
-      // Return transformed entry with added field if updatedDate exists
+      // The folder decides the Diátaxis type; it overrides anything an author writes in frontmatter.
```
decepulis, line 85: question: Is that still true?

site/src/styles/globals.css
```diff
@@ -187,6 +194,74 @@
...
+  --color-accent: #e0608c;
+}
+
+html[data-accent="red"] {
+  --color-accent: var(--color-red);
+}
+
+html.dark[data-accent="red"] {
+  --color-accent: #ff5d5e;
+}
```
decepulis, line 263: suggestion(non-blocking)

I see the intent here. You seem to want gold to look darker in light mode (#a86a00) and magenta and red to look lighter in dark mode (#e0608c and #ff5d5e0). I don't love the colors you ended up with; the gold feels muddy and the red and magenta have this pastel feeling that goes a bit off brand. 

Did some experimentation... and... I think we can still achieve what you're looking for, but with a slightly more conservative take. And we can use CSS to make sure we maintain an in-code association with the brand.

Here's a palette I generated, aiming for an APCA lightness contrast of 35, suitable for the design elements we're altering
```suggestion
html[data-accent="gold"] {
  --color-accent: oklch(from var(--color-gold) calc(l - 0.06) c h);
}

html.dark[data-accent="gold"] {
  --color-accent: var(--color-gold);
}

html[data-accent="red"] {
  --color-accent: var(--color-red);
}

html.dark[data-accent="red"] {
  --color-accent: oklch(from var(--color-red) calc(l + 0.03) c h);
}

html[data-accent="magenta"] {
  --color-accent: var(--color-magenta);
}

html.dark[data-accent="magenta"] {
  --color-accent: oklch(from var(--color-magenta) calc(l + 0.07) c h);
}
```

site/src/styles/globals.css
decepulis: nit(non-blocking) I don't remember if I said this out loud, but I don't love the use of `muted` to establish text hierarchy. mostly because the gray looks awkward on the manila background. I prefer our previous approach of establishing hierarchy just through size (though it was limiting).

decepulis (commented):
believe it or not I reviewed every gosh darned file in this pr. (except the tests lol)

a few blocking suggestions, a few non-blocking. still approving since I trust you to carry that over the line and I don't want you to wait on me to approve.

----

also got an error in dev mode while working on the home page:
```
[plugin:astro] No guide found with slug &quot;reference/feature-playback&quot;
```
idk if that's preexisting or not

site/src/styles/globals.css
decepulis: Looks a little better if we use the palette I suggest in the other comment. Still worth reconsidering.

site/src/styles/globals.css
decepulis: suggestion(non-blocking) If you adopt this change, I'd also swap the order around in the palette. gold/orange/red/magenta, even if orange is default.

site/src/layouts/Base.astro
sampotts: I don’t think it’s just a font-weight issue. Forcing `-webkit-font-smoothing: antialiased` can make text look softer/fuzzier and less crisp, particularly on Retina/high-DPI displays and even on older displays (I hung onto Cinema Displays for ages) I tended to leave it disabled for similar reasons. 

Modern displays already have enough pixel density for the OS/browser to rasterize text cleanly without us forcing a particular smoothing mode. The default rendering tends to preserve sharper glyph edges and better perceived definition, whereas antialiased can introduce more partially transparent edge pixels, making text feel slightly blurred as well as thinner (depending on the typefaces hinting of course).

Personally, I’d rather leave font rendering to the platform and use the actual font weight/typography settings to control how the type looks. That said, some typefaces have better hinting than others and some actually look better with it forced.

site/src/components/home/Header.astro
mihar-22: Done in 1ef745a49 — the hero video now hydrates with `client:load`.

site/src/components/Aside.astro
mihar-22: Good catch. All three caution spots (ring, stripe, icon) are back on orange in be7a171f3.

site/src/components/FooterEasterEgg.astro
mihar-22: Fixed in be7a171f3 — gold/orange in light, orange/red in dark, as before.

site/src/styles/globals.css
mihar-22: Agreed. The home page is back on literal orange everywhere (Get Started band, built-different bar, roadmap, sponsors, data bars) in be7a171f3; docs keep the accent. The only thing still following the accent on home is the nav logo, since the nav is shared — happy to pin that too if you'd prefer.

site/src/styles/globals.css
mihar-22: Renamed back to `data-bar-orange` / `horizontal-bars-orange` in be7a171f3.

site/src/styles/globals.css
mihar-22: Checked what actually loads: Instrument Sans comes in as a variable font (400–600) via Astro's font provider, so `font-semibold` on body text, cards, and the sidebar is covered. The real issue was the `--text-h*` weight tokens resolving to semibold while nearly all of them sit on Eurostile Extended, which we only have in bold. Those are back on bold in dfe3005f5, plus `Th`, which used semibold on the display face directly. Browsers were already snapping 600 to the bold face there, so nothing should look different.

site/src/components/blog/BlogPostCard.astro
mihar-22: Yes for this one — Instrument Sans loads 400–600 as a variable font, so semibold here is real. Eurostile is bold-only; headings are back on bold, see the thread on globals.css.

site/src/styles/globals.css
mihar-22: Adopted as suggested in c67ecf3db — the three variants now derive from the brand tokens with just a lightness nudge, and the menu order is gold / orange / red / magenta with orange still the default. Thanks for doing the APCA legwork.

site/src/components/AppearanceMenu.tsx
mihar-22: Done in c67ecf3db — swatches now paint from `var(--color-gold)` etc. Side effect worth knowing: the gold swatch shows the real brand gold now rather than the darker amber it was using for the swatch only.

site/src/stores/appearance.ts
mihar-22: Went with the comment route in c189ea687: ThemeInit runs in the head before the stylesheet is available (in dev, globals.css lands after it), so it can't read the tokens at that point. The hexes now live once in `consts.ts` as `THEME_COLORS`, shared by ThemeInit, the appearance store, and the default meta tag, with a note pointing at `--color-faded-black` and the deep-tone override in globals.css.

site/src/styles/globals.css
mihar-22: Fair on both. Plan: keep `muted` for metadata only (eyebrows, captions, timestamps) and stop using it on anything a reader actually has to read, then do the uniform text-token pass as a follow-up once this review settles, so it doesn't balloon this PR further. Will link the follow-up here.

site/src/styles/globals.css
mihar-22: Tried them back on to be sure — but manila thumbs in dark mode fight the user's own preference: with `color-scheme: dark` the browser darkens native scrollbars to match, and a light thumb sitting on dark surfaces reads as us overriding that. Keeping the subtle dark thumb. RIP manila.

site/src/pages/blog/[...slug].astro
Comment (mihar-22) [reply site/src/pages/blog/[...slug].astro]: Agreed, back to the old `sm:text-h15` size in 6e9a2a046.

site/src/styles/shiki-transformers.css
mihar-22: Good call. Went em-based in 0b755c8c8, with one tweak: a plain `1em` would have grown the snippets that sit in 16px paragraphs on component pages past normal inline-code size, so the rule is scoped to `:is(td, th) [data-code-inline] { font-size: 1em }`. Table snippets follow the cell text, prose snippets keep stepping down like regular inline code. Measured across a component page, an API page, and a guide: no rendered size changed.

site/src/components/docs/api-reference/SimpleRow.astro
mihar-22: Moved onto the cells in da7db889e.

site/src/components/docs/api-reference/DetailRow.astro
mihar-22: Did this for SimpleRow (da7db889e), but kept it here on purpose: the summary cells arrive through the slot from PropRow, StateRow, DataAttrRow, etc., so padding on the row is what lets those stay bare `<td>`s. Moving it down would mean the same padding classes in five row components. Happy to flip if you feel strongly.

site/src/components/docs/api-reference/DetailRow.astro
mihar-22: We didn't — so now we do: `--ease-soft` in the theme (2ce20a02a), used by the hero reveal, this panel, and the overscroll spring. The dial nav uses Tailwind's built-in `ease-in-out`. One place to change for a brand pass.

site/src/components/docs/api-reference/InlineCode.astro
mihar-22: Yes — folded into `MarkdownCode` in 68f990d23 via an optional `code` prop that highlights; `InlineCode` is gone and the reference rows use the typography component directly.

site/src/components/docs/DocsLinkCards.astro
mihar-22: On board. Added a generic `Grid` (fba0f1643) with a `columns` prop, registered for MDX so no import needed. `DocsLinkCards` is now `<Grid class="… [&_.docs-link-card]:contents">`, so the unwrap stays a one-liner on the wrapper and the grid itself is reusable for images etc.

site/src/components/docs/RelatedLinks.astro
mihar-22: Now built on the shared `Grid` (fba0f1643). Kept the group labels for the moment since the rest of the markup is tiny either way — easy to drop later if we decide they're noise.

site/src/components/typography/Hr.astro
mihar-22: `line` is semantic (`--line` = manila-75 in light, a translucent manila in dark). Hr now uses `border-line` (44c07b018). Swept the rest: that was the only mixed usage; the remaining `border-manila-dark` uses are deliberate strong borders on cards in light mode.

site/src/components/docs/DocsSidebar.astro
mihar-22: Real case: `Docs.astro` renders `DocsSidebar` twice — the desktop `<aside>` and the mobile drawer. Only one may own the indicator's view-transition name or Safari cancels the transition, hence the opt-in.

site/src/components/docs/api-reference/AnchorLink.astro
mihar-22: Ran the audit: every hover style on the site either goes through `intent:` or pairs `hover:` with `focus-visible:` (this permalink is the latter, since it's revealed by the row's `group-hover`). Nothing hover-only left. FYI `intent` is defined as hover + focus-within rather than focus-visible; fine for current uses.

site/src/components/Tabs.tsx
mihar-22: Intentional and cautious — the changes are the compact/expanded `variant` (nav tabs vs code frame) and the "Show more" collapse for long code panels; the id wiring, initial-tab logic, and Astro hydration handling are untouched. Tabs were re-tested across MDX, framework switching, and the installation picker.

site/src/components/Tabs.tsx
mihar-22: Fair question. It lives in `TabsPanel` because the collapse has to measure the active panel and the code frame *is* Tabs today. If we grow a proper `CodeFrame` wrapper I'd move it there. Leaving as is for this PR unless you object.

site/src/components/installation/SkinPickerSection.astro
mihar-22: Taken as suggested in ce07f9003.

site/src/components/installation/MediaSourceCase.tsx
mihar-22: Reproduced #1224 on Astro 7 to be sure: with `{shouldRender && <slot />}` the html captions and fullscreen guides lose the `astro-island` bootstrap entirely (0 of 18 / 0 of 22 islands hydrate) because the first framework block on those pages is the react one and its scripts get deduped away. So FrameworkCase keeps `hidden` for now. MediaSourceCase is a different mechanism — a React island gating on a client store, so its `return null` is ordinary React rendering and never touches Astro's slot bookkeeping. Verified on the installation page: pick HLS, the gated aside appears, island hydrated, no errors.

site/src/components/installation/MediaSourcePicker.tsx
mihar-22: Checked. `@videojs/cli docs` takes framework / preset / skin / media / source-url / install-method, which map one-to-one onto the six picks in the installation UI; the only UI-only path is the Mux uploader, which just resolves to a source URL the CLI accepts directly. It reads bundled markdown rather than the live site, so the slug moves here don't affect it, and its 70 tests pass. Bonus: the installation page now keeps its picks in the URL using the CLI's exact vocabulary (`?preset=live-video&skin=minimal&media=hls&install-method=npm`), so a link and a CLI command describe the same player.

site/src/components/installation/MediaSourcePicker.tsx
mihar-22: Someday. Noted for the pile.

site/src/assets/icons/paintbrush.svg
mihar-22: Agreed it reads lighter. Octicons only draws it as an outline; Fluent has a filled paintbrush and the icon README already allows Fluent for missing glyphs, so swapping is a two-minute change if we want it. Leaving as is for this PR.

site/integrations/docs-redirects.ts
mihar-22: You win on taste — moved to netlify.toml in 73518f5dc. All 136 generated rules now sit in the toml ahead of the section-wide rules (so a page that changed section and slug still lands in one hop), and the integration, the sidebar field, and its test are gone. Simulated every old URL through the new file after a build: all land in one hop. On the RC question: beta links to `how-to/…` are already in issues, Slack, and the blog post, so keeping them alive is cheap; we just don't grow the table beyond what already moved.

site/src/components/docs/DocsSidebarRestoration.astro
mihar-22: Recorded in the writing guide (6aaa84707): a short "Why the sidebar has no folders" section explaining the flat-list choice and what to do when a section outgrows a screen (split, don't nest).

site/src/components/docs/DocsNavigation.astro
mihar-22: You're right, it was noise. Removed the section label from the prev/next cards entirely in 1dd7342a2 — direction and title only.

site/src/content/docs/writing-style/write-guides.mdx
mihar-22: Hidden pages already work (`hidden: true` on the sidebar entry — the Vue/Svelte install pages use it), and orphan pages with no entry are now allowed too: the test that required a sidebar entry is gone and the guide says so (6aaa84707).

.agents/skills/write-api-reference/references/mdx-structure.md
mihar-22: Yes. Now `unstable` everywhere — frontmatter, schema, badge, skill reference (92e343bc8).

site/src/content/docs/reference/components/dash-video.mdx
mihar-22: Done in 92e343bc8: pages with `stability: unstable` open with an "Unstable API" caution saying the API works but its shape may still change in a minor release, and that the implementation underneath is production quality — so dash-video and shaka-video read the way you wanted. Programmatic, so any future unstable page gets it for free.

site/src/content/docs/writing-style/write-references.mdx
mihar-22: Rewritten to be evergreen in 92e343bc8: it now describes what `unstable` means rather than where v10 is in its release cycle.

site/src/content/blog/2026-03-10-videojs-v10-beta-hello-world-again.mdx
mihar-22: Relative now (7704a70e6). Framework-agnostic links are already in via your PR.

site/src/content.config.ts
mihar-22: Still true — the parser sets `type` from the folder after spreading the frontmatter, so a hand-written `type` never wins. The comment stays accurate.

site/src/layouts/Docs.astro
mihar-22: Yes: it's rendered by the three layouts that serve a markdown twin — Docs, blog post, changelog entry — and nowhere else, so nothing is missing a `<link rel=alternate>`.

site/src/content/docs/guides/architecture.mdx
mihar-22: Agreed they're a maintenance risk. Kept for now since they're short and hand-picked; generating them from similarity is a fun follow-up if they start rotting.

site/src/layouts/Base.astro
mihar-22: Agreed, and out of scope here. The installation islands genuinely need React (shared store, hydration), but several small ones (appearance menu, copy button) could be astro+script. Worth its own pass.

site/src/layouts/Docs.astro
mihar-22: Noted — same pattern is in a few places on this branch; we can sweep them later if the perf story changes.

site/src/content/docs/concepts/custom-element-lifecycle.mdx
mihar-22: 🫡

site/src/pages/blog/[...slug].astro
Comment (mihar-22) [reply site/src/pages/blog/[...slug].astro:58]: 👍

site/src/assets/icons/paintbrush.svg
mihar-22: Update: swapped it after all — Fluent's filled 24px brush (`fluent:paint-brush-24-filled`) now sits there, same weight as the Discord and GitHub marks. Credited in the icons README.

site/src/styles/globals.css
mihar-22: Done in 914b69c19. Swept all 40 `text-muted` usages: 7 were text a reader has to read (API reference descriptions, related-link descriptions, picker status/helper sentences) — those keep full contrast in light mode now, with hierarchy by size, and soften only in dark where grey reads fine. The other 33 are metadata/chrome (eyebrows, captions, group labels, key hints, placeholders, inactive nav, icons, empty-value dashes) and stay muted. Rule recorded in `site/AGENTS.md` and next to the token in globals.css so it sticks.

mihar-22:
## Review status

All 48 review threads addressed, 47 resolved with a reply each, 28 commits since the review. Preview is rebuilding and the PR description is current.

### Blocking
- Home page, caution asides, and the footer easter egg are back on literal orange; the `data-bar` / `horizontal-bars` utilities are named `-orange` again (be7a171f3)
- Hero video hydrates with `client:load` (1ef745a49)
- Heading tokens are back on Eurostile bold, the only weight we license; Instrument Sans keeps semibold since it loads 400–600 (dfe3005f5)

### Colour and type
- Accent palette adopted as suggested: gold, red, and magenta derive from the brand tokens with a lightness nudge per theme; menu order gold / orange / red / magenta with orange the default; swatches read the colour tokens (c67ecf3db)
- Theme-colour hexes live in one constant shared by ThemeInit, the appearance store, and the meta tag (c189ea687)
- Blog title back to `sm:text-h15` (6e9a2a046); inline code in tables sized with `1em` (0b755c8c8)
- `muted` is metadata-only in light mode: 7 readable spots switched to full contrast with size-based hierarchy, 33 metadata/chrome usages kept; the rule is recorded in `site/AGENTS.md` and beside the token (914b69c19, c78b95896)
- Dark scrollbars stay dark; manila thumbs fight `color-scheme: dark`

### Components
- Generic `Grid`, available in MDX, behind link cards and related links (fba0f1643)
- `InlineCode` folded into `MarkdownCode` via a `code` prop (68f990d23)
- `--ease-soft` token replaces the inline cubic-bezier in the hero, the API detail panel, and the overscroll spring (2ce20a02a)
- SimpleRow padding moved onto the cells (da7db889e); Hr uses `border-line` (44c07b018); `intent:` audit clean; DetailRow keeps its `*:` selector because its cells arrive through a slot

### Docs IA
- Redirects moved to `netlify.toml` as the single source of truth: 136 rules ahead of the section-wide wildcards, every old URL verified one-hop against a build; the integration and sidebar field are gone (73518f5dc)
- `unstable` everywhere, with an explainer note at the top of any unstable API page (92e343bc8)
- Prev/next cards dropped the section label (1dd7342a2)
- Flat-sidebar rationale recorded in the writing guide; hidden and orphan pages allowed (6aaa84707)
- Beta post links are relative (7704a70e6)

### Installation page
- Picks live in the URL using the CLI's flag names (`?preset=…&skin=…&media=…&install-method=…&source-url=…`); reload, back, shared links, and framework switches keep them (81dd907c9, 66e53105a)
- The store fits skin and media to the preset, which fixes audio picks resetting on reload; islands hydrate against the defaults so URL picks never cause mismatch warnings (60e614631)
- Section gaps evened out at 32px (d42ec8378, 0fb9eefc3); skin copy taken as suggested (ce07f9003)
- FrameworkCase keeps `hidden`: #1224 reproduced on Astro 7.1.3 and the evidence posted on the issue
- CLI parity checked: the `docs` flags map one-to-one onto the UI picks

### Found along the way
- ⌘K chips no longer flash between pages; they are painted before hydration from a platform flag (af806f956, 3e7732170)
- The legacy banner no longer shifts the nav on navigation; it hides per document once scrolled past (028de4ff9)
- Filled Fluent paintbrush in the nav (ecab84d91); home page logo pinned to orange (edcfc6202)

### Left open on purpose
- Antialiasing (`Base.astro`): Darius and Sam's call

decepulis (approved):

decepulis:
Genuinely exceptional work. I feel so much better about the docs after your hard efforts here. Thank you!!

PR #2572, opened 2026-08-31
feat(html)!: make thumbnail images composable

mihar-22:
Builds on #2568, now merged.

## Summary

Let HTML thumbnails compose like the new React API. Authors may supply a native image child, which the thumbnail element adopts and controls; without one, the element draws a fallback `` in its shadow DOM so `<media-thumbnail>` keeps working on its own.

## Changes

- Adopt a supplied light-DOM image on `media-thumbnail` and `media-slider-thumbnail`, including images added, replaced, or assigned through a forwarding slot later, and own `src` and `srcset` on it; the adopted image's attributes are observed on the image itself, so a slotted image outside the subtree cannot keep a later `srcset`
- Draw a fallback image in the shadow root only while no image is supplied, so a supplied image never sits beside a hidden one; the fallback is exposed as `::part(image)`
- Keep source selection, loading state, sprite sizing, CORS inheritance, and cleanup on the thumbnail root; a reconnect requests an update so a moved element binds its image again
- Make `<media-slider-thumbnail>` the HTML root in the VJSC target with a plain `` image part, so generated HTML skins carry the spinner as a sibling of the controlled image
- Drop the descendant loading selectors from the thumbnail styles now that both targets report `data-loading` on the root, and close the tracked gap; the gradient overlay draws from the root, since the image part is a real `` now
- Update the Sandbox E2E thumbnail assertions and reference docs for the new markup
- Render the standalone `Thumbnail.Image` as a light-DOM `` in the HTML target too, since `<media-thumbnail>` adopts it now
- Share the slot-aware image lookup: `getComposedChildren`, `findComposedElement`, and `isHTMLImageElement` join `@videojs/utils/dom`, replacing the private copies in the poster and thumbnail elements

## Testing

- `pnpm -F @videojs/utils test src/dom/tests/children.test.ts src/dom/tests/predicates.test.ts`
- `pnpm -F @videojs/html test src/ui/poster src/ui/thumbnail src/ui/slider/tests/slider-thumbnail-element.test.ts`
- `pnpm -F @videojs/skins test`
- `pnpm vp run @videojs/html#build`
- `pnpm typecheck`
- `pnpm lint`

---

> [!NOTE]
> **Medium Risk**
> Changes thumbnail/poster image adoption and generated skin HTML, which can affect custom CSS and integrations that assumed shadow-only images; behavior is covered by expanded unit and E2E tests.
> 
> **Overview**
[...]

PR #2693, opened 2026-09-09
fix(skin): style poster and slider thumbnail shadow hosts in webkit

luwes:
Refs #2563
Refs #2572

## Summary

In Safari, the poster no longer hides once playback starts and the time slider's storyboard thumbnail renders unstyled inside the preview. WebKit does not match an `@scope` rule whose subject hosts a shadow root or is slotted into one, and both `<media-poster>` (#2563) and `<media-slider-thumbnail>` (#2572) recently became shadow hosts carrying skin classes. Chromium and Firefox are unaffected.

## Changes

- Flag the poster root and image, the slider thumbnail root, image, and spinner, and the shared `previewContent` and `popup.surface` rules as `shadowHost`, so they are also emitted outside the `@scope` block for WebKit.
- The `vjsc` renderer detects shadow-host rules by the selector's subject rather than its first component, so relationship selectors such as `:where(.media-slider-preview)[data-pointing] .media-slider-preview-content` are covered.
- Shadow-host rules are repeated outside the scope rather than moved out of it.

<details>

Repeating rather than moving keeps the cascade identical in engines that match `@scope` correctly: a scoped declaration always outranks an equal-specificity unscoped one on scope proximity regardless of source order (verified in Chromium, WebKit, and Firefox), so the `:where(scope)`-prefixed copy can never beat the original. Moving would have changed tie-breaking for shared classes like `media-popup-surface` on tooltips and menus.

`shadowHost` is now documented as covering slotted light-DOM children too; the thumbnail image and spinner are not hosts themselves but are slotted through one, and WebKit skips scoped rules for them as well.

</details>

## Testing

- `pnpm -F vjsc test src/styles` (updated shadow-host compile test) and `pnpm -F @videojs/skins test`.
- Manual, via Playwright against the sandbox `/html-video/` and `/html-mux-video/` pages in WebKit, across the package, registry, and authored skin sources and the minimal skin: `media-poster` fades to `opacity: 0` after play, and `media-slider-thumbnail` is absolutely positioned at 190×106 with rounded corners, matching Chromium. Chromium output unchanged.

---

> [!NOTE]
> **Medium Risk**
> Touches global skin CSS compilation and cascade for all engines; duplication is designed not to outrank scoped rules, but styling regressions are possible if scope proximity behaves differently than expected.
> 
> **Overview**
> Fixes **Safari/WebKit** styling for `<media-poster>` and `<media-slider-thumbnail>` (and related preview/popup surfaces) after those elements became shadow hosts carrying skin classes. WebKit does not apply `@scope` rules when the subject hosts a shadow root or is slotted through one.
> 
> **Skin definitions** now mark poster/thumbnail (root, image, spinner), slider `previewContent`, and popup `surface` with `shadowHost: true` so the compiler emits the extra unscoped copy.
> 
> **`vjsc` CSS emission** changes how those rules are handled: shadow-host rules **stay inside** `@scope` and are **duplicated** outside with a `:where(.media-skin-video)` ancestor (instead of being removed from the scoped block), preserving cascade behavior in Chromium/Firefox while giving WebKit a matching rule. Detection now keys off the selector **subject** (last compound), so relationship selectors like preview → preview-content are included. Slotted `::slotted` rules are still moved out only.
> 
> Compile tests were updated to assert the repeat-outside-scope behavior and nested relationship scopes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed228ac625c51ab03ee2dfed2f86c3114cc31dc8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/skins/src/styles/layout/poster.styles.ts
+++ b/packages/skins/src/styles/layout/poster.styles.ts
@@ -4,7 +4,9 @@ export default styles({
   file: 'poster.css',
   prefix: 'media-poster',
   rules: {
+    // `<media-poster>` hosts a shadow root, and the skin's image is slotted through it.
     root: {
+      shadowHost: true,
       utilities: [
         'pointer-events-none layer-media',
         'transition-opacity duration-media-slower not-data-visible:opacity-0',
@@ -18,6 +20,7 @@ export default styles({
       },
     },
     image: {
+      shadowHost: true,
       utilities: ['layer-media object-media', '[&:not([src]):not([srcset])]:invisible'],
     },
   },
--- a/packages/skins/src/styles/popups/popup.styles.ts
+++ b/packages/skins/src/styles/popups/popup.styles.ts
@@ -36,6 +36,8 @@ export default styles({
       utilities: ['transition-media-popup data-ending-style:duration-media-instant'],
     },
     surface: {
+      // Also carried by `<media-slider-thumbnail>`, which hosts a shadow root.
+      shadowHost: true,
       utilities: 'bg-media-popover text-media-popover-foreground surface-media after:surface-media-inset',
       variants: {
         minimal: 'after:hidden',
--- a/packages/skins/src/styles/sliders/slider.styles.ts
+++ b/packages/skins/src/styles/sliders/slider.styles.ts
@@ -105,6 +105,8 @@ export default styles({
       },
     },
     previewContent: {
+      // Also carried by `<media-slider-thumbnail>`, which hosts a shadow root.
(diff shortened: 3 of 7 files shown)


decepulis (approved):
based on the fact that this fixes the problem, is easily reversible, and doesn't affect the user-facing api, I'm inclined to approve!

mihar-22 (approved):

PR #2695, opened 2026-09-09
chore(root): upgrade workspace and ci to pnpm 12

mihar-22:
## Summary

Move the workspace and CI from pnpm 11.17.0 to pnpm 12.3.4, the current `latest`. CI installs it through `pnpm/action-setup@v6.1.0`, the first release of that action that can install pnpm 12.

## Changes

- `packageManager` pins `pnpm@12.3.4`. Any pnpm since 9.7 switches to the pinned version automatically, so contributors do not need to upgrade their global install.
- pnpm 12 records its own binary and per-platform executables in a second YAML document at the top of `pnpm-lock.yaml`. A frozen install fails without it, so the lockfile changes with the pin. The dependency graph itself is untouched.
- Every workflow installs pnpm via `pnpm/action-setup@v6.1.0`. That release adds pnpm 12 support (pnpm/action-setup#288), but the moving `v6` tag still points at 6.0.10, so the patch release is pinned until the major tag catches up.
- `CONTRIBUTING.md` prerequisite bumped to pnpm ≥ 12.3.4.

<details>
<summary>Notes for reviewers</summary>

- pnpm 12 fails installs on unrecognized `pnpm-workspace.yaml` keys. Every key in ours passes.
- pnpm 12 re-keys peer variants of cyclic dependencies the first time it re-resolves. This PR leaves the lockfile as-is rather than forcing that, so the first dependency bump after it lands will carry some unrelated lockfile churn.
- The bundle-size base job on this PR runs pnpm 11 at the root and auto-switches to 12 for the PR checkout. Verified locally that 11.17 downloads and runs 12.3.4 on its own.
- Netlify picks pnpm from `packageManager`. Since 12.3.3 pnpm runs through Node when installed without lifecycle scripts, but the deploy preview is the real check.
- pnpm now recommends `pnpm/setup`, which replaces both this action and `actions/setup-node`. That restructure is out of scope; the 6.1.0 README states action-setup plus setup-node remains supported for pnpm 11 and 12.

</details>

## Testing

- `pnpm install --frozen-lockfile` in a fresh worktree under 12.3.4 succeeds and leaves the lockfile byte-identical.
- `vp run @videojs/utils#build` and `vp run @videojs/utils#test:ci` on that install: 58 files, 483 tests pass.
- Flag forms CI relies on still parse under 12: `--config.x=y`, `-C`, `--dir`, `-r --filter --fail-if-no-match`.
- `pnpm lint` is clean.
- CI on this PR exercises action-setup 6.1.0 for real.

---

> [!NOTE]
> **Medium Risk**
> Touches all CI/CD install paths and the lockfile format; a pnpm 12 install or resolver quirk could break builds or deploys until CI and Netlify validate.
> 
> **Overview**
> Upgrades the monorepo from **pnpm 11.17.0** to **pnpm 12.3.4** by updating the root `packageManager` field and refreshing `pnpm-lock.yaml` with pnpm 12’s extra lockfile document (pinned pnpm binary / platform executables) so `--frozen-lockfile` still passes—the dependency graph itself is unchanged.
> 
> Every GitHub Actions workflow that installs pnpm now uses **`pnpm/action-setup@v6.1.0`** (pinned instead of `v4` / floating `v6`) so CI can install pnpm 12. **CONTRIBUTING.md** documents the new minimum pnpm version (≥ 12.3.4).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3e1eca5e49ef4cd20c425f097398d23dc3a01c15. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


luwes (approved):
👍

sampotts (approved):

PR #2705, opened 2026-09-10
fix(skin): resolve blurry slider rendering

sampotts:
## Summary

Previously, `clip-path` did two jobs: **hide the unfilled portion** and **draw the rounded shape**. At fractional screen positions, the rounded clip antialiased the top edge. Nested clips made that edge darker still.

The fix separates those jobs:

- **A `::before` element paints each layer**—empty track, buffer, and progress—with normal `border-radius`. Its dimensions include the chapter boundaries and gaps.
- **A rectangular `clip-path` controls visibility.** It sits 1px outside the painted shape, so it doesn’t cut through the rounded edges and antialias them again.

That 1px clearance changes only the invisible clipping boundary. The painted track remains 4px tall, with its original radius. Our earlier experiment expanded the *rounded shape itself*, which is why its caps looked wrong.

The player can now retain its natural fractional height. We’re fixing how the track is painted rather than rounding the container to hide the problem.

### Before

CleanShot 2026-09-10 at 6 59 03 PM@2x

### After

CleanShot 2026-09-10 at 6 59 52 PM@2x

---

> [!NOTE]
> **Medium Risk**
> Touches shared slider styling and clip utilities used by seek and volume controls across skins; visual parity is covered by updated e2e tests but chapter/progress edge cases depend on the new layer model.
> 
> **Overview**
> Fixes **blurry seek/volume slider edges** by splitting **painting** from **visibility**: track, buffer, and progress now render on **`::before`** layers with normal `border-radius`, while parent elements use a **rectangular `clip-path` inset 1px outside** the painted shape so clipping no longer antialiases rounded caps. `clip-media-x/y` utilities compute new internal tokens **`--media-slider-layer-start`** and **`--media-slider-layer-end`** (chapter gaps included); chapter track clipping moves to **`::before`** with positioning via `left`/`right` or `top`/`bottom` instead of rounded clip-path alone.
> 
> The sandbox **drops the forced `round(nearest, …)` video height** in `PLAYER_FRAME_CLASSES` and `styles.css`, since fractional `aspect-video` height is acceptable once the slider is fixed.
> 
> E2E contracts now assert accent fill and chapter track clipping on **`::before`** pseudo-elements.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0cf8607d50852198da02ab2a3c1a5ceec9ec1352. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/e2e/suites/sandbox/tests/sandbox-skin-styling.spec.ts
+++ b/apps/e2e/suites/sandbox/tests/sandbox-skin-styling.spec.ts
@@ -68,7 +68,7 @@ for (const { platform, skin, styling, skins } of CASES) {
     const styles = await root.evaluate((element) => {
       const accent = 'rgb(18, 52, 86)';
       const fillUsesAccent = [...element.querySelectorAll<HTMLElement>('[data-orientation]')].some(
-        (part) => getComputedStyle(part).backgroundColor === accent
+        (part) => getComputedStyle(part, '::before').backgroundColor === accent
       );
       const style = getComputedStyle(element);
 
--- a/apps/e2e/suites/skin-parity/tests/vjsc-video-skin-styling.spec.ts
+++ b/apps/e2e/suites/skin-parity/tests/vjsc-video-skin-styling.spec.ts
@@ -485,7 +485,7 @@ test('React chapter segments match across styles and retain their generated rang
             orientation: element.getAttribute('data-orientation'),
             segment: getComputedStyle(element).clipPath,
             start: element.style.getPropertyValue('--media-slider-chapter-start'),
-            track: track && getComputedStyle(track).clipPath !== 'none' ? 'clipped' : 'none',
+            track: track && getComputedStyle(track, '::before').clipPath !== 'none' ? 'clipped' : 'none',
           };
         })
       )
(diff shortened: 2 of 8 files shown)


mihar-22 (approved):
I'm hesitantly approving. I'm not a fan of how much additional styling is added, particularly per track and the `clip-*` utilities have become much more complicated. It blows up a lot in the final Tailwind CSS output today because it's per track and we have to map it. Maybe I can find a way to clean it up in the compiled output later.

I can't see the difference in the images, maybe due to compression? I'll trust you to make the right call here whether it's worth it or not, maybe there's a simpler alternative? Either way, your call! :)

sampotts:
> I'm hesitantly approving. I'm not a fan of how much additional styling is added, particularly per track and the `clip-*` utilities have become much more complicated. It blows up a lot in the final Tailwind CSS output today because it's per track and we have to map it. Maybe I can find a way to clean it up in the compiled output later.
> 
> I can't see the difference in the images, maybe due to compression? I'll trust you to make the right call here whether it's worth it or not, maybe there's a simpler alternative? Either way, your call! :)

Yeah, I’m not thrilled with the extra styling either. The difference is subtle in screenshots, but visible in the browser at widths that produce fractional player heights: the track’s top edge looks soft and uneven. I'm a bit of a perfectionist when it comes to that sort of thing. I hate soft edges / half pixel stuff. 

I tried rounding the container and adjusting the clip geometry, but those either depended on the consumer’s layout or distorted the rounded ends. This keeps the rounding on the pseudo-element and moves the clip just outside it, which was the first approach that consistently fixed both.

I’ve simplified the shared positioning a little, but the compiled output is still a fair concern. I think the visual fix is worth keeping, and I’d welcome a way to reduce the generated CSS without losing it. I'll take another stab at it today and see if I can reduce it further - I couldn't yesterday but maybe with a fresh mind it'll come up with something new.

PR #2721, opened 2026-09-11
fix(skin): animate slider progress continuously across chapters

sampotts:
## Summary

Fix seek and buffer transitions that animate each chapter independently. Register the shared fill and buffer percentages with `@property` and animate them on the slider before applying chapter bounds, so progress moves continuously through the track.

## Changes

- Keep the thumb aligned with the animated fill and dragging immediate.
- Preserve the existing chapter clipping and rounded paint geometry.
- Add midpoint regression coverage for fill and buffer progression across chapters, and update audio and video transition assertions.

## Testing

- 20 focused Chromium and WebKit E2E checks passed across HTML/React and CSS/Tailwind.
- `pnpm -F @videojs/skins test`: 48 passed.
- Formatting, lint and `git diff --check` passed.

---

> [!NOTE]
> **Medium Risk**
> Touches core seek/buffer rendering and transition behavior across video/audio skins; regressions would show as janky seeks or chapter-boundary jumps, though coverage was expanded in E2E.
> 
> **Overview**
> Seek and buffer progress now animate via registered **`--media-slider-fill`** and **`--media-slider-buffer`** custom properties instead of per-layer **`inset`** transitions, so playback moves smoothly across chapter segments.
> 
> **`base.css`** adds **`@property`** rules so those percentages interpolate. **`slider.styles.ts`** moves **`transition-[--media-slider-fill,--media-slider-buffer]`** to the slider root (instant while **`data-dragging`**), drops inset animation from track layers, and stops transitioning the thumb’s **`left`/`top`** at rest while keeping pointer-based positioning during drag.
> 
> Skin-parity E2E tests assert **`--media-slider-fill`** transitions on the root, add a chaptered HLS case that fill/buffer widths stay in sync mid-animation, and refresh the shared motion contract (fill/buffer no longer **`inset`**-driven; thumb omits position properties).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4c471e5e1d493a70093c3c5d86bb11bc7a48f80b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/skins/src/styles/base.css
+++ b/packages/skins/src/styles/base.css
@@ -6,6 +6,19 @@
 @import "./themes/minimal.css";
 @import "./themes/preferences.css";
 
+/* Interpolate progress before chapter bounds are applied. */
+@property --media-slider-fill {
+  syntax: "<percentage>";
+  inherits: true;
+  initial-value: 0%;
+}
+
+@property --media-slider-buffer {
+  syntax: "<percentage>";
+  inherits: true;
+  initial-value: 0%;
+}
+
 @layer base {
   @scope (.media-skin) {
     :scope {
--- a/apps/e2e/suites/skin-parity/tests/vjsc-audio-skin-styling.spec.ts
+++ b/apps/e2e/suites/skin-parity/tests/vjsc-audio-skin-styling.spec.ts
@@ -410,14 +410,9 @@ async function audioSeekContract(root: Locator) {
   await expect
     .poll(() =>
       slider.evaluate((element) => {
-        const fills = [...element.querySelectorAll('*')]
+        const fills = [element]
           .map((target) => getComputedStyle(target))
-          .filter((style) =>
-            style.transitionProperty
-              .split(',')
-              .map((value) => value.trim())
-              .includes('inset')
-          );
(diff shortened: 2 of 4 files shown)


packages/skins/src/styles/base.css
```diff
@@ -6,6 +6,19 @@
...
+  syntax: "<percentage>";
+  inherits: true;
+  initial-value: 0%;
+}
+
+@property --media-slider-buffer {
+  syntax: "<percentage>";
+  inherits: true;
+  initial-value: 0%;
+}
```
mihar-22, line 20: nice! if this list grows we could create a new `properties.css`

mihar-22 (approved):

packages/skins/src/styles/base.css
sampotts: Yeah, good idea - I had it before but removed it in a cleanup. 

I tried as hard as I could to not require these `@property`s but out of all the options it seemed like the most sensible.

mihar-22: i like it, think it's a small perf boost too since the engine can know upfront that these are percentages and not some arbitrary value

PR #2729, opened 2026-09-11
fix(react): keep media attached when the composed ref changes identity

### LINKED ISSUE #2725: Bug: Inline `ref` on Any Media Component Detaches and Re-attaches the Engine on Every Parent Render

React (@videojs/react)

@videojs/react@10.0.0-rc.2 (main at 68bcf5b0), react@19.2.8, hls.js@1.6.15, Chromium (Edge 140 headless and Brave) on macOS 15

## Summary

Every media component composes the ref that attaches the engine with the consumer's forwarded ref through `useComposedRefs(attachRef, ref)`. The hook memoizes on the refs themselves, so the composed callback takes a new identity whenever the forwarded ref does. An inline `ref={(el) => ...}` is a new function every render, which React allows and player UIs do all the time, and React runs the cleanup of a ref callback whose identity changed before calling the new one. The cleanup composed from `useAttachMedia` is `media.detach()`, the new callback runs `media.attach(element)`. For hls.js that is `detachMedia()` (removes `src`, calls `load()`) followed by `attachMedia()`: the MediaSource is torn down and rebuilt on every parent render.

This is the churn from #1296 coming back through the other input. That fix memoized the Provider value and replaced `attachMediaElement`, the internal sources of a fresh callback ref, but the composed callback is still coupled to whatever ref the consumer passes, and the JSDoc still promises "Memoized for stable reference".

## Reproduction

Two `HlsJsVideo` elements, same Mux test stream, `autoPlay muted playsInline preload="auto"`, inside a parent that re-renders every 500 ms (a `setInterval` state tick standing in for a time display). Left: `ref={useCallback((el) => track(el), [])}`. Right: `ref={(el) => track(el)}`. The event counters are bound once per element, so the probe itself never restarts anything.

```tsx
function App() {
  const [, setTick] = useState(0);
  const videoRef = useRef<HTMLVideoElement | null>(null);

  useEffect(() => {
    const id = setInterval(() => setTick((t) => t + 1), 500);
    return () => clearInterval(id);
  }, []);

  return (
    <HlsJsVideo
      ref={(el) => { videoRef.current = el; }}
      src="https://stream.mux.com/VcmKA6aqzIzlg3MayLJDnbF55kX00mds028Z65QxvBYaA.m3u8"
      autoPlay muted playsInline preload="auto"
    />
  );
}
```

After 20 s, one player per page (headless Edge, same stream):

| | stable ref | inline ref |
|---|---|---|
| `currentTime` | 19.80 | 8.62 |
| `emptied` / `loadstart` | 0 / 1 | 39 / 40 |
| `waiting` | 0 | 39 |
| playlists / segments fetched | 2 / 12 | 3 / 467 |

Every parent render logs `emptied, loadstart, waiting, loadedmetadata, playing`. The picture blanks for the reset, playback advances at less than half speed and the CDN gets 39x the segment traffic. Side by side on one page it is the same picture (18.78 s vs 8.26 s after 20 s).

stable ref and inline ref side by side after 20 s

inline ref alone after 20 s: 39 resets, 467 segments, black frame during a reset

With the fix, same page and stream:

inline ref with the fix after 20 s: 0 resets, 12 segments

## Root cause

`packages/react/src/utils/use-composed-refs.ts`:

```ts
export function useComposedRefs<T>(...refs: OptionalRef<T>[]): RefCallback<T> {
  return useCallback(composeRefs(...refs), [...refs]);
}
```

`composeRefs` returns a cleanup whenever any composed ref returned one, and the `useAttachMedia` ref always does (`() => media.detach?.()`). React 19 treats a ref callback with a new identity as unmount plus mount of the ref: cleanup, then the new callback with the element. React 18 gets to the same place through `ref(null)` then `ref(element)`, and `useAttachMedia` maps `null` to `media.detach()`. Either way, a new identity of any one composed ref releases and rebinds all of them.

Every adapter composes `attachRef` with the forwarded `ref` this way: hlsjs-video, hls-video, hls-audio, native-hls-video, dash-video, shaka-video, mux-video, mux-audio, hls-background-video, and the embeds (youtube, vimeo, twitch, tiktok, wistia, cloudflare, spotify) through `eventsRef, attachRef, ref`.

## Fix

Branch: https://github.com/videojs/v10/compare/main...rennanribeiro:v10:fix/react-composed-ref-identity

The first version of this branch changed `useComposedRefs`; after the review below it was replaced. `useComposedRefs` is now untouched and the change lives where the churn actually comes from:

- `useAttachMedia` remembers what the adapter is attached to and only attaches when the element or the adapter actually changed, so the `null` then same-element pair React hands a composed ref that changed identity is a no-op. Attaching still happens while the refs run, ahead of any effect, exactly where it did before. A layout effect that runs after the refs of each commit detaches when the element went away with nothing to follow (and reattaches after a StrictMode simulated unmount); another detaches on unmount.
- `useAttachIframe`, a copy of the old hook used by the six embed adapters (YouTube, Vimeo, Twitch, TikTok, Cloudflare, Spotify), now delegates to it. There the churn was worse: a re-attach destroys and rebuilds the embed player.
- Reference page sentence updated to the new contract.
- Tests: nine cases in `use-attach-media.test.tsx` (mount/unmount, identity churn, attach before the mounting commit's effects, listeners composed ahead run first, element swap, adapter swap with detach-before-attach order, element removed while mounted, adapter that fails to attach, StrictMode) plus `media/tests/hlsjs-video.test.tsx` against the real `HlsJsAdapter`. Four fail on main; the `@videojs/react` suite is green (583 tests), lint, typecheck and check:workspace pass. Re-measured on the same page: inline ref now plays 19.3 s in 20 s, 0 resets, 12 segments.

Related, not in this branch: `renderElement` -> `mergeRefs` -> `composeRefs` still builds a fresh callback every render for the 36 `ref: [forwardedRef, xRef]` call sites (VolumeSlider's `wheelRef` re-adds its listener on every render), and `HlsJsAdapter.attach` forwards a repeat attach of the same element to `hls.attachMedia`, which hls.js answers with a detach and re-attach. Happy to take both next.

Happy to open this as a PR if that can be enabled, or for a collaborator to pick the branch up.

luwes:
Fixes #2725

## Summary

Every React media component composes the ref that attaches the engine with the consumer's forwarded ref. When that forwarded ref is an inline `ref={(el) => ...}`, the composed callback gets a new identity each render, and React hands the attach ref `null` then the same element again. `useAttachMedia` mapped that to `detach()` + `attach()`, so hls.js tore down and rebuilt the MediaSource on every parent render, and the iframe embeds destroyed and rebuilt their player. `useAttachMedia` now only attaches when the element or adapter actually changed.

Takes the direction of @rennanribeiro's branch in the issue (idempotent `useAttachMedia`, `useComposedRefs` untouched) with a smaller implementation: a single `sync()` reconciler and the redundant `useAttachIframe` removed.

## Changes

- `useAttachMedia` tracks what the adapter is attached to and treats a `null` followed by the same element as a no-op. Detach is deferred to a layout effect that runs after the commit's refs, so a genuinely removed element, an adapter swap, and unmount still detach (in the right order relative to the new attach).
- Attach still happens while the refs run, ahead of the commit's effects, so `useMediaEvents` listeners composed ahead of it keep seeing `loadstart`.
- `useAttachMedia` accepts any `Element`; `useAttachIframe` (an internal copy of the old hook) is deleted and the six embed adapters use `useAttachMedia` directly.
- Reference page updated to the new contract.

<details>

React cannot tell a ref whose identity changed apart from an element that was removed at the moment it calls the ref with `null`, so the ref only records the element and attaches; a `useLayoutEffect` with no deps reconciles afterwards (this is also what re-attaches after React 18's StrictMode simulated unmount, where refs are not double-invoked). Unmount runs the parent's layout cleanup before the child's ref receives `null`, so the unmount cleanup detaches unconditionally rather than reconciling.

Verified against both React 19.2.8 (the suite) and React 18.3.1 (same test bodies with `react`/`react-dom` aliased to the 18 copies in the store); all pass on both.

</details>

## Testing

- `pnpm -F @videojs/react test`: 583 pass. New `use-attach-media.test.tsx` (9 cases: mount/unmount, identity churn, attach-before-effects, composed order, element swap, element removed, adapter swap, failed attach, StrictMode) and `hlsjs-video.test.tsx` against the real `HlsJsAdapter`. Four of the new tests fail on `main`.
- `pnpm lint`, `pnpm typecheck` pass.
- Manual: the reproduction in #2725 (parent re-rendering every 500ms with an inline ref on `HlsJsVideo`) should play at full speed with a single `loadstart`.

## Follow-ups (not in this PR)

- [ ] `renderElement` → `mergeRefs` → `composeRefs` builds a fresh callback every render for the `ref: [forwardedRef, xRef]` call sites, so any ref in that chain with a side effect re-runs it each render (e.g. `VolumeSlider`'s `wheelRef` re-adds its listener).
- [ ] `HlsJsAdapter.attach` forwards a repeat attach of the same element to `hls.attachMedia`, which hls.js answers with a detach and re-attach; guarding at the adapter would make it robust regardless of the caller.

---

> [!NOTE]
> **Medium Risk**
> Changes core ref/attach lifecycle for all React media components; incorrect detach timing could leak adapters or skip cleanup, but behavior is heavily tested including StrictMode and real HlsJsAdapter.
> 
> **Overview**
> Fixes spurious **detach/reattach** when a parent passes an inline `ref` composed with the media attach ref—React calls the callback with `null` then the same DOM node on every identity change, which previously tore down hls.js MediaSource and rebuilt iframe embed players on each parent re-render.
> 
> **`useAttachMedia`** now tracks the attached element and adapter, attaches only when those actually change, and **defers detach** to `useLayoutEffect` so a transient `null` from ref churn is not treated as removal. Attach still runs during the ref callback (before effects) so listeners composed ahead of it still see `loadstart` in order. The hook accepts any **`Element`**; **`useAttachIframe`** is removed and Cloudflare, Spotify, TikTok, Twitch, Vimeo, and YouTube adapters call **`useAttachMedia`** instead.
> 
> Adds **`use-attach-media.test.tsx`** (mount/unmount, identity churn, StrictMode, element/adapter swaps, etc.) and **`hlsjs-video.test.tsx`** asserting a single attach under inline ref rerenders. Docs describe the updated contract for video, audio, and iframe embeds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea486bd6f2d6025632fce3f7d8edfcbfb8bbe286. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cjpillsbury (approved):
Looks like a solid improvement! Also love that we got to unify for iframe and "original recipe" medias! Thanks also @rennanribeiro for surfacing this and giving a good starting point for the fix!

PR #2750, opened 2026-09-15
fix(html): register skin properties in the host document

sampotts:
Packaged HTML skins declare `@property` inside shadow roots, where browsers do not register those properties for interpolation. This makes slider progress jump even though the skin defines a transition. Generate a property-only stylesheet from the compiled skin CSS and include it in the existing host-document stylesheet, preserving conditional groups and avoiding a hard-coded list of property names.

Also reset slotted video and audio margins in the shared shadow stylesheet so page typography does not introduce gaps inside the skin. The reset uses `!important` to override light-DOM theme styles without maintaining a list of player element names or adding attributes.

Validation:

- Compiler and generated-output tests: 11 passed.
- Packaged-player interpolation and margin regression against an important theme rule: Chromium, WebKit, and Firefox passed.
- HTML package build, workspace type check, focused formatting/lint, and workspace checks passed.
- Rebuilt the WordPress plugin against this branch and confirmed document-level registrations, zero video margins, and the unchanged 200 ms delay without plugin workarounds.

---

> [!NOTE]
> **Low Risk**
> CSS and HTML package build output changes with targeted tests; no auth, data, or API surface changes, though incorrect extraction could affect slider animation across packaged players.
> 
> **Overview**
> Packaged HTML skins declare `@property` (e.g. `--media-slider-fill`) inside shadow roots, where browsers do not register them for interpolation—so progress sliders could jump instead of animating. The skin build now **extracts** only `@property` rules (keeping nested `@media` / `@supports` wrappers) via a new `propertyStyles` lightningcss pass, emits **`internal/skins/properties.css`**, and **`global.css` imports** that file so registrations live on the host document.
> 
> **Shadow layout:** `shared.css` forces **`margin: 0 !important`** on slotted `video`/`audio` so page/theme typography cannot add gaps inside the player.
> 
> **Coverage:** Vitest for the extractor; Playwright asserts margin isolation against hostile page CSS and ~50% interpolation on `--media-slider-fill` during a Web Animations API sweep.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f5b523843ac28c300a660b78ff64e948412991bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/html/src/define/shared.css
mihar-22, line 1: Random question: should this file be replaced by our base CSS files in skins package?

mihar-22 (approved):

packages/html/src/define/shared.css
sampotts: Agreed that the general resets should come from the skins package. They’re already included, so we can remove the duplication here and keep this file limited to shadow-DOM integration (:host and ::slotted()). Renaming it to `shadow.css` would make that responsibility clearer.

I'll address in a follow up PR.

PR #566, opened 2026-02-18
feat(site): new home page, docs, and design system

ronald-urbina:
This PR implements the beta version of the **Video.js** site homepage along with a site-wide facelift as outlined in issue [#343](https://github.com/videojs/v10/issues/343).

image

Changes include:
- Homepage new structure and styling for desktop  

- resolves [552](https://github.com/videojs/v10/issues/552) 
- resolves [553](https://github.com/videojs/v10/issues/553) 
- resolves [554](https://github.com/videojs/v10/issues/554) 
- resolves [555](https://github.com/videojs/v10/issues/555) 
- resolves [556](https://github.com/videojs/v10/issues/556) 
- resolves [557](https://github.com/videojs/v10/issues/557) 
- resolves [558](https://github.com/videojs/v10/issues/558) 
- resolves [559](https://github.com/videojs/v10/issues/559) 
- resolves [560](https://github.com/videojs/v10/issues/560) 
- resolves [563](https://github.com/videojs/v10/issues/563)
- resolves [564](https://github.com/videojs/v10/issues/564)
- resolves [562](https://github.com/videojs/v10/issues/562)

Update: partially resolves [562](https://github.com/videojs/v10/issues/561) missing, some titles and code-blocks styles not reflected on Figma. Also dark mode on docs is missing some styling.

PR #2821, opened 2026-09-15
fix(site): restore the 450 body weight and antialiasing

decepulis:
Refs #2645

## Summary

The pre-beta design system set body copy one step above regular — `--font-weight-normal: 450` — while holding code and other mono text at 400, with `antialiased` smoothing on the body to keep the heavier weight from reading thick. #2645 drops both, so body copy renders at 400 with default smoothing.

This restores them, as a suggestion against that branch rather than `main`.

## Changes

- Body copy renders at weight 450 again. Mono and code text stay at 400 via a separate `--font-weight-code-normal` token, which `--text-code--font-weight` points at — without it, code would inherit 450 and round up to 500, since IBM Plex Mono ships static instances rather than a variable range.
- The body carries `antialiased` again, so strokes thin back out and the extra weight lands where it did pre-beta.

Both tokens are the same shape they had before #566 removed them.

## Testing

Manual. Run the site and compare body copy against code spans and the mono labels in the appearance menu — body should sit visibly heavier, code unchanged. Worth a look on both light and dark grounds, since the smoothing change reads differently on each.

---

> [!NOTE]
> **Low Risk**
> Visual-only global CSS and layout class changes; no auth, data, or API behavior.
> 
> **Overview**
> Restores the pre-beta typography treatment for body copy sitewide via global theme tokens and the root layout.
> 
> **Body weight and code separation:** `--font-weight-normal` is set back to **450** so paragraph scales (`text-p1`–`text-p4`) render slightly heavier than regular. Code/mono keeps **400** through a new `--font-weight-code-normal` on `--text-code--font-weight`, avoiding IBM Plex Mono rounding 450 up to 500.
> 
> **Letter-spacing:** New `--tracking-body` / `--tracking-body-tight` tokens replace the generic tracking on paragraph sizes so the heavier weight doesn’t read too wide.
> 
> **Rendering:** The document `body` in `Base.astro` gets **`antialiased`** again, and sans font feature settings enable ligatures alongside non-zero body tracking.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f8ef49fec44167dcda16ef03e6b287eac717f57e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


luwes (approved):
pre approving, have a look at codebot's comment. not sure if that makes sense

Issue #1316, opened 2026-04-13
Bug: FrameworkCase Conditional Rendering Breaks Astro Hydration


decepulis:
## Problem

The fix in #1223 (commit `55a07559`) introduced true conditional rendering in `FrameworkCase.astro` using `Astro.slots.render()` + `set:html`. While this successfully removed hidden DOM content, it breaks Astro hydration — but only on **HTML framework pages**. React pages are unaffected.

Symptoms (reported by @mihar-22):
- Switching frameworks sometimes freezes the page
- Style selector renders empty
- File selector in demos is broken
- Navigating to an HTML component page and trying to switch back to React fails

Astro silently fails to hydrate `client:*` islands. This affects all browsers, not just Safari.

Thanks to @mihar-22 for the report.

## Cause

Commit `55a07559` (`fix(site): true conditional rendering in FrameworkCase (#1223)`). Confirmed via manual bisect.

## Reverting

Reverting that commit restores hydration. The previous `hidden`-attribute workaround is the safe fallback until a proper fix is found.

Reopens #1224

PR #2837, opened 2026-09-16
fix(site): leave non-matching FrameworkCase content out of the html

### LINKED ISSUE #1224: Bug: FrameworkCase Renders Non-Matching Content to the DOM

decepulis:
Closes #1224
Refs #1316

Rebased onto the docs redesign (#2645). Re-verified there; details in Testing.

## Summary

`<FrameworkCase>` now leaves non-matching framework content out of the HTML entirely instead of rendering it behind a `hidden` wrapper. The `hidden` workaround existed because true conditional rendering broke hydration on HTML pages. This change finally explains why and fixes it.

## The problem, in plain terms

A docs page is written once but built twice, once for HTML and once for React. `<FrameworkCase frameworks={["react"]}>` means "only include this chunk on the React version." Ideally the HTML version simply wouldn't contain it.

You'd expect Astro to build the stuff inside a component only when the component asks for it. It doesn't. The moment Astro creates a `FrameworkCase`, it builds everything inside it right away, before `FrameworkCase` gets a chance to say "skip this." So on an HTML page, the React block's interactive islands get fully rendered, and then `FrameworkCase` throws the result away.

Normally that's just wasted work. The catch is that Astro also plays "first one pays": the first island on a page carries the shared setup script that brings every island on the page to life. Astro ticks a box saying "setup script already emitted" and never emits it again. On HTML pages the React block comes first in the file, so its island got the setup script. Then that block went in the trash, taking the script with it. The box stayed ticked. Every island in the HTML block rendered as inert markup with nothing to wake it up.

That's why `{shouldRender && <slot />}` broke only HTML pages, and why it looked like a mystery: it depended purely on which block appeared first in the file.

## The fix, in plain terms

`FrameworkCase` can't stop Astro from building the discarded block, but it can look at what got built before tossing it. When a block doesn't match, it keeps only the `<script>` and `<style>` tags from the discarded output and drops all the actual content. The setup script survives, the non-matching content is gone, and the page hydrates normally.

## Changes

- Non-matching `<FrameworkCase>` blocks no longer produce any markup, so the `data-search-ignore` and `data-llms-ignore` escape hatches on the wrapper are gone with them.
- Matching blocks still render inside `<div class="contents">`, which the heading spacing utilities still rely on after the redesign.
- New `extractScriptsAndStyles` helper with unit tests does the "keep only scripts and styles" part.
- The guide-writing docs note that `<FrameworkCase>` resolves at build time while `<StyleCase>` stays a CSS toggle.

<details>
<summary>Technical details and alternatives considered</summary>

Astro's `AstroComponentInstance` constructor calls every slot function immediately, before the component's own frontmatter runs. For MDX children that means every `client:*` island inside a `<FrameworkCase>` is server-rendered whether or not the component ever renders its `<slot />`. While rendering, Astro dedupes page-level assets by flags on the render result: the first island to stringify gets the `<astro-island>` custom-element definition, its `<style>`, and the directive script (`Astro.idle = …`); hoisted component `<script>` tags are deduped by id the same way.

The fix reads the already-rendered slot output with `Astro.slots.render('default')` (this returns the eagerly rendered string; it does not render again) and re-emits only its `<script>` and `<style>` elements. Those are the only assets Astro dedupes, and the island runtime is order-independent: elements upgrade when the definition arrives and islands wait for their `astro:<directive>` event, so it doesn't matter where on the page the preserved scripts land.

Alternatives that don't work:

- **Reset Astro's dedupe flags instead.** Sibling blocks render concurrently under `Promise.all`, so this races.
- **Upgrade Astro.** 7.3.2 (latest) still has the eager slot evaluation and byte-identical dedupe code.
- **Prune at MDX compile time.** The same compiled page module serves both framework routes, so the compiler can't know which block to drop.

The real upstream fix would be lazy slot evaluation in Astro. That's worth an issue on `withastro/astro`.

</details>

## What this is and is not worth

Measured across the full docs build, comparing this branch against its base.

| Metric | before | after | change |
|---|---|---|---|
| Raw HTML | 84.12 MB | 75.04 MB | −10.8% |
| Brotli | 6.11 MB | 5.90 MB | −3.4% |
| Astro islands | 13,998 | 9,429 | −32.6% |

Brotli already collapses the hidden duplicate, so this is **not** a meaningful bandwidth win. The real wins are DOM weight and hydration work: on the heaviest pages the DOM node count drops up to 42%, and a third of all islands across the docs stop being created. Plus the correctness win the issue was actually filed for, no duplicate framework content reaching search and LLM pipelines, and no `data-search-ignore` scaffolding needed to prevent it.

For context, #1223 reported "42% smaller" for this same change. That number measured the `<article>` element only, uncompressed, on a single page. All three scopes on that page: article uncompressed 37.9%, whole document uncompressed 11.8%, whole document brotli 3.6%.

## Testing

Re-run after rebasing onto the redesign:

- `pnpm -F site test` (45 files, 636 tests, includes the new `extractScriptsAndStyles.test.ts`) and a production `astro build` both pass.
- 306 built docs pages, zero `class="contents" hidden` wrappers remaining.
- Headless Chromium against the static build: every island hydrates (`astro-island:not([ssr])` equals the island count) on both frameworks for `guides/build-your-own-component`, `guides/installation`, `guides/migrate-from-media-chrome`, and `reference/components/mute-button`. Every page keeps exactly one `astro-island` definition and one directive script.
- Confirmed the underlying Astro bug still reproduces on the redesigned site: with the naive conditional slot, `guides/build-your-own-component` renders 46 islands on HTML with zero island definitions, while React is unaffected. See #2838 for a clickable preview of that failure.

To review manually: open any reference page on the deploy preview in both frameworks, confirm the demo tabs and framework switcher work, and check the DOM for `.contents[hidden]` (there should be none).

https://claude.ai/code/session_01H4GB5kE1fCAUeU2VcyjRZ3


site/src/utils/extractScriptsAndStyles.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Returns only the `<script>` and `<style>` elements from an HTML fragment, in document order.
+ *
+ * Astro renders every slot eagerly when it creates a component instance, before the component's own frontmatter runs,
+ * and it deduplicates page-level assets at that moment: the `<astro-island>` definition and its `<style>`, `client:*`
+ * directive scripts, and hoisted component `<script>` tags are each emitted into the first output that needs them and
+ * never again. A component that then discards that slot output would discard the only copy, so callers use this to keep
+ * those assets while dropping the markup around them.
+ */
+export function extractScriptsAndStyles(html: string): string {
+  return html.match(/<(script|style)\b[^>]*>[\s\S]*?<\/\1\s*>/gi)?.join('') ?? '';
```
luwes, line 11: feels a little brittle but lets give it a go, Astro doesn't mark these somehow with specific attributes?

luwes (approved):
LGTM 🏎️ !

PR #2842, opened 2026-09-16
fix(react): forward refs to plain function render targets on react 18

mihar-22:
Refs #2836

## Summary

Alternative to #2836. On React 18 and Preact compat a plain function component passed to `render` never receives the refs its host composes onto it, so the generated skins' `Button`, slider parts, and `MuteButton` lost the media button, popup trigger, and slider thumb refs. `renderElement` now wraps such components in a cached `forwardRef` at render time instead of compiling `forwardRef` into every skin component.

## Changes

- Plain function render targets, bare or wrapped in `memo`, go through one cached `forwardRef` wrapper per component that passes `ref` inside props, matching React 19's ref-as-prop. Intrinsic tags, class components, and existing `forwardRef` components are untouched.
- React 19 keeps the existing `cloneElement` path; the wrapper only activates when the renderer reports a major version below 19.
- Consumers' own plain function components passed to `render` are covered as well, not only vjsc output.
- No compiler, schema, or generated-code changes. Emitted skins stay plain functions, so React 19 users do not inherit deprecated `forwardRef` wrappers in ejected code.

<details>
<summary>Why not the compiler</summary>

#2836 adds an `element: false` flag to the framework-neutral core schema, but the HTML target resolves those same `Root` parts to real custom elements, so the flag is only true for React. It also threads one question, "does this React component take a ref", through the schema, a lookup helper, a target hook, and a rewrite of authored function declarations. The runtime already owns the `render` contract, so the fix sits in one function.

Known limit: a ref placed directly on a generated skin component, not through `render`, still resolves only where the renderer passes `ref` as a prop. That is the same contract Base UI and Radix document for custom render components.

</details>

## Testing

- `pnpm -F @videojs/react test`. The new `use-render-ref-forwarding.test.tsx` reports React 18 through a `react` module mock so the wrapper path runs on the React 19 test renderer: wrapper shape, per-component caching, key preservation, pass-through of tags, classes, and `forwardRef`, `memo` unwrapping, no remount across host renders, and the `Popover.Trigger` to `MuteButton` to `<button>` chain including a click that reaches `toggleMuted`.
- Locally with the workspace pinned to React 18.3.1 through a pnpm override: the three failing scenarios from #2836 (media button `render={<Button />}`, `Slider.Thumb render={<SliderThumb />}`, `Popover.Trigger render={<MuteButton />}`) return `null` refs on main and resolve with this change. The full `@videojs/react` suite shows the same four pre-existing React 18 failures (`wistia-video` ×3, `menu/root` ×1) with and without it.
- A standalone harness on React 18.3.1, Preact 10.29.8 compat, and React 19.3.0 confirms the same, including hook state surviving parent re-renders. On Preact, `memo` around a plain function is not covered.
- There is no React 18 or Preact matrix in CI; `pnpm typecheck` and lint pass.

---

> [!NOTE]
> **Medium Risk**
> Touches shared ref composition in `renderElement`, which all `render`-based UI uses, but behavior is gated to pre-React 19 and covered by new integration-style tests.
> 
> **Overview**
> Fixes broken refs when skins and consumers pass **plain function** components to `render` on React 18 and Preact compat, where refs only work through `forwardRef` (React 19 still uses ref-as-prop and skips wrapping).
> 
> `renderElement` now runs element render targets through `ensureRefForwarding`: it wraps bare or `memo`-wrapped plain functions in a **cached** `forwardRef` that injects `ref` into props, preserves element keys, and recreates the element when the type changes; intrinsic tags, class components, and existing `forwardRef` types are unchanged.
> 
> Adds `use-render-ref-forwarding.test.tsx`, mocking React as 18.3.1 so the wrapper path is exercised on the React 19 test renderer, covering caching, keys, `memo`, ref delivery, stable mounts, and `Popover.Trigger` → `MuteButton`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 18935427cdf1d0b139f2820adca7bb101aae1615. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sampotts (approved):
Much cleaner 👍🏼

PR #2870, opened 2026-09-17
feat(spf): send credentials with hls requests for crossorigin="use-credentials"

### LINKED ISSUE #2859: Feature: Send Request Credentials (Cookies) With Cross-Origin HLS in `<hls-video>` / `<hlsjs-video>`
### Should this be a feature proposal discussion?

- [x] This is not a substantial change

- [x] I have searched the existing issues and discussions (nothing on request credentials / cookies for the HLS media elements; #1404 and #1873 are unrelated)

### Which package(s) does this apply to?

HTML (`@videojs/html`), SPF (`@videojs/spf`), and the `hlsjs-video` adapter

### Motivation

Access-controlled HLS is commonly served from a CDN on a *different origin* than the page — signed cookies scoped to the parent domain (CloudFront signed cookies are the usual case; Cloudflare and Akamai have equivalents). The manifest and every segment then have to be requested `credentials: "include"` (fetch) / `withCredentials = true` (XHR), or the CDN answers 403. That is how our Video.js 8 player streams today: `html5.vhs.withCredentials: true` plus `crossorigin="use-credentials"` on the element.

Migrating that page to v10 (`<video-player>` + `<hls-video>`) I could not find a way to make the engine send credentials, and measured that it does not.

### Description

Measured with `@videojs/cdn@10.0.0-rc.2` in Chromium 153 (headless, Playwright 1.63) against a small fixture server on a second port of the same host — cross-origin, same-site, the CDN shape:

- the fixture answers `OPTIONS` with `Access-Control-Allow-Origin: <origin>` + `Access-Control-Allow-Credentials: true`, sets a cookie on `POST /login`, and returns `403` for `index.m3u8` / `seg0.ts` unless the cookie is present; it logs every request with whether the cookie was present.
- a plain `fetch(url, {credentials: "include"})` from the page gets the playlist (200); a default-mode `fetch(url)` is refused (403) — so the cookie and CORS are right.
- `<video-player><hls-video src="http://host:8766/harness/hls/gated/index.m3u8"></hls-video></video-player>` → manifest request logged with **no cookie** (403).
- the same with `crossorigin="use-credentials"` on `<hls-video>` → still **no cookie**.
- `<hlsjs-video>` (bare, and with `crossorigin="use-credentials"`) → **no cookie** either.
- `<native-hls-video crossorigin="use-credentials">` → the cookie IS sent on the manifest and the segment (the browser's own fetch honours the attribute; Chromium 153 reports native HLS support here), which is consistent with Safari — but it leaves Firefox and older Chromium with no credentialed path.
- element surface: `hls-video` exposes `engine` (`{state, context, destroy}`), no `credentials` / request-init / config property; `hlsjs-video` has no `config` property, so hls.js's `xhrSetup` cannot be reached.

Reading the SPF bundle, the fetcher builds `new Request(url, {method: "GET", headers, ...init})` with an internal init — the plumbing for a `credentials` value exists, it is just not reachable from the element. The migration guide says `html5.vhs.withCredentials` has "no direct equivalent", which matches.

### Expected Behavior

A documented way to have `<hls-video>` (and `<hlsjs-video>`) send credentials on manifest and segment requests — for example honouring the media element's own `crossorigin="use-credentials"` (the attribute native HLS already honours, so one markup works for both paths), or a `credentials="include"` attribute / property on the media element that is passed through to the engine's `RequestInit` (and to hls.js's `xhrSetup`).

### Alternatives and Workarounds

- Wrap the global `fetch` before the module loads and add `credentials: "include"` for the CDN host — works in principle but is invisible to the library and fragile across releases.
- Serve HLS from the page's own origin — not possible for a CDN.
- Use `<native-hls-video>` only — excludes browsers without native HLS.
- Stay on Video.js 8, which is what we are doing until this exists.

### Prior art

- Video.js 8 / VHS: `html5.vhs.withCredentials` (per-player) and `xhr.onRequest` hooks.
- hls.js: `xhrSetup(xhr, url) { xhr.withCredentials = true }` / `fetchSetup` with `credentials: "include"`.
- Shaka: `player.getNetworkingEngine().registerRequestFilter` / `allowCrossSiteCredentials`.
- The native media element: `crossorigin="use-credentials"`.

luwes:
Closes #2859

## Summary

Cookie-gated cross-origin HLS (CloudFront signed cookies and the like) had no way to make `<hls-video>` or `<hlsjs-video>` send credentials, so every manifest and segment request came back 403. Both now honor the media element's standard `crossorigin="use-credentials"`, the attribute native HLS playback already honors, so one markup works for every path. This is the v10 answer to Video.js 8's `html5.vhs.withCredentials`.

## Changes

- `<hls-video>` / `<hls-audio>` (SPF): `crossorigin="use-credentials"` sends cookies (`credentials: 'include'`) on manifest, media playlist, segment, and chapters requests. The attribute is read per request, so changing it applies to the next fetch without rebuilding the engine.
- `<hlsjs-video>`: the same attribute sets `withCredentials` on hls.js's XHR loader and `credentials: 'include'` on its fetch loader. A consumer's own `xhrSetup` / `fetchSetup` from `source.engine.hlsJs` still runs afterwards on the same request and can override it.
- Only `use-credentials` changes anything. `anonymous`, an empty value, or no attribute all leave the platform default (`same-origin`), so streams behind `Access-Control-Allow-Origin: *` keep working exactly as today.
- Docs: the migration guide maps `html5.vhs.withCredentials` to the attribute, and the HlsVideo / HlsJsVideo references gain a cookie-gated streams section with the CORS response the CDN must send.

<details>

- SPF gets a `requestCredentials` input slot on the video and audio engines (materialized via `shareSignals`, like `disableRemotePlayback`). A `credentialsFetch` decorator, sibling of `failoverFetch`, wraps each fetch innermost so a failed credentialed request still trips CDN failover. The slot is optional on the decorator's state type, following the `failedCdns` contract, so behaviors don't need to declare it.
- The SPF mixins gain a `crossOrigin` accessor mirroring the `disableRemotePlayback` pattern: it reflects onto the element through the base (guarded so the standalone core doesn't end up with a shadowing data property) and writes the slot. On `attach()` it applies pre-attach intent or adopts an element's existing attribute, which covers React-rendered `<video crossOrigin>` and standalone attach. `crossOrigin` joins `defaultProps` so `useSyncProps` routes the React prop to the adapter.
- "Attribute absent" deliberately does not map to `include`, even though the browser's own no-CORS media fetches send cookies: `fetch` can't use no-CORS here (bodies must be readable), and a credentialed CORS request is refused by any server answering `*`.
- Not in this PR: SPF text-track segments still load through a `<track>` on an `anonymous` dummy video (`resolve-vtt-segment.ts`) and a raw `fetch` in `resolve-vtt-metadata.ts`; wiring credentials there means changing the `TextTrackSegmentResolver` signature. `hls-background-video` is also not wired yet.

</details>

## Testing

- `pnpm -F @videojs/spf test`: new unit tests for `credentialsFetch` and `crossOriginToRequestCredentials`, adapter tests pinning the `crossOrigin` → `requestCredentials` contract (including a chromium test asserting the manifest `Request.credentials` is `include`).
- `pnpm -F @videojs/hlsjs-video test`: unit tests for the layered `xhrSetup` / `fetchSetup`, plus adapter tests confirming the installed hook follows the attribute without rebuilding the engine and coexists with a configured `xhrSetup`.
- `pnpm typecheck`, `pnpm lint`, and the React / HTML / Mux adapter media tests pass.
- Manual: point `<hls-video crossorigin="use-credentials">` at a CDN that requires a cookie and answers CORS with `Access-Control-Allow-Credentials: true`; the manifest and segments carry the cookie. Without the attribute, requests stay uncredentialed as before.

---

> [!NOTE]
> **Medium Risk**
> Touches all HLS network fetches and credentialed CORS semantics; default behavior is unchanged unless authors set `use-credentials`, but misconfigured CDNs could still break credentialed playback.
> 
> **Overview**
> **Cookie-gated cross-origin HLS** can now use the standard `crossorigin="use-credentials"` on `<hls-video>`, `<hls-audio>`, and `<hlsjs-video>` so manifest, playlist, segment, and (on SPF) chapters requests send cookies—replacing Video.js 8’s `html5.vhs.withCredentials`.
> 
> On **SPF**, adapters expose `crossOrigin`, map `use-credentials` → `credentials: 'include'` on engine `requestCredentials`, and a new `credentialsFetch` wrapper applies that mode per request on manifest/playlist/segment (and chapters) fetches, inside CDN failover. Only `use-credentials` changes behavior; other values keep today’s defaults.
> 
> On **hls.js**, `withRequestCredentials` layers `xhrSetup` / `fetchSetup` to read the element’s `crossOrigin` at request time; consumer hooks from `source.engine.hlsJs` still run afterward.
> 
> Docs add migration and component guidance for cookie-gated streams and required CORS headers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df887186906df226df15611e966608305be62d68. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


luwes:
Follow-up in b24555b: `requestCredentials` is now an engine **config policy** rather than a state slot.

The slot was only ever read untracked per request, so nothing reacted to it — a signal with no subscribers. It's now `HlsVideoEngineConfig.requestCredentials?: RequestCredentials | ((resource) => RequestCredentials | undefined)`, in the same shape as `getCdnId` / `canPlayTrack`. The adapters install a policy that reads their `crossOrigin` mirror, so it stays live across attribute changes without the engine being rebuilt — the policy is consulted per request, not captured at construction. Direct engine consumers get a per-host policy for free, and a consumer-supplied policy wins over the `crossorigin` mapping.

The PR description's "Implementation details" still describes the slot; I'll leave it as history unless you'd prefer it updated.

packages/spf/src/playback/adapters/hls-audio/mixin.ts
```diff
@@ -25,6 +26,11 @@ import {
 export interface HlsAudioAdapterProps {
   src: string;
   preload: '' | 'none' | 'metadata' | 'auto';
+  /**
+   * The element's CORS-settings attribute. `use-credentials` also sends cookies with every manifest, playlist, and
+   * segment request the engine makes; any other value leaves those requests at the platform default.
+   */
+  crossOrigin: string | null;
```
cjpillsbury: nitpick(blocking): use string literals here like we do for e.g. `preload`.

packages/spf/src/playback/adapters/hls-video/mixin.ts
```diff
@@ -125,10 +132,17 @@ const FATAL_SVTA_CODES: ReadonlySet<number> = new Set<number>([
  * @fires error - Fired when a fatal condition is reported. Read `error` for it.
  */
 export function HlsVideoMixin<Base extends Constructor<any>>(BaseClass: Base) {
+  // Whether the base reflects `crossOrigin` onto the media element (the
+  // `HTMLMediaAdapter` family does; the standalone core has no base). Guards
+  // the `super` write: without a setter up the chain it would create a data
+  // property on the instance and shadow the accessor below.
+  const baseReflectsCrossOrigin = 'crossOrigin' in BaseClass.prototype;
```
cjpillsbury: issue(blocking): This feels hacky. Solve this at the class definition level without busting TS and/or base class assumptions (same for audio).

packages/spf/src/playback/adapters/hls-video/mixin.ts
```diff
@@ -335,6 +350,16 @@ export function HlsVideoMixin<Base extends Constructor<any>>(BaseClass: Base) {
     attach(mediaElement: HTMLMediaElement): void {
       super.attach?.(mediaElement);
       this.#signals.context.mediaElement.set(mediaElement);
+
+      // Author intent set before attach reaches the element now (the base only
+      // reflects onto an attached target); otherwise an element that already
+      // carries `crossorigin` (a React-rendered `<video>`, a standalone attach)
+      // is the intent, and the engine adopts it.
+      if (this.#crossOrigin !== null) {
```
cjpillsbury: question(blocking): Is this `null` check right? This is making assumptions. e.g. how would someone "unset" `#crossOrigin` here?

packages/spf/src/playback/adapters/hls-video/mixin.ts
```diff
@@ -370,6 +395,28 @@ export function HlsVideoMixin<Base extends Constructor<any>>(BaseClass: Base) {
...
+    // -------------------------------------------------------------------------
+
+    get crossOrigin(): string | null {
+      return this.#crossOrigin;
+    }
+
+    set crossOrigin(value: string | null) {
+      this.#crossOrigin = value;
+
+      if (baseReflectsCrossOrigin) super.crossOrigin = value;
```
cjpillsbury: issue(blocking): It looks like we're being inconsistent about propagating `crossOrigin` down to the underlying `HTMLMEdiaElement` instance...

packages/spf/src/playback/adapters/hls-video/mixin.ts
```diff
@@ -492,6 +539,10 @@ export function HlsVideoMixin<Base extends Constructor<any>>(BaseClass: Base) {
     #createEngine(): Composition<HlsVideoEngineState, HlsVideoEngineContext> {
       return createHlsVideoEngine({
         ...this.#config,
+        // A policy, not a value: the engine outlives `crossOrigin` changes and
+        // consults this on every request. A consumer-supplied policy wins.
+        requestCredentials:
+          this.#config?.requestCredentials ?? (() => crossOriginToRequestCredentials(this.#crossOrigin)),
```
cjpillsbury, line 545: praise: LGTM!

packages/spf/src/network/request-credentials.ts
```diff
@@ -0,0 +1,18 @@
...
+ * resources with, so a cookie-gated cross-origin stream works the same whether the browser or the engine loads it.
+ * Every other value — `anonymous`, the empty string, an unknown keyword, or no attribute — yields `undefined`, leaving
+ * the platform default (`same-origin`).
+ *
+ * The attribute-absent case deliberately does not map to `'include'` even though the browser's own no-CORS media
+ * fetches send cookies: `fetch` cannot use no-CORS here (bodies must be readable), and a credentialed CORS request is
+ * refused by any server answering `Access-Control-Allow-Origin: *`, which would break every stream that works today.
+ */
+export function crossOriginToRequestCredentials(
+  crossOrigin: string | null | undefined
```
cjpillsbury: nitpick(blocking): use string literals here like we do for e.g. preload.

packages/spf/src/playback/primitives/credentials-fetch.ts
```diff
@@ -0,0 +1,34 @@
...
+ * so this wraps the manifest, playlist, and segment fetches alike; with no policy the fetch is returned as is.
+ *
+ * The policy runs per request rather than once at decoration: it mirrors the media element's `crossorigin` attribute,
+ * which an author can change after the fetch was built. An explicit per-call `credentials` still wins.
+ */
+export function credentialsFetch<Fetch extends CredentialableFetch>(
+  baseFetch: Fetch,
+  policy: RequestCredentialsPolicy | undefined
+): Fetch {
+  if (policy === undefined) return baseFetch;
```
cjpillsbury, line 30: suggestion(non-blocking): Consider adding a code comment here saying "this means if the config-provided policy is itself `undefined` (and not e.g. whether or not to use credentials), we can simply return `baseFetch`.

packages/spf/src/playback/primitives/credentials-fetch.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  return ((addressable: Resource, options?: FetchOptions) => {
+    const credentials = isFunction(policy) ? policy(addressable) : policy;
```
cjpillsbury: nitpick(blocking): Either make `RequestCredentialsPolicy` always a function or promote this normalization to the higher order function closure so we can assume it's always a function here.

packages/spf/src/playback/primitives/credentials-fetch.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (credentials === undefined) return baseFetch(addressable, options);
```
cjpillsbury: nitpick(non-blocking): Ideally this could be a type-safe implementation that is simply:
```
return baseFetch(addressable, { credentials: policy(addressable), ...options });
```
or similar. 

It looks like `"same-origin"` is default per MDN (https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials#value). This would shave off some extra code.

cjpillsbury (changes requested):
Small asks, but this is def an improvement on the SPF side.

packages/spf/src/playback/adapters/hls-audio/mixin.ts
luwes: Done in 1bd0593: `crossOrigin: MediaCrossOrigin | null` where `MediaCrossOrigin = '' | 'anonymous' | 'use-credentials'` (the `''` is the bare attribute, and matches React's own `CrossOrigin` type so `<HlsVideo crossOrigin="">` still typechecks).

packages/spf/src/playback/adapters/hls-video/mixin.ts
luwes: Agreed, that was working around the base rather than with it. Removed in 1bd0593. The setter now reflects straight onto the attached element from `context.mediaElement` — which the mixin already owns — so there's no prototype probe and no assumption about what the base does or doesn't define. Works identically for `HlsVideoAdapterCore` (no base) and `HlsVideoAdapter` (`HTMLVideoAdapter` base); both are now pinned by tests.

packages/spf/src/playback/adapters/hls-video/mixin.ts
luwes: Fair — the old check conflated "never set" with "set to null", and the precedence was backwards relative to `preload`. In 1bd0593 attach follows `syncPreload`'s most-recent-wins rule: an element authored with `crossorigin` is adopted; otherwise a value set before attach reaches the element now. Unsetting is `media.crossOrigin = null`, which removes the attribute from the attached element exactly like the element's own IDL attribute does — pinned in the new "null removes it" test.

packages/spf/src/playback/adapters/hls-video/mixin.ts
luwes: You're right. In 1bd0593 there's one path: the setter always writes `mediaElement.crossOrigin = value` when attached (`null` removes the attribute), and attach reconciles the two directions once. No base-dependent branch anymore.

packages/spf/src/network/request-credentials.ts
luwes: Done in 1bd0593. The function takes `MediaCrossOrigin | null | undefined` and compares with `===`; the `toLowerCase()` is gone since the IDL attribute is limited to known values and can't reflect anything else.

packages/spf/src/playback/primitives/credentials-fetch.ts
luwes: Added in 1bd0593.

packages/spf/src/playback/primitives/credentials-fetch.ts
luwes: Went with the second option in 1bd0593: normalized once in the outer closure (`const resolve = isFunction(policy) ? policy : () => policy`), so the per-request path just calls `resolve(addressable)`. Kept the plain-value form in the type since `requestCredentials: 'include'` is the natural spelling for direct engine consumers.

packages/spf/src/playback/primitives/credentials-fetch.ts
luwes: Yes — it's now exactly that in 1bd0593: `baseFetch(addressable, { credentials: resolve(addressable), ...options })`. `Request` treats an `undefined` init member as absent, so `same-origin` falls out with no branch; the test asserts that via `new Request(url, options).credentials`.

packages/spf/src/network/request-credentials.ts
```diff
@@ -0,0 +1,23 @@
...
+ * `anonymous`, the bare attribute, and no attribute yield `undefined`, leaving the platform default (`same-origin`).
+ *
+ * The attribute-absent case deliberately does not map to `'include'` even though the browser's own no-CORS media
+ * fetches send cookies: `fetch` cannot use no-CORS here (bodies must be readable), and a credentialed CORS request is
+ * refused by any server answering `Access-Control-Allow-Origin: *`, which would break every stream that works today.
+ */
+export function crossOriginToRequestCredentials(
+  crossOrigin: MediaCrossOrigin | null | undefined
+): RequestCredentials | undefined {
+  return crossOrigin === 'use-credentials' ? 'include' : undefined;
```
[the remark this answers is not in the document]
luwes: Good catch — dropping the case fold left the raw attribute path (custom element → setter) unnormalized. Fixed in 1789444: the setter now runs `normalizeCrossOrigin`, which reflects "limited to only known values" the way the element's IDL attribute does (any-case `use-credentials` → itself, everything else → `anonymous`, missing → `null`), while the attribute keeps the author's spelling. Attach adoption uses the same normalizer. Tests cover `USE-CREDENTIALS` → `include` and `bogus` → `anonymous` on the adapter.

packages/spf/src/playback/adapters/hls-video/mixin.ts
```diff
@@ -335,6 +344,17 @@ export function HlsVideoMixin<Base extends Constructor<any>>(BaseClass: Base) {
...
       this.#signals.context.mediaElement.set(mediaElement);
+
+      // Most-recent-wins on attach, as with `preload`: an element authored with
+      // `crossorigin` (a React-rendered `<video>`, a standalone attach) is the
+      // intent; otherwise what was set before attach reaches the element now.
+      const authored = toMediaCrossOrigin(mediaElement.crossOrigin);
+
+      if (authored !== null) {
+        this.#crossOrigin = authored;
+      } else if (this.#crossOrigin !== null) {
```
cjpillsbury, line 355: question(blocking): I still don't understand why we need this `else if`? can we not just always do `mediaElement.crossOrigin = this.#crossOrigin` Why would it matter if we set it to `null` here? In fact, why can't we simply:
```ts
this.#crossOrigin = toMediaCrossOrigin(mediaElement.crossOrigin) ?? this.#crossOrigin;
mediaElement.crossOrigin = this.#crossOrigin;
```

The purpose of this is to prioritize a non-nullish `mediaElement.crossOrigin` value as the `this.#crossOrigin` value and then keep `mediaElement.crossOrigin`'s value in sync with whatever `this.#crossOrigin`'s value is, right?

(same for hls-audio/mixin)

packages/spf/src/playback/primitives/credentials-fetch.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Decorate a fetch so every request carries the mode the policy resolves to. The decorated fetch's type is preserved,
+ * so this wraps the manifest, playlist, and segment fetches alike.
+ *
+ * The policy runs per request rather than once at decoration: it mirrors the media element's `crossorigin` attribute,
+ * which an author can change after the fetch was built. An explicit per-call `credentials` still wins, and a policy
+ * yielding `undefined` leaves the request at the platform default — `Request` treats an `undefined` init member as
+ * absent.
+ */
+export function credentialsFetch<Fetch extends CredentialableFetch>(
```
cjpillsbury, line 24: nitpick(blocking): Sorry I missed this last pass. Unless I'm missing something, this should move to `/network` since it has no `core` or `playback` dependencies (aka it's just a generic "network" function definition.

cjpillsbury (changes requested):
Sorry a couple more small asks here, but it's close!

PR #2879, opened 2026-09-18
fix(site): respond to system color scheme change

sampotts:
Currently the docs require a refresh after the system color scheme changes. This instead listens for the media query change and updates accordingly. 

---

> [!NOTE]
> **Low Risk**
> Theme/appearance behavior only; no auth, data, or API changes. Small risk of duplicate listeners or edge cases around Astro view transitions, but scope is limited to client-side theming.
> 
> **Overview**
> **Fixes live updates when the OS switches light/dark while theme is set to System**, without requiring a page refresh.
> 
> `ThemeInit.astro` now listens for `prefers-color-scheme` changes and re-runs `applyStoredAppearance` on every page load path, so the `<html>` class, datasets, and `theme-color` meta stay in sync even when the appearance menu never mounts.
> 
> The appearance store stops applying the document on OS changes from `themePreference`’s mount hook; instead it exposes a **`systemDark`** atom for UI state. **`AppearanceControls`** uses that atom (with the stored preference) to decide whether dark mode is active—e.g. enabling the dark tone control—when the system scheme flips.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce1895b9450ea7ff97f99ac0395a103b38c79b02. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


decepulis (approved):

decepulis:
There used to be a media query listener in the head. I don't remember seeing that removed...

Issue #2885, opened 2026-09-21
Bug: Skin Stylesheets No Longer Cover UI Components Outside the Skin Composition


luwes:
## Context

User feedback after upgrading `10.0.0-beta.25` → `10.0.0-rc.2`: the packaged skin stylesheet no longer includes CSS for "optional" UI components, i.e. components the default video skin does not itself compose (`SeekButton` / `<media-seek-button>`, `PlaybackRateButton`, `LiveButton`, and any `MuteButton` outside the volume popover). Anyone who placed those beside a packaged skin, or in a beta-era ejected skin, and relied on the shipped CSS now sees them unstyled.

This is a deliberate consequence of the VJSC skin generation that shipped in rc.1 (#2545, #2546), but it was not called out as a breaking change and HTML consumers currently have no supported way to get that CSS back. This issue exists to decide what, if anything, to do about it and to capture the evidence so the decision is not repeated.

## What changed

**beta.25 → beta.32.** `packages/html/src/define/video/skin.css` and `packages/react/src/presets/video/skin.css` both imported `@videojs/skins/default/css/video.css`, a hand-written component library. Its `components/button.css` styled `.media-button--seek`, `.media-button--playback-rate`, `.media-button--live`, and so on regardless of whether the skin rendered them.

**rc.1 onward.** Skin CSS is generated from the skin's module closure:

- `packages/skins/build/variants.ts` → `skinRoots()` collects `collectModules(graph, root.id)`.
- `packages/skins/build/packages/html.ts` and `react.ts` call `bundleStyles(graph, skin.modules, …)` (`packages/vjsc/src/graph/styles.ts`), which emits only the style assets owned by those modules.
- The default video composition (`packages/skins/src/skins/default/video/skin.tsx` + `layout/controls.tsx`) does not include `SeekButton`, `PlaybackRateButton`, or `LiveButton`. Seek buttons left the default video skin in #1832 (beta.26), and #1949 / #2111 explicitly kept `SeekButton` "independently installable" but absent from Default and Minimal video skins.

Rule counts in the generated `packages/html/src/internal/skins/*/skin.css` (React `presets/*/skin.css` matches):

| Skin | `media-seek-button` | `media-playback-rate-button` | `media-live-button` |
| --- | --- | --- | --- |
| default-video, minimal-video | 0 | 0 | 0 |
| default-audio | 5 | 2 | 0 |
| default-live-video, default-live-audio | 0 | 0 | 4 |

[...]

Issue #2526, opened 2026-08-28
Regression: Latest @videojs/html Release (v10.0.0-rc1) Is missing built Zip and Tar.gz


joshsedl:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which package(s) or projects are affected?

HTML (@videojs/html)

### Browser/OS/Node environment

Chrome

### Description

The latest @videojs/html release (v10.0.0-beta.32) is missing built zip and tar.gz. The release before still had it. See: https://github.com/videojs/v10/releases/download/%40videojs%2Fhtml%4010.0.0-beta.31/videojs-html-10.0.0-beta.31.zip:

<img width="400" height="288" alt="Image" src="https://github.com/user-attachments/assets/aeed4212-b4db-409d-bf6a-3f7882122906" />

The build step was originally added through https://github.com/videojs/v10/issues/828

### Steps to reproduce

Check the beta 32 assets. There is no built zip nor tar.gz.

### Current Behavior

There is none.

### Expected Behavior

There should be one.

JPustkuchen:
Would be great to have a new (fixed) release and ensure it works in the future. This is a blocker for the Drupal CMS module once again.

Thanks! :)

joshsedl:
Still relevant for latest rc1 release!

JPustkuchen:
Still seems to be missing in rc2 - is anyone seeing this? @luwes @decepulis? (Sorry for the ping but this is a real issue for the integrations)

luwes:
Thanks for the report. This turned out to be two separate things, and both are addressed on `main`.

### beta.32 (the missing assets you linked)

A race in `cd.yml`. The workflow runs under `concurrency: production`, so the run for the commit *before* the release PR was queued and only started after `chore: release main (#2323)` had merged. release-please reads live `main` through the API and created the `@videojs/html@10.0.0-beta.32` release, but `actions/checkout` had no `ref`, so it checked out the triggering SHA where `package.json` still said `beta.31`. That run built `videojs-html-10.0.0-beta.31.zip` and `gh release upload … --clobber` re-uploaded it onto the **beta.31** release (its asset timestamps are all `2026-08-26T14:56:27Z`). beta.32 got nothing, and the run for the actual release commit was cancelled while still pending. Run: https://github.com/videojs/v10/actions/runs/32982534149

[...]

Issue #2384, opened 2026-08-24
Add a Generic Dialog Component


mihar-22:
## Context

Video.js 10 has specialized alert and error dialogs, but it does not provide a generic modal dialog for application content such as opening a player from a thumbnail. Consumers otherwise need to recreate focus management, background isolation, dismissal behavior, and playback integration themselves.

This was raised in Discussion #2158 while looking for the Video.js 10 replacement for the legacy `ModalDialog` API.

## Scope

Provide a generic dialog primitive for HTML and React that can host arbitrary content, including a Video.js player. Alert and error dialogs should compose the same dialog behavior instead of maintaining parallel modal implementations.

Document how to open a player from an external trigger, start playback when the dialog opens, pause playback when it closes, and migrate from the Video.js 8 `ModalDialog` API.

## Acceptance Criteria

- HTML and React expose a generic Dialog API with controlled and uncontrolled open state.
- The dialog traps focus, restores focus on close, isolates background content, and supports Escape dismissal.
- Dialog title and description parts provide the accessible name and description relationships.
- AlertDialog and ErrorDialog build on the generic dialog primitives while retaining their specialized semantics and behavior.
- Documentation includes a player modal example and migration guidance for the legacy `ModalDialog` API.

## Related

- Discussion #2158

JPustkuchen:
Great work @mihar-22 that looks very promising!

I'm not totally sure, is this based on https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog and it's API and properties? I think that would make sense for unification and web standards?

PR #2887, opened 2026-09-21
feat(cdn): publish one ui bundle per @videojs/html ui definition

### LINKED ISSUE #2886: Bug: Generic Dialog Elements Are Unavailable From @videojs/cdn
## Context

Raised in https://github.com/videojs/v10/issues/2526#issuecomment-5760545155. #2384 shipped the generic Dialog in `@videojs/core`, `@videojs/html`, and `@videojs/react`, and the v8 migration guide points `ModalDialog` users at it. Script-tag installs cannot reach it:

- `@videojs/cdn` publishes preset, `media/*`, `extensions/*`, and `locales/*` bundles only (`packages/cdn/vite.config.ts`). There is no `ui/*` surface, so no `ui/dialog.js` exists to load.
- The skin preset bundles register the elements their template composes. `packages/html/src/internal/skins/*/register.ts` is generated by `packages/skins/build/packages/html.ts` from the `<media-*>` tags in the rendered template. Every skin composes `<media-error-dialog>` with the shared `media-dialog-backdrop`, `-popup`, `-title`, `-description`, and `-close` parts, but none compose `<media-dialog>` or `<media-alert-dialog>`. That is correct: those are application dialogs, not player UI, so they do not belong in a skin's closure.

Verified against the `10.0.0-rc.2` CDN output: a shared chunk defines `media-error-dialog` and every `media-dialog-*` part; no file in the package defines `media-dialog` or `media-alert-dialog`.

**Impact.** With `video.js` from the CDN, the documented HTML anatomy in `site/src/content/docs/reference/components/dialog.mdx` upgrades its parts but never its root, so the trigger and `open` state do nothing and no error is reported. The CDN guide's only advice for adding components is "install from npm instead", which is the audience most likely to be migrating from a v8 `ModalDialog`.

## Reproduction

```html
<script type="module" src="https://cdn.jsdelivr.net/npm/@videojs/cdn@10.0.0-rc.2/video.js"></script>

<button commandfor="video-dialog">Open video</button>
<media-dialog id="video-dialog">
  <media-dialog-backdrop></media-dialog-backdrop>
  <media-dialog-popup>
    <media-dialog-title>Title</media-dialog-title>
  </media-dialog-popup>
</media-dialog>

<script type="module">
  await customElements.whenDefined('media-dialog-popup'); // resolves
  console.log(customElements.get('media-dialog')); // undefined
</script>
```

## Acceptance criteria

- `@videojs/cdn` publishes one bundle per `@videojs/html/ui/*` definition under `ui/`, with the same URL shape, package export, and html-sync test that `media/` and `extensions/` use, so any install can load `<media-dialog>`, `<media-alert-dialog>`, or any other element a skin does not compose.
- Skin bundles keep registering only what their skin composes; the generic dialogs are not added to them.
- The CDN guide documents the `ui/` bundles and no longer states that the CDN publishes no per-component bundles.
- The v8 migration guide gives `ModalDialog` users the CDN and package-manager registrations for the Dialog, not just a link to the reference page.

## Related

- #2384 generic Dialog
- #2526 where this was reported
- #2885 skin stylesheets for components outside the skin composition. That issue owns the CSS half of the ejected-skin story; this one covers element registration only.

luwes:
Closes #2886
Refs #2384, #2526

## Summary

Script-tag installs could not use the generic Dialog shipped in #2384: the skins register only their own error dialog, and the CDN had no per-element bundle to load `<media-dialog>` from. The CDN now publishes one bundle per `@videojs/html` UI definition, and lists all of its media, extension, and UI bundles straight from html's published output instead of mirroring them in source.

## Changes

- New `ui/<element>.js` bundles, one per `@videojs/html/ui/*` definition, with the same URL shape and `exports` entry as `media/` and `extensions/`. The release archive picks them up through the shared entry list.
- The `media/`, `extensions/`, and `ui/` entries are derived from html's `dist/default/define/` listing and built from bare `@videojs/html/...` specifiers. The 90-odd one-line source files and the test that kept them in sync with html are gone; adding a definition to html publishes its CDN bundle with no cdn-side change.
- Skin bundles are unchanged. They keep registering exactly what each skin composes, which includes the error dialog and the shared `media-dialog-*` parts but not the application-owned `<media-dialog>` or `<media-alert-dialog>`.
- The CDN guide documents the `ui/` bundles and no longer sends ejected-skin users to npm for element registration.
- The `implement-spf-use-case` skill no longer tells agents to create a cdn-side media entry.

<details>

Rolldown resolves bare specifiers passed as `input`, and tsdown's object-entry handling passes non-glob values through untouched, so `entry: { 'ui/dialog': '@videojs/html/ui/dialog' }` needs no file on disk. The dev build still follows the `development` condition to html's `dist/dev`; source maps of the built output confirm prod resolves `dist/default` and dev resolves `dist/dev`.

The config is loaded for every workspace task, including before html is built, so the listing returns an empty set rather than throwing when html's output is absent. The pack plugin throws at `buildStart` in that case, so a stale checkout cannot publish a CDN with no media bundles. `build:cdn` already depends on html's build through `workspaceTaskDependencies`, so in practice the guard never fires.

A side effect of building the html modules directly is fewer facade chunks: 699 output files where the source-mirror approach produced 823, with an identical set of 96 entry bundles.

An earlier revision also registered the two dialogs in the skin preset entries so `video.js` alone would upgrade `<media-dialog>`. That was reverted: the skins compose only player-owned UI, and adding application dialogs to every skin consumer to fix a CDN delivery gap was the wrong layer.

CSS for elements outside a skin's composition is a separate question and stays with #2885.

</details>

## Testing

- `pnpm -F @videojs/cdn test` (new `scripts/tests/cdn-html-entries.test.ts` covers the dist-path → entry mapping, flavor `index` collapsing, subpath ordering, and the missing-output case)
- `pnpm build:cdn && pnpm -F @videojs/cdn check:cdn` — 699 bundles self-contained; 8 skins pass
- Diffed the set of prod entry files under `media/`, `extensions/`, `ui/` before and after the refactor: identical 96 files
- `pnpm -F @videojs/cdn build:archive` — 347 bundles from 163 entries, `ui/` included
- `pnpm lint`, `pnpm check:workspace`
- Verified against the built output that `ui/dialog.js` defines `media-dialog`, `ui/alert-dialog.js` defines `media-alert-dialog`, and no skin bundle defines either.

---

> [!NOTE]
> **Medium Risk**
> Changes how all CDN media/extension/UI bundles are discovered and built; a failed html build or mapping bug could omit or misname published entries, though tests and buildStart guards aim to prevent drift.
> 
> **Overview**
> The CDN now ships **`ui/<element>.js`** bundles (same URL/export pattern as `media/` and `extensions/`), so script-tag installs can register UI pieces such as Dialog without pulling everything from npm.
> 
> **Media, extension, and UI entries are no longer hand-maintained** under `packages/cdn/src/`. A new `cdn-html-entries` helper globs `@videojs/html`’s published `define/` output and the Vite pack build uses bare `@videojs/html/...` module specifiers as rolldown inputs. Adding a definition in html automatically gets a CDN bundle; the old one-line facades and the sync test that compared them to html are removed.
> 
> Packaging and checks pick up **`ui/`** (`.gitignore`, `package.json` `files`/`exports`, clean task, self-contained bundle check). Docs and the SPF use-case skill now say CDN media/UI follow html’s define output—no cdn-side entry file—and the CDN guide documents per-element `ui/` loads for player-only or ejected-skin setups.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 165a9c073e8afa75f24b7052267761222895d113. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


site/src/content/docs/guides/migrate-from-video-js-8.mdx
```diff
@@ -463,6 +463,45 @@ Those hotkey and gesture declarations are where `userActions` went. v8's `userAc
 
 Ejecting is also where you take on the skin's CSS. The packaged skins ship styles for everything above; a bare layout renders unstyled until you bring those along, so the copied source includes the skin's stylesheet next to its markup.
 
+## Modal dialogs
```
mihar-22: I don't think this section (the "Modal dialogs subsection") is technically relevant for migration from v8, we can remove it?

mihar-22 (approved):
Awesome! Would be nice if we could just pull directly from `@videojs/html` package source so we don't have to sync these or create these files at all 🙏

site/src/content/docs/guides/migrate-from-video-js-8.mdx
luwes: it's relevant if you want to create app level dialogs with a player inside it I believe
https://github.com/videojs/v10/discussions/2158

thoughts?

luwes:
Both addressed in 93d452027 and 165a9c073:

- **Entries from `@videojs/html`** — yes, it works: rolldown resolves bare specifiers as `input`, and tsdown passes non-glob object entry values straight through. The config now lists html's published `dist/default/define/{media,extensions,ui}` and builds each module from its `@videojs/html/...` specifier. The 90-odd one-liner files and the sync test are gone; adding a definition to html publishes its CDN bundle with no cdn-side change. Dev still resolves the `development` condition to `dist/dev` (checked via source maps). Same 96 entry bundles as before, with fewer facade chunks (699 vs 823 files).
- **Modal dialogs section** — removed; the migration guide is back to the original `ModalDialog` row.

